PR #3987, opened 2022-07-26
feat: collect user agent in normalized span tag `http.useragent`

avara1986:
## Description
Collect user agent in normalized span tag `http.useragent`

## Checklist

## Reviewer Checklist
- [ ] Title is accurate.
- [ ] Add to milestone.


ddtrace/contrib/trace_utils.py
```diff
@@ -102,7 +102,9 @@ def _store_headers(headers, span, integration_config, request_or_response):
 
     for header_name, header_value in headers.items():
         tag_name = integration_config._header_tag_name(header_name)
-        if tag_name is None:
+        if header_name.lower() == "user-agent":
```
brettlangdon, line 105: we'll want to performance test this change just to be on the safe side, since this will get called for every header on every request

ddtrace/contrib/trace_utils.py
```diff
@@ -102,6 +102,7 @@ def _store_headers(headers, span, integration_config, request_or_response):
 
     for header_name, header_value in headers.items():
         tag_name = integration_config._header_tag_name(header_name)
+
```
brettlangdon, line 105:
```suggestion
```

brettlangdon: unrelated change

releasenotes/notes/feat-collect-user-agent-630b2c10f83ac063.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    Collect user agent in normalized span tag `http.useragent`.
```
brettlangdon, line 4:
```suggestion
    Collect user agent in normalized span tag ``http.useragent``.
```

tests/contrib/httpx/test_httpx.py
```diff
@@ -274,8 +274,12 @@ async def test_request_headers(snapshot_context):
     """
     url = get_url("/response-headers?Some-Response-Header=Response-Value")
 
+    # Ceveat(avara1986): Docker imagen "httpbin" set as user-agent "python-httpx/0.23.0".
```
brettlangdon, line 277:
```suggestion
    # Caveat(avara1986): Docker image "httpbin" set as user-agent "python-httpx/0.23.0".
```

ddtrace/contrib/trace_utils.py
```diff
@@ -288,7 +289,11 @@ def set_http_meta(
         span._set_str_tag(http.QUERY_STRING, query)
 
     if request_headers is not None and integration_config.is_header_tracing_configured:
+        """We should store both http.<request_or_response>.headers.<header_name> and http.<key>. The last one
+        is the DD standardized tag for user-agent"""
         _store_request_headers(dict(request_headers), span, integration_config)
+        if request_headers.get("user-agent", False):
+            span.set_tag(http.USER_AGENT, request_headers["user-agent"])
```
brettlangdon, line 296: I would do:

```suggestion
        if "user-agent" in request_headers:
            span.set_tag(http.USER_AGENT, request_headers["user-agent"])
```

Or:

```suggestion
        user_agent = request_headers.get("user-agent")
        if user_agent:
            span.set_tag(http.USER_AGENT, user_agent)
```

The second might be preferred to avoid the double look-up in the dict, but "key in dict" is pretty quick check so I am ok with whatever we think is easier to read.

brettlangdon (changes requested):
We should update the tests for `set_http_meta` as well so we can have good coverage there.

vdeturckheim (approved):

tests/tracer/test_trace_utils.py
```diff
@@ -446,6 +447,44 @@ def test_set_http_meta_no_headers(mock_store_headers, span, int_config):
...
+)
+def test_set_http_meta_headers_useragent(
+    mock_store_headers, user_agent_value, expected_keys, expected, span, int_config
+):
+    int_config.myint.http._header_tags = {"enabled": True}
+    assert int_config.myint.is_header_tracing_configured is True
+    trace_utils.set_http_meta(
+        span,
+        int_config.myint,
+        request_headers={"user-agent": user_agent_value},
```
brettlangdon, line 481: do we normalize headers at some point? do all integrations use this casing (or provide case-insensitive access)?

gnufede (approved):

ddtrace/contrib/trace_utils.py
```diff
@@ -49,6 +49,24 @@
 # starting a "new object" on the UI.
 NORMALIZE_PATTERN = re.compile(r"([^a-z0-9_\-:/]){1}")
 
+# Permutation of possible User Agent header. Mix of upper, camel case and lower of words
+# "user", "agent", separators such as "-" and "_". the result would be like:
+# ['User-Agent', 'USER-AGENT', 'User_Agent', 'USER_AGENT', 'UserAgent', 'USERAGENT', ...]
+# Its important that 'User-Agent' would be the first element because is the most common case, for this reason
+# this variable is a list instead a set (orders matters)
+USER_AGENT_PATTERNS = [
```
ZStriker19, line 57: I hadn't thought of generating patterns this way. I learned from this!

ddtrace/contrib/trace_utils.py
```diff
@@ -102,13 +120,28 @@ def _store_headers(headers, span, integration_config, request_or_response):
         return
 
     for header_name, header_value in headers.items():
+        """config._header_tag_name get a element of the dictionary in config.http._header_tags
```
ZStriker19, line 123:
```suggestion
        """config._header_tag_name gets an element of the dictionary in config.http._header_tags
```

ddtrace/contrib/trace_utils.py
```diff
@@ -102,13 +120,28 @@ def _store_headers(headers, span, integration_config, request_or_response):
...
+        witch get the value from DD_TRACE_HEADER_TAGS environment variable."""
```
ZStriker19, line 124:
```suggestion
        which gets the value from DD_TRACE_HEADER_TAGS environment variable."""
```

ZStriker19: Should this doc string go under the _header_tag_name method declaration? I don't know what the correct practice is, but it seems a bit odd to say what a method does where you use it, but not where it's actually defined.

ZStriker19 (approved):
Just a couple of nits but LGTM!

juanjux (dismissed):

ddtrace/contrib/trace_utils.py
```diff
@@ -102,13 +120,28 @@ def _store_headers(headers, span, integration_config, request_or_response):
...
         span.set_tag(tag_name or _normalize_tag_name(request_or_response, header_name), header_value)
 
 
+def _get_request_header_user_agent(headers):
+    # type: (Dict[str, str], Span, IntegrationConfig, str) -> str
+    """Get user agent from request headers
+    :param headers: A dict of http headers to be stored in the span
+    :type headers: dict or list
+    """
+    for key_pattern in USER_AGENT_PATTERNS:
```
brettlangdon, line 138: 🤔 I am worried about the performance impact of this. it should be worse case `O(n)` where today `n = 18`, but 18 dict gets _could_ (hypothesis) be a little heavy on every web request.

It might be worth us adding performance benchmarks for `set_http_meta` first so we can compare before/after impact.

What do you think?

https://github.com/DataDog/dd-trace-py/tree/1.x/benchmarks


We also have some basic benchmarks for Django and Flask, but we've found them to have mixed results (no performance difference whether tracing is enabled or not, which is not accurate).

brettlangdon: I know we are being this exhaustive to be sure we don't miss it, but aren't there only a finite set of possible values?

We should probably have either WSGI (`HTTP_USER_AGENT`) or do most frameworks offer case insensitive lookups? (I know Flask/Werkzeug is case insensitive, Django has `request.META` which has `HTTP_USER_AGENT`, etc).

I wonder if we need this to be this exhaustive, or if we can/should optimize for known best case scenarios first?

avara1986: At first look could be weird, but I measured the performance of this PR (and other failed attempts) and a list checking the key is faster than regex:
https://gist.github.com/avara1986/07a312d7ca5e2a6e1547eb60c2cd6ca3

Common cases, this list is x10 faster than regex

I like your idea about benchmarks :smiley: what do you think about below plan: 
- WIP This PR 
- Open new PR with benchmark tests on set_http_meta 
- Measure this PR
I like your idea about benchmarks :smiley: what do you think about below plan: 
- WIP This PR 
- Open new PR with benchmark tests on set_http_meta 
- Measure this PR

brettlangdon: that plan sounds good. a separate PR for the benchmark 👍🏻 

then we can create a bunch of scenarios... like what if we have no headers, with and without a matching user-agent value, what about common vs uncommon variants (`HTTP_USER_AGENT`, for example), crazy large set of headers with/without matches, etc, etc...

thanks!

avara1986: Benchmark results of this branch https://github.com/DataDog/dd-trace-py/pull/4047

```
+-------------------------------------------+----------+--------------------------+
| Benchmark                                 | /1.4.0   | /artifacts/2372bcb0-f1b1 |
+===========================================+==========+==========================+
| sethttpmeta-no-useragentvariant           | 251 ms   | 227 ms: 1.11x faster     |
+-------------------------------------------+----------+--------------------------+
| sethttpmeta-all-disabled                  | 86.8 ms  | 89.7 ms: 1.03x slower    |
+-------------------------------------------+----------+--------------------------+
| sethttpmeta-useragentvariant_exists_1     | 228 ms   | 243 ms: 1.06x slower     |
+-------------------------------------------+----------+--------------------------+
| sethttpmeta-useragentvariant_exists_2     | 255 ms   | 261 ms: 1.02x slower     |
+-------------------------------------------+----------+--------------------------+
| sethttpmeta-useragentvariant_exists_3     | 230 ms   | 243 ms: 1.05x slower     |
+-------------------------------------------+----------+--------------------------+
| sethttpmeta-useragentvariant_not_exists_1 | 231 ms   | 235 ms: 1.02x slower     |
+-------------------------------------------+----------+--------------------------+
| sethttpmeta-useragentvariant_not_exists_2 | 232 ms   | 235 ms: 1.02x slower     |
+-------------------------------------------+----------+--------------------------+
| Geometric mean                            | (ref)    | 1.01x slower             |
+-------------------------------------------+----------+--------------------------+
```

juanjux (approved):

majorgreys (approved):
With the addition of a scenario in #4047, I was able to get the following results for this PR (commit 26affbeb50c40454e52b4d88465e670a77c24d08). We see a pretty consistent worst case increase of 7% (2-3 us) when one of the user agent variants is present in the request headers. This a best case increase of 2-3% when there is no match is important. I would conclude that this PR does not introduce a performance regression. 

| Benchmark                                 | 1.3.2 | 1.4.0rc2.dev81+g26affbeb |
|-------------------------------------------|--------------------------------------|-----------------------|
| sethttpmeta-no-useragentvariant           | 33.7 us                              | 34.6 us: 1.03x slower |
| sethttpmeta-all-disabled                  | 13.6 us                              | 14.2 us: 1.04x slower |
| sethttpmeta-useragentvariant_exists_1     | 34.4 us                              | 36.7 us: 1.07x slower |
| sethttpmeta-useragentvariant_exists_2     | 37.4 us                              | 40.2 us: 1.07x slower |
| sethttpmeta-useragentvariant_exists_3     | 34.3 us                              | 36.8 us: 1.07x slower |
| sethttpmeta-useragentvariant_not_exists_1 | 34.3 us                              | 35.2 us: 1.03x slower |
| sethttpmeta-useragentvariant_not_exists_2 | 34.5 us                              | 35.3 us: 1.02x slower                       |

brettlangdon (approved):

PR #3992, opened 2022-07-27
fix(internal): span sampling rule match logic

ZStriker19:
## Description
Previous logic did not cover the possibility of spans not having a service name or not having an operation name.  Previously if the span only had a service or only had and operation name, we would get an error when the GlobMatcher tried to take the length of the subject, which for whichever was missing (the name or the service) was None `len(None)`: https://github.com/DataDog/dd-trace-py/blob/1.x/ddtrace/internal/glob_matching.py#L24

## Testing strategy
I've added tests for the span sampling rule to test matching on spans with only a service name and spans with only an operation name.

## Checklist

## Reviewer Checklist
- [ ] Title is accurate.
- [ ] Add to milestone.


majorgreys (commented):
I understand the bug being with spans missing attributes that there are glob matchers in a span sampling rule. The tests look good as does the addition of checks on the span attributes. But I'm not sure the motivation for replacing the catch-all glob matcher with the need to bring back the `hasattr` for the service and name matchers. Was that change necessary?

ZStriker19:
To me it seems cleaner, because otherwise we'd have a service_matcher and name_matcher and the only way to check if we actually wanted to use them would be to make sure their pattern isn't `*`  which is unreliable, because a user could enter `*` as their pattern.

Maybe a better option than what I currently have would be to just set the matcher to None if there's no pattern for it? That avoids having to do hasattr() every time, which seems better. I'll make that change now, but let me know if you have a different opinion of how it should look.

ddtrace/internal/sampling.py
```diff
@@ -155,7 +148,28 @@ def _sample(self, span):
...
+            if span.service is None:
+                return False
+            else:
+                service_match = self._service_matcher.match(span.service)
+        if self._name_matcher:
+            if span.name is None:
+                return False
+            else:
+                name_match = self._name_matcher.match(span.name)
+        return service_match and name_match
```
mabdinur, line 172: Would this work?

```suggestion
        service_match = False
        if self._service_matcher and span.service:
                service_match = self._service_matcher.match(span.service)
        name_match = False
        if self._name_matcher and span.name:
                name_match = self._name_matcher.match(span.name)
        return service_match and name_match
```

mabdinur (commented):
Left one comment. Aside from that LGTM

ddtrace/internal/sampling.py
```diff
@@ -155,7 +148,28 @@ def _sample(self, span):
...
+        if service is None and name is None:
+            return False
+
+        # Default to True, as the rule may not have a name or service rule
+        # For whichever rules it does have, it will attempt to match on them
+        service_match = True
+        name_match = True
+
+        if self._service_matcher:
+            if span.service is None:
```
brettlangdon, line 163:
```suggestion
            if service is None:
```

ddtrace/internal/sampling.py
```diff
@@ -155,7 +148,28 @@ def _sample(self, span):
...
+                return False
+            else:
+                service_match = self._service_matcher.match(span.service)
+        if self._name_matcher:
+            if span.name is None:
```
brettlangdon, line 168:
```suggestion
            if name is None:
```

ddtrace/internal/sampling.py
```diff
@@ -155,7 +148,28 @@ def _sample(self, span):
...
+                return False
+            else:
+                name_match = self._name_matcher.match(span.name)
```
brettlangdon, line 171:
```suggestion
                name_match = self._name_matcher.match(name)
```

ddtrace/internal/sampling.py
```diff
@@ -155,7 +148,28 @@ def _sample(self, span):
...
+        # Default to True, as the rule may not have a name or service rule
+        # For whichever rules it does have, it will attempt to match on them
+        service_match = True
+        name_match = True
+
+        if self._service_matcher:
+            if span.service is None:
+                return False
+            else:
+                service_match = self._service_matcher.match(span.service)
```
brettlangdon, line 166:
```suggestion
                service_match = self._service_matcher.match(service)
```

brettlangdon (commented):
nits, but otherwise lgtm

Kyle-Verhoog (approved):
lgtm with @brettlangdon's suggestions

ddtrace/internal/sampling.py
ZStriker19: Unfortunately not, because for example if you only set a name pattern e.g `rule = SpanSamplingRule(name="test_name")` then you end up with the service_match being False and the evaluation always being False, when it should be true as long as the span has name="test_name".

brettlangdon (approved):

mabdinur (approved):

PR #4042, opened 2022-08-03
feat(internal): span sampling file config envar

ZStriker19:
## Description
This PR introduces the ability to configure span sampling rules via a file. The file path should be specified with `DD_SPAN_SAMPLING_RULES` and must be json.

## Motivation
This is required in order for users to specify the rules they would like to be applied for single span sampling.

## Design 
This is mostly accomplished by modifying get_span_sampling_rules(). If both `DD_SPAN_SAMPLING_RULES` and `DD_SPAN_SAMPLING_RULES_FILE` are specified, then we should default to `DD_SPAN_SAMPLING_RULES`.

## Testing strategy
Most of the testing was done to make sure configurations were as expected.

## Checklist

## Reviewer Checklist
- [ ] Title is accurate.
- [ ] Add to milestone.


ddtrace/internal/sampling.py
```diff
@@ -208,44 +208,60 @@ def apply_span_sampling_tags(self, span):
...
+        # If sample_rate not specified default to 100%
+        sample_rate = float(rule.get("sample_rate", 1.0))
+        service = rule.get("service")
+        name = rule.get("name")
+        # If max_per_second not specified default to no limit
+        max_per_second = int(rule.get("max_per_second", -1))
+        if service is None and name is None:
+            raise ValueError("Neither service or name specified for single span sampling rule:%r" % rule)
+        if service:
+            _check_unsupported_pattern(service)
```
brettlangdon, line 253: I didn't catch this before, but we should validate that service and name are actually strings, we have our typing say they are, but they could be anything.

brettlangdon: reason why this would be nice is because if someone gives you a number then `for char in 10:` will fail and they'll get a weird `TypeError: 'int' object is not iterable`

tests/tracer/test_single_span_sampling_rules.py
```diff
@@ -119,6 +119,169 @@ def test_single_span_rules_do_not_tag_if_tracer_samples():
...
+    file.write('[{"sample_rate":1.0}]')
+    with override_env(dict(DD_SPAN_SAMPLING_RULES_FILE=str(file))):
+        with pytest.raises(ValueError):
+            sampling_rules = get_span_sampling_rules()
+            assert sampling_rules is None
+
+
+def test_exception_thrown_service_unsupported_char_via_file(tmpdir):
+    file = tmpdir.mkdir("data").join("rules.json")
+    file.write('[{"sample_rate":1.0}]')
```
brettlangdon, line 189: is this the correct rules for "unsupported_char" ?

tests/tracer/test_single_span_sampling_rules.py
```diff
@@ -119,6 +119,169 @@ def test_single_span_rules_do_not_tag_if_tracer_samples():
...
+
+        assert sampling_rules[0].match(span) is True
+
+        assert_sampling_decision_tags(span, sample_rate=None, mechanism=None, limit=None, trace_sampling=True)
+
+
+def test_wrong_file_path(tmpdir):
+    """Test that single span sampling tags are not applied to spans that do not match rules via file"""
+    file = tmpdir.mkdir("data").join("rules.json")
+    file.write('[{"service":"test_ser","name":"test_na"}]')
```
brettlangdon, line 260:
```suggestion
```

no need to create this if we aren't using it

tests/tracer/test_single_span_sampling_rules.py
```diff
@@ -119,6 +119,169 @@ def test_single_span_rules_do_not_tag_if_tracer_samples():
         assert_sampling_decision_tags(span, sample_rate=None, mechanism=None, limit=None, trace_sampling=True)
 
 
+def test_sampling_rule_init_sngl_rule_via_file(tmpdir):
+    file = tmpdir.mkdir("data").join("rules.json")
```
brettlangdon, line 123: you probably don't need `.mkdir("data")` here, `tmpdir` should be some random temporary directory so you can be safe to just do `tmpdir.join("rules.json")`

not that big of a deal, more FYI.

brettlangdon (commented):
minor nits, otherwise lgtm

ZStriker19:
@Mergifyio backport 1.4.0

tests/tracer/test_single_span_sampling_rules.py
ZStriker19: thanks for the catch!

ddtrace/internal/sampling.py
```diff
@@ -208,44 +210,62 @@ def apply_span_sampling_tags(self, span):
...
+    for rule in json_rules:
+        if not isinstance(rule, dict):
+            raise TypeError("rule specified via DD_SPAN_SAMPLING_RULES is not a dictionary:%r" % rule)
+        # If sample_rate not specified default to 100%
+        sample_rate = float(rule.get("sample_rate", 1.0))
+        service = rule.get("service")
+        name = rule.get("name")
+        # If max_per_second not specified default to no limit
+        max_per_second = int(rule.get("max_per_second", -1))
+        if not isinstance(service, (string_types, type(None))):
```
brettlangdon, line 249:
```suggestion
        if service is not None and not isinstance(service, string_types):
```

?

ZStriker19: Ok I'll change it, but what's the difference between these two?

mabdinur (commented):
aside from that LGTM

ddtrace/internal/sampling.py
```diff
@@ -208,44 +210,62 @@ def apply_span_sampling_tags(self, span):
 
 def get_span_sampling_rules():
```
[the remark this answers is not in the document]
ZStriker19: I'm usually kind of partial to breaking things up into multiple methods too, but since we're not reusing code, doesn't it make more sense to have it all in one method? @Kyle-Verhoog what do you think?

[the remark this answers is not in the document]
jd: splitting might make it easier to test. my 2c.

ddtrace/internal/sampling.py
```diff
@@ -208,44 +210,62 @@ def apply_span_sampling_tags(self, span):
...
+
+    try:
+        json_rules = json.loads(raw_json_rules)
+        if not isinstance(json_rules, list):
+            raise TypeError("DD_SPAN_SAMPLING_RULES is not list, got %r" % json_rules)
+    except JSONDecodeError:
+        raise ValueError("Unable to parse DD_SPAN_SAMPLING_RULES=%r" % env_json_rules)
+
+    sampling_rules = []
+    for rule in json_rules:
```
jd, line 240: this whole thing should be written with a schema library at this point, pydantic or the like 😢

jd (dismissed):
I am not a fan of the testing logic, it seems we're testing both the file-reading code and the logical JSON-reading code.
Except the json-reading-parsing code is not changed here so the tests added should just be added if needed in a different PR to validate the logic of parsing the JSON.

I agree that we should just have a `file -> json` function that fills the existing code so we can rely on the existing tests, not duplicating or adding a bunch of tests like here. :)

ddtrace/internal/sampling.py
```diff
@@ -208,44 +210,62 @@ def apply_span_sampling_tags(self, span):
...
+            (
+                "DD_SPAN_SAMPLING_RULES and DD_SPAN_SAMPLING_RULES_FILE detected. "
+                "Defaulting to DD_SPAN_SAMPLING_RULES value."
+            )
+        )
+        raw_json_rules = env_json_rules
+    elif env_json_rules:
+        raw_json_rules = env_json_rules
+    elif file_json_rules:
+        with open(file_json_rules) as f:
```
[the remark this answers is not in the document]
ZStriker19: I think I was told it's ok to crash at startup? Do you think we shouldn't?

ddtrace/internal/sampling.py
ZStriker19: Pydantic looks cool, I'd be happy to give that a go but wondering if there's a reason we don't use it anywhere else? Since it uses Python 3.6 type hinting, could that cause issues for 2.7?

ZStriker19: Will take a look at json schema validator

ZStriker19:
after this is merged we can merge the json schema validation https://github.com/DataDog/dd-trace-py/pull/4064

tests/tracer/test_single_span_sampling_rules.py
```diff
@@ -119,6 +122,56 @@ def test_single_span_rules_do_not_tag_if_tracer_samples():
...
+    with override_env(dict(DD_SPAN_SAMPLING_RULES_FILE=str(file))):
+        sampling_rules = get_file_json()
+        assert sampling_rules[0]["service"] == "xy?"
+        assert sampling_rules[0]["name"] == "a*c"
+
+        assert sampling_rules[1]["sample_rate"] == 0.5
+        assert sampling_rules[1]["service"] == "my-service"
+        assert sampling_rules[1]["name"] == "my-name"
+        assert sampling_rules[1]["max_per_second"] == "20"
+        assert len(sampling_rules) == 2
```
brettlangdon, line 141: this is probably better as a direct equality of the list, e.g.

```python
assert sampling_rules == [
  {"service": "xy?", "name": "a*c"},
  {"sample_rate": 0.5, ..... },
]
```

since we want to ensure these are the only properties present.

brettlangdon (commented):
nit for one test, otherwise lgtm

tests/tracer/test_single_span_sampling_rules.py
brettlangdon: this was marked as resolved, but not updated?

ddtrace/internal/sampling.py
```diff
@@ -208,44 +211,90 @@ def apply_span_sampling_tags(self, span):
 
 def get_span_sampling_rules():
     # type: () -> List[SpanSamplingRule]
-    json_rules_raw = os.getenv("DD_SPAN_SAMPLING_RULES")
-    if json_rules_raw is None:
+    json_rules = get_span_sampling_json()
```
majorgreys, line 214: All but the `get_span_sampling_rules` function can be private so prefixed with underscore.

ddtrace/internal/sampling.py
```diff
@@ -208,44 +211,90 @@ def apply_span_sampling_tags(self, span):
...
+    sampling_rules = []
+    for rule in json_rules:
+        if not isinstance(rule, dict):
+            raise TypeError("rule specified via DD_SPAN_SAMPLING_RULES is not a dictionary:%r" % rule)
+        # If sample_rate not specified default to 100%
+        sample_rate = float(rule.get("sample_rate", 1.0))
+        service = rule.get("service")
+        name = rule.get("name")
+        # If max_per_second not specified default to no limit
+        max_per_second = int(rule.get("max_per_second", -1))
```
majorgreys, line 224: We should add a constant for this special value.

tests/tracer/test_single_span_sampling_rules.py
```diff
@@ -119,6 +122,56 @@ def test_single_span_rules_do_not_tag_if_tracer_samples():
...
 
+def test_sampling_rule_init_config_multiple_sampling_rule_json_via_file(tmpdir):
+    file = tmpdir.join("rules.json")
+    file.write(
+        '[{"service":"xy?","name":"a*c"}, \
+            {"sample_rate":0.5,"service":"my-service","name":"my-name", "max_per_second":"20"}]'
+    )
+
+    with override_env(dict(DD_SPAN_SAMPLING_RULES_FILE=str(file))):
+        sampling_rules = get_file_json()
```
majorgreys, line 133: Do we want to be calling `get_span_sampling_rules` instead?

ddtrace/internal/sampling.py
```diff
@@ -208,44 +211,90 @@ def apply_span_sampling_tags(self, span):
...
+        return None
+
+
+def get_env_json():
+    # type: () -> Optional[List[Dict[str, Any]]]
+    env_json_raw = os.getenv("DD_SPAN_SAMPLING_RULES")
+    if env_json_raw:
+        return load_span_sampling_json(env_json_raw)
+    else:
+        return None
```
mabdinur, line 285:
```suggestion
    return None
```

tests/tracer/test_single_span_sampling_rules.py
```diff
@@ -119,6 +122,56 @@ def test_single_span_rules_do_not_tag_if_tracer_samples():
...
+        assert sampling_rules[1]["service"] == "my-service"
+        assert sampling_rules[1]["name"] == "my-name"
+        assert sampling_rules[1]["max_per_second"] == "20"
+        assert len(sampling_rules) == 2
+
+
+def test_wrong_file_path(tmpdir):
+    """Test that single span sampling tags are not applied to spans that do not match rules via file"""
+    with override_env(dict(DD_SPAN_SAMPLING_RULES_FILE="data/this_doesnt_exist.json")):
+        if sys.version_info.major > 3:
```
majorgreys, line 147: Avoid repeating the test code by setting a variable for the exception class we should expect.

ddtrace/internal/sampling.py
```diff
@@ -208,44 +211,90 @@ def apply_span_sampling_tags(self, span):
...
-
-            try:
-                sampling_rule = SpanSamplingRule(
-                    sample_rate=sample_rate, service=service, name=name, max_per_second=max_per_second
-                )
-            except Exception as e:
-                raise ValueError("Error creating single span sampling rule {}: {}".format(json.dumps(rule), e))
-            sampling_rules.append(sampling_rule)
-        return sampling_rules
+        return None
```
mabdinur, line 276: You don't need the else here either

tests/tracer/test_single_span_sampling_rules.py
```diff
@@ -119,6 +122,56 @@ def test_single_span_rules_do_not_tag_if_tracer_samples():
...
+            with pytest.raises(FileNotFoundError):
+                sampling_rules = get_span_sampling_rules()
+                assert sampling_rules is None
+        else:
+            with pytest.raises(IOError):
+                sampling_rules = get_span_sampling_rules()
+                assert sampling_rules is None
+
+
+def test_default_to_env_if_both_env_and_file_config(tmpdir):
```
majorgreys, line 157: Add a check for the warning log.

ddtrace/internal/sampling.py
```diff
@@ -208,44 +211,90 @@ def apply_span_sampling_tags(self, span):
...
+            )
+        )
+        return env_json_rules
+    elif env_json_rules:
+        return env_json_rules
+    elif file_json_rules:
+        return file_json_rules
+    # No rules specified
+    else:
         return []
```
mabdinur, line 266: This could be simplified to:
```suggestion
        return env_json_rules or file_json_rules or []
```

mabdinur (approved):
small nits but feel free to ignore

tests/tracer/test_single_span_sampling_rules.py
```diff
@@ -119,6 +122,56 @@ def test_single_span_rules_do_not_tag_if_tracer_samples():
...
+        )
+    ):
+        sampling_rules = get_span_sampling_rules()
+        assert sampling_rules[0]._sample_rate == 0.5
+        assert sampling_rules[0]._service_matcher.pattern == "xyz"
+        assert sampling_rules[0]._name_matcher.pattern == "abc"
+        assert sampling_rules[0]._max_per_second == 100
+        assert len(sampling_rules) == 1
+
+
```
majorgreys, line 174: Add tests for json decode errors when loading from file or env.

tests/tracer/test_single_span_sampling_rules.py
ZStriker19: Given prior feedback about breaking the function up a bit more, I thought it made sense just to test that get_file_json() returned what we expected.

tests/tracer/test_single_span_sampling_rules.py
ZStriker19: Ooops sorry, I updated in the wrong branch before.

ddtrace/internal/sampling.py
ZStriker19: Ah it seemed like I was getting mypy errors from this before, so I added the else and they went away... but I guess it must've been something else, because it works now. Thanks!

ddtrace/internal/sampling.py
ZStriker19: 😮

tests/tracer/test_single_span_sampling_rules.py
```diff
@@ -119,6 +122,56 @@ def test_single_span_rules_do_not_tag_if_tracer_samples():
...
+    """Test that single span sampling tags are not applied to spans that do not match rules via file"""
+    with override_env(dict(DD_SPAN_SAMPLING_RULES_FILE="data/this_doesnt_exist.json")):
+        if sys.version_info.major > 3:
+            with pytest.raises(FileNotFoundError):
+                sampling_rules = get_span_sampling_rules()
+                assert sampling_rules is None
+        else:
+            with pytest.raises(IOError):
+                sampling_rules = get_span_sampling_rules()
+                assert sampling_rules is None
```
mabdinur, line 154:
```suggestion
        exception = IOError 
        if sys.version_info.major > 3:
            exception = FileNotFoundError

        with pytest.raises(exception):
            sampling_rules = get_span_sampling_rules()
            assert sampling_rules is None
```

ZStriker19: I did this instead
`exception = FileNotFoundError if sys.version_info.major > 3 else IOError
        with pytest.raises(exception):
            sampling_rules = get_span_sampling_rules()
            assert sampling_rules is None`

brettlangdon (approved):

majorgreys (dismissed):

mabdinur (approved):

tests/tracer/test_single_span_sampling_rules.py
```diff
@@ -75,7 +78,23 @@ def test_sampling_rule_init_via_env():
...
+        with pytest.raises(TypeError):
+            get_span_sampling_rules()
+
+
+def test_json_decode_error_throws_ValueError():
+    with override_env(
+        dict(DD_SPAN_SAMPLING_RULES='{"sample_rate":0.5,"service":"xyz","name":"abc","max_per_second":100')
+    ):
+        with pytest.raises(ValueError):
+            get_span_sampling_rules()
```
P403n1x87, line 94: These tests are probably not necessary as presumably `get_span_sampling_rules` is tested elsewhere.

tests/tracer/test_single_span_sampling_rules.py
```diff
@@ -119,6 +138,56 @@ def test_single_span_rules_do_not_tag_if_tracer_samples():
...
+        ]
+
+
+def test_wrong_file_path(tmpdir):
+    """Test that single span sampling tags are not applied to spans that do not match rules via file"""
+    with override_env(dict(DD_SPAN_SAMPLING_RULES_FILE="data/this_doesnt_exist.json")):
+        exception = FileNotFoundError if sys.version_info.major > 3 else IOError
+        with pytest.raises(exception):
+            sampling_rules = get_span_sampling_rules()
+            assert sampling_rules is None
```
P403n1x87, line 162:
```suggestion
```
we probably never hit this line?

tests/tracer/test_single_span_sampling_rules.py
ZStriker19: They were tested before but not like this, since adding the file I ended up doing some refactoring and also arguably wasn't strict enough going through testing with the envar config introduction.

mabdinur (approved):

brettlangdon (approved):

majorgreys:
@Mergifyio backport 1.4

PR #4047, opened 2022-08-04
feat(perf): add set_http_meta scenario

avara1986:
## Description
Add benchmark tests to `ddtrace.contrib.trace_utils.set_http_meta`

Results between branch #3987  and 1.4.0rc1
```
+-------------------------------+--------------------------------------------------------------------------------------+-------------------------------------------------------------------------------------------------------+
| Benchmark                     | /artifacts/05a38bf3-a628-46d2-93f4-1c397cd3ac85/set_http_meta/1.4.0rc1//results.json | /artifacts/05a38bf3-a628-46d2-93f4-1c397cd3ac85/set_http_meta/1.3.0rc2.dev113+gbcf523c5//results.json |
+===============================+======================================================================================+=======================================================================================================+
| sethttpmeta-start             | 13.5 us                                                                              | 13.9 us: 1.03x slower                                                                                 |
+-------------------------------+--------------------------------------------------------------------------------------+-------------------------------------------------------------------------------------------------------+
| sethttpmeta-all-enabled       | 29.3 us                                                                              | 28.2 us: 1.04x faster                                                                                 |
+-------------------------------+--------------------------------------------------------------------------------------+-------------------------------------------------------------------------------------------------------+
| sethttpmeta-all-disabled      | 12.4 us                                                                              | 13.8 us: 1.12x slower                                                                                 |
+-------------------------------+--------------------------------------------------------------------------------------+-------------------------------------------------------------------------------------------------------+
| sethttpmeta-no-useragentkey   | 43.8 ns                                                                              | 43.2 ns: 1.01x faster                                                                                 |
+-------------------------------+--------------------------------------------------------------------------------------+-------------------------------------------------------------------------------------------------------+
| sethttpmeta-useragentvariants | 151 us                                                                               | 155 us: 1.02x slower                                                                                  |
+-------------------------------+--------------------------------------------------------------------------------------+-------------------------------------------------------------------------------------------------------+
| Geometric mean                | (ref)                                                                                | 1.02x slower                                                                                          |
+-------------------------------+--------------------------------------------------------------------------------------+-------------------------------------------------------------------------------------------------------+
```

## Checklist

## Reviewer Checklist
- [ ] Title is accurate.
- [ ] Add to milestone.


benchmarks/set_http_meta/scenario.py
```diff
@@ -0,0 +1,132 @@
...
+            if self.useragentvariants:
+                user_agent_keys = ["HTTP_USER_AGENT", "USER_AGENT", "UserAgent", "User-Agent", "user-agent"]
+
+        def bm(loops):
+            for _ in range(loops):
+                for user_agent_key in user_agent_keys:
+                    data["request_headers"][user_agent_key] = "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 " \
+                                                              "(KHTML, like Gecko) Chrome/92.0.4515.159 Safari/537.36"
+                    span = utils.gen_span(str("test"))
+                    span._local_root = utils.gen_span(str("root"))
```
brettlangdon, line 129: do we need a local root?

benchmarks/set_http_meta/scenario.py
```diff
@@ -0,0 +1,132 @@
...
+            user_agent_keys = ["HTTP_USER_AGENT"]
+            if self.useragentvariants:
+                user_agent_keys = ["HTTP_USER_AGENT", "USER_AGENT", "UserAgent", "User-Agent", "user-agent"]
+
+        def bm(loops):
+            for _ in range(loops):
+                for user_agent_key in user_agent_keys:
+                    data["request_headers"][user_agent_key] = "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 " \
+                                                              "(KHTML, like Gecko) Chrome/92.0.4515.159 Safari/537.36"
+                    span = utils.gen_span(str("test"))
```
brettlangdon, line 128: is the `str()` needed?

brettlangdon: it'll be better if we can create the span once outside of this loop to avoid measuring the cost of creating the span (some isinstance checks, and random number generator, and getting current timestamp.

benchmarks/set_http_meta/scenario.py
brettlangdon: would be better to avoid creating a second span if we don't need it.

benchmarks/set_http_meta/scenario.py
```diff
@@ -0,0 +1,132 @@
...
+
+        user_agent_keys = []
+        if self.useragentkey:
+            user_agent_keys = ["HTTP_USER_AGENT"]
+            if self.useragentvariants:
+                user_agent_keys = ["HTTP_USER_AGENT", "USER_AGENT", "UserAgent", "User-Agent", "user-agent"]
+
+        def bm(loops):
+            for _ in range(loops):
+                for user_agent_key in user_agent_keys:
```
brettlangdon, line 125: I wonder if these should be different variants?

If we run them all in one benchmark will it just show case the average case instead of highlighting the best v worse case scenarios?

e.g. if `HTTP_USER_AGENT` is always the fastest to lookup because we check it first, vs `user-agent` being the slowest because it is the last... for Django META/WSGI environ it'll always be fast, but for a framework only provides `user-agent` it'll always be slow.

benchmarks/set_http_meta/scenario.py
majorgreys: We can follow a similar approach taken in other scenarios if we are trying to isolate regressions in a specific function call, as is the case here.

https://github.com/DataDog/dd-trace-py/blob/d32de4e02e8038a49c9c8aa9e33cc6335408eeb4/benchmarks/http_propagation_inject/scenario.py#L24-L30

benchmarks/set_http_meta/scenario.py
majorgreys: In general, I agree with @brettlangdon though would avoid adding more variants than we expect to yield meaningful differences.

benchmarks/set_http_meta/scenario.py
avara1986: yes, because set_http_meta calls to _context.set_item and this functions is:

```python
def set_items(kvs, span=None):
    # type: (Dict[str, Any], Optional[Span]) -> None
    """Set multiple items in the context of a trace."""
    ctx = span if span is not None else _DD_CONTEXTVAR.get()  # type: Optional[Union[Context, Span]]
    if not isinstance(ctx, Span) or ctx._local_root is None:
        raise ValueError("No context found")
    for k, v in kvs.items():
        ctx._local_root._set_ctx_item(k, v)
```

brettlangdon: Ah! I completely missed that, thanks for sharing, having a root is smart

benchmarks/set_http_meta/config.yaml
```diff
@@ -0,0 +1,24 @@
+start: &defaults
```
brettlangdon, line 1: this and `no-useragentvariant` seem identical, we can probably just rename this one and remove the duplicate?

no strong opinion/no change is actually necessary, just thinking about how we should be comparing the results between these different configs.

brettlangdon (commented):
looking good! would love to see updated numbers for this!

benchmarks/set_http_meta/config.yaml
avara1986: ok! I thought the "start" test works as warm up process, I remove it :+1:

avara1986: I thought that because some results look like:
```
sethttpmeta-start  |  31.5 us   26.8 | us: 1.17x faster   
sethttpmeta-no-useragentvariant | 26.6 us  | 27.1 us: 1.02x slower
```

brettlangdon: oh.... that might just be the variability of running it locally.

pyperf should handle the warm up per-scenario.

benchmarks/set_http_meta/scenario.py
```diff
@@ -0,0 +1,130 @@
...
+                "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 "
+                "(KHTML, like Gecko) Chrome/92.0.4515.159 Safari/537.36"
+            )
+
+        span = utils.gen_span(str("test"))
+        span._local_root = utils.gen_span(str("root"))
+
+        def bm(loops):
+            for _ in range(loops):
+                for _ in range(10000):
```
majorgreys, line 127: Can you explain why you do this 10k times within the inner loop? This might be better pulled out as a parameter for the scenario.

More generally, do we expect the timing to be the same the first time this function is called as it is for the last time it is called?

majorgreys: Ah was it a matter of having nanosecond timings that had random variations? I wonder if this has to do with how it can be hard to get good results when running locally during development.

juanjux (approved):

juanjux (approved):

gnufede (approved):

PR #4051, opened 2022-08-05
chore(debugging): clean up snapshot duration field

P403n1x87:
## Description

The semantics of the duration field was ambiguous and not aligned with other runtimes. This change cleans up the data model and ensures that the definition conforms to the new standard, which is that of using the duration field only as a subfield of the snapshot structure, bearing the value of the duration of the wrapped function.

## Checklist

## Reviewer Checklist
- [ ] Title is accurate.
- [ ] Add to milestone.


ddtrace/debugging/_async.py
```diff
@@ -3,19 +3,23 @@
...
 from ddtrace.debugging._snapshot.collector import SnapshotContext
+from ddtrace.internal import compat
 
 
 async def dd_coroutine_wrapper(coro, contexts):
     # type: (CoroutineType, List[SnapshotContext]) -> CoroutineType
+    start_time = compat.monotonic_ns()
     try:
         retval = await coro
+        end_time = compat.monotonic_ns()
```
ZStriker19, line 14: would it make sense to put this in a `finally` block? Or should we not because of timing?

tests/debugging/test_debugger.py
```diff
@@ -664,3 +665,25 @@ def count_status(queue):
 
         sleep(0.5)
         assert count_status(queue) == {"INSTALLED": 3, "RECEIVED": 2, "ERROR": 3}
+
+
+@pytest.mark.parametrize("duration", [1e5, 1e6, 1e7])
```
ZStriker19, line 670: Could you comment this to explain it a bit? I'm having trouble understanding how it's testing the duration.

ZStriker19 (approved):
Two questions but looks good!

ddtrace/debugging/_async.py
P403n1x87: I think this way we are more in control of when `compat.monotonic_ns` is called.

tests/debugging/test_debugger.py
P403n1x87: We are varying the duration of the wrapped function to ensure that the value that gets stored in the `duration` field is compatible with the actual function call duration. Here, `durationstuff` takes a duration as argument and will sleep for that amount of microseconds. We then check that we we snapshotted in the `duration` field is roughly the duration we passed to `durationstuff`.

Issue #4062, opened 2022-08-08
Wrong service name when using split_by_domain for requests


goodspark:
### Which version of dd-trace-py are you using?
1.2.1

### Which version of pip are you using?
20.3.4

### How can we reproduce your problem?
I assume this is caused by the `text@domain.com` at the _end_ of the URL.

### What is the result that you get?
![image](https://user-images.githubusercontent.com/29210237/183526731-05796a95-de6b-4765-9d2f-f2bfa130be4f.png)

### What is the result that you expected?
The service should be 'sendgrid.com'.

goodspark:
I'm guessing it's caused by this code: https://github.com/DataDog/dd-trace-py/blob/fb8dfa2f33fff37d21df9728d8386c0260df9744/ddtrace/contrib/requests/connection.py#L29-L36

Should probably use something like urllib.parse or only strip out user creds from the host after first narrowing down the possible hostname between `://` and the first `/` or `?` characters.

goodspark:
Yup that seems to be it.
```
>>> _extract_hostname('https://asd.com/blah/hello@zxc.com')
'zxc.com'
```

mabdinur:
Hi @goodspark, 

Thanks for calling out this issue. I opened a pull request to address this bug.

PR #4067, opened 2022-08-09
fix(requests): set correct service name when using split_by_domain

mabdinur:
## Description

When ``ddrace.config.requests['split_by_domain'] == True`` the service name of the `request.request` span should always be the hostname. 

Currently ’extract_hostname’ fails to correctly return the hostname in a url when an ampersats is present in the query string. There might be other unknown issues as well. Using `urlparse.parse()` is preferred over implementing our own url parser.

## Checklist

## Relevant issue(s)

Fixes: https://github.com/DataDog/dd-trace-py/issues/4062

## Testing strategy

Ensure `_extract_hostname` set the correct host name when an email is present in the url.

## Reviewer Checklist
- [ ] Title is accurate.
- [ ] Add to milestone.


brettlangdon:
@mabdinur can you add a better description to the PR?

it also seems like you included the templates for both a fix and a feature.

releasenotes/notes/fix-split-by-domain-requests-978e1f51a5b3f77d.yaml
```diff
@@ -0,0 +1,4 @@
+---
+fixes:
+  - |
+    requests: fixes split by domain
```
brettlangdon, line 4: is there a better description we can use here? what was fixed?

tests/contrib/requests/test_requests.py
```diff
@@ -302,6 +302,23 @@ def test_user_service_name_precedence(self):
         assert s.name == "requests.request"
         assert s.service == "clients"
 
+    def test_split_by_domain_with_email(self):
```
brettlangdon, line 305: does the email matter? or is it just the presence of a query string?

mabdinur: It's the presence of `@` in the url. The previous parsing logic did not account for this case. With this change we will use `urllib.parse.urlparse(url)` to get the hostname so this should not be an issue

releasenotes/notes/fix-split-by-domain-requests-978e1f51a5b3f77d.yaml
mabdinur: I updated the release note

tests/contrib/requests/test_requests.py
brettlangdon: can you include that context here and maybe in the PR description?

releasenotes/notes/fix-split-by-domain-requests-978e1f51a5b3f77d.yaml
```diff
@@ -0,0 +1,4 @@
+---
+fixes:
+  - |
+    requests: ensures the correct service name is set when ``ddrace.config.requests['split_by_domain'] == True``
```
brettlangdon, line 4:
```suggestion
    requests: fix split-by-domain service name when multiple ``@`` signs are present in the url
```

wdyt?

brettlangdon (approved):

Kyle-Verhoog (dismissed):

mabdinur:
@Mergifyio backport 1.4 1.3 0.61

mabdinur:
https://github.com/Mergifyio backport 1.4 1.3

ddtrace/contrib/requests/connection.py
```diff
@@ -17,23 +18,10 @@
...
-    i = uri.find("@", start, end) + 1
-    if i != 0:
-        start = i
-    j = uri.find("/", start, end)
-    if j != -1:
-        end = j
-
-    return uri[start:end]
+    parsed_uri = parse.urlparse(uri)
+    if parsed_uri.port:
```
mabdinur, line 22: The requests framework test [fails](https://github.com/DataDog/dd-trace-py/runs/7755898156?check_suite_focus=true) on this line. If parsed_uri.port is an invalid value then a [value error is raised](https://github.com/python/cpython/blob/main/Lib/urllib/parse.py#L175). We should wrap this code block in a try/except.

ddtrace/contrib/requests/connection.py
```diff
@@ -17,23 +18,10 @@
...
+        return "%s:%s" % (parsed_uri.hostname, str(parsed_uri.port))
```
mabdinur, line 23:
```suggestion
    port = None
    try:
        port = parsed_uri.port
    except ValueError:
        pass
    if port:
        return "%s:%s" % (parsed_uri.hostname, str(port))
```

ddtrace/contrib/requests/connection.py
```diff
@@ -17,23 +18,16 @@
...
-    if j != -1:
-        end = j
-
-    return uri[start:end]
+    parsed_uri = parse.urlparse(uri)
+    port = None
+    try:
+        port = parsed_uri.port
+    except ValueError:
+        pass
```
mabdinur, line 26:
```suggestion
        return "invalid_url"
```

majorgreys: I would suggest we use `?` in something like the service as it will be in the UI where a longer string is less desirable. So in the case that a requests get is to a url with an invalid port, we'd have a request span marked as an error and the service, if split by domain is enabled, would be `hostname:?` rather than `hostname:8888` otherwise.

ZStriker19 (approved):

brettlangdon (approved):

PR #4070, opened 2022-08-10
chore(gevent): run after-in-child hooks after reinit

P403n1x87:
## Description

Threads created too early in an application that uses gevent end up not running when the gevent hub reinit is executed after fork in the child process. This change ensures that we trigger the after-in-child fork hooks after the call to `gevent.hub.reinit` to ensure that threads created at any time will run as expected after fork.

### More details

In its implementation, `gevent` wraps around `os.fork` and calls `gevent.hub.reinit` to re-initialise the state of the child process after fork. The `forksafe` hooks that are registered by the library end up running after the call to the OS `fork` syscall, but *before* `gevent.hub.reinit`, which causes the threads to not work as expected in the child process. Other frameworks, like `gunicorn` make their own call to `os.fork` and then call `gevent.hub.reinit` in the child process to obtain the same effect. This poses the same problem as in plain `gevent`. With this change, we register an import hook on `gevent.hub` to detect the possibility that `gevent.hub.reinit` might be called, and we patch this function to re-run the after-in-child fork hook to ensure that they run *after* the gevent hub has been reinitialised fully in the child process after fork. In particular, this ensures that threads are recreated at the right time and can then work as expected in child processes.

## Checklist

## Reviewer Checklist
- [ ] Title is accurate.
- [ ] Add to milestone.


P403n1x87:
> are we able to reproduce this behavior in a test?

I think yes, but not easily. I'll give it a try.

P403n1x87:
Based on some of the CI failures, this is likely to require #4049.

ddtrace/internal/module.py
```diff
@@ -308,11 +303,7 @@ def find_module(self, fullname, path=None):
         self._finding.add(fullname)
 
         try:
-            if PY2:
-                __import__(fullname)
-                return _ImportHookLoader(self.after_import)
-
-            loader = getattr(find_spec(fullname), "loader", None)
+            loader = find_loader(fullname)
```
P403n1x87, line 306: This change doesn't seem to fix the `pytest` issues on Python 2 😢

P403n1x87:
It seems that `ModuleWatchdog` conflicts with the `testdir` fixture on Python 2.

tests/tracer/test_forksafe.py
```diff
@@ -285,3 +286,59 @@ def fn():
     pid, status = os.waitpid(child, 0)
     exit_code = os.WEXITSTATUS(status)
     assert exit_code == 42
+
+
+@pytest.mark.subprocess(
+    out="C\nT\nC\nT\nC\nT\n" if sys.platform == "darwin" else "C\nC\nC\nT\nT\nT\n",
```
P403n1x87, line 292: 🤷

P403n1x87:
@brettlangdon I've added a test case.

ddtrace/internal/module.py
```diff
@@ -122,26 +122,24 @@ def _resolve(path):
 # https://github.com/GrahamDumpleton/wrapt/blob/df0e62c2740143cceb6cafea4c306dae1c559ef8/src/wrapt/importer.py
 
 if PY2:
+    import pkgutil
```
P403n1x87, line 125: Probably worth making these changes in a dedicated PR

pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools >= 40.6.0", "setuptools_scm[toml] >=4,<6.1", "cython", "cmake >= 3.14", "ninja"]
+requires = ["setuptools >= 40.6.0,<64", "setuptools_scm[toml] >=4,<6.1", "cython", "cmake >= 3.14", "ninja"]
```
P403n1x87, line 2: This delta will go away after a rebase/merge

ddtrace/internal/module.py
P403n1x87: Done in #4083

P403n1x87:
#4083 is a pre-requisite for this PR.

ddtrace/__init__.py
```diff
@@ -1,4 +1,9 @@
-from ._logger import configure_ddtrace_logger
+from ddtrace.internal.module import ModuleWatchdog
+
+
+ModuleWatchdog.install()
```
brettlangdon, line 4: is this related?

this is needed now because we need it in order to set a callback for gevent loading?

should we enable `ModuleWatchdog` by default in a separate PR? seems "unrelated", or something that can be pulled out for clarity. wdyt?

tests/internal/test_module.py
```diff
@@ -9,6 +9,21 @@
...
+    # early stage. This fixture ensures that the watchdog is not installed
+    # before the tests start.
+    was_installed = ModuleWatchdog.is_installed()
+    if was_installed:
+        ModuleWatchdog.uninstall()
+
+    yield
+
+    if was_installed:
+        ModuleWatchdog.install()
```
brettlangdon, line 24:
```suggestion
    try:
        yield
    finally:
        if was_installed:
            ModuleWatchdog.install()
```

tests/internal/test_module.py
```diff
@@ -65,7 +80,10 @@ def test_import_origin_hook_for_module_not_yet_imported():
     path = os.getenv("MODULE_ORIGIN")
     hook = mock.Mock()
 
-    ModuleWatchdog.install()
+    try:
+        ModuleWatchdog.install()
+    except RuntimeError:
+        pass
```
brettlangdon, line 86: when/why could this happen? why do we not care?

tests/profiling/collector/test_threading.py
```diff
@@ -72,7 +72,7 @@ def test_lock_acquire_events():
     # It's called through pytest so I'm sure it's gonna be that long, right?
     assert len(event.frames) > 3
     assert event.nframes > 3
-    assert event.frames[0] == (__file__, 65, "test_lock_acquire_events")
+    assert event.frames[0] == (__file__.replace(".pyc", ".py"), 65, "test_lock_acquire_events")
```
brettlangdon, line 75: related?

tests/profiling/collector/test_traceback.py
```diff
@@ -14,11 +14,9 @@ def test_check_traceback_to_frames():
         exc_type, exc_value, traceback = sys.exc_info()
     frames, nframes = _traceback.traceback_to_frames(traceback, 10)
     assert nframes == 2
+
+    this_file = __file__.replace(".pyc", ".py")
```
brettlangdon, line 18: why this change? I am not sure how it is related

tests/tracer/test_forksafe.py
```diff
@@ -285,3 +286,59 @@ def fn():
     pid, status = os.waitpid(child, 0)
     exit_code = os.WEXITSTATUS(status)
     assert exit_code == 42
+
+
+@pytest.mark.subprocess(
+    out="" if (3,) < sys.version_info < (3, 7) else ("CTCTCT" if sys.platform == "darwin" else "CCCTTT"),
```
brettlangdon, line 292: is this consistent? I'd guess this was flaky.

do we just care that stdout contains 3 "C" and 3 "T" regardless of the order?

ddtrace/__init__.py
P403n1x87: > this is needed now because we need it in order to set a callback for gevent loading?
Correct!
> should we enable `ModuleWatchdog` by default in a separate PR? seems "unrelated", or something that can be pulled out for clarity. wdyt?
Yes, we could do this.

tests/internal/test_module.py
P403n1x87: If `ModuleWatchdog` is already installed we get a `RuntimeError`. This makes the call to `install` idempotent

tests/profiling/collector/test_threading.py
P403n1x87: Probably related to now having a `ModuleWatchdog` installed.

tests/profiling/collector/test_traceback.py
P403n1x87: Python 2.7 started failing on these tests because `__file__` had a `.pyc` extension. Not sure why this is happening, but I suspect it's the import mechanism from the `ModuleWatchdog`

tests/profiling/collector/test_threading.py
brettlangdon: this might be a good argument for moving enabling `ModuleWatchdog.install()` by default it's own PR, since these things seem un-related to the gevent change.

tests/internal/test_module.py
brettlangdon: 🤔 since we install it by default should it always be available, do we need to change these tests to assume it is already available? or to use it's own instance of `ModuleWatchdog` instead of the globally installed one?

tests/tracer/test_forksafe.py
brettlangdon: wonder if it is just how darwin handles the scheduling

P403n1x87: It looks like it, but it "shouldn't be" 😕

tests/internal/test_module.py
P403n1x87: > or to use it's own instance of ModuleWatchdog instead of the globally installed one?

The fixture should be removing the global one if it's installed and then reinstalling it at the end.

brettlangdon: ok, so if that is the case, then this shouldn't fail ever right?

P403n1x87: Ah, my bad! Here we are in a subprocess, where the fixture does not apply. That's why we have this pattern only in subprocess tests.

jd:
@P403n1x87 do you have an idea of the side effect this has so far in the real world?
I'm curious as we're detecting this issue a bit "late".

tests/internal/test_module.py
brettlangdon: oh, I missed that these were in subprocesses as well

so that should mean these are running in isolation then and can probably just assume that `ModuleWatchdog` is already installed

P403n1x87:
> @P403n1x87 do you have an idea of the side effect this has so far in the real world? I'm curious as we're detecting this issue a bit "late".

The side-effect should be that we now properly restart threads in e.g. gunicorn worker processes. I have done some manual testing with a Flask application and a modified tracer that starts the writer thread immediately. I was able to see the writer thread running correctly in each process with this fix, whereas without it the thread would be running only in the master process.

tests/internal/test_module.py
P403n1x87: done in #4098

jd (approved):

brettlangdon (approved):

PR #4098, opened 2022-08-16
chore: enable ModuleWatchdog globally

P403n1x87:
## Description

This change installs an instance of the `ModuleWatchdog` for registering module import hooks.

## Checklist

## Reviewer Checklist
- [ ] Title is accurate.
- [ ] Add to milestone.


tests/internal/test_module.py
```diff
@@ -9,6 +9,21 @@
...
+    # early stage. This fixture ensures that the watchdog is not installed
+    # before the tests start.
+    was_installed = ModuleWatchdog.is_installed()
+    if was_installed:
+        ModuleWatchdog.uninstall()
+
+    yield
+
+    if was_installed:
+        ModuleWatchdog.install()
```
brettlangdon, line 24:
```suggestion
    try:
        yield
    finally:
        if was_installed:
            ModuleWatchdog.install()
```

tests/internal/test_module.py
```diff
@@ -65,7 +80,10 @@ def test_import_origin_hook_for_module_not_yet_imported():
     path = os.getenv("MODULE_ORIGIN")
     hook = mock.Mock()
 
-    ModuleWatchdog.install()
+    try:
+        ModuleWatchdog.install()
+    except RuntimeError:
+        pass
```
brettlangdon, line 87: I'd like to see us avoid this if we can. I wonder if this test should just assume that ModuleWatchdog is installed and therefore we can just register a hook and everything works as expected (test the expected default behavior)

P403n1x87: I'd like to avoid this verbosity too, but `test_module` should have just unit tests for the stuff in `ddtrace.internal.module`. Assuming that `ModuleWatchdog` is already installed is more like an integration test IMO. If for some reasons we decide to revert installing `ModuleWatchdog` globally, we would have to change these tests back.

A slightly less verbose alternative would be to define a subclass in each of the subprocess tests, like so
~~~
class TestModuleWatchdog(ModuleWatchdog):
    pass
~~~
but perhaps it's still worth testing the base class.

Yet another option would be to either change the behaviour of `install` to be idempotent, or pass a `raise=False` optional argument.

jd (approved):

mabdinur (approved):

PR #4147, opened 2022-09-01
refactor(profiling): remove lock from _ThreadLink

jd:
The _ThreadLink data structure historically used a lock to make sure there was
no threads overlapping each other. For example, this could happen:

1. Main thread from the application starts a trace and `link_object` is being
   called
2. Stack profiler would run and clear the thread list or get an spans from the
   list

Removing the lock makes it possible that, while the stack profiler thread
clears the mapping, a user's thread links a thread to a span.

That's a trade-off we are willing to accept for performance and safeness
reason.

### DIFF
--- a/ddtrace/profiling/_threading.pyx
+++ b/ddtrace/profiling/_threading.pyx
@@ -89,7 +89,7 @@ class _ThreadLink(_thread_link_base):
         """
         # This code clears the thread/object mapping by clearing a copy and swapping it in an atomic operation This is
         # needed to be able to have this whole class lock-free and avoid concurrency issues.
-        # The fact that is is lock free means we might lose some accuracy, but it's worth the trade-off for speed and simplicity.
+        # The fact that it is lock free means we might lose some accuracy, but it's worth the trade-off for speed and simplicity.
         new_thread_id_to_object_mapping = self._thread_id_to_object.copy()
         # Iterate over a copy of the list of keys since it's mutated during our iteration.
         for thread_id in list(new_thread_id_to_object_mapping):
--- a/ddtrace/profiling/_threading.pyx
+++ b/ddtrace/profiling/_threading.pyx
@@ -69,9 +69,6 @@ class _ThreadLink(_thread_link_base):
     # Key is a thread_id
     # Value is a weakref to an object
     _thread_id_to_object = attr.ib(factory=dict, repr=False, init=False, type=typing.Dict[int, _weakref_type])
-    # Note that this lock has to be reentrant as spans can be activated unexpectedly in the same thread
-    # ex. during a gc weakref finalize callback
-    _lock = attr.ib(factory=nogevent.RLock, repr=False, init=False, type=nogevent.RLock)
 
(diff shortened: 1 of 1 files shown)


ddtrace/profiling/_threading.pyx
```diff
@@ -92,11 +87,17 @@ class _ThreadLink(_thread_link_base):
...
-            for thread_id in list(self._thread_id_to_object.keys()):
-                if thread_id not in existing_thread_ids:
-                    del self._thread_id_to_object[thread_id]
+        # This code clears the thread/object mapping by clearing a copy and swapping it in an atomic operation This is
+        # needed to be able to have this whole class lock-free and avoid concurrency issues.
+        # The fact that is is lock free means we might lose some accuracy, but it's worth the trade-off for speed and simplicity.
+        new_thread_id_to_object_mapping = self._thread_id_to_object.copy()
+        # Iterate over a copy of the list of keys since it's mutated during our iteration.
+        for thread_id in list(new_thread_id_to_object_mapping):
+            if thread_id not in existing_thread_ids:
```
brettlangdon, line 96: does `existing_thread_ids` always have something in them?

if we think it might be often we have cases where it is an empty set, we could always have a short-circuit of:

```python
if not existing_thread_ids:
    self._thread_id_to_object.clear()
```

but maybe not worth it.

brettlangdon (approved):
random thought, but otherwise lgtm

ddtrace/profiling/_threading.pyx
jd: That's never empty because it's the list of all running threads, and you have at least 2 threads: the main thread, and the stack profiler thread calling this function.

ddtrace/profiling/_threading.pyx
```diff
@@ -92,11 +87,17 @@ class _ThreadLink(_thread_link_base):
...
         :param existing_thread_ids: A set of thread ids to keep.
         """
-        with self._lock:
-            # Iterate over a copy of the list of keys since it's mutated during our iteration.
-            for thread_id in list(self._thread_id_to_object.keys()):
-                if thread_id not in existing_thread_ids:
-                    del self._thread_id_to_object[thread_id]
+        # This code clears the thread/object mapping by clearing a copy and swapping it in an atomic operation This is
+        # needed to be able to have this whole class lock-free and avoid concurrency issues.
+        # The fact that is is lock free means we might lose some accuracy, but it's worth the trade-off for speed and simplicity.
```
majorgreys, line 92:
```suggestion
        # The fact that it is lock free means we might lose some accuracy, but it's worth the trade-off for speed and simplicity.
```

majorgreys (approved):

ddtrace/profiling/_threading.pyx
brettlangdon: 👍🏻 makes sense.

majorgreys:
@Mergifyio backport 1.4 1.3 0.x 0.61

PR #4169, opened 2022-09-06
chore(debugging): use new snapshot document structure

P403n1x87:
## Checklist

## Reviewer Checklist
- [x] Title is accurate.
- [ ] Add to milestone.


ddtrace/debugging/_encoding.py
```diff
@@ -289,36 +287,98 @@ def _get_fields(obj):
...
+                        _captured_value_v2(v, level=level - 1, maxlen=maxlen, maxsize=maxsize, maxfields=maxfields),
+                    )
+                    for _, (k, v) in zip(range(maxsize), value.items())
+                ],
+                "size": len(value),
+            }
+
+        elif _type in BUILTIN_CONTAINER_TYPES:
+            # Sequence
+            if level < 0:
```
shatzi, line 331: remove duplication by taking those two if level < 0 outside if is dict

ddtrace/debugging/_encoding.py
```diff
@@ -289,36 +287,98 @@ def _get_fields(obj):
...
+
+        if len(value) > maxsize:
+            data["notCapturedReason"] = "collectionSize"
+
+        return data
+
+    # Arbitrary object
+    fields = _get_fields(value)
+    data = (
+        {
```
shatzi, line 354: maybe return depth error before fetching the fields?

ddtrace/debugging/_encoding.py
```diff
@@ -289,36 +287,98 @@ def _get_fields(obj):
...
-            {n: _captured_value_v2(v, level - 1) for n, v in _get_fields(value).items()} if level > 0 else None
+
+    if _type in BUILTIN_SIMPLE_TYPES:
+        value_repr = repr(value)
+        return (
+            {
+                "type": _qualname(_type),
+                "value": value_repr,
+            }
+            if len(value_repr) <= maxlen
```
shatzi, line 301: not sure this is what it was is expected for strings.
We need to clarify the RFC around this. there are two examples on with "notCapturedReason":"length" and another is truncated: true.

So, two comments:
1. in terms of usability - long strings should be truncated and not omitted ("notCapturedReason":"length").
2. if a value was truncated, I would also expect to have original length of the string - I recall we chatted about this but I forgot if we decided to not have it at the end.

/cc @jpbempel @OmerRaviv

ddtrace/debugging/_encoding.py
```diff
@@ -206,11 +206,9 @@ def _qualname(_type):
             except AttributeError:
                 return repr(_type)
         try:
```
shatzi, line 208: this logic now seems to be same as if module == "__builtin__" - maybe this can be simplified?

shatzi (commented):
looks awesome. few comments but only one is important - regarding how to we deal with long strings

ddtrace/debugging/_encoding.py
jpbempel: yes, long strings must be truncated and not omitted.
I have fixed one place in the RFC, but it shouldn't be any `"notCapturedReason": length` in the the snapshot, if truncated, it's `"truncated": true`

for the length/size, definitely I don't want to put it for all the strings. only if truncated

P403n1x87: Thanks for clarifying. Originally the string was also truncated, but I only noticed the example with the "notCapturedReason" in the RFC and missed the one with "truncated". I'l fix this!

ddtrace/debugging/_encoding.py
```diff
@@ -396,16 +485,10 @@ def logs_track_upload_request_v2(
...
-            ", ".join(("=".join((n, a["value"])) for n, a in arguments.items())),
-        )
+        message = format_message(top_frame["function"], arguments)
     elif isinstance(snapshot.probe, FunctionProbe):
         arguments = snapshot_data["captures"]["entry"]["arguments"]
-        message = "%s(%s)" % (
-            snapshot.probe.func_qname,
-            ", ".join(("=".join((n, a["value"])) for n, a in arguments.items())),
-        )
+        message = format_message(cast(str, snapshot.probe.func_qname), arguments)
```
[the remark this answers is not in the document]
P403n1x87: We could do. What are the other runtimes doing? It looks like Java at least is adding the return value to the message?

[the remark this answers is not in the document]
jpbempel: Yes we are adding the return value in Java, and for me makes sense as we may want to tsee pattern for the return values

jpbempel (dismissed):

ddtrace/debugging/_encoding.py
```diff
@@ -386,6 +445,40 @@ def add_tags(payload):
...
+
+def format_message(function, args, retval=None):
+    # type: (str, Dict[str, Any], Optional[Any]) -> str
+    message = "%s(%s)" % (
+        function,
+        ", ".join(("=".join((n, format_captured_value(a))) for n, a in args.items())),
+    )
+
+    if retval is not None:
+        return "\n".join((message, "=".join(("@return", format_captured_value(retval)))))
```
cimi, line 477: In the Java tracer we also print the locals in the summary message - we should do that here as well 🙂 

Since we print the return value separately, we should exclude `@return` from the serialised locals - https://github.com/DataDog/dd-trace-java/pull/4011#discussion_r994473993

ddtrace/debugging/_encoding.py
```diff
@@ -289,17 +290,82 @@ def _get_fields(obj):
...
+
+    if _type in BUILTIN_SIMPLE_TYPES:
+        value_repr = repr(value)
+        return (
+            {
+                "type": _qualname(_type),
+                "value": value_repr,
+            }
+            if len(value_repr) <= maxlen
+            else {
```
cimi, line 305: Nit: isn't it clearer to invert the condition and drop the `else`? I've never seen blank if statements before 🙂

ddtrace/debugging/_encoding.py
```diff
@@ -289,17 +290,82 @@ def _get_fields(obj):
...
         "type": _qualname(_type),
-        "value": _serialize(value, level),
-    }  # type: Dict[str, Any]
-    if _type not in BUILTIN_TYPES:
-        data["fields"] = (
-            {n: _captured_value_v2(v, level - 1) for n, v in _get_fields(value).items()} if level > 0 else None
-        )
+        "fields": {
+            n: _captured_value_v2(v, level=level - 1, maxlen=maxlen, maxsize=maxsize, maxfields=maxfields)
+            for _, (n, v) in zip(range(maxfields), fields.items())
```
cimi, line 363: Is the order of the fields deterministic?

cimi (dismissed):
:+1: Thanks!

Left some minor comments, not blocking - you can address them in another PR if you prefer.

ddtrace/debugging/_encoding.py
P403n1x87: No. Should it be? In general we retrieve them from looking into a dictionary, so any form of determinism would require computation 🙁

ddtrace/debugging/_encoding.py
P403n1x87: Sure, we can do that, but I suppose this only applies to line probes, in which case we don't have a `@return`? We don't have access to locals for function probes in Python.

ddtrace/debugging/_encoding.py
P403n1x87: Ah I think the formatting is to be blamed here. The `if`  is not empty, but evaluates the stuff that comes before, if the condition is met (`A if COND else B`)

ddtrace/debugging/_encoding.py
cimi: The order we put the fields in the output JSON does not matter but if we limit the number of fields we capture to `maxfields` it would be nice to have snapshots consistent with each other (i.e. get the same x fields all the time, rather than having different fields in different snapshots).

It's not a big deal, if you're concerned about the overhead of sorting the field names alphabetically before deciding which ones we keep we can leave this as is.

ddtrace/debugging/_encoding.py
cimi: Ah, ok - makes sense!

ddtrace/debugging/_encoding.py
cimi: Didn't think of that, but now I remember we're wrapping the function for method probes. In Java we have the locals so the comment about `@return` applies there.

IMO we should print the locals in the summary when we have them, but it's fine to do that change in another PR.

ddtrace/debugging/_encoding.py
```diff
@@ -289,17 +290,82 @@ def _get_fields(obj):
         return {s: _safe_getattr(obj, s) for s in get_slots(obj)}
 
 
-def _captured_value_v2(value, level=1):
-    # type: (Any, int) -> Dict[str, Any]
+def _captured_value_v2(value, level=MAXLEVEL, maxlen=MAXLEN, maxsize=MAXSIZE, maxfields=MAXFIELDS):
+    # type: (Any, int, int, int, int) -> Dict[str, Any]
     _type = type(value)
+
+    if _type in BUILTIN_SIMPLE_TYPES:
```
shatzi, line 297: Maybe we want to check for None as well and set "isNull":true as in spec?

shatzi (approved):
LGTM

ddtrace/debugging/_encoding.py
P403n1x87: Ah good catch! 👀 I'll make the change.

brettlangdon (dismissed):

shatzi (approved):

Kyle-Verhoog (dismissed):

Kyle-Verhoog:
@P403n1x87 looks like tests are failing: https://app.circleci.com/pipelines/github/DataDog/dd-trace-py/23144/workflows/53e4706d-bd93-47e7-ab00-f7a8383fa42c/jobs/1574040

shatzi (approved):

Kyle-Verhoog (approved):
just one nit, otherwise lgtm

PR #4185, opened 2022-09-12
ci(boto): remove test_kinesis_put_records_base64_exceeds_max_size

mabdinur:
## Description

With the latest [moto release](https://github.com/spulec/moto/issues/5249), adding a record which exceeds the max size constraint is not supported. Since this test validates unsupported behavior it was removed.

New [Boto Validation Error](https://app.circleci.com/pipelines/github/DataDog/dd-trace-py/21184/workflows/33375cfd-6f3f-446e-aae7-2e800994e9df/jobs/1440546):
```
def test_kinesis_put_records_base64_exceeds_max_size(self):
        client = self.session.create_client("kinesis", region_name="us-east-1")
    
        stream_name = "test"
        client.create_stream(StreamName=stream_name, ShardCount=1)
        stream = client.describe_stream(StreamName=stream_name)["StreamDescription"]
        shard_id = stream["Shards"][0]["ShardId"]

        partition_key = "1234"
        sample_string = json.dumps({"Hello": "x" * (1 << 20)})
        sample_string_bytes = sample_string.encode("ascii")
        base64_bytes = base64.b64encode(sample_string_bytes)
        data_str = base64_bytes.decode("ascii")
        data = [
            {"Data": data_str, "PartitionKey": partition_key},
            {"Data": data_str, "PartitionKey": partition_key},
        ]
    
        Pin(service=self.TEST_SERVICE, tracer=self.tracer).onto(client)
>       client.put_records(StreamName=stream_name, Records=data)

tests/contrib/botocore/test.py:1676: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
        if http.status_code >= 300:
            error_code = parsed_response.get("Error", {}).get("Code")
            error_class = self.exceptions.from_code(error_code)
>           raise error_class(parsed_response, operation_name)
E           botocore.errorfactory.ValidationException: An error occurred (ValidationException) when calling the PutRecords operation: 1 validation error detected: Value at 'records.1.member.data' failed to satisfy constraint: Member must have length less than or equal to 1048576

.riot/venv_py3615_moto[all]/lib/python3.6/site-packages/botocore/client.py:911: ValidationException
```

## Checklist
- [ ] Title must conform to [conventional commit](https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional).
[...]

PR #4188, opened 2022-09-13
fix(ci): pin moto<4.0.3 for botocore tests

majorgreys:
The botocore tests make a call to an API that as of v4.0.3 of moto raises an exception.

https://github.com/DataDog/dd-trace-py/blob/fde91fd2571f6d9de1a3b32ab136281041a3bc69/tests/contrib/botocore/test.py#L1676

See https://github.com/spulec/moto/issues/5249 for more details on the change in moto.

### DIFF
--- a/riotfile.py
+++ b/riotfile.py
@@ -1118,7 +1118,9 @@ venv = Venv(
             command="pytest {cmdargs} tests/contrib/botocore",
             pkgs={"botocore": latest},
             venvs=[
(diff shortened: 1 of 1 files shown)


majorgreys:
The failing test has been removed in https://github.com/DataDog/dd-trace-py/pull/4185. We should reintroduce the failing test and make the necessary changes to have it pass with `moto>=4.0.3`.

PR #4199, opened 2022-09-16
chore(debugging): support run module

P403n1x87:
## Description

When a Python script is started as a module, it doesn't go through the import machinery and it might then go under the radar for the debugger. With this change, we capture the run module event and translate it into an import event so that the debugger can be aware of the run module.

## Checklist

## Reviewer Checklist
- [ ] Title is accurate.
- [ ] Add to milestone.


tests/debugging/run_module/target.py
```diff
@@ -0,0 +1,25 @@
...
+    sleep(0.5)
+
+    foo()
+
+    (snapshot,) = TestDebugger._instance.test_queue
+
+    assert snapshot.probe.probe_id == "run_module_test"
+    assert snapshot.frame.f_locals["concealed"] == "secret"
+
+    print("OK")
```
ZStriker19, line 22: why do we print here?

tests/debugging/run_module/target.py
```diff
@@ -0,0 +1,25 @@
...
+from tests.debugging.mocking import TestDebugger
+
+
+def foo():
+    concealed = "secret"  # noqa
+    return
+
+
+def test():
+    sleep(0.5)
```
ZStriker19, line 13: why do we need to sleep here? To wait for `foo` to be registered?

tests/debugging/run_module/target.py
ZStriker19: Oh I see, it's because we want to see that calling the program passed.

ZStriker19 (approved):

tests/debugging/run_module/target.py
P403n1x87: We give time to the module to finish loading, so that we are sure that the debugger watchdog triggered the instrumentation hooks and the target function is instrumented. When we call it, we then test that it generates a snapshot, as expected. Without the sleep, the thread would be racing with the module loading process.

tests/debugging/test_debugger.py
```diff
@@ -699,3 +701,18 @@ def test_debugger_function_probe_duration(duration):
 
         (snapshot,) = d.test_queue
         assert 0.9 * duration <= snapshot.duration <= 2.5 * duration, snapshot
+
+
+def test_debugger_run_module():
```
majorgreys, line 706: Do we want to also check the debugger snapshot?

P403n1x87: The emitted snapshot is checked in the `target.py` source, where essentially the testing logic is, if this is what you meant?

ZStriker19 (dismissed):

majorgreys (commented):
Should we add a release note and change this to a feature?

ZStriker19 (dismissed):

P403n1x87:
> Should we add a release note and change this to a feature?

In theory yes, but considering that the whole DI has no official users, we might spare the "noise" in the change log. WDYT?

emmettbutler (approved):

majorgreys (approved):

P403n1x87:
The new test no longer passes 😢 . Investigating

majorgreys (approved):

emmettbutler (approved):

PR #4255, opened 2022-10-03
ci: run pre checks and coverage with Python 3.10

jd:


riotfile.py
```diff
@@ -129,7 +129,7 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
             ],
         ),
         Venv(
-            pys=["3.9"],
+            pys=["3.10"],
```
mabdinur, line 132: We'll probably need to bump this to py310 as well: https://github.com/DataDog/dd-trace-py/blob/1.x/.circleci/config.yml#L294

jd: Good catch, thank you!

ddtrace/debugging/_function/discovery.py
```diff
@@ -9,7 +9,7 @@
 try:
     from collections.abc import Iterator
 except ImportError:
-    from collections import Iterator
+    from collections import Iterator  # type: ignore[attr-defined,no-redef]
```
P403n1x87, line 12: 🤔 I wonder why `attr-defined` is needed here. Perhaps something to do with the import above?

P403n1x87 (approved):

ddtrace/debugging/_function/discovery.py
jd: Yes, because in Python 3.10 there is no `collections.Iterator` anymore so mypy is saying that this is not going to work 😉

majorgreys (approved):

PR #4318, opened 2022-10-14
feat(profiler): extract class names for memory profiles

jd:
This adds some C code in the memalloc profiler module to extract class names
from the stack traces. This work as the stack profiler, by introspecting the
`self` and `cls` variables when they are present in the local frame.

Depends-On: https://github.com/DataDog/dd-trace-py/pull/4304


jd:
The Unicode memory is allocated using PyObject in Python < 3.10, so this will crash until we merge https://github.com/DataDog/dd-trace-py/pull/4304 in earlier Python version. Python 3.10 uses `PyMem` domain which is not (yet) profiled.

ddtrace/profiling/collector/_memalloc_tb.c
```diff
@@ -189,9 +228,19 @@ traceback_to_tuple(traceback_t* tb)
...
-        /* Class name */
-        PyTuple_SET_ITEM(frame_tuple, 3, empty_string);
-        Py_INCREF(empty_string);
+
+        PyObject* class_name;
+
+        if (frame->type) {
+            PyTypeObject* type = (PyTypeObject*)frame->type;
+            /* Improvement for Python 3.11+? */
+            // class_name = _PyType_GetQualName(type);
```
P403n1x87, line 237: 👍  code objects have a `co_qualname` in Python 3.11.

jd: We don't support it yet though so…

P403n1x87 (approved):

ddtrace/profiling/collector/_memalloc_tb.c
```diff
@@ -189,9 +228,19 @@ traceback_to_tuple(traceback_t* tb)
...
+            class_name = PyUnicode_FromString(type->tp_name);
+        } else {
+            class_name = empty_string;
```
P403n1x87, line 240: nit: what's the issue with setting this to `Py_None`?

jd: This is mostly to avoid having to handle the `None` vs `str` in the pprof encoding. Having everything as a string right away makes it easier to serialize everything in the end.

PR #4319, opened 2022-10-14
fix(telemetry): register after fork hooks on initialization

mabdinur:
## Description

app-started, app-closed, and app-heartbeat telemetry events should not be sent in a child processes. To avoid this `forksafe.register(TelemetryWriter._fork_writer)` needs to be called as early as possible. 

Currently `TelemetryWriter._fork_writer` is registered in `TelemetryWriter.enabled()` and this method is called right before the first trace is sent to the agent. Since the global [telemetry_writer](https://github.com/DataDog/dd-trace-py/blob/1.x/ddtrace/internal/telemetry/__init__.py#L12) is initialized on `import ddtrace`, registering `TelemetryWriter._fork_writer` here will reduce the likelihood that telemetry events get sent in child processes.

Note - This does not prevent the ddtrace library from sending telemetry in all child processes. If ddtrace is imported after the main process forks this issue can still occur.

## Checklist

## Reviewer Checklist
- [x] Title is accurate.
- [ ] Add to milestone.


mehulsonowal (dismissed):

ddtrace/internal/telemetry/writer.py
```diff
@@ -66,6 +67,9 @@ def __init__(self, agent_url=None):
         # _sequence is a counter representing the number of requests sent by the writer
         self._sequence = 1  # type: int
 
+    def __del__(self):
```
mabdinur, line 70: Is this needed?

jd: Yes and no: this method is never going to be called because the forksafe hook is going to hold a reference to this object forever.

mabdinur: Since a reference to this object will live forever we shouldn't need to deregister the this hook

mabdinur:
@Mergifyio backport 1.6

Yun-Kim (approved):

PR #4321, opened 2022-10-14
chore(dbm): generate sql comments for DBM

mabdinur:
## Description

Adds utils for generating DBM comments

## Checklist

## Reviewer Checklist
- [ ] Title is accurate.
- [ ] Add to milestone.

### DIFF
--- a/ddtrace/context.py
+++ b/ddtrace/context.py
@@ -95,6 +95,15 @@ class Context(object):
             self._metrics[SAMPLING_PRIORITY_KEY] = value
 
     @property
+    def traceparent(self):
+        # type: () -> str
+        if self.trace_id is None or self.span_id is None:
+            return ""
+
+        sampled = 1 if self.sampling_priority and self.sampling_priority > 0 else 0
+        return "00-{:032x}-{:016x}-{:02x}".format(self.trace_id, self.span_id, sampled)
+
+    @property
     def dd_origin(self):
         # type: () -> Optional[Text]
         """Get the origin of the trace."""
--- a/ddtrace/context.py
+++ b/ddtrace/context.py
@@ -95,7 +95,7 @@ class Context(object):
             self._metrics[SAMPLING_PRIORITY_KEY] = value
 
     @property
-    def traceparent(self):
+    def _traceparent(self):
         # type: () -> str
         if self.trace_id is None or self.span_id is None:
             return ""
--- a/ddtrace/settings/database_monitoring.py
+++ b/ddtrace/settings/database_monitoring.py
@@ -1,5 +1,21 @@
+from typing import TYPE_CHECKING
+
 from envier import En
 from envier import validators
+
+from ddtrace.vendor.sqlcommenter import generate_sql_comment as _generate_sql_comment
+
+from . import _config as dd_config
+
+
+if TYPE_CHECKING:
+    from ddtrace import Span
+
+DBM_PARENT_SERVICE_NAME_KEY = "ddps"
+DBM_DATABASE_SERVICE_NAME_KEY = "dddbs"
+DBM_ENVIRONMENT_KEY = "dde"
+DBM_VERSION_KEY = "ddpv"
+DBM_TRACE_PARENT_KEY = "traceparent"
 
 
 class DatabaseMonitoringConfig(En):
@@ -14,4 +30,23 @@ class DatabaseMonitoringConfig(En):
     )
 
 
-config = DatabaseMonitoringConfig()
+dbm_config = DatabaseMonitoringConfig()
+
+
+def get_dbm_comment(db_span):
+    # type: (Span) -> str
+    if dbm_config.injection_mode == "disabled":
+        return ""
+
+    dbm_tags = {
+        DBM_PARENT_SERVICE_NAME_KEY: dd_config.service,
+        DBM_ENVIRONMENT_KEY: dd_config.env,
+        DBM_VERSION_KEY: dd_config.version,
+        DBM_DATABASE_SERVICE_NAME_KEY: db_span.service,
+    }
+
+    if dbm_config.injection_mode == "full":
+        # TODO: add _dd.dbm_trace_injected tag to db_span
+        dbm_tags[DBM_TRACE_PARENT_KEY] = db_span.context.traceparent
+
+    return _generate_sql_comment(**dbm_tags)
--- a/ddtrace/settings/database_monitoring.py
+++ b/ddtrace/settings/database_monitoring.py
@@ -47,6 +47,6 @@ def _get_dbm_comment(db_span):
 
     if dbm_config.injection_mode == "full":
         # TODO: add _dd.dbm_trace_injected tag to db_span
-        dbm_tags[DBM_TRACE_PARENT_KEY] = db_span.context.traceparent
+        dbm_tags[DBM_TRACE_PARENT_KEY] = db_span.context._traceparent
 
     return _generate_sql_comment(**dbm_tags)
--- a/ddtrace/settings/database_monitoring.py
+++ b/ddtrace/settings/database_monitoring.py
@@ -33,7 +33,7 @@ class DatabaseMonitoringConfig(En):
 dbm_config = DatabaseMonitoringConfig()
 
 
-def get_dbm_comment(db_span):
+def _get_dbm_comment(db_span):
     # type: (Span) -> str
     if dbm_config.injection_mode == "disabled":
         return ""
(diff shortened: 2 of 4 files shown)


ddtrace/settings/database_monitoring.py
```diff
@@ -14,4 +31,23 @@ class DatabaseMonitoringConfig(En):
     )
 
 
-config = DatabaseMonitoringConfig()
+dbm_config = DatabaseMonitoringConfig()
+
+
+def get_dbm_comment(db_span):
```
mabdinur, line 37: Not sure if this function should live here or be move to a util module. Leaving here simplifies imports and is easier to read.

ddtrace/span.py
```diff
@@ -493,6 +493,14 @@ def context(self):
             self._context = Context(trace_id=self.trace_id, span_id=self.span_id)
         return self._context
 
+    @property
+    def traceparent(self):
```
mabdinur, line 497: The traceparent property should be consistent with the following spec:  https://www.w3.org/TR/trace-context/#traceparent-header

ddtrace/settings/database_monitoring.py
ZStriker19: Leaving it here for now seems fine to me.

ddtrace/span.py
ZStriker19: From the spec: _This tag is a formatted composite value that includes the trace id, span id and a sampled flag. This is compliant with the w3c trace context spec. One example would be:_ `00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01`. 

So I think we'll need to convert the span_id and trace_id into hexadecimal values here.

tests/internal/test_database_monitoring.py
```diff
@@ -27,3 +27,64 @@ def test_injection_mode_configuration():
...
+    from ddtrace.settings import database_monitoring
+
+    dbspan = tracer.trace("dbname", service="orders-db")
+
+    sqlcomment = database_monitoring.get_dbm_comment(dbspan)
+    # tags in sqlcomments are comma delimited
+    # assert 4 tags are present in the sqlcomment
+    assert len(sqlcomment.split(",")) == 4
+    # assert tags sqlcomment contains the correct value
+    assert "dddbs='orders-db'" in sqlcomment
```
ZStriker19, line 61: Nit: Would we want to import and use the keys we specified in database_monitoring.py instead of dddbs, dde, ddps, and ddpv?

ddtrace/span.py
ZStriker19: E.g: https://github.com/DataDog/dd-trace-py/blob/11dddbd05152c9493d7da7b9fd5316d38ea47696/ddtrace/propagation/http.py#L546

ZStriker19 (dismissed):
LGTM!

ddtrace/context.py
```diff
@@ -94,6 +94,15 @@ def sampling_priority(self, value):
                 return
             self._metrics[SAMPLING_PRIORITY_KEY] = value
 
+    @property
+    def traceparent(self):
```
majorgreys, line 98: Unless we intend to make this public, we can add it to the internal interface.
```suggestion
    def _traceparent(self):
```

ddtrace/context.py
```diff
@@ -94,6 +94,15 @@ def sampling_priority(self, value):
...
+        # type: () -> str
+        if self.trace_id is None or self.span_id is None:
+            return ""
+
+        sampled = 1 if self.sampling_priority and self.sampling_priority > 0 else 0
+        return "00-{:032x}-{:016x}-{:02x}".format(self.trace_id, self.span_id, sampled)
```
majorgreys, line 104: Do we expect this property to be accessed multiple times for the same span? trace id? Are there any performance concerns with having an internal representation that needs to be converted during the lifecycle of a span?

mabdinur: The following benchmark converts traceid, spanid, and sample rate to hex then generates a traceparent. This ran 1 million operations in 0.433 seconds (0.43 microseconds per string): 
```
import timeit
timeit.timeit(setup='from ddtrace import Span; s = Span("a");', stmt='"00-{:032x}-{:016x}-{:02x}".format(s.trace_id, s.span_id, 0)', number=1000000)
```

I updated the benchmark to only convert the sample rate to hex and then generate the traceparent. This ran in 0.245 seconds (0.245 microseconds per string): 
```
import timeit
timeit.timeit(setup='from ddtrace import Span; s = Span("a");', stmt='"00-00000000000000009ab1134a468e7936-7473f04b601bdbc0-{:02x}".format(0)', number=1000000)
```

majorgreys: Yes, this is helpful!

I was wondering what benefit we gain from not repeatedly converting the `trace_id` for multiple spans.

For a trace with 100 db spans, constructing the `traceparent` header from the `trace_id` and `span_id` took on my machine 24.5us.

```
root@b1754db4c2d8:/# python -m timeit --number=100000 --setup='from ddtrace.internal._rand import rand64bits; trace_id = rand64bits(); trace_id_hex = "{:032x}".format(trace_id)' '[("{:032x}-{:016x}".format(trace_id, rand64bits())) for _ in range(100)]'
100000 loops, best of 5: 24.5 usec per loop
```

If instead we already have the hex formatted `trace_id`, we get close to the same timing as we get if we do not include the `trace_id` in `traceparent`:

```
root@b1754db4c2d8:/# python -m timeit --number=100000 --setup='from ddtrace.internal._rand import rand64bits; trace_id = rand64bits(); trace_id_hex = "{:032x}".format(trace_id)' '[("-{:016x}".format(rand64bits())) for _ in range(100)]'
100000 loops, best of 5: 17 usec per loop
root@b1754db4c2d8:/# python -m timeit --number=100000 --setup='from ddtrace.internal._rand import rand64bits; trace_id = rand64bits(); trace_id_hex = "{:032x}".format(trace_id)' '[(trace_id_hex + "-{:016x}".format(rand64bits())) for _ in range(100)]'
100000 loops, best of 5: 18.7 usec per loop
```

tests/internal/test_database_monitoring.py
```diff
@@ -27,3 +27,64 @@ def test_injection_mode_configuration():
...
+    from ddtrace.settings import database_monitoring
+
+    dbspan = tracer.trace("dbname", service="orders-db")
+
+    sqlcomment = database_monitoring._get_dbm_comment(dbspan)
+    # tags in sqlcomments are comma delimited
+    # assert 4 tags are present in the sqlcomment
+    assert len(sqlcomment.split(",")) == 5
+    # assert tags sqlcomment contains the correct value
+    assert "dddbs='orders-db'" in sqlcomment
```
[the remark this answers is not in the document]
mabdinur: I did not notice this earlier but the keys should already be sorted. This is defined in the sqlcommenter spec: https://github.com/DataDog/dd-trace-py/blob/1.x/ddtrace/vendor/sqlcommenter/__init__.py#L45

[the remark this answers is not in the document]
mabdinur: I updated the tests to validate the order

Yun-Kim (approved):

tests/tracer/test_context.py
```diff
@@ -48,3 +48,30 @@ def test_eq(ctx1, ctx2):
...
+def test_traceparent():
+    def validate_traceparent(context, sampled_expected):
+        version_hex, traceid_hex, spanid_hex, sampled_hex = context._traceparent.split("-")
+        assert version_hex == "00"
+
+        assert len(traceid_hex) == 32
+        assert traceid_hex == "{:032x}".format(context.trace_id)
+
+        assert len(spanid_hex) == 16
+        assert spanid_hex == "{:016x}".format(context.span_id)
```
ZStriker19, line 62: Why do we pull the `span_id` and `trace_id` directly off of the `context`, rather than the `span`? It looks like they should always be the same so maybe it doesn't matter? I'm just unsure if there's ever and instance where they could not be the same. If that's not an issue, please resolve.

ZStriker19 (approved):

tests/tracer/test_context.py
mabdinur: Sampling priority is set on the context and not on the span. Also the context object is used to propagate trace and span ids so this seemed like a logical place to add the traceparent. We could also move this property to the Span object. I don't have strong feelings either way.

ddtrace/settings/database_monitoring.py
```diff
@@ -1,6 +1,22 @@
...
 
+from ddtrace.vendor.sqlcommenter import generate_sql_comment as _generate_sql_comment
+
+from . import _config as dd_config
+
+
+if TYPE_CHECKING:
+    from ddtrace import Span
+
+DBM_PARENT_SERVICE_NAME_KEY = "ddps"
```
Kyle-Verhoog, line 14: These are all included in the public api is this intentional?

PR #4329, opened 2022-10-18
chore(dbm): add dbm injection tag to database spans

mabdinur:
## Description

Sets ` _dd.dbm_trace_injected` tag on database spans. This tag will be used to correlate Database Monitoring profiles to APM spans. 

## Checklist
- [X] Title must conform to [conventional commit](https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional).

## Reviewer Checklist
- [x] Title is accurate.
- [x] Add to milestone.


ZStriker19 (approved):
So we need to tag our spans with _dd.dbm_trace_injected so that the backend can grab those spans and relate them to DBM? I'm a bit surprised that it doesn't go the other direction, with the DBM profiles having a trace id that they then search for? Anyways, if this is with the spec and makes sense, looks good to me.

majorgreys (approved):

Yun-Kim (dismissed):

ddtrace/settings/_database_monitoring.py
```diff
@@ -46,7 +48,7 @@ def _get_dbm_comment(db_span):
     }
 
     if dbm_config.injection_mode == "full":
-        # TODO: add _dd.dbm_trace_injected tag to db_span
+        db_span.set_tag_str(DBM_TRACE_INJECTED_TAG, "true")
```
brettlangdon, line 51: this seems like a weird side effect of this method. especially given that it is written where it seems like calling this only returns generated data.

is there another common place we could consider setting this tag?

mabdinur: I left the tag here to keep things concise. I don't see an alternative a side from moving this logic to it's own function and calling it in dbapi/aiomysql/aiopg/asyncpg. 

I have no strong feelings

brettlangdon: 🤔 yeah, I am not sure the best action to take here. 

one other way to look at it is... this isn't _actually_ injecting the context into the sql statement, it is just generating the sql comment that would be injected. so are we sure it is going to get used and the injection will occur?

mabdinur: I'll add a docstring to note the side effect

majorgreys (approved):

majorgreys (approved):

brettlangdon (dismissed):

majorgreys (approved):

Yun-Kim (approved):

PR #4339, opened 2022-10-20
Revert "chore(asm): enable ip collection if not disabled and appsec is enabled"

juanjux:
Reverts DataDog/dd-trace-py#4333

Looks like the spec is changing and we have to wait for another change in system tests before merging.


jd (approved):

juanjux:
Closing. This will be updated on 1.x instead of reverted. Will only be reverted on 1.6.

PR #4340, opened 2022-10-20
Revert 4333 juanjux/appsec 6386 enable ip collection if not disabled and appsec enabled

juanjux:
Revert #4333 from the 1.6 branch.

Looks like the spec is changing and we have to wait for another change in system tests before merging.

So this will probably be a change for the 1.7 version.


juanjux:
Cancelled, will backport the new version with the latest spec once merged on 1.x.

PR #4344, opened 2022-10-20
fix(httpx): fix support for older httpx versions

majorgreys:
## Description
Blocks #4294 due to patch-on-import exposing a failing sanic test suite due to [importing](https://github.com/sanic-org/sanic/blob/v19.12.0/setup.py#L87) an older version of httpx (`httpx==0.9.3`).

We currently support `httpx>=0.14`. This PR introduces support for older versions of httpx up to `httpx>=0.9` by adding compatibility for converting httpx URL components into a single string, as older httpx versions do not have the `URL.raw` property and we have to construct it manually. Second, we include correct patching for asynchronous/synchronous clients depending on the httpx version:
- `httpx==0.9.3` does not have a `AsyncClient`, but its `Client` class is actually asynchronous, meaning we need to patch the asynchronous send function for the `Client` class. 
- `httpx==0.10` has both `AsyncClient` and `Client`, but they are actually synonymous (`Client is just a reference to `AsyncClient`) and so we only need to patch the asynchronous send function for the `Client` class once.
- `httpx>=0.11` introduces a synchronous `Client` class, meaning we need to patch both sync/async send functions for `Client` and `AsyncClient` respectively.

## Testing
The test cases in `test_httpx` were structured to test both synchronous `Client` and asynchronous `AsyncClient` in each test, meaning this was incompatible for httpx versions < 0.11 because:
- `httpx<0.10` does not have `AsyncClient` 
- The `Client` class in `httpx<=0.10` is actually asynchronous (requiring `await` keywords to be called). 

To resolve this, I wrote a new test file with the same tests but added compatibility with asynchronous `Client` calls and removed references to `AsyncClient` (note we don't need to test `AsyncClient` for older versions since it is synonymous with the`Client` class).

## Checklist
- [X] Title must conform to [conventional commit](https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional).
- [X] Add additional sections for `feat` and `fix` pull requests.
- [X] [Library documentation](https://github.com/DataDog/dd-trace-py/tree/1.x/docs) and/or [Datadog's documentation site](https://github.com/DataDog/documentation/) is updated. Link to doc PR in description.

## Relevant issue(s)

We find older versions used as test dependencies and this mitigation was introduced in #4294.

## Reviewer Checklist
- [x] Title is accurate.
- [x] Add to milestone.


P403n1x87 (approved):

Kyle-Verhoog (dismissed):

Kyle-Verhoog (approved):

P403n1x87:
I think this PR is very close to completion. Can we try to get it merged?

emmettbutler (dismissed):
@Yun-Kim, this change is awesome. Nice work on navigating this tricky annoying situation.

ddtrace/contrib/httpx/patch.py
```diff
@@ -42,14 +44,19 @@ def _url_to_str(url):
     """
     Helper to convert the httpx.URL parts from bytes to a str
     """
-    # httpx==0.23.1 removed URL.raw, must construct it manually
-    if HTTPX_VERSION >= (0, 23, 1):
+    # httpx==0.13.0 added url.raw, removed in httpx==0.23.1. Otherwise must construct manually
+    if HTTPX_VERSION < (0, 13, 0):
+        scheme = url.scheme.encode("ascii")
```
mabdinur, line 49: Can we document why we need to use `ascii` encoding?

ddtrace/contrib/httpx/patch.py
```diff
@@ -153,13 +160,21 @@ def patch():
...
+        _w(httpx.Client, "send", _wrapped_sync_send)
+    else:
+        # httpx==0.9 Client class was asynchronous, httpx==0.10 made Client synonymous with AsyncClient
+        _w(httpx.Client, "send", _wrapped_async_send)
 
     pin = Pin()
-    pin.onto(httpx.AsyncClient)
     pin.onto(httpx.Client)
 
+    if HTTPX_VERSION >= (0, 11):
```
mabdinur, line 173: nit: we can consolidate this if block with the one above.

emmettbutler (approved):

mabdinur (approved):

PR #4356, opened 2022-10-24
feat(propagation):  DD_TRACE_PROPAGATION_STYLE_EXTRACT order matters

ZStriker19:
## Description
Going forward, we want the order of the propagation styles specified with `DD_TRACE_PROPAGATION_STYLE_EXTRACT` to matter. Previously we had hardcoded order. This gives users greater control over propagation.

## Example
Before this change, the order was hardcoded, so if we had `DD_TRACE_PROPAGATION_STYLE_EXTRACT=b3,datadog` then datadog extraction would be attempted first. Now if someone puts `DD_TRACE_PROPAGATION_STYLE_EXTRACT=b3,datadog` b3 will be evaluated first.

## Testing strategy
There is already a test that has all of the propagation styles, with datadog as the first value. Therefore I didn't think it was necessary to add further tests.

## Reviewer Checklist
- [ ] Title is accurate.

### DIFF
--- a/ddtrace/propagation/http.py
+++ b/ddtrace/propagation/http.py
@@ -544,19 +544,22 @@ class HTTPPropagator(object):
         try:
             normalized_headers = {name.lower(): v for name, v in headers.items()}
             # Check all styles until we find the first valid match
-            # DEV: We want to check them in this specific priority order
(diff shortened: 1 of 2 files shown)


ZStriker19:
Closing since this would be a breaking change will instead change the hardcoded evaluation order.

PR #4474, opened 2022-11-04
refactor(stack): do not try to account CPU time on gevent tasks

jd:
The current stack collector maps the CPU usage of a thread to its main
greenlet. This means that the following happens:
```
  MainThread +-> main gevent task (CPU + wall time)
             |-> other gevent task 1 (wall time)
             `-> other gevent task 2 (wall time)
```
The issue with this approach is that when using an application like gunicorn,
the end result is suboptimal: the main gevent task appears to be sleeping in
gunicorn main loop (because it's waiting for I/O) and hides potential real code
running in the main thread.

The original idea was to hide the MainThread frames completely as they
represent the gevent hub running most of the time, and replace it with the
gevent main task. However, it turns out that with software like gunicorn, this
loses information.

This new approach does not try to be smarter, and take a step back. It will now
accounts the CPU time on the MainThread, and list all the tasks only for the
wall time:
```
  MainThread (CPU + wall time) +-> main gevent task (wall time)
                               |-> other gevent task 1 (wall time)
                               `-> other gevent task 2 (wall time)
```
This does not try anymore to assign the CPU time to any gevent task, but keeps
the CPU time on the main thread. However, it shows the details of the gevent
tasks running in the wall time, making it clear for a user what's running.

As there is no way to account for the CPU time of the different gevent task
(they run in the same thread), this is a good representation of the reality
anyway.

This also has the upside to bring gevent tasks representation closer to what is
done for asyncio.


P403n1x87:
> As there is no way to account for the CPU time of the different gevent task
(they run in the same thread), this is a good representation of the reality
anyway.

Can't we tell which task is currently running when we sample?

P403n1x87 (approved):
Nice 🧹 . Just one question, otherwise LGTM.

jd:
> Can't we tell which task is currently running when we sample?

We can (we have the code in `ddtrace.profiling.collector._tasks.get_task` but that does not tell you how much CPU each tasks used between 2 polling intervals. So we could use the task frames to assigned the CPU time, but that circles back to what we do currently where we'd hide the main thread frames, which are sometimes running the real code we want to see.

P403n1x87:
> So we could use the task frames to assigned the CPU time, but that circles back to what we do currently where we'd hide the main thread frames, which are sometimes running the real code we want to see.

Ah so does this mean that there always is a running gevent task, which then ends up "masking" the main thread work? Otherwise one could assign the CPU time to the main thread when there is no active gevent task?

jd:
> Ah so does this mean that there always is a running gevent task, which then ends up "masking" the main thread work? Otherwise one could assign the CPU time to the main thread when there is no active gevent task?

Yes, there's always at least one gevent task (the main one). This is typically what happens in the gunicorn case.

majorgreys:
@jd Can you give an example of MainThread frames that are currently "hidden" that will be present after this change?

Also, this strikes me as a fix for how the stack collector works with gevent/gunicorn rather than a refactor. wdyt?

jd:
> @jd Can you give an example of MainThread frames that are currently "hidden" that will be present after this change?
> 
> Also, this strikes me as a fix for how the stack collector works with gevent/gunicorn rather than a refactor. wdyt?

Typically a MainThread runs Python code that does not involve gevent at all. I've seen this happening in with gunicorn where you basically only see the code sleeping in the `gr_frame` but the interesting code is actually in the main thread.
Looks better to not hide anything at this point.

I've little opinion on fix vs refactor :)

Kyle-Verhoog (approved):
lgtm but I think we need a release note as this should be a noticeable change for a user

Issue #4397, opened 2022-10-27
Trace middleware in Fastapi


Meemaw:
### Summary of problem

It seems middlewares are not traced in fastapi.

Moreover, trying to use `tracer.get_log_correlation_context()` within a middleware will not return any active context, which means we can't operate with `traceId` (e.g to enrich response/logs within a middleware).

https://fastapi.tiangolo.com/tutorial/middleware/

### Which version of dd-trace-py are you using?

1.5.2

brettlangdon:
@Meemaw thank you for opening this issue.

We do not support creating spans automatically from FastAPI middleware.

But I would have expected there to be an active root `fastapi.request` span that is accessible from the middleware.

Do you have a minimal reproducible setup? I assume just FastAPI with a route, and then a custom middleware that tries to access the current active span (`get_log_correlation_context()` is one way) ?

Meemaw:
I would have expected that as well. Should be reproducible with just any route, and just adding a custom middleware.

PR #4478, opened 2022-11-04
fix(fastapi): apply tracing to user middlewares in fastapi

mabdinur:
## Description

Resolve: https://github.com/DataDog/dd-trace-py/issues/4397

In Starlette and FastAPI applications the last middleware configured is the first one called ([here](https://github.com/encode/starlette/blob/0.21.0/starlette/applications.py#L168)). The fastapi trace middleware is applied when a fastapi application is created ([fastapi.FastAPI()](https://github.com/DataDog/dd-trace-py/blob/a21c8dbc433577245f5c145fd46f2a3dc4540402/ddtrace/contrib/fastapi/patch.py#L38)). If another middleware is configured after an application is initialized it will not be traced.

Here's an example of a fastapi middleware being configured after `fastapi.FastApi()` is called: https://fastapi.tiangolo.com/tutorial/middleware/#create-a-middleware. 

Note - This PR only ensures Fastapi middlewares are traced. Supporting all starlette middlewares is out of scope. 

## Reviewer Checklist
- [x] Title is accurate.


tests/contrib/fastapi/test_fastapi.py
```diff
@@ -589,3 +589,36 @@ def test_host_header(client, tracer, test_spans, host):
     assert test_spans.spans
     request_span = test_spans.spans[0]
     assert request_span.get_tag("http.url") == "http://%s/asynctask" % (host,)
+
+
+def test_tracing_in_middleware(tracer, test_spans):
+    """Tests if background tasks have been excluded from span duration"""
```
mabdinur, line 595:
```suggestion
    """"""
```

mabdinur:
@Mergifyio backport 1.6 .1.5

tests/contrib/fastapi/test_fastapi.py
```diff
@@ -51,6 +51,20 @@ def application(tracer):
     yield application
 
 
+@pytest.fixture
+def app_with_middleware(application, tracer):
+    @application.middleware("http")
+    async def traced_middlware(request, call_next):
+        response = await call_next(request)
+        response.headers["MIDDLEWARE-CALLED"] = "true"
+        span = tracer.current_span()
```
brettlangdon, line 60: does it matter if we call this before or after the `await call_next(request)` ?

tests/contrib/fastapi/test_fastapi.py
```diff
@@ -589,3 +603,20 @@ def test_host_header(client, tracer, test_spans, host):
     assert test_spans.spans
     request_span = test_spans.spans[0]
     assert request_span.get_tag("http.url") == "http://%s/asynctask" % (host,)
+
+
+def test_tracing_in_middleware(app_with_middleware, test_spans):
```
brettlangdon, line 608: Would a snapshot test work better here?

Basically we want to assert that the normal `fastapi.request` span exists along with the customer user middleware span?

ddtrace/contrib/fastapi/patch.py
```diff
@@ -42,6 +42,15 @@ def traced_init(wrapped, instance, args, kwargs):
     wrapped(*args, **kwargs)
 
 
+def trace_add_middleware(wrapped, instance, args, kwargs):
+    # remove ddtrace trace middleware
```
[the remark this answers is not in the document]
mabdinur: Nope. It's leftover from an approach I scrapped. I will remove it

tests/contrib/fastapi/test_fastapi.py
```diff
@@ -589,3 +603,20 @@ def test_host_header(client, tracer, test_spans, host):
...
+        assert r.status_code == 200
+
+        spans = test_spans.pop_traces()
+        assert len(spans) == 1
+        assert len(spans[0]) == 2
+        request_span, _ = spans[0]
+
+        assert r.headers["MIDDLEWARE-CALLED"] == "true"
+        assert "DD-TRACE-CURRENT-SPAN" in r.headers
+        assert r.headers["DD-TRACE-CURRENT-SPAN"] == request_span.name
```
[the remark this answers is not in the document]
mabdinur: We do not create middleware spans in fastapi. This test just asserts that the active span in user middlewares is `fastapi.request`. 

I'll make this test a snapshot test so it's easier to understand.

ddtrace/contrib/fastapi/patch.py
```diff
@@ -42,6 +42,18 @@ def traced_init(wrapped, instance, args, kwargs):
...
 
 
+def trace_add_middleware(wrapped, instance, args, kwargs):
+    # Ensures user_middlewares are added after ddtrace TraceMiddleware
+    if instance.user_middleware and instance.user_middleware[0].cls is TraceMiddleware:
+        # Overrides FastApi.add_middlware(). Note - fastapi.applications.FastAPI is the
+        # child class of starlette.applications.Starlette():
+        # https://github.com/encode/starlette/blob/0.13.2/starlette/applications.py#L115
+        instance.user_middleware.insert(1, Middleware(*args, **kwargs))
+        instance.middleware_stack = instance.build_middleware_stack()
```
brettlangdon, line 52: I worry about this a little bit, since we are basically duplicating the logic from Starlette, if that behavior ever changes then this will break things.

I am not sure a better way at the moment, but calling out that this gives me pause.

releasenotes/notes/fastapi-fix-middlewares-705975a535daaea8.yaml
```diff
@@ -0,0 +1,4 @@
+---
+fixes:
+  - |
+    fastapi: Ensure spans are created for all user middlewares.
```
ZStriker19, line 4: We don't actually generate a span for each middleware do we? Would it be more accurate to say "Ensure traced middleware span duration encompasses all middlewares"? Or something like that?

mabdinur, line 4:
```suggestion
    fastapi: Ensure all fastapi middlewares are captured in the `fastapi.request` span.
```

ZStriker19 (dismissed):

releasenotes/notes/fastapi-fix-middlewares-705975a535daaea8.yaml
```diff
@@ -0,0 +1,4 @@
+---
+fixes:
+  - |
+    fastapi: Ensure all fastapi middlewares are captured in the `fastapi.request` span.
```
Yun-Kim, line 4:
```suggestion
    fastapi: Previously, custom fastapi middlewares configured after application startup were not traced. This fix ensures that all fastapi middlewares are captured in the `fastapi.request` span.
```

ddtrace/contrib/fastapi/patch.py
Yun-Kim: I think this is worth discussing, as the alternative option is to change the instance's middleware stack after the middleware has been added. Not sure which one is better, as both rely on the internal mechanism of `FastAPI.add_midleware()`.

mabdinur: Whenever user middlewares are added [build_middleware_stack](https://github.com/encode/starlette/blob/0.13.2/starlette/applications.py#L58) is called. Instead of patching` Fastapi.__init__()` and `add_middleware()` we can wrap the asgi app returned by `build_middleware_stack` in the ddtrace TraceMiddleware. However this approach patches fastapi "internals" and could break at any point. Also I think it's harder to follow and maintain

ZStriker19 (dismissed):

ddtrace/contrib/fastapi/patch.py
emmettbutler: If I'm following the conversation correctly, there are three options being discussed:

1. This PR's current approach, patching the two-line function `add_middleware` and duplicating its internal logic
2. Wrap the return value of `build_middleware_stack`, a `starlette.routing.Router`
3. Wrap `add_middleware` but use @mabdinur's steps above to avoid duplicating internal logic at the cost of decreased performance

Assuming I understand the options correctly, I'm actually most interested in number 2. This is because my read of the [Starlette documentation suggests](https://www.starlette.io/routing/#working-with-router-instances) that the `starlette.routing.Router` class is part of the public API. Thus, wrapping the instance returned by `build_middleware_stack` appears to be the approach least dependent on the private API.

emmettbutler (changes requested):

emmettbutler (approved):

ZStriker19 (dismissed):

ZStriker19 (approved):

emmettbutler (approved):

PR #4479, opened 2022-11-04
chore(test): automatic circleci configuration generated from the riotfile.py

christophe-papazian:
## Description

This PR introduce a new way to configure the circleci tests. Instead of touching several files to configure or add new tests, only the riotfile.py is edited and those changes are used to automatically generate the right configuration for the circleci workflows:

## Motivation
 - easier and simpler maintenance
 - open the way to migrate all tests from tox to riot. This will be a subsequent PR.
 - open the way to add new optional test workflow. We can fix all the versions in the mandatory tests and add  optional workflow to test the latest version of each dependencies. This will prevent to break the tests when a new version of a dependency is released, while still being able to monitor the changes. This will be a subsequent PR.

## How ?
- most of the yaml configuration into the .circleci/config.yml have been moved to python format into other files. 
- Each riot test configuration is in riotfile.py directly, so each riot test is only defined here.
- The main skeleton of the configuration is in circle_config_init.py. It still contains configuration for non riot tests for now.
- The script generate-circleci-config.py automatically generates the configuration from the 2 previous files.

## Reviewer Checklist
- [ ] Title is accurate.


gnufede (dismissed):

juanjux (approved):

majorgreys:
I labeled this as manual merge so we could review this solution, which makes significant changes to the library testing.

PR #4505, opened 2022-11-08
fix(docs): fix configuration options overflowing the sidebar

brettlangdon:
## Description
**Before:**
image

**After:**
image

We will not be able to see this locally or from the artifacts in CircleCI since RTD is using a newer version of Sphinx (RTD: 5.3.0 Us: ~=4.3.2) which is what is causing these options to show up in the sidebar and overflow.

I made an attempt to upgrade Sphinx, but it is non-trivial. This fix should mitigate the issue for now until we can revisit and reconcile the differences between RTD and CircleCI builds.

This approach allows us to use our own CSS file inherited from the base Alabaster CSS file with whatever overrides we want.

## Reviewer Checklist
- [ ] Title is accurate.

### DIFF
--- a/docs/_static/css/ddtrace.css
+++ b/docs/_static/css/ddtrace.css
@@ -1,0 +1,10 @@
+@import '../alabaster.css';
+
+/*
+  Clip any text in the sidebar.
+
+  Originally added to avoid configuration names overflowing into the page content.
+*/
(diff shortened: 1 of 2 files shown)


brettlangdon:
Closing since preference is to hide these items instead.

PR #4470, opened 2022-11-03
refactor(dbm): decouple dbm propagation from dbapi

mabdinur:
## Description

Previously we were relying on dbapi internals to intercept and add dbm comments to sql statements. 

Limitations of the current design:
- It highly couples DBAPI tracing code with DBM injection. Supporting async libraries will require significant duplication.
- It did not support intercepting sql statements that are not strings (ex: psycopg2.sql.Composable). This functionality was achieved by overriding methods in TracedCursor().
- It required args to be passed to dbapi.TracedCursor._tracemethod() in very specific way. If this code was refactored DBM support could break and it would be hard to debug.

## Proposal

- Introduce a DBM Propagator to extract and replace sql statements from args/kwargs
- - Move DBM args/kwargs parsing from `ddtrace.contrib.dbapi.TracedCursor()` to `ddtrace.propagation._database_monitoring._DBM_Propagator()`
- Moves the sql injection support from DBAPI to the DBM Propagator
- Updates dbapi tests to use the new _DBM_Propagator()
   - psycopg tests do not touch dbm internals so they don't need to be updated. 

## Reviewer Checklist
- [x] Title is accurate.
- [x] Description motivates each change.
- [x] No unnecessary changes were introduced in this PR.
- [x] Avoid breaking [API](https://ddtrace.readthedocs.io/en/stable/versioning.html#interfaces) changes unless absolutely necessary.
- [x] Tests provided or description of manual testing performed is included in the code or PR.
- [x] Release note has been added for fixes and features, or else `changelog/no-changelog` label added.
- [x] All relevant GitHub issues are correctly linked.
- [x] Backports are identified and tagged with Mergifyio.

PR #4508, opened 2022-11-08
chore(dbm): support composable queries in psycopg

mabdinur:
## Description

DBM context propagation only supports sql statements with the type: string. This PR extends support to Postgres sql statements which execute python objects containing sql statements.

## Testing strategy

Run psycopg tests with DBM context propagation enabled. 
Update snapshot test.

## Reviewer Checklist
- [x] Title is accurate.


brettlangdon (commented):
no tests?

ddtrace/contrib/dbapi/__init__.py
```diff
@@ -154,6 +154,29 @@ def _set_post_execute_tags(self, span):
         if isinstance(row_count, six.integer_types) and row_count >= 0:
             span.set_tag(sql.ROWS, row_count)
 
+    def _propagate_dbm_context(self, dbspan, args):
+        # type: (...) -> Tuple[Tuple[Any, ...]]
+        dbm_comment = _database_monitoring._get_dbm_comment(dbspan)
+        if dbm_comment is None:
+            return args
+        # add DBM comment to query arg
+        sql_with_dbm_tags = self._dbm_sql_injector(dbm_comment, args[0])
```
ZStriker19, line 163: we can always rely on this order?

ddtrace/contrib/psycopg/patch.py
```diff
@@ -65,13 +65,17 @@ def _trace_method(self, method, name, resource, extra_tags, dbm_operation, *args
...
 
         return super(Psycopg2TracedCursor, self)._trace_method(
             method, name, resource, extra_tags, dbm_operation, *args, **kwargs
         )
 
+    def _dbm_sql_injector(self, dbm_comment, sql_statement):
+        if isinstance(sql_statement, Composable):
+            composable_dbm_comment = psycopg2.sql.SQL(dbm_comment)
+            return composable_dbm_comment + sql_statement
+        return super(Psycopg2TracedCursor, self)._dbm_sql_injector(dbm_comment, sql_statement)
```
ZStriker19, line 77: I'm not sure I understand this return, can you explain why we need to use super here?

ddtrace/contrib/dbapi/__init__.py
mabdinur: Yes due to how we call TracedCursor._traced_method(…) in the execute/executemany/callproc. The sql statement is never passed as a keyword argument. It’s always the first argument. 

Relying on this behavior was the motivation for this refactor PR: https://github.com/DataDog/dd-trace-py/pull/4470.

ddtrace/contrib/psycopg/patch.py
mabdinur: This function checks if the sql statement is a composable object. If it’s not then the query must be a string and the default dbm sql injector is used. The default sql injector is implemented in the parent class and we call it using super.  It simply adds strings. I can add a comment to make this more clear

ZStriker19 (dismissed):

ZStriker19 (approved):

ddtrace/contrib/dbapi/__init__.py
```diff
@@ -315,18 +339,3 @@ def _get_vendor(conn):
...
     return conn.__class__.__module__.split(".")[0]
-
-
-def _propagate_dbm_context(dbspan, args):
-    # type: (...) -> Tuple[Tuple[Any, ...]]
-    dbm_comment = _database_monitoring._get_dbm_comment(dbspan)
-    if dbm_comment is None:
-        return args
-    # If dbm propagation supported and enabled the first arg must be a sql query or a procedure.
-    assert len(args) > 0 and isinstance(args[0], str)
```
brettlangdon, line 326: This assertion is being removed, is this correct?

tests/contrib/psycopg/test_psycopg.py
```diff
@@ -408,11 +413,34 @@ def test_postgres_dbm_propagation_comment(self):
...
+        cursor.executemany(SQL("select %s"), (("foo",), ("bar",)))
+        dbm_comment = " /*dddbs='postgres',dde='staging',ddps='orders-app',ddpv='v7343437-d7ac743'*/"
+        cursor.__wrapped__.execute.assert_called_once_with(
+            Composed(
+                [
+                    SQL(" /*dddbs='postgres',dde='staging',ddps='orders-app',ddpv='v7343437-d7ac743'*/"),
+                    SQL("select 'blah'"),
+                ]
+            )
+        )
```
brettlangdon, line 434: Do we have a test that when dbm is disabled that this is `Composed([SQL("select 'blah'")])` ?

I want to be sure the data type passed to psycopg2 is the exact same (`Composed` and not `SQL`)

ddtrace/contrib/dbapi/__init__.py
mabdinur: Yup with this change the assertion is not required. Previously the first argument had be a string now it can be overritten to be anything (ex: psycopg composable).

tests/contrib/psycopg/test_psycopg.py
mabdinur: Yup there's this preexisting test: https://github.com/DataDog/dd-trace-py/blob/v1.6.2/tests/contrib/psycopg/test_psycopg.py#L267. Here the `DD_DBM_PROPAGATION_MODE` is not set so the sql statement is not changed.

brettlangdon: I am not sure that tests what I expect/want it to test. Or at least it doesn't assert it.

That is testing that with a `SQL()` that we still properly trace and collect the right resource name.

But we are passing a `SQL()` (correct?) into `cur.execute()`. With the DBM stuff that is being changed to a `Composed()` correct? Is that a problem? Would/could psycopg2 behavior differently if we are passing in a different type than the user provided?

mabdinur: We pass in a SQL() object in the test. In the DBM logic we generate another SQL() object containing the dbm comment and we add those two SQL objects together. Adding two SQL() objects produces the composed object in the assertion. According to the psycopg spec creating a composed object from two SQL object is the preferred way of modifying a query: https://www.psycopg.org/docs/sql.html#sql-objects.

ex: `SQL("/* dbm comment */") + SQL("/* select *; */") -> Composed([SQL("/* dbm comment */"), SQL("/* select *; */"])`

Both SQL() and Composed() implement the Composable interface. But you're right they're not the same type.
`isinstance(psycopg.sql.SQL('h')), psycopg.sql.Composable)== True`
`isinstance(psycopg.sql.Composed([psycopg.sql.SQL('h'), psycopg.sql.SQL('h')]), psycopg.sql.Composable)== True`
`isinstance(psycopg.sql.SQL('h'), psycopg.sql.Composed)== False`

To answer your questions:
1. Yes,  this logic converts SQL() objects to Composed(). But if a composed object is passed to `cursor.execute()` then we don't change the type we simply append to it. 
2. This behavior is expected and shouldn't cause any unexpected side effects.

Alternative approach:

-  SQL() is passed to cursor.execute() get the raw string and create a new SQL() object with the DBM object
- If Composed() is passed to cursor.execute() create a SQL() object with the DBM comment and append it to the existing Composed. 

The con with this alternative approach is that it goes against psycopg composable conventions.


@brettlangdon  do you have any strong feelings? We can move forward with the alternative approach and put this PR on hold for now.

brettlangdon: Ok, I don't think I understood the class hierarchy with `Composed`/`SQL` and `Composable`.

Given that they are both `Composable`... and the preferred way from psycopg2, then seems fine to me.


My concern was if we changed a `str` to an `int` type of change, which this doesn't seem to be.

brettlangdon (approved):

PR #4526, opened 2022-11-10
feat(sci): add git metadata tags support

sashacmc:
## Description
Add four environment variables:

- DD_GIT_REPOSITORY_URL string, default: empty
- DD_GIT_COMMIT_SHA string, default: empty
- DD_TRACE_GIT_METADATA_ENABLED bool, default: true
- DD_MAIN_PACKAGE string, default empty

Add following steps:

- Check if DD_TRACE_GIT_METADATA_ENABLED is True
- Check if git metadata is already provided by environment variables:
  - Check DD_GIT_REPOSITORY_URL and DD_GIT_COMMIT_SHA, if they exist, use metadata from them.
  - Check DD_TAGS, if it contains git.repository_url and git.commit.sha, use metadata from them.
  - Continue with metadata from artifacts.
- Read package name from DD_MAIN_PACKAGE, if it empty try detect package (will work for direct import)
- Read git metadata from the package metadata field Project-URL: source_code_link, split the value by "#" and use first part as repository URL and second part as commit hash

During trace telemetry sending:

- Add metadata tags to the first span of each chunk (NB: if [V7](https://github.com/DataDog/datadog-agent/blob/3ce2edbd37f3815102e31f59926699930bacdfc2/pkg/trace/api/api.go#L439) payload support already implemented send tags on payload level):
  - _dd.git.repository_url for repository URL
  - _dd.git.commit.sha for commit hash

During profile telemetry sending:
- Add metadata tags to each profile:
  - git.repository_url for repository URL
  - git.commit.sha to commit hash

## Checklist

## Motivation
Add new git metadata tags and extraction support
RFC: [SCI Embedding Git references in built artifacts RFC](https://docs.google.com/document/d/1vxuRUNzHqd6sp1lnF3T383acbLrG0R-xDGi8cdZ4cs8/edit#)
Jira: [SC-1159](https://datadoghq.atlassian.net/browse/SC-1159)

## Testing strategy
1. Build app from the cloned git repo with tracer/profiler with ddgitmetadata module in setup.py
1. Check git.repository_url and git.commit.sha tags on traces/profiles DD pages
1. Same as 1, but with DD_GIT_REPOSITORY_URL and DD_GIT_COMMIT_SHA with different values (values from env variables must present)
1. Same as 1, but with DD_TAGS="git.repository_url:some_repo git.commit.sha:some_sha" (values from env variables must present)
1. Same as 1, but with DD_TRACE_GIT_METADATA_ENABLED=false, tags must not be presented

## Reviewer Checklist
- [x] Title is accurate.

[SC-1159]: https://datadoghq.atlassian.net/browse/SC-1159?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ddtrace/tracer.py
```diff
@@ -192,6 +193,9 @@ def __init__(
         # globally set tags
         self._tags = config.tags.copy()
 
+        # update git metadata tags
+        self._tags.update(gitmetadata.get_tags())
```
brettlangdon, line 197: We should not be adding this metadata onto every span in the tracer.

Instead we can set it once per-payload and have intake expand this onto every tag.

The reasoning is encoding data is very expensive, adding some additional bytes once per-trace isn't too bad, but once per-span will have a measurable performance overhead.

ddtrace/gitmetadata.py
```diff
@@ -0,0 +1,88 @@
...
+            attrs["project_urls"] = {}
+        attrs["project_urls"]["source_code_link"] = source_code_link
+        return setup(**attrs)
+
+    return patch
+
+
+def __find_package():
+    frm = inspect.stack()[-1]
+    module = inspect.getmodule(frm[0])
```
brettlangdon, line 39: this is probably going to be error prone. I am concerned about our abilities to properly get the user's package name.

have we tested this code with `ddtrace-run python my_script.py`? if I understand this correctly, it would mean it would get our `sitecustomize.py` in this case and then resolve to our `ddtrace` module ?


I don't love the idea... but an alternative is we can scan all available packages and get a list of ones which have the extra urls metadata you are adding.

We already iterate over all packages anyways so we should be able to capture that. The one downside is doing that is expensive, and blocking startup for it is probably not what we want (`pkg_resources` is soooo slow to import... seriously... really slow to just import)

ddtrace/gitmetadata.py
```diff
@@ -0,0 +1,88 @@
...
+            metadata_lines = pkg.get_metadata_lines("PKG-INFO")
+
+        source_code_link = ""
+        for metadata in metadata_lines:
+            name_val = metadata.split(": ")
+            if name_val[0] == "Project-URL":
+                capt_val = name_val[1].split(", ")
+                if capt_val[0] == "source_code_link":
+                    source_code_link = capt_val[1].strip()
+                    break
```
brettlangdon, line 72: I _think_ there is a better way to do this, I cannot remember what it is, but leaving a comment as a reminder to research a better way.

`importlib` might parse this already, or else I think it is a valid MIME message, so can be parsed with the email module to get something easier to work with. (the performance of what you have is likely better than parsing)

ddtrace/gitmetadata.py
```diff
@@ -0,0 +1,88 @@
+import os
+import inspect
+import subprocess
+import pkg_resources
```
brettlangdon, line 4: we should avoid importing `pkg_resources`. by default when imported it builds a working set of all packages, so it iterates all packages and parses their metadata. it is really really really slow. the overhead it adds to startup makes it basically a non-starter for serverless environments.

sashacmc:
From RFC discussion:

I'd like to see the test scenarios laid out a bit more clearly. I think there are a few ways people can setup their application, and a few ways they can use the `ddtrace` package to load the tracer.

It looked like the POC depends on the tracer being imported for the first time from a specific module/package, which might not be the case. I want to be sure all scenarios are properly laid out and considered when implementing the solution.

Some initial ones are:

- Manually importing ddtrace
- Using ddtrace-run
- uwsgi using --import=ddtrace.bootstrap.sitecustomize
- New k8s admission controller injection
- Coming soon: agent runc shim injection



ddtrace-run: [https://ddtrace.readthedocs.io/en/stable/installation_quickstart.html#tracing](https://www.google.com/url?q=https://ddtrace.readthedocs.io/en/stable/installation_quickstart.html%23tracing&sa=D&source=docs&ust=1671130471326535&usg=AOvVaw3Ou-F9km3mJyMd5mYOBFUd)

uwsgi: [https://ddtrace.readthedocs.io/en/stable/advanced_usage.html#uwsgi](https://www.google.com/url?q=https://ddtrace.readthedocs.io/en/stable/advanced_usage.html%23uwsgi&sa=D&source=docs&ust=1671130471326670&usg=AOvVaw1O-ssW6oscwzyhpfnaEN5q)

k8s admission controller: [https://docs.datadoghq.com/tracing/trace_collection/admission_controller/#pagetitle](https://www.google.com/url?q=https://docs.datadoghq.com/tracing/trace_collection/admission_controller/%23pagetitle&sa=D&source=docs&ust=1671130471326720&usg=AOvVaw3kF3cIsAQRVMW5ZGug3Qd3)

I don't think I have the links for the runc shim injection work, but it isn't done for Python anyways.

ddtrace/internal/gitmetadata.py
```diff
@@ -0,0 +1,112 @@
+import os
+import typing
+
+from ddtrace.internal.utils import formats
+
+
+__GITMETADATA_TAGS = None  # type: typing.Optional[typing.Dict[str, str]]
```
P403n1x87, line 7: Is there a reason why we are using double underscore prefixes for private variables here?

ddtrace/internal/gitmetadata.py
```diff
@@ -0,0 +1,112 @@
...
+ENV_GLOBAL_TAGS = "DD_TAGS"
+
+TAG_REPOSITORY_URL = "git.repository_url"
+TAG_COMMIT_SHA = "git.commit.sha"
+
+TRACE_TAG_REPOSITORY_URL = "_dd.git.repository_url"
+TRACE_TAG_COMMIT_SHA = "_dd.git.commit.sha"
+
+
+def __get_tags_from_env():
```
P403n1x87, line 22: We already have [`envier`](https://github.com/datadog/envier) as a dependency in `ddtrace`. I wonder if we could use that to fetch from the environment and avoid custom caching logic here.

ddtrace/internal/gitmetadata.py
```diff
@@ -0,0 +1,112 @@
...
+        if source_code_link != "" and "#" in source_code_link:
+            repository_url, commit_sha = source_code_link.split("#")
+            commit_sha = commit_sha.split("&")[0]
+            return {TRACE_TAG_REPOSITORY_URL: repository_url, TRACE_TAG_COMMIT_SHA: commit_sha}
+        return {}
+    except importlib_metadata.PackageNotFoundError:
+        return {}
+
+
+def get_tracer_tags():
```
P403n1x87, line 72: I'd be more inclined to leave product-specific logic out of this generic module and let the tracer and profiler use the (internal) API exposed here. WDYT?

ddtrace/internal/gitmetadata.py
sashacmc: When I studied python (long long time ago) it means that variable is private and nobody outside the module have access (except some tricks), is something changed?

ddtrace/internal/gitmetadata.py
sashacmc: Hmm, interesting, never know about this module. Thanks!

But here, I used the same way, as [tracer ](https://github.com/DataDog/dd-trace-py/blob/1.x/ddtrace/settings/config.py#L199) and this seems to be more convenient because we need custom logic to parse tags.

ddtrace/internal/gitmetadata.py
sashacmc: Yes, I thought about it, but it depends on what you consider a product. If we consider the delivery of git metadata as a product, then this solution looks more nice, since it encapsulates all the logic in this module.

Moreover, when the transition to the payload format V7 is made, only the get_tags method should remain here (now we seprate it because of tracer's workaround to reduce traffic)

ddtrace/internal/gitmetadata.py
P403n1x87: A single underscore is generally enough. A double underscore prefix (with no double underscore suffix) is generally used in class methods and attributes and causes the variable names to be mangled. We very rarely use double underscore prefixes and we document that anything starting with a single underscore is private. So I think we could perhaps align with the rest of the library and use a single underscore prefix here.

ddtrace/internal/gitmetadata.py
P403n1x87: Envier variable declarations allow for a custom parser, so it should be flexible enough. Our long-term plan is to eventually replace every direct env access with envier, so if we could avoid any direct env accesses going forward it will make our life easier later 🙂.

ddtrace/internal/gitmetadata.py
sashacmc: Ok! Thanks!

sashacmc:
NB: Don't merge until APM Intake team deploy corresponded changes!

emmettbutler:
I added the Stale tag because this branch has conflicts with the base.

brettlangdon:
Since there are conflicts and no activity, I am going to close for now. Please feel free to update and reopen whenever you are ready!

ddtrace/internal/gitmetadata.py
sashacmc: Done!

PR #4534, opened 2022-11-11
chore(tracing): change system.pid to process_id

wconti27:
## Description
Changing the system.pid span attribute to process_id.  Tests were also updated to have new process_id assertion.

## Motivation
Part of a larger effort to make tag naming more consistent among Datadog tracing libraries. 'process_id' is a required attribute for all spans.

## Design 
No real design choices to be made. Changed constant string for PID and updated snapshots

## Testing strategy
Snapshot tests updated. Also changed a util test asserting on system.pid before to now assert on process_id

## Relevant issue(s)
Related PR's:

## Reviewer Checklist
- [x] Title is accurate.


brettlangdon:
Context for others, our backend already remaps `system.pid` to `process_id`, so the `system.pid` tag doesn't show up for anyone right now anyways.

brettlangdon:
I labeled as not needing a changelog, and updated title to be a `chore` since there shouldn't be any customer facing impact from this change since we already remap the tag.

Anyone feeling differently?

brettlangdon (approved):

Yun-Kim:
> I labeled as not needing a changelog, and updated title to be a `chore` since there shouldn't be any customer facing impact from this change since we already remap the tag.
> 
> Anyone feeling differently?

Technically this is a breaking change since `ddtrace.constants` is a public module. If we're confident this shouldn't affect users in any way, we should still consider adding a release note just in case?

brettlangdon:
> > I labeled as not needing a changelog, and updated title to be a `chore` since there shouldn't be any customer facing impact from this change since we already remap the tag.
> > Anyone feeling differently?
> 
> Technically this is a breaking change since `ddtrace.constants` is a public module. If we're confident this shouldn't affect users in any way, we should still consider adding a release note just in case?

🤔 hmm. The variable name isn't change, the value changing _should_ be ok ... ?

I don't think I have a strong opinion about adding a note for:

> tracing: ``ddtrace.constants.PID`` is now ``process_id`` instead of ``system.pid``

?

mabdinur (approved):

PR #4535, opened 2022-11-11
feat(tracing): add language tag to first span within each encoder payload chunk

wconti27:
## Description
This change sets the language tag within the encoder. The tag is set on the first span of every chunk within a payload.  Within backend trace intake, the tag will be expanded to all spans.

NOTE: Agent 7.33 and above adds the tracer language within the headers that are sent to intake. Python 1.x supports agents 7.25 and above so those agent versions between 7.25 and 7.33 cannot send the tracer language within headers. The result of this is that all spans must have the language tag encoded as the agent `/info` endpoint is not implemented yet within Python, leaving no way to detect agent version. Therefore, the tag must be set by the Python tracer for every payload sent to an agent.

## Motivation
Part of a larger effort to make tag naming more consistent among tracing libraries.

## Design 
Set the tag within the encoder for best performance. Only set it on the first span of each chunk within a trace payload.

## Testing strategy
Changed all snapshots to have language tag. Added general testing for the language tag as well within the tracer tests.  Changed runtime metrics test assertions to no longer assert the language field is none. Assert child does not have language tag as well.

## Relevant Issues / PR's
Relevant related issues: 

## Reviewer Checklist
- [ ] Title is accurate.


ddtrace/tracer.py
```diff
@@ -694,6 +694,8 @@ def _start_span(
                 # We must always mark the span as sampled so it is forwarded to the agent
                 span.sampled = True
 
+        span.set_tag_str("language", "python")
```
brettlangdon, line 697: We might want to performance test this change, since this will add encoding overhead to all spans.


Given that there are different ways for us to get this data to the trace agent/intake we might want to defer the implementation detail to the encoder/writer.

By that I mean instead of setting the language tag on every span, add it as a default tag on each `Context` object.

Then we can either:

- Ignore the tag if we have trace agent >= 7.33 since we get the value from setting the header
- Have the encoder add this to every span, similar to what we do for `_dd.origin` tag.


cc @Kyle-Verhoog who I am guessing will have some good thoughts/opinions here.

Yun-Kim: I'm not sure how the first option works with setting the header, could you explain? Thanks!

brettlangdon: The trace agent in version 7.33 will also pass along the value of `Datadog-Meta-Lang` header to the backend in a new payload format. The backend will then ensure all spans in that payload have the language tag present and set to value of `Datadog-Meta-Lang`.

This is the commit in the agent which adds it: https://github.com/DataDog/datadog-agent/commit/4b005758cdf6b5a63c9f3957ad8d82c17b9c558a

This is the code that adds the header value to the payloads: https://github.com/DataDog/datadog-agent/blob/4b005758cdf6b5a63c9f3957ad8d82c17b9c558a/pkg/trace/api/api.go#L401-L445

ddtrace/context.py
```diff
@@ -57,6 +57,8 @@ def __init__(
         self._meta = meta if meta is not None else {}  # type: _MetaDictType
         self._metrics = metrics if metrics is not None else {}  # type: _MetricDictType
 
+        self._meta["language"] = meta.language if (meta is not None and "language" in meta) else "python"
```
brettlangdon, line 60:
```suggestion
        if "language" not in self._meta:
            self._meta["language"] = "python"
```

or maybe just having L57 as `self._meta = meta if meta is not None else {"language": "python"}` would also work?

ddtrace/internal/_encoding.pyx
```diff
@@ -509,7 +510,7 @@ cdef class MsgpackEncoderBase(BufferedEncoder):
     cpdef flush(self):
         raise NotImplementedError()
 
-    cdef int pack_span(self, object span, void *dd_origin) except? -1:
+    cdef int pack_span(self, object span, void *dd_origin, int i) except? -1:
```
brettlangdon, line 513: I wonder if having something more explicit like `bool is_first_span` or something

Or maybe, we can just change `i` to be a better name like `index_in_chunk` or... idk, naming is hard.

ddtrace/internal/_encoding.pyx
```diff
@@ -534,6 +535,8 @@ cdef class MsgpackEncoderV03(MsgpackEncoderBase):
             L = len(d)
             if dd_origin is not NULL:
                 L += 1
+            if i == 0:
```
brettlangdon, line 538: comment on why would be good, less obvious here why we are increasing the length by 1 when `i` is `0`

ddtrace/internal/_encoding.pyx
```diff
@@ -548,6 +551,11 @@ cdef class MsgpackEncoderV03(MsgpackEncoderBase):
                     ret = pack_bytes(&self.pk, _ORIGIN_KEY, _ORIGIN_KEY_LEN)
                     if ret == 0:
                         ret = pack_bytes(&self.pk, dd_origin, strlen(dd_origin))
+                if i == 0:
+                    ret = pack_text(&self.pk, <str> "language")
+                    if ret != 0: return ret
+                    ret = pack_text(&self.pk, <str> "python")
```
brettlangdon, line 557: we will want to use a `const char*` here like for origin (L79) for both the key and the value.

`pack_text` has some conversion that may happen to get bytes and then append, where if we start with a properly formed `const char*`/bytes anyways, they can just be appended to the payload with minimal overhead (we just need to mem copy the bytes).

ddtrace/internal/_encoding.pyx
```diff
@@ -524,7 +525,7 @@ cdef class MsgpackEncoderV03(MsgpackEncoderBase):
     cdef void * get_dd_origin_ref(self, str dd_origin):
         return string_to_buff(dd_origin)
 
-    cdef inline int _pack_meta(self, object meta, char *dd_origin) except? -1:
+    cdef inline int _pack_meta(self, object meta, char *dd_origin, int i) except? -1:
```
brettlangdon, line 528: same here, instead of `i` maybe a more explicit name like `bool include_global_meta`

ddtrace/internal/_encoding.pyx
```diff
@@ -732,8 +740,15 @@ cdef class MsgpackEncoderV05(MsgpackEncoderBase):
...
 
-        ret = msgpack_pack_map(&self.pk, len(span._meta) + (dd_origin is not NULL))
+        ret = msgpack_pack_map(&self.pk, len(span._meta) + (dd_origin is not NULL) + (i == 0))
         if ret != 0: return ret
+
+        if i == 0:
+            ret = self._pack_string("language")
+            if ret != 0: return ret
+            ret = self._pack_string("python")
+            if ret != 0: return ret
```
brettlangdon, line 750: we should try to mimic the dd origin code below.

the v0.5 encoder uses a string table to store key names and values. so we actually end up packing in integers instead of strings/bytes here.

since `language` and `python` are going to always show up in every single payload, we can pre-create the string table with entries for `language` and `python`, then we can just pack in the hard coded string values here (see the origin key below)

ddtrace/internal/encoding.py
```diff
@@ -51,8 +51,8 @@ def encode(self, obj):
         raise NotImplementedError()
 
     @staticmethod
-    def _span_to_dict(span):
-        # type: () -> Dict[str, Any]
+    def _span_to_dict(span, i):
```
brettlangdon, line 54: same here, a better name than `i` would be great

tests/snapshots/tests.contrib.aiohttp.test_aiohttp_client.test_trace_query_string.json
```diff
@@ -15,6 +15,7 @@
       "http.status_code": "404",
       "http.status_msg": "NOT FOUND",
       "http.url": "http://localhost:8001/status/200/?k1=v1&k2=v2&k3=v3&k4=v4",
+      "language": "python",
```
brettlangdon, line 18: aside: I wonder if we should add this validation to the test agent itself.

We an have the test agent validate that the first span in every chunk has a language tag which matches the `X-Datadog-Header` tag.

(we still need to update these snapshots, but we can also add the validation elsewhere)

emmettbutler:
I added the Stale tag because this branch has conflicts with the base

brettlangdon:
Give we have another approach, I am hesitant to make changes to the encoder at this time.

Since this hasn't seen any activity for the last 2 months closing for now.

Feel free to resolve conflicts and re-open if this PR is still needed.

PR #4536, opened 2022-11-11
chore(tracing): add span.kind tag 

wconti27:
## Description
Changed general span attributes. Added a span.kind attribute that will be populated for relevant integration spans with one of client, producer, consumer, or server. Spans without the span.kind attribute will be considered internal. 

** This branch was created by branching off of the branch adding component attribute within the Python tracer. Once the component tag change PR is merged, this PR will be changed to be merged into 1.x **

Also providing an Excel file with all integrations, labeled with what integrations were changed, what tests were changed, etc
Document can be found [here](https://docs.google.com/spreadsheets/d/1F8yqzdi9AncVRy--k6Evw235PAudRZ4lYu5HSojXM8A/edit?usp=sharing).

## Motivation
Part of a larger effort to make tag naming more consistent among Datadog tracing libraries. Component is a required attribute for spans created from an integration.

## Design 
Span.Kind tag was set directly within each integrations after initially starting to trace a span if span was tracing a relevant span.kind operation.

Drawbacks of the change are that this PR is huuuuuuuuge. Sorry. 

## Testing strategy
Tested span.kind field by adding assertions on span.kind field to unit/integration tests. Also fixed snapshots to contain newly added span.kind field. 

## Relevant issue(s)
Relevant related issues: 

## Testing strategy
? Detailed above

## Reviewer Checklist
- [x] Title is accurate.


ddtrace/constants.py
```diff
@@ -50,6 +50,14 @@
 
 PID = "process_id"
 
+SPAN_KIND = "span.kind"
+
+# Types of span.kind
+SPAN_CLIENT = "client"
+SPAN_SERVER = "server"
```
mabdinur, line 57: Do we expect ddtrace users to use these constants? If these constants will only used in integrations we should define these values in an internal module (ex: ddtrace/internal/constants.py).

Suggestion: We can create a SpanKind class that defines these constants. This better aligns with how we handle [spantypes](https://github.com/DataDog/dd-trace-py/blob/v1.7.2/ddtrace/ext/__init__.py):

ex:
```
Class SpanKind:
    SPAN_CLIENT = "client"
    SPAN_SERVER = "server"
    SPAN_PRODUCER = "producer"
    SPAN_CONSUMER = "consumer"
```

We can define SpanKind in a public or internal module depending on the expected usage.

wconti27: Though these constants are mainly for our internal use, it's not out of the question IMO that a user would want to utilize these constants when manually instrumenting their own code. I like the class idea!

wconti27: Done. Added the SpanKind class like SpanTypes

ddtrace/contrib/aiohttp/patch.py
```diff
@@ -73,6 +87,9 @@ async def _traced_clientsession_request(aiohttp, pin, func, instance, args, kwar
         # set component tag equal to name of integration
         span.set_tag_str("component", config.aiohttp_client.integration_name)
 
+        # set span.kind tag equal to type of request
+        span.set_tag_str(SPAN_KIND, SpanKind.SERVER)
```
brettlangdon, line 91:
```suggestion
        span.set_tag_str(SPAN_KIND, SpanKind.CLIENT)
```

ddtrace/contrib/aiohttp/patch.py
```diff
@@ -45,13 +47,25 @@ def __init__(self, obj, pin):
...
     async def _create_connection(self, req, *args, **kwargs):
         pin = Pin.get_from(self)
-        with pin.tracer.trace("%s._create_connection" % self.__class__.__name__):
+        with pin.tracer.trace("%s._create_connection" % self.__class__.__name__) as span:
+            # set component tag equal to name of integration
+            span.set_tag("component", config.aiohttp.integration_name)
+
+            # set span.kind tag equal to type of request
+            span.set_tag(SPAN_KIND, SpanKind.SERVER)
+
```
brettlangdon, line 68:
```suggestion
```

I don't think we want these to have span kind ?

ddtrace/contrib/aiohttp/patch.py
```diff
@@ -45,13 +47,25 @@ def __init__(self, obj, pin):
...
     async def connect(self, req, *args, **kwargs):
         pin = Pin.get_from(self)
-        with pin.tracer.trace("%s.connect" % self.__class__.__name__):
+        with pin.tracer.trace("%s.connect" % self.__class__.__name__) as span:
+            # set component tag equal to name of integration
+            span.set_tag("component", config.aiohttp.integration_name)
+
+            # set span.kind tag equal to type of request
+            span.set_tag(SPAN_KIND, SpanKind.SERVER)
+
```
brettlangdon, line 56:
```suggestion
```

I am not sure if this should have span kind

ddtrace/contrib/aiohttp/patch.py
brettlangdon: everything in this file is for the client, and everything in `middlewares` is for the server (as far as I can tell)

ddtrace/contrib/asgi/middleware.py
```diff
@@ -123,6 +125,9 @@ async def __call__(self, scope, receive, send):
         # set component tag equal to name of integration
         span.set_tag_str("component", self.integration_config.integration_name)
 
+        # set span.kind to the type of request being performed
+        span.set_tag_str(SPAN_KIND, SpanKind.SERVER)
```
brettlangdon, line 129: if we set this here, then we probably don't need to set it in the span modifier for any integration which uses this as it's base.

Same goes for component tag since we are using the passed in integration_config (I am not 100% confident on this component part, but seems like it?)

ddtrace/contrib/flask/patch.py
```diff
@@ -135,6 +137,9 @@ def _request_span_modifier(self, span, environ):
 
         span.set_tag_str(FLASK_VERSION, flask_version_str)
 
+        # set span.kind to the type of request being performed
+        span.set_tag_str(SPAN_KIND, SpanKind.SERVER)
```
brettlangdon, line 141: I don't think we need this is WSGI does this for us?

brettlangdon (commented):
sorry I didn't get 100% of the way through the file, but thought I'd leave the comments I do have so far (I made it up to Flask)

ddtrace/contrib/aiohttp/patch.py
wconti27: Yup, good catch, fixed

ddtrace/contrib/aiohttp/patch.py
wconti27: Agreed, removed any span kind tagging unless the span is a request span.

ddtrace/contrib/flask/patch.py
wconti27: Yup, I removed this

ddtrace/contrib/asgi/middleware.py
wconti27: Yeah you're correct, Django and Starlette use this class for middleware spans. Currently we are already using this class for setting the component and span.kind tags. None of that work is happening in the patches for middleware spans specifically.

ddtrace/contrib/requests/connection.py
```diff
@@ -82,6 +84,9 @@ def _wrap_send(func, instance, args, kwargs):
         # set component tag equal to name of integration
         span.set_tag_str("component", config.requests.integration_name)
 
+        # set span.kind to the type of operation being performed
+        span.set_tag_str(SPAN_KIND, SpanKind.SERVER)
```
mabdinur, line 88: Client?

ddtrace/contrib/urllib3/patch.py
```diff
@@ -99,6 +101,9 @@ def _wrap_urlopen(func, instance, args, kwargs):
         # set component tag equal to name of integration
         span.set_tag_str("component", config.urllib3.integration_name)
 
+        # set span.kind to the type of operation being performed
+        span.set_tag_str(SPAN_KIND, SpanKind.SERVER)
```
mabdinur, line 105: Client?

riotfile.py
```diff
@@ -156,22 +156,22 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
...
+        #         "mypy": latest,
+        #         "envier": latest,
+        #         "types-attrs": latest,
+        #         "types-docutils": latest,
+        #         "types-protobuf": latest,
+        #         "types-PyYAML": latest,
+        #         "types-setuptools": latest,
+        #         "types-six": latest,
+        #     },
+        # ),
```
mabdinur, line 174: We should revert this change. This PR should pass mypy checks

riotfile.py
```diff
@@ -186,16 +186,16 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
...
+        # Venv(
+        #     pys=["3"],
+        #     pkgs={"slotscheck": latest},
+        #     venvs=[
+        #         Venv(
+        #             name="slotscheck",
+        #             command="python -m slotscheck -v {cmdargs}",
+        #         ),
+        #     ],
+        # ),
```
mabdinur, line 198: revert

mabdinur (commented):
It is always easier to add a tag than to remove it or update its value (Updating tags is disruptive to users. It can break metrics and facets). We should be very cautious about which spans have the kind tag and which value is set, especially in helper functions that create spans for different operations.

ddtrace/contrib/fastapi/patch.py
```diff
@@ -63,7 +63,10 @@ async def traced_serialize_response(wrapped, instance, args, kwargs):
     if not pin or not pin.enabled:
         return await wrapped(*args, **kwargs)
 
-    with pin.tracer.trace("fastapi.serialize_response"):
+    with pin.tracer.trace("fastapi.serialize_response") as span:
+        # set component tag equal to name of integration
+        span.set_tag_str("component", config.fastapi.integration_name)
+
         return await wrapped(*args, **kwargs)
```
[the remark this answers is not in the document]
wconti27: Asked this same question to @brettlangdon awhile ago when this PR was made and he confirmed that it should be internal. I have it noted in my personal Excel notes for this PR

ddtrace/contrib/httplib/patch.py
```diff
@@ -79,6 +81,9 @@ def _wrap_request(func, instance, args, kwargs):
         # set component tag equal to name of integration
         span.set_tag_str("component", config.httplib.integration_name)
 
+        # set span.kind to the type of operation being performed
+        span.set_tag_str(SPAN_KIND, SpanKind.SERVER)
```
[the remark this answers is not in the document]
wconti27: Correct, fixed this for all http clients that were miscorrectly labeled.

ddtrace/contrib/httplib/patch.py
```diff
@@ -122,6 +127,9 @@ def _wrap_putrequest(func, instance, args, kwargs):
             # set component tag equal to name of integration
             span.set_tag_str("component", config.httplib.integration_name)
 
+            # set span.kind to the type of operation being performed
+            span.set_tag_str(SPAN_KIND, SpanKind.SERVER)
```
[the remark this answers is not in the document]
wconti27: resolved

riotfile.py
wconti27: Reverted this change, didn't mean for this to sneak into the PR

riotfile.py
wconti27: same as above, reverted

ddtrace/ext/__init__.py
```diff
@@ -12,3 +12,10 @@ class SpanTypes(object):
     TEST = "test"
     WEB = "web"
     WORKER = "worker"
+
+
+class SpanKind(object):
+    CLIENT = "client"
+    SERVER = "server"
+    PRODUCER = "producer"
+    CONSUMER = "consumer"
```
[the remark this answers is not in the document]
wconti27: Considering that we don't tag `span.kind` for internal spans, is there any use in adding a constant? Though I guess we can add it to match OTel. @brettlangdon any thoughts?

majorgreys (changes requested):
Following recent discussions on this, I would ask that the type of change here is a chore as no public interface changes are being made, nor are users impacted. If a user can opt in or out of this new behavior, then we'd want it to be flagged as such. But as it is, it is a change to the internal data model.

ddtrace/contrib/aiohttp/patch.py
```diff
@@ -46,13 +48,19 @@ def __init__(self, obj, pin):
 
     async def connect(self, req, *args, **kwargs):
         pin = Pin.get_from(self)
-        with pin.tracer.trace("%s.connect" % self.__class__.__name__):
+        with pin.tracer.trace("%s.connect" % self.__class__.__name__) as span:
+            # set component tag equal to name of integration
+            span.set_tag("component", config.aiohttp.integration_name)
```
majorgreys, line 53: Can the additions of component tag be pulled out?

ddtrace/contrib/aiohttp/patch.py
```diff
@@ -46,13 +48,19 @@ def __init__(self, obj, pin):
...
+
             result = await self.__wrapped__.connect(req, *args, **kwargs)
             return result
 
     async def _create_connection(self, req, *args, **kwargs):
         pin = Pin.get_from(self)
-        with pin.tracer.trace("%s._create_connection" % self.__class__.__name__):
+        with pin.tracer.trace("%s._create_connection" % self.__class__.__name__) as span:
+            # set component tag equal to name of integration
+            span.set_tag("component", config.aiohttp.integration_name)
```
majorgreys, line 62: Same as above

ddtrace/contrib/aiohttp/patch.py
wconti27: yea will add this to another PR

majorgreys (dismissed):

ZStriker19 (approved):

mabdinur (approved):

ZStriker19 (dismissed):

emmettbutler (dismissed):

mabdinur (approved):

ZStriker19 (dismissed):

ddtrace/contrib/flask/patch.py
```diff
@@ -636,6 +638,9 @@ def _set_request_tags(span):
 
         span.set_tag_str(COMPONENT, config.flask.integration_name)
 
+        if span.name.split(".")[-1] == "request":
```
majorgreys, line 641: Would this not always be true for this function?

majorgreys (commented):
one question with the flask change

ddtrace/contrib/flask/patch.py
majorgreys, line 641: not-blocking: We should not be calling this function when the span is not the flask request. This is a result of how the wsgi middleware is used to update the span resource.

majorgreys (approved):

majorgreys (approved):

PR #4539, opened 2022-11-12
chore(dbm): enable dbm for only psycopg2

mabdinur:
## Description

- Makes DBM comment propagation opt in. With this change integrations which use dbapi to execute database queries need to set the following configuration to enable DBM propagation: `_dbapi_propagation_supported=True` (default value is False).
- Limits DBM propagation for psycopg2. 

## Motivation

Currently when `DD_DBM_PROPAGATION_MODE=full` all database integrations which use `ddtrace.contrib.dbapi.TracedCursor` will attempt to add DBM comments to database queries. This PR adds a mechanism to distinguish between integrations which should apply DBM tags and which integrations shouldn't. In essence it keeps the core implementation in dbapi while making DBM propagation opt-in.  

## Reviewer Checklist
- [x] Title is accurate.


majorgreys (commented):
We should add test coverage for this new internal configuration option and ensure that the query executed is unchanged if in the disabled setting.

tests/contrib/psycopg/test_psycopg.py
```diff
@@ -386,6 +386,15 @@ def test_contextmanager_connection(self):
             conn.cursor().execute("""select 'blah'""")
             self.assert_structure(dict(name="postgres.query", service=service))
 
+    @snapshot()
+    @TracerTestCase.run_in_subprocess(env_overrides=dict(DD_DBM_PROPAGATION_MODE="full"))
+    def test_postgres_dbm_propagation(self):
+        """generates snapshot to check whether execution of SQL string sets dbm propagation tag"""
```
brettlangdon, line 392: where do we check that the SQL comment was added? it should be getting it right?

mabdinur: In the dbapi tests we mock the cursor object and assert that the sql comment was set: https://github.com/DataDog/dd-trace-py/blob/1.x/tests/contrib/dbapi/test_dbapi.py#L54. We could do the same here but I wanted this test to:
1. make a real database query
2. assert that the query successed without raising a syntax error
3. the `_dd.dbm_trace_injected` tag was set in the snapshot test

We plan to check the sql comment added in the system test

tests/contrib/dbapi/test_dbapi.py
```diff
@@ -39,9 +51,10 @@ def test_execute_wrapped_is_called_and_returned(self):
             DD_VERSION="v7343437-d7ac743",
         )
     )
-    def test_curser_execute_with_dbm_injection(self):
+    def test_cursor_execute_with_dbm_injection(self):
```
ZStriker19, line 54: don't this test and test at 531 have the same name? https://github.com/DataDog/dd-trace-py/blob/58d1b172ddf6e51bafe9951cf2d23f287fd97114/tests/contrib/dbapi/test_dbapi.py#L531

mabdinur: Yes but they're in different classes. One test is defined in TestTracedCursor and the other is in TestFetchedTracedCursor

ZStriker19 (approved):

tests/contrib/dbapi/test_dbapi.py
```diff
@@ -31,6 +31,18 @@ def test_execute_wrapped_is_called_and_returned(self):
         assert "__result__" == traced_cursor.execute("__query__", "arg_1", kwarg1="kwarg1")
         cursor.execute.assert_called_once_with("__query__", "arg_1", kwarg1="kwarg1")
 
+    @TracerTestCase.run_in_subprocess(env_overrides=dict(DD_DBM_PROPAGATION_MODE="full"))
+    def test_dbm_propagation_not_supported(self):
+        cursor = self.cursor
+        cfg = IntegrationConfig(Config(), "dbapi", service="dbapi_service")
+        # By default _dbm_propagation_supported attribute should not be True.
+        # DBM context propagation should be opt in.
+        assert not getattr(cfg, "_dbm_propagation_supported", False)
```
majorgreys, line 40:
```suggestion
        # By default _dbm_propagation_supported attribute should be false.
        # DBM context propagation should be opt in.
        assert not getattr(cfg, "_dbm_propagation_supported")
```

tests/contrib/dbapi/test_dbapi.py
```diff
@@ -515,9 +528,10 @@ def test_callproc_can_handle_arbitrary_args(self):
             DD_VERSION="v7343437-d7ac743",
         )
     )
-    def test_curser_execute_with_dbm_injection(self):
+    def test_cursor_execute_with_dbm_injection(self):
```
majorgreys, line 531:
```suggestion
    def test_cursor_execute_fetch_with_dbm_injection(self):
```

tests/contrib/psycopg/test_psycopg.py
majorgreys: Is there a similar test for psycopg with the mocked cursor?

majorgreys (commented):
We seem to be using the following phrases interchangeably `dbm.(comment|propagation|injection)`. Do we need all three?

tests/contrib/dbapi/test_dbapi.py
mabdinur: This will raise an AttributeError. `cfg._dbm_propagation_supported` isn't always defined. This is a pattern we use in database integrations to add attributes to the integration config object ([example](https://github.com/DataDog/dd-trace-py/blob/v1.6.0/ddtrace/contrib/psycopg/patch.py#L27)).  Should we do `assert not getattr(cfg, "_dbm_propagation_supported", None)` instead?

tests/contrib/psycopg/test_psycopg.py
mabdinur: Nope just the dbapi test. I'll add it.

Yun-Kim (approved):

tests/contrib/psycopg/test_psycopg.py
```diff
@@ -2,7 +2,7 @@
 import time
 from unittest import skipIf
 
-# 3p
```
mabdinur, line 5: This comment was removed because it is useless and messes with isort linting. Here's where it was introduced: https://github.com/DataDog/dd-trace-py/pull/39/files#diff-41f07c055df91a25d2f3c15e89367264e67c74167119355a7b1f8f9da3bf9546

PR #4547, opened 2022-11-15
ci: add prune workflow runs action

majorgreys:
There is no method for delete workflow runs from GitHub for older workflow runs or workflows that are no longer run. The action added here is scheduled to perform regular deletion of older workflow runs. A manual trigger can also be used to specify the types of workflow runs to be deleted, which is useful for those workflows that have become inactive or have been disabled.

### DIFF
--- a/.github/workflows/prune_workflow.yml
+++ b/.github/workflows/prune_workflow.yml
@@ -9,10 +9,6 @@ on:
         description: 'Number of days.'
         required: true
         default: 30
-      minimum_runs:
-        description: 'The minimum runs to keep for each workflow.'
-        required: true
-        default: 6
       delete_workflow_pattern:
         description: 'The name or filename of the workflow. if not set then it will target all workflows.'
         required: false
--- a/.github/workflows/prune_workflow.yml
+++ b/.github/workflows/prune_workflow.yml
@@ -53,7 +53,6 @@ jobs:
           token: ${{ github.token }}
           repository: ${{ github.repository }}
           retain_days: ${{ inputs.days || 30}}
-          keep_minimum_runs: ${{ inputs.minimum_runs || 6 }}
           delete_workflow_pattern: ${{ inputs.delete_workflow_pattern }}
           delete_workflow_by_state_pattern: ${{ inputs.delete_workflow_by_state_pattern || 'All' }}
           delete_run_by_conclusion_pattern: ${{ inputs.delete_run_by_conclusion_pattern || 'All' }}
--- a/.github/workflows/prune_workflow.yml
+++ b/.github/workflows/prune_workflow.yml
@@ -1,0 +1,62 @@
+name: Prune workflow runs
+on:
+  schedule:
+    # Run monthly, at 00:00 on the 1st day of month.
+    - cron: '0 0 1 * *'
+  workflow_dispatch:
+    inputs:
+      days:
+        description: 'Number of days.'
+        required: true
+        default: 30
+      minimum_runs:
+        description: 'The minimum runs to keep for each workflow.'
+        required: true
+        default: 6
+      delete_workflow_pattern:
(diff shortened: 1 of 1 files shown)


.github/workflows/prune_workflow.yml
```diff
@@ -0,0 +1,60 @@
+name: Prune workflow runs
+on:
+  schedule:
+    # Run monthly, at 00:00 on the 1st day of month.
+    - cron: '0 0 1 * *'
```
brettlangdon, line 5: daily?

.github/workflows/prune_workflow.yml
```diff
@@ -0,0 +1,60 @@
...
+
+jobs:
+  del_runs:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Delete workflow runs
+        uses: Mattraks/delete-workflow-runs@v2
+        with:
+          token: ${{ github.token }}
+          repository: ${{ github.repository }}
```
brettlangdon, line 54:
```suggestion
```

.github/workflows/prune_workflow.yml
```diff
@@ -0,0 +1,60 @@
...
+          retain_days: ${{ inputs.days || 30}}
+          keep_minimum_runs: ${{ inputs.minimum_runs || 6 }}
+          delete_workflow_pattern: ${{ inputs.delete_workflow_pattern }}
+          delete_workflow_by_state_pattern: ${{ inputs.delete_workflow_by_state_pattern || 'All' }}
+          delete_run_by_conclusion_pattern: ${{ inputs.delete_run_by_conclusion_pattern || 'All' }}
+          dry_run: ${{ inputs.dry_run }}
```
brettlangdon, line 60: if you hard code this to `true` for now, we can probably merge and make sure it runs properly before enabling the destructive part ?

.github/workflows/prune_workflow.yml
majorgreys: This is required: https://github.com/marketplace/actions/delete-workflow-runs#2-repository

.github/workflows/prune_workflow.yml
majorgreys: Good idea!

.github/workflows/prune_workflow.yml
majorgreys: Daily seems too aggressive as these runs would be consulted for PRs being reviewed. A month seems like a fair start for when to perform pruning. If we need to prune sooner, we run manually.

.github/workflows/prune_workflow.yml
brettlangdon: but the default is `${{ github.repository }}` so it shouldn't matter (it really doesn't matter either way, feel free to ignore)

.github/workflows/prune_workflow.yml
```diff
@@ -0,0 +1,62 @@
...
+  del_runs:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Delete workflow runs
+        uses: Mattraks/delete-workflow-runs@v2
+        with:
+          token: ${{ github.token }}
+          repository: ${{ github.repository }}
+          retain_days: ${{ inputs.days || 30}}
+          keep_minimum_runs: ${{ inputs.minimum_runs || 6 }}
```
brettlangdon, line 56:
```suggestion
```

I think we can get rid of this, otherwise old workflows won't age out and get removed automatically (which might be ok?)

brettlangdon (approved):

brettlangdon (approved):

Kyle-Verhoog (approved):
Is the motivation just that there's noise when looking at the actions page?

brettlangdon:
> Is the motivation just that there's noise when looking at the actions page?

yes! we cannot delete old/unused workflows anymore, this is one way to make unrelated workflows "fall off" the actions page.

PR #4558, opened 2022-11-16
feat(iast): add env var DD_IAST_WEAK_HASH_ALGORITHMS

gnufede:
## Description
The list of weak hashing algorithms to report is configurable through the env var `DD_IAST_WEAK_HASH_ALGORITHMS`

## Motivation
This allows customers to configure which algorithms they want reported.

## Design 
`DD_IAST_WEAK_HASH_ALGORITHMS` is a comma-separated list of weak hashing algorithms, if unset, it will default to the ones that we detect currently: MD5 and SHA1

## Testing strategy
Unit tests with the env var missing, configured to each single one of the algorithms, both and other invalid algorithms.

## Reviewer Checklist
- [ ] Title is accurate.


ddtrace/appsec/iast/_patch.py
```diff
@@ -1,11 +1,15 @@
 import ctypes
 import gc
+from typing import Any
```
avara1986, line 3: add TYPE_CHECKING:


```python
if TYPE_CHECKING:  # pragma: no cover
    from typing import Any
    from typing import Dict
```

juanjux (dismissed):

ddtrace/appsec/iast/_patch.py
gnufede: IIUC that makes sense [if you're importing from an expensive module](https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING), but I don't see the point here as we're importing TYPE_CHECKING from typing as well.

avara1986 (approved):

christophe-papazian (approved):

PR #4590, opened 2022-11-18
feat(settings): fuzzy env variable matching

paullegranddc:
## Description

This PR replaces `os.environ.get` and `os.getenv` calls with a proxy object that checks if the env variable is not set wether another variable with a close name is set. 

## Checklist

## Reviewer Checklist
- [ ] Title is accurate.

### DIFF
--- a/ddtrace/settings/matching.py
+++ b/ddtrace/settings/matching.py
@@ -112,7 +112,7 @@ class FuzzyEnvMatcher(object):
             max_dist = min(max((len(key) - len(self.SCANNED_PREFIX)) / 3, 1), 2)
             match = self.matcher.match_damreau_levhenstein(key, max_dist)
             if match is not None:
-                log.warning("Env variable %s not recognized, did you mean %s", key, match.match)
+                log.warning("Env variable %s not recognized, did you mean %s", match.match, key)
         return default
 
 
--- a/ddtrace/settings/matching.py
+++ b/ddtrace/settings/matching.py
@@ -54,7 +54,7 @@ class Trie(object):
             current_node = current_node.branches[c]
         current_node.match = s
 
-    def _match_levhenstein(self, s, idx, current_dist, max_dist=None):
+    def _match_damreau_levhenstein(self, s, idx, current_dist, max_dist=None):
         # type: (str, int, int, Optional[int]) -> Optional[SearchResult]
         best_match = None
         if max_dist is not None and current_dist > max_dist:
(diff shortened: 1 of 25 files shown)


ddtrace/settings/matching.py
```diff
@@ -0,0 +1,123 @@
+import os
```
brettlangdon, line 1: We will want this to be a private module, maybe something in `ddtrace.internal`, or just renaming as `_matching.py` ?

ddtrace/settings/matching.py
```diff
@@ -0,0 +1,123 @@
...
+        return self._match_damreau_levhenstein(s, 0, 0, max_dist)
+
+
+class FuzzyEnvMatcher(object):
+    SCANNED_PREFIX = "DD_"
+
+    def __init__(self):
+        self.matcher = Trie()
+        for env_name in os.environ.keys():
+            if env_name.startswith(self.SCANNED_PREFIX):
```
brettlangdon, line 101: so we'll only be able to fuzzy match if they have properly started with `DD_`?

ddtrace/settings/matching.py
```diff
@@ -0,0 +1,123 @@
...
+        # type: (str, Optional[str]) -> Optional[str]
+        if key in os.environ:
+            return os.environ.get(key, default)
+        if key.startswith(self.SCANNED_PREFIX) and key not in self.seen:
+            self.seen.add(key)
+            # We bound the number errors we look for between min 1 error and max 2 errors
+            max_dist = min(max((len(key) - len(self.SCANNED_PREFIX)) / 3, 1), 2)
+            match = self.matcher.match_damreau_levhenstein(key, max_dist)
+            if match is not None:
+                log.warning("Env variable %s not recognized, did you mean %s", key, match.match)
```
brettlangdon, line 115:
```suggestion
                log.warning("Env variable %s not recognized, did you mean %s", key, match.match)
```

Then we'll get quotes around them.

setup.cfg
```diff
@@ -2,7 +2,7 @@
 universal=1
 
 [codespell]
-skip = *.json,*.cpp,*.c,.riot,.tox,.mypy_cache,.git,*ddtrace/vendor
+skip = *.json,*.cpp,*.c,.riot,.tox,.mypy_cache,.git,*ddtrace/vendor,tests/settings/test_levhenstein_distance.py
```
brettlangdon, line 5: why do we need to skip the whole file here?

brettlangdon (commented):
this is cool, let some initial minor thoughts/comments

setup.cfg
paullegranddc: The codespell has an errror on of of the test input, and from a quick look it looks like codespell doesn't support inline ignore https://github.com/codespell-project/codespell/issues/1212

ddtrace/settings/matching.py
paullegranddc: I was thinking that `DD_` was probably something people wouldn't mistype (or the mistake is quite obvious) so it's not really worth taking into account variables without the prefix.

Plus if a user forgets to put DD_ (like `TRACE_ENABLED=true`) the matching method doesn't work that well since the distance to the actual env is at least 3.

ddtrace/settings/matching.py
```diff
@@ -0,0 +1,123 @@
...
+    if a is None:
+        return b
+    elif b is None:
+        return a
+    if a.cost > b.cost:
+        return b
+    return a
+
+
+class Trie(object):
```
P403n1x87, line 35: Can we leverage the `difflib` module here somehow?

paullegranddc: I've taken a look at the `difflib` module, and it's based on another algorithm (substring matching).
I think the one currently implemented works better to correct typos for two reasons:
* The algorithm difflib uses returns a score that depends on the length of the strings matched, whereas we mostly care about the absolute number of differences, and it doesn't consider character swap as a single change.  
* The current implementation should be more efficient on long strings because we can prune results early by giving a max distance of 1 or 2. That makes the matching mostly O(len(text)) whereas difflib is O(len(text) * len(pattern) * nb(pattern))

brettlangdon:
This is a cool POC, but closing the PR for now until we are able to come back to it!

PR #4609, opened 2022-11-22
chore(tracing): change error.msg span tag to error.message

wconti27:
## Description
This PR changes the span tag: `error.msg` to `error.message`. The PR also changed the constant `ERROR_MSG` to `ERROR_MESSAGE` within the repo for consistency reasons.

## Motivation
Part of a larger effort to make tag naming more consistent among tracer libraries. 

## Reviewer Checklist
- [ ] Title is accurate.


mabdinur (commented):
We recently add 2 new snapshot tests to the wsgi test suite. Those files will need to be updated. 

Aside from that we should be good to go 🚢

mabdinur (approved):

ddtrace/constants.py
```diff
@@ -44,7 +44,7 @@
 MANUAL_KEEP_KEY = "manual.keep"
 
 
-ERROR_MSG = "error.msg"  # a string representing the error message
+ERROR_MESSAGE = "error.message"  # a string representing the error message
```
brettlangdon, line 47: We cannot change this name since it is part of the public API.

We need to keep the variable name as `ERROR_MSG`... or we need to have it aliased.

```python
ERROR_MSG = ERROR_MESSAGE  # Add deprecation warning?
```

brettlangdon (changes requested):

ddtrace/constants.py
mabdinur, line 47: Add these imports: 
```
from ddtrace.vendor.debtcollector import removals
from ddtrace.internal.utils.deprecations import DDTraceDeprecationWarning
```

Replace line 47 with:
```suggestion
ERROR_MESSAGE = "error.message"  # a string representing the error message
@removals.remove(message="Use ERROR_MESSAGE instead." category=DDTraceDeprecationWarning)
ERROR_MSG  = ERROR_MESSAGE
```

Test the deprecation warning locally to make sure it's logged when `ERROR_MSG ` and the message is formatted correctly.

Here's how you can generate the deprecation warning:

1. Make sure ddtrace is installed locally and is an editable package. If not run the following command in a virtual environment:
 - pip install -e  [path to dd-trace-py/]

2. Generate the deprecation warning by accessing ERROR_MSG. Run:
 - python -c "from ddtrace.constants import ERROR_MSG; print(ERROR_MSG)"

wconti27: I was able to do this in a different way since you can't add a decorator in Python unless its a class or function. Instead just created another simple removal function that deprecates a constant and called it within constants.py as a normal function.

wconti27: Also I would assume that those test failures for the agent/testagent are valid and should be fixed correct? As in we are now expecting the deprecation warning when building?

ddtrace/constants.py
```diff
@@ -44,7 +48,12 @@
...
+ERROR_MSG = ERROR_MESSAGE
+removals.removed_constant(
```
brettlangdon, line 53: ... does this function/helper exist? I cannot find it in the debtcollector vendored code

wconti27: See below

ddtrace/constants.py
mabdinur: We need to fix those tests. We shouldn't use the deprecated constant internally.

ddtrace/vendor/debtcollector/removals.py
```diff
@@ -57,14 +57,23 @@ class removed_property(object):
...
-        'delete': "Deleting the '%s' property is deprecated",
+        "set": "Setting the '%s' property is deprecated",
+        "get": "Reading the '%s' property is deprecated",
+        "delete": "Deleting the '%s' property is deprecated",
     }
 
-    def __init__(self, fget=None, fset=None, fdel=None, doc=None,
-                 stacklevel=3, category=DeprecationWarning,
-                 version=None, removal_version=None, message=None):
+    def __init__(
```
brettlangdon, line 65: we should avoid all of these formatting changes

ddtrace/vendor/debtcollector/removals.py
```diff
@@ -319,16 +315,44 @@ def removed_module(module, replacement=None, message=None,
...
-                                          removal_version=removal_version)
-    _utils.deprecation(out_message,
-                       stacklevel=stacklevel, category=category)
+    out_message = _utils.generate_message(
+        prefix, postfix=postfix, message=message, version=version, removal_version=removal_version
+    )
+    _utils.deprecation(out_message, stacklevel=stacklevel, category=category)
+
+
+def removed_constant(
```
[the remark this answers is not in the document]
brettlangdon: is this going to show the deprecation warning always on startup/import of `ddtrace.constants`?

if so, it would be better to only show the message if someone imports/uses `ddtrace.constants.ERROR_MSG`

I cannot remember how we did that in the past 🤔 

I think @P403n1x87 might know (I think they did it before)

ddtrace/vendor/debtcollector/removals.py
wconti27: Fixed

ddtrace/vendor/debtcollector/removals.py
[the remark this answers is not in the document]
P403n1x87: I think we did it with PEP 562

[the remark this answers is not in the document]
wconti27: @brettlangdon Yup, I did something similar to PEP 562 that can be used to deprecate a variable / constant. should be all set

ddtrace/constants.py
wconti27: @mabdinur This is fixed and tested, resolving

ddtrace/vendor/debtcollector/removals.py
```diff
@@ -300,8 +300,8 @@ def removed_module(module, replacement=None, message=None,
                    version=None, removal_version=None, stacklevel=3,
                    category=None):
     """Helper to be called inside a module to emit a deprecation warning
-
-    :param str replacment: A location (or information about) of any potential
+    
+    :param str replacement: A location (or information about) of any potential
```
brettlangdon, line 304: unrelated change?

brettlangdon (approved):

mabdinur (approved):

PR #4613, opened 2022-11-23
feat(asm): add IP Blocking capabilities

juanjux:
## Description
This adds IP blocking for the Python agent. It also sets the base for the other blocking features (UserId blocking and Suspicious Request blockins). Also adds a barebones ASM context manager.

## Checklist

## Reviewer Checklist

- [ ] Title is accurate.


juanjux:
Note: rebased against the libddwaf 1.6.0 branch so it's easier to review, but once that's merged it should go to 1.x obviously.

ddtrace/tracer.py
```diff
@@ -102,7 +132,8 @@ def _default_span_processors_factory(
     single_span_sampling_rules,  # type: List[SpanSamplingRule]
     agent_url,  # type: str
 ):
-    # type: (...) -> List[SpanProcessor]
+    # type: (...) -> Tuple[List[SpanProcessor], Optional[Any]]
+    # FIXME: type should be AppsecSpanProcessor but we have a cyclic import here
```
avara1986, line 136: the `#FIXME` is not needed anymore, right?

ddtrace/tracer.py
```diff
@@ -62,6 +55,21 @@
...
+if TYPE_CHECKING:  # pragma: no cover
+    from typing import Any
+    from typing import Dict
+    from typing import List
+    from typing import Optional
+    from typing import Set
+    from typing import Union
+    from typing import Tuple
+
+    # from .appsec.processor import AppSecSpanProcessor
```
avara1986, line 67: Remove commented line :upside_down_face:

ddtrace/appsec/processor.py
```diff
@@ -204,9 +207,46 @@ def __attrs_post_init__(self):
...
+        if peer_ip or headers:
+            ip = trace_utils._get_request_header_client_ip(span, headers, peer_ip, headers_case_sensitive)
+            # Save the IP and headers in the context so the retrieval can be skipped later
+            _context.set_item("http.request.remote_ip", ip, span=span)
+            if ip and self._is_needed(_Addresses.HTTP_CLIENT_IP):
+                data = {_Addresses.HTTP_CLIENT_IP: ip}
+                ddwaf_result = self._run_ddwaf(data)
+
+                if ddwaf_result and ddwaf_result.actions:
+                    if "block" in ddwaf_result.actions:
```
gnufede, line 237: nit: looks like these two if could be merged

ddtrace/appsec/utils.py
```diff
@@ -1,25 +1,92 @@
...
+
+
 def _appsec_rc_features_is_enabled():
     # type: () -> bool
     if asbool(os.environ.get("DD_REMOTE_CONFIGURATION_ENABLED", "true")):
         return APPSEC_ENV not in os.environ
     return False
 
 
+def to_bytes_py2(n, length=1, byteorder="big", signed=False):
```
gnufede, line 22: nit: Maybe move this to `ddtrace.internal.compat` and add some comments on what does it do?

ddtrace/tracer.py
juanjux: The FIXME is about the typing with Any above.

ddtrace/appsec/processor.py
juanjux: Yup. Originally it has an else for debugging but now that it doesn't have it they can be joined.

ddtrace/appsec/utils.py
juanjux: Yup. Good call.

ddtrace/contrib/trace_utils.py
```diff
@@ -433,15 +436,17 @@ def set_http_meta(
     :param url: the HTTP URL
     :param status_code: the HTTP status code
     :param status_msg: the HTTP status message
-    :param query: the HTTP query part of the URI as a string
-    :param parsed_query: the HTTP query part of the URI as parsed by the framework and forwarded to the user code
+    :param query: the HTTP query part of the_context.get_item("http.request.headers", span=span), URI as a string
+    :param parsed_query: the HTTP query part_context.get_item("http.request.headers", span=span), of the URI as parsed
```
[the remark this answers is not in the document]
juanjux: Interesting... fixed, thanks.

ddtrace/tracer.py
juanjux: Done!

avara1986 (approved):

gnufede (dismissed):

ddtrace/appsec/processor.py
```diff
@@ -204,9 +207,45 @@ def __attrs_post_init__(self):
...
-    def on_span_start(self, span):
-        # type: (Span) -> None
-        pass
+    def update_rules(self, new_rules):
+        # type: (List[Dict[str, Any]]) -> None
+        self._ddwaf.update_rules(new_rules)
+
+    def on_span_start(self, span, *args, **kwargs):
+        # type: (Span, Any, Any) -> None
+        peer_ip = kwargs.get("peer_ip")
```
brettlangdon, line 216: what are the other options you can think of for us to get this information here?

I don't really like that we are extending the public API for creating traces to meet an internal only need.


I am assuming we are not using span tags because this is information we don't want to collect in tags? (would always have to be removed before encoding the span whether we had ASM enabled or not?)

brettlangdon (changes requested):

ddtrace/appsec/processor.py
juanjux: PR updated, I switched to pass the early data using contextvars until the new context API allows for passing data without a span, please take a look.

scripts/run-tox-scenario
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+
```
brettlangdon, line 2: unnecessary change

riotfile.py
```diff
@@ -178,7 +178,7 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
             venvs=[
                 Venv(
                     name="codespell",
-                    command='codespell --skip="ddwaf.h" ddtrace/ tests/',
+                    command='codespell --skip="ddwaf.h" ddtrace/ tests/ -I docs/spelling_wordlist.txt',
```
brettlangdon, line 181: can you introduce this change in it's own PR?

ddtrace/internal/processor/trace.py
```diff
@@ -101,7 +101,7 @@ class TopLevelSpanProcessor(SpanProcessor):
 
     """
 
-    def on_span_start(self, _):
+    def on_span_start(self, span):
```
brettlangdon, line 104: unnecessary change?

ddtrace/internal/remoteconfig/__init__.py
```diff
@@ -63,6 +63,13 @@ def register(cls, product, handler):
         except Exception:
             log.warning("error starting the RCM client", exc_info=True)
 
+    @classmethod
+    def unregister(cls, product):
+        try:
+            cls._worker._client.unregister_product(product)
```
brettlangdon, line 69: how does this relate to the IP blocking change?

would it be cleaner to introduce a PR just for registering/unregistering products with RCM?

ddtrace/tracer.py
```diff
@@ -808,13 +829,13 @@ def trace(self, name, service=None, resource=None, span_type=None):
             parent2.finish()
         """
         return self.start_span(
-            name,
+            name=name,
```
brettlangdon, line 832:
```suggestion
            name,
```

ddtrace/tracer.py
```diff
@@ -808,13 +829,13 @@ def trace(self, name, service=None, resource=None, span_type=None):
...
             child_of=self.context_provider.active(),
             service=service,
             resource=resource,
             span_type=span_type,
             activate=True,
-        )
+        )  # type: ignore[misc]
```
brettlangdon, line 838:
```suggestion
        )
```

ddtrace/tracer.py
```diff
@@ -224,9 +243,11 @@ def __init__(
         self._partial_flush_enabled = asbool(os.getenv("DD_TRACE_PARTIAL_FLUSH_ENABLED", default=True))
         self._partial_flush_min_spans = int(os.getenv("DD_TRACE_PARTIAL_FLUSH_MIN_SPANS", default=500))
         self._appsec_enabled = config._appsec_enabled
+        # Direct link to the appsec processor
+        self._appsec_processor = None
         self._iast_enabled = config._iast_enabled
 
-        self._span_processors = _default_span_processors_factory(
+        self._span_processors, self._appsec_processor = _default_span_processors_factory(
```
brettlangdon, line 250: is `self._appsec_processor` only used by tests?

ddtrace/debugging/_function/discovery.py
```diff
@@ -9,7 +9,7 @@
 try:
     from collections.abc import Iterator
 except ImportError:
-    from collections import Iterator  # type: ignore[attr-defined,no-redef]
+    from collections import Iterator  # type: ignore
```
brettlangdon, line 12: is this change needed?

brettlangdon (commented):
Looking over the PR it seems like there are a handful of things changing that can probably be broken into their own PRs to make review clearer. (I've found myself distracted reviewing things that seem unrelated, I haven't re-reviewed the contextvar change yet)


Also, is there a reason why `MANIFEST.in` is being added? (I couldn't comment on the file directly)

riotfile.py
juanjux: Yup. Removed.

ddtrace/internal/processor/trace.py
juanjux: Yes, was needed when there were more args but not anymore.

ddtrace/internal/remoteconfig/__init__.py
juanjux: I will let @avara1986 comment on this but I think is needed since we register for the first time a RCM feature and thus have to get these functions.

ddtrace/tracer.py
juanjux: Fixed.

ddtrace/debugging/_function/discovery.py
juanjux: No idea at what point it went away... restored.

ddtrace/contrib/trace_utils.py
```diff
@@ -156,15 +156,17 @@ def _get_request_header_user_agent(headers, headers_are_case_sensitive=False):
     :param headers: A dict of http headers to be stored in the span
     :type headers: dict or list
     """
+    user_agent = None
+
     for key_pattern in USER_AGENT_PATTERNS:
-        if not headers_are_case_sensitive:
-            user_agent = headers.get(key_pattern)
-        else:
+        if headers_are_case_sensitive or type(headers) == dict:
```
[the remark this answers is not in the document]
juanjux: Not strictly, will remove and do a separate PR.

ddtrace/appsec/processor.py
```diff
@@ -204,9 +211,47 @@ def __attrs_post_init__(self):
...
         self._mark_needed(_Addresses.SERVER_RESPONSE_HEADERS_NO_COOKIES)
 
-    def on_span_start(self, span):
-        # type: (Span) -> None
-        pass
+    def update_rules(self, new_rules):
+        # type: (List[Dict[str, Any]]) -> None
+        self._ddwaf.update_rules(new_rules)
+
+    def on_span_start(self, span, *args, **kwargs):
```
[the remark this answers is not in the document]
juanjux: Yup, missed that one when manually reverting, thanks.

ddtrace/tracer.py
juanjux: It's also used in `remoteconfiguration.py`.

ddtrace/internal/remoteconfig/__init__.py
avara1986: @brettlangdon we add this feature in this PR because IP blocking is the only Remote Configuration product that needs the unregister function

ddtrace/tracer.py
```diff
@@ -91,6 +97,28 @@
 AnyCallable = TypeVar("AnyCallable", bound=Callable)
 
 
+def _start_appsec_processor():
+    # type: () -> Optional[Any]
+    # FIXME: type should be AppsecSpanProcessor but we have a cyclic import here
```
brettlangdon, line 102: you might be able to do `'AppsecSpanProcessor'` (a string) and have it work... but I am not sure since it is imported inside the function 🤔

tests/snapshots/tests.contrib.django.test_django_snapshots.test_appsec_enabled.json
```diff
@@ -19,12 +19,14 @@
       "django.response.class": "django.http.response.HttpResponse",
       "django.user.is_authenticated": "False",
       "django.view": "tests.contrib.django.views.index",
+      "http.client_ip": "",
```
brettlangdon, line 22: should this be an empty string?

should we not set it if we don't have a value?

ddtrace/contrib/flask/patch.py
```diff
@@ -113,28 +120,14 @@ def _traced_start_response(self, start_response, req_span, app_span, status_code
...
-        # We will override this below in `traced_dispatch_request` when we have a `
-        # RequestContext` and possibly a url rule
-        span.resource = u" ".join((request.method, request.path))
-
-        span.set_tag(SPAN_MEASURED_KEY)
-        # set analytics sample rate with global config enabled
-        sample_rate = config.flask.get_analytics_sample_rate(use_global_config=True)
-        if sample_rate is not None:
-            span.set_tag(ANALYTICS_SAMPLE_RATE_KEY, sample_rate)
+        if config._appsec_enabled and _context.get_item("http.request.blocked", span=req_span):
```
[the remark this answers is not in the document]
juanjux: Yes. Christophe found a better place to block on flask on his branch and I'll merge it on monday.

ddtrace/contrib/flask/patch.py
```diff
@@ -560,15 +576,22 @@ def _traced_request(pin, wrapped, instance, args, kwargs):
...
+        try:
+            with pin.tracer.trace(
+                ".".join(("flask", name)),
+                service=trace_utils.int_service(pin, config.flask, pin),
+            ) as request_span:
+                # set component tag equal to name of integration
+                request_span.set_tag_str("component", config.flask.integration_name)
+
+                request_span._ignore_exception(werkzeug.exceptions.NotFound)
+                return wrapped(*args, **kwargs)
```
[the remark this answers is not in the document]
juanjux: Ditto.

ddtrace/contrib/flask/patch.py
```diff
@@ -560,15 +576,22 @@ def _traced_request(pin, wrapped, instance, args, kwargs):
...
+        finally:
+            _reset_contextvars()
```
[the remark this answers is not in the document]
juanjux: Yep. It's in the appsec package so it's namespaced but since we usually do symbolic imports it'll be better to rename it to get a clearer picture that is an ASM thing when reading the code. Didn't want to complicate it much since it's supossed to be a temporary thing but I'll take your suggestion.

Commited.

ddtrace/contrib/django/patch.py
```diff
@@ -345,11 +357,17 @@ def traced_get_response(django, pin, func, instance, args, kwargs):
...
+        if _context.get_item("http.request.blocked", span=span):
+            func = HttpResponseForbidden
+            args = [appsec_utils._get_blocked_template(request_headers.get("Accept"))]
+
         try:
             response = func(*args, **kwargs)
             return response
         finally:
             # DEV: Always set these tags, this is where `span.resource` is set
+            _reset_contextvars()
```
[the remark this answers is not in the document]
juanjux: I didn't want to call the response object instantiation before `func` in case there were some side effects to it (now or in the future) but your code look clearer and they don't seem to be any so go with your suggestion.

ddtrace/tracer.py
juanjux: Yep, even with a string it fails without having the import before (which would case the cycle).

tests/snapshots/tests.contrib.django.test_django_snapshots.test_appsec_enabled.json
juanjux: Maybe, but we're already returning the empty string for this tag in previous versions (probably went for empty because there are other empty strings in the jsons) so I don't want to change it now.

ddtrace/internal/remoteconfig/client.py
```diff
@@ -33,7 +33,7 @@
     from typing import Tuple
     from typing import Union
 
-    ProductCallback = Callable[[Optional["ConfigMetadata"], Union[Mapping[str, Any], bool, None]], None]
+    ProductCallback = Optional[Callable[[Optional["ConfigMetadata"], Union[Mapping[str, Any], bool, None]], None]]
```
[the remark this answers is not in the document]
juanjux: Looks reasonable, @avara1986

ddtrace/appsec/_remoteconfiguration.py
```diff
@@ -29,6 +37,40 @@ def enable_appsec_rc():
...
+
+        if not (APPSEC_ENV not in os.environ and rc_appsec_enabled is True) and (
+            asbool(os.environ.get(APPSEC_ENV)) is False or rc_appsec_enabled is False
+        ):
+            _appsec_enabled = False
+            RemoteConfig.unregister(ASM_DATA_PRODUCT)
+        else:
+            RemoteConfig.register(ASM_DATA_PRODUCT, appsec_rc_reload_features(tracer))
+
+        tracer.configure(appsec_enabled=_appsec_enabled)
```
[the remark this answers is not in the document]
juanjux: @avara1986, WDYT?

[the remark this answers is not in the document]
juanjux: Personally I think always registering the `AppsecSpanProcessor` even disabled would make tests a lot cleaner with lots less reconfigurations... but on the other hand I'm not sure if it makes sense to add a processor that does nothing and will be iterated. Maybe we could add an `enabled` member to one of the parent classes and use it for all processors (and would also allow for a processor to be checked for "enablement" without knowing the specific type but I agree this is something for another PR (and discussion), do you agree if we go with this for this version? (there are other tracer.configure calls in the codebase outside of this one so it's not exactly new).

ddtrace/contrib/trace_utils.py
```diff
@@ -178,10 +180,11 @@ def _get_request_header_client_ip(span, headers, peer_ip=None, headers_are_case_
     global _USED_IP_HEADER
 
     def get_header_value(key):  # type: (str) -> Optional[str]
-        if not headers_are_case_sensitive:
-            return headers.get(key)
-
-        return _get_header_value_case_insensitive(headers, key)
+        # Sanic client will use a dict for headers even if under Django headers are case
```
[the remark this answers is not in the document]
juanjux: What the comment states, it uses a dict for headers thus some tests from this PR will break without this change but I'll split this change to another micro PR.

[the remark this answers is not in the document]
juanjux: Update: did the separate PR and removed from this one: https://github.com/DataDog/dd-trace-py/pull/4947

ddtrace/contrib/trace_utils.py
```diff
@@ -442,6 +445,7 @@ def set_http_meta(
     :param request_path_params: the parameters of the HTTP URL as set by the framework: /posts/<id:int> would give us
          { "id": <int_value> }
     """
+
```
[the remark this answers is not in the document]
juanjux: Fixed, thanks.

ddtrace/internal/remoteconfig/client.py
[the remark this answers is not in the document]
avara1986: We added Optional because `ProductCallback` is optional in all cases we  return a `ProductCallback` or use it as parameter. It's a semantic debate ;) we could use:

```
ProductCallback = ...
OptionalProductCallback = Optional[ProductCallback]
```

ddtrace/appsec/utils.py
```diff
@@ -13,13 +17,59 @@ def _appsec_rc_features_is_enabled():
...
+
+    if need_html_template:
+        template_path = os.getenv("DD_APPSEC_HTTP_BLOCKED_TEMPLATE_HTML")
+    else:
+        template_path = os.getenv("DD_APPSEC_HTTP_BLOCKED_TEMPLATE_JSON")
+
+    if template_path and os.path.exists(template_path) and os.path.isfile(template_path):
+        try:
+            with open(template_path, "r") as template_file:
+                return template_file.read()
```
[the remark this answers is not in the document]
juanjux: It's from a user-defined env var but catching won't hurt so I added it.

ddtrace/contrib/trace_utils.py
```diff
@@ -156,15 +156,17 @@ def _get_request_header_user_agent(headers, headers_are_case_sensitive=False):
     :param headers: A dict of http headers to be stored in the span
     :type headers: dict or list
     """
+    user_agent = None
+
     for key_pattern in USER_AGENT_PATTERNS:
-        if not headers_are_case_sensitive:
-            user_agent = headers.get(key_pattern)
-        else:
+        if headers_are_case_sensitive:
```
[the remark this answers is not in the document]
juanjux: Change undoed and moved to the other PR.

ddtrace/appsec/context_vars.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+# FIXME: remove these and use the new context API once implemented and allowing
+# contexts without spans
+
+_DD_EARLY_IP_CONTEXTVAR = contextvars.ContextVar("datadog_early_ip_contextvar", default=None)
+_DD_EARLY_HEADERS_CONTEXTVAR = contextvars.ContextVar("datadog_early_headers_contextvar", default=None)
+_DD_EARLY_HEADERS_CASE_SENSITIVE_CONTEXTVAR = contextvars.ContextVar(
+    "datadog_early_headers_casesensitive_contextvar", default=False
+)
```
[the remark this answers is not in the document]
juanjux: I kept the functions separate in case we need them in the future (I could see some blocking based on user agent or other header values as a future feature) but took your first idea and created the context manager with all the parameters (then it individually calls each function inside).

avara1986 (approved):

avara1986 (approved):

ddtrace/appsec/processor.py
juanjux: @brettlangdon could you remove the required change from here if you are ok with the new context manager for contextvars?

avara1986 (dismissed):

ddtrace/contrib/wsgi/wsgi.py
```diff
@@ -115,45 +117,44 @@ def __call__(self, environ, start_response):
...
 
-        self._response_span_modifier(resp_span, result)
+            # set component tag equal to name of integration
+            resp_span.set_tag_str("component", self._config.integration_name)
+            self._response_span_modifier(resp_span, result)
 
-        return _TracedIterable(iter(result), resp_span, req_span)
+            return _TracedIterable(iter(result), resp_span, req_span)
+        finally:
+            _asm_context.reset()
```
brettlangdon, line 157: are we able to push enough context from the wsgi environ to make a blocking decision?

if someone wrapped their app with the wsgi middleware then they could prevent the request from even calling their web app in the first place.

("nice to have"? feel free to resolve with no action)

tests/appsec/test_processor.py
```diff
@@ -200,6 +202,87 @@ def test_appsec_body_no_collection_snapshot(tracer):
...
+            assert "triggers" in json.loads(span.get_tag(APPSEC_JSON))
+            assert _context.get_item("http.request.remote_ip", span) == "8.8.4.4"
+            assert _context.get_item("http.request.blocked", span)
+            assert "block" in _context.get_item("http.request.waf_actions", span)
+
+
+def test_ip_not_block(tracer):
+    with override_env(dict(DD_APPSEC_RULES=RULES_GOOD_PATH)), override_global_config(dict(_appsec_enabled=True)):
+        _enable_appsec(tracer)
+        with _asm_context.asm_request_context_manager("8.8.8.4", {}):
```
brettlangdon, line 224:
```suggestion
        with _asm_context.asm_request_context_manager("1.1.1.1", {}):
```

or please add a comment or something... I stared at this and `test_ip_block` for about 5 minutes before I was able to spot the difference between the two tests.

or maybe have some constants at top like `BLOCKED_IP = "8.8.4.4"; GOOD_IP = "8.8.8.4"` so make these tests more legible.

tests/appsec/test_processor.py
```diff
@@ -200,6 +202,87 @@ def test_appsec_body_no_collection_snapshot(tracer):
...
+
+
+def test_ip_update_rules_expired_no_block(tracer):
+    with override_global_config(dict(_appsec_enabled=True)):
+        _enable_appsec(tracer)
+        tracer._appsec_processor.update_rules(
+            [
+                {
+                    "data": [
+                        {"expiration": 1662804872, "value": "8.8.4.4"},
```
brettlangdon, line 268: I assume this is a date in the past? a comment might help

tests/appsec/test_remoteconfiguration.py
```diff
@@ -1,16 +1,30 @@
...
 from ddtrace.ext import SpanTypes
+from ddtrace.internal import _context
+from ddtrace.internal.remoteconfig import RemoteConfig
+from tests.appsec.test_processor import Config
 from tests.utils import override_env
 
 
+@pytest.fixture
+def remote_config_worker(tracer):
+    RemoteConfig._worker = None
```
brettlangdon, line 24: do you need to reset this after the fixture?

```python
old_worker = RemoteConfig._worker
RemoteConfig._worker = None
try:
    yield
finally:
    RemoteConfig._worker = old_worker
```

brettlangdon: also... why set it to `None` ?

tests/appsec/test_remoteconfiguration.py
```diff
@@ -30,36 +44,166 @@ def test_rc_enabled_by_default(tracer):
...
+    assert RemoteConfig._worker._client._products["ASM_DATA"]
+
+
+def test_rc_activation_ip_blocking_data(tracer, remote_config_worker):
+    with override_env({APPSEC_ENV: "true"}):
+        tracer.configure(appsec_enabled=True, api_version="v0.4")
+        rc_config = {
+            "rules_data": [
+                {
+                    "data": [{"expiration": 1755346879, "value": "user8"}],
```
brettlangdon, line 128: future date?

brettlangdon: I wonder if we should just do `int(time.time()) + 20000)` or something similar to what you did below.

tests/appsec/test_remoteconfiguration.py
```diff
@@ -30,36 +44,166 @@ def test_rc_enabled_by_default(tracer):
...
+
+
+def test_rc_activation_ip_blocking_data_not_expired(tracer, remote_config_worker):
+    with override_env({APPSEC_ENV: "true"}):
+        tracer.configure(appsec_enabled=True, api_version="v0.4")
+        rc_config = {
+            "rules_data": [
+                {
+                    "data": [
+                        {"expiration": int(time.time()), "value": "8.8.4.4"},
```
brettlangdon, line 190: is there a possible race condition here? a chance that we check this rule after `time.time()` has incremented by 1 ?

tests/contrib/django/test_django_appsec.py
```diff
@@ -380,3 +380,74 @@ def test_django_weak_hash(client, test_spans, tracer):
...
+
+
+def test_request_ipblock_match_403(client, test_spans, tracer):
+    with override_global_config(dict(_appsec_enabled=True)), override_env(dict(DD_APPSEC_RULES=RULES_GOOD_PATH)):
+
+        tracer._appsec_enabled = True
+        # # Hack: need to pass an argument to configure so that the processors are recreated
+        tracer.configure(api_version="v0.4")
+
+        result = client.get("/?a=1&b&c=d", HTTP_X_REAL_IP="8.8.4.4", HTTP_ACCEPT="text/html")
```
brettlangdon, line 403: this is another place where a constant might just make these tests easier to read, it is easy to miss that `8.8.8.8` and `8.8.4.4` are meant to behave differently.

tests/contrib/django/test_django_appsec.py
```diff
@@ -380,3 +380,74 @@ def test_django_weak_hash(client, test_spans, tracer):
...
+        tracer._appsec_enabled = True
+        # Hack: need to pass an argument to configure so that the processors are recreated
+        tracer.configure(api_version="v0.4")
+
+        result = client.get("/?a=1&b&c=d", HTTP_X_REAL_IP="8.8.8.8")
+        assert result.status_code == 200
+        assert result.content == b"Hello, test app."
+
+
+def test_request_ipblock_match_403(client, test_spans, tracer):
```
brettlangdon, line 396: I wonder if we should make this a snapshot test as well. It would be good confirmation to know exactly which spans get created. Are only the code paths we want getting executed?

brettlangdon: (for all of these three new ones?)

tests/contrib/flask/test_flask_appsec.py
```diff
@@ -253,3 +255,76 @@ def test_flask_body_xml_empty_logs_warning(self):
             self._aux_appsec_prepare_tracer()
             self.client.post("/", data="", content_type="application/xml")
             assert "Failed to parse werkzeug request body" in self._caplog.text
+
+    def test_request_ipblock_nomatch_200(self):
```
brettlangdon, line 259: same as for Django, I think snapshot tests would be good for these to verify only expected code paths are called/spans are created.

good way to ensure that if someone refactors the Flask integration that they don't accidentally move the blocking to the end of the request.

ddtrace/contrib/trace_utils.py
```diff
@@ -481,10 +482,15 @@ def set_http_meta(
...
-                span.set_tag_str("network.client.ip", ip)
+            # Retrieve the IP if it was calculated on AppSecProcessor.on_span_start
+            request_ip = _context.get_item("http.request.remote_ip", span=span)
+
+            if not request_ip:
+                # Not calculated: framework does not support IP blocking or testing env
+                request_ip = _get_request_header_client_ip(span, request_headers, peer_ip, headers_are_case_sensitive)
+
+            span.set_tag_str(http.CLIENT_IP, request_ip)
+            span.set_tag_str("network.client.ip", request_ip)
```
brettlangdon, line 493: 🤔 I wonder if we should be skipping these if there is no value rather than setting an empty string... but maybe this is a change we should consider for the encoder instead.

Not sure, just still feels weird that we have empty string tag values in the snapshots.

brettlangdon (changes requested):
do we need any tests for the `_asm_context`?

mainly, the important bit is that the context doesn't leak between two web requests as that'll cause some trouble for sure.

ddtrace/contrib/trace_utils.py
juanjux: Even empty it gives the information that we tried to get the IP but could not, so the user knows he hasn't misconfigured something and it's not retrieving IPs. On the real world this would never be empty since we fallback to the peer IP, that should be always present, if we can't find one in the headers but I prefer it to be empty if IP retrieval is activated either by ASM or the environment var.

ddtrace/appsec/_asm_context.py
```diff
@@ -0,0 +1,59 @@
...
+
+def set_ip(ip):  # type: (str) -> None
+    _DD_EARLY_IP_CONTEXTVAR.set(ip)
+
+
+def get_ip():  # type: () -> str
+    return _DD_EARLY_IP_CONTEXTVAR.get()
+
+
+def set_headers(headers):  # type: (Any) -> None
```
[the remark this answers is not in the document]
juanjux: I used `Any` because different frameworks use different objects for headers and it's really opaque to these functions since they just store or retrieve them without changing them or iterating or reading the internal data so it also doesn't make sense to force them into a dict or Mapping at this stage like the `trace_utils` do later (since they need to read and iterate). I'll add a comment to explain this.

ddtrace/appsec/_asm_context.py
```diff
@@ -0,0 +1,59 @@
...
+    _DD_EARLY_IP_CONTEXTVAR.set(None)
+    _DD_EARLY_HEADERS_CONTEXTVAR.set(None)
+    _DD_EARLY_HEADERS_CASE_SENSITIVE_CONTEXTVAR.set(False)
+
+
+def set_ip(ip):  # type: (str) -> None
+    _DD_EARLY_IP_CONTEXTVAR.set(ip)
+
+
+def get_ip():  # type: () -> str
```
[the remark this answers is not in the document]
juanjux: Yup.

ddtrace/appsec/_asm_context.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def asm_request_context_set(remote_ip=None, headers=None, headers_case_sensitive=False):
+    set_ip(remote_ip)
+    set_headers(headers)
+    set_headers_case_sensitive(headers_case_sensitive)
+
+
+@contextlib.contextmanager
+def asm_request_context_manager(remote_ip=None, headers=None, headers_case_sensitive=False):
```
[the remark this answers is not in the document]
juanjux: Thanks, will check.

ddtrace/contrib/django/patch.py
```diff
@@ -331,26 +336,35 @@ def traced_get_response(django, pin, func, instance, args, kwargs):
...
 
-        response = None
-        try:
-            response = func(*args, **kwargs)
-            return response
-        finally:
-            # DEV: Always set these tags, this is where `span.resource` is set
-            utils._after_request_tags(pin, span, request, response)
+            try:
+                if _context.get_item("http.request.blocked", span=span):
```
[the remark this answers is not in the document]
juanjux: If appsec is not enabled the context won't have anything for that value but I guess adding the check will be slightly faster, which is good, and also documents better that is an ASM feature.

ddtrace/contrib/flask/patch.py
```diff
@@ -560,14 +569,20 @@ def _traced_request(pin, wrapped, instance, args, kwargs):
         # This call may be unnecessary since we try to add the tags earlier
         # We just haven't been able to confirm this yet
         _set_request_tags(span)
+        request = flask.request
 
+        _asm_context.asm_request_context_set(request.remote_addr, request.headers)
```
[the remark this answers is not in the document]
juanjux: For Flask is a little different than Django since we need to clear the outside this [here](https://github.com/DataDog/dd-trace-py/pull/4613/files/7b8853882dac1ea7ef10b4db5582e7dea1e21586#diff-83b4eda0004838863b6f460c5fb4f05ba87dd6d16192db38108c1973e2937383R157); clearing the context here would reset it before getting the response tags. So we set the values here, when we have them, and reset outside this. Thus we can't use the context manager here. But maybe there is a better way? (I admit I'm less familiar with Flask than Django).

ddtrace/contrib/wsgi/wsgi.py
juanjux: We get the information on `flask_patch::request_tracer` (called before this) since that's the point where we have the request information.

ddtrace/appsec/processor.py
```diff
@@ -204,9 +208,46 @@ def __attrs_post_init__(self):
...
     def on_span_start(self, span):
         # type: (Span) -> None
-        pass
+
+        peer_ip = _asm_context.get_ip()
+        headers = _asm_context.get_headers()
+        headers_case_sensitive = _asm_context.get_headers_case_sensitive()
+        _context.set_items(
+            {
+                "http.request.headers": headers,
```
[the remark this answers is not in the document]
juanjux: Realistically, not in a real application right now, but I'll add an if for correctness and to make sure it doesn't store None if we add options to disable things atomically in the future.

ddtrace/appsec/processor.py
```diff
@@ -216,8 +257,12 @@ def _is_needed(self, address):
         # type: (str) -> bool
         return address in self._addresses_to_keep
 
+    def _run_ddwaf(self, data):
```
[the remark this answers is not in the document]
juanjux: Yup. However I've seen that the type used in the `run` function called by this one is pretty complex so I'll do a small PR before to create an alias.

ddtrace/appsec/utils.py
```diff
@@ -13,13 +18,76 @@ def _appsec_rc_features_is_enabled():
...
+        try:
+            with open(template_path, "r") as template_file:
+                content = template_file.read()
+            if need_html_template:
+                _HTML_USERDEFINED_BLOCKED_TEMPLATE_CACHE = content
+            else:
+                _JSON_USERDEFINED_BLOCKED_TEMPLATE_CACHE = content
+            return content
+        except OSError:
+            pass
```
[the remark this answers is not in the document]
juanjux: Yes, good idea.

ddtrace/appsec/utils.py
```diff
@@ -13,13 +18,76 @@ def _appsec_rc_features_is_enabled():
...
+
+    # No user-defined template at this point
+    if need_html_template:
+        return constants.APPSEC_BLOCKED_RESPONSE_HTML
+
+    return constants.APPSEC_BLOCKED_RESPONSE_JSON
```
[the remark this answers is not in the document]
juanjux: They're catched :) (both the user template and the default ones).

ddtrace/appsec/processor.py
```diff
@@ -204,9 +208,46 @@ def __attrs_post_init__(self):
         # we always need the response headers
         self._mark_needed(_Addresses.SERVER_RESPONSE_HEADERS_NO_COOKIES)
 
+    def update_rules(self, new_rules):
+        # type: (List[Dict[str, Any]]) -> None
+        self._ddwaf.update_rules(new_rules)
```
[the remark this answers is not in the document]
juanjux: Yes, good idea, also added some debug logging.

tests/appsec/test_processor.py
juanjux: Done

ddtrace/appsec/utils.py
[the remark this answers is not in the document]
brettlangdon: no, I mean cached as the value of `_JSON_USERDEFINED_BLOCKED_TEMPLATE_CACHE` so we avoid trying to call `with open(...)` on every request if the files are not there.

There is a scenario where the env vars are defined, but the files do not exist, so we will attempt to read the files and get an OSError on every single blocked request.

Is that the expected behavior, or do we want to cache the ultimate end result of this function (obviously html vs json need to be different), so that we avoid any future resolution/logic?

ddtrace/appsec/utils.py
```diff
@@ -13,13 +18,76 @@ def _appsec_rc_features_is_enabled():
...
+        template_path = os.getenv("DD_APPSEC_HTTP_BLOCKED_TEMPLATE_HTML")
+    else:
+        template_path = os.getenv("DD_APPSEC_HTTP_BLOCKED_TEMPLATE_JSON")
+
+    if template_path and os.path.exists(template_path) and os.path.isfile(template_path):
+        if need_html_template and _HTML_USERDEFINED_BLOCKED_TEMPLATE_CACHE:
+            return _HTML_USERDEFINED_BLOCKED_TEMPLATE_CACHE
+
+        if not need_html_template and _JSON_USERDEFINED_BLOCKED_TEMPLATE_CACHE:
+            return _JSON_USERDEFINED_BLOCKED_TEMPLATE_CACHE
```
brettlangdon, line 77: could this check be moved up, like to L66

if we have the cached values, we don't need to get the env vars, or check if the files exist, etc.

tests/contrib/django/test_django_appsec.py
juanjux: Done!

tests/contrib/django/test_django_appsec.py
juanjux: Good idea. Never did a snapshot test so I'll take a look at the existing ones.

ddtrace/appsec/utils.py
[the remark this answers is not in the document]
juanjux: I think I understand now, you mean something like "_FILE_NOT_FOUND_PREVIOUSLY" or something like that (will think a better name) and doing:

```       
        except OSError as e:
            log.warning("Could not load custom template at %s: %s", template_path, str(e))  # noqa: G200
            _FILE_NOT_FOUND_PREVIOUSLY = True

```

and then checking it at the start and if `True` avoiding the `open` and directly log the warning (or maybe not even logging the warning after the first one or one in every mod N requests so to not dump too much noise?) right?

ddtrace/appsec/utils.py
juanjux: Not in 66 since we need to check the context type to know if we must return json or html, but could be moved after that (line 69).

ddtrace/appsec/utils.py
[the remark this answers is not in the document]
juanjux: Found a better way: I also store in the cache the default templates, which are set if the user defined ones are not set or not found, so if `open` fails the default template will be catched and returned and on next requests it will directly return the catched ones without opening.

juanjux:
> do we need any tests for the `_asm_context`?
> 
> mainly, the important bit is that the context doesn't leak between two web requests as that'll cause some trouble for sure.

Indirectly, since it's used on the django and flask tests, but I've added some more testing the context directly before/after the request.

tests/appsec/test_remoteconfiguration.py
juanjux: Talked with @avara1986 just to be sure he confirmed me that field is of no consequence for this specific test, the `ip_with_expiration` one is the one we check here.

tests/contrib/flask/test_flask_appsec.py
juanjux: Done.

tests/appsec/test_remoteconfiguration.py
avara1986: No, this date isn't needed, is part or `data_with_expiration` object. We use in this test `ip_with_expiration` but we want to check the rules with two or more objects to be sure the function hasn't unexpected overrides

tests/appsec/test_remoteconfiguration.py
avara1986: Remote Config Worker has an initial state of None when the application start. We want to use a "clean" Remote Config worker for each unit test

ddtrace/appsec/processor.py
```diff
@@ -275,14 +329,28 @@ def on_span_finish(self, span):
...
-        try:
-            ddwaf_result = self._ddwaf.run(data, self._waf_timeout)  # res is a serialized json
-        except OSError:
-            log.warning("Error executing Appsec In-App WAF: ", exc_info=True)
-        except Exception as e:
-            log.warning("Error executing Appsec In-App WAF: %s", repr(e))
+        total_runtime = 0.0
+        total_overall_runtime = 0.0
+        blocked_request = _context.get_item("http.request.blocked", span=span)
+        if not blocked_request:
```
brettlangdon, line 342: if the request wasn't originally blocked, we want to run the waf again to double check?

I assume this is in case additional information came up during the lifetime of the request that would make the request be blocked?


if a request is valid/good then we will always run the waf twice, once at the beginning of the request, and once at the end to see if it should be blocked ?


(this seems valid, just typing it out so confirm my understanding)

ddtrace/appsec/processor.py
```diff
@@ -313,10 +381,18 @@ def on_span_finish(self, span):
...
                 _set_headers(span, data[_Addresses.SERVER_RESPONSE_HEADERS_NO_COOKIES], kind="response")
+
+            if not blocked_request and ddwaf_result:
+                # Partial DDAS-011-00
+                log.debug("[DDAS-011-00] AppSec In-App WAF returned: %s", ddwaf_result.data)
+                span.set_tag_str(APPSEC_JSON, '{"triggers":%s}' % (ddwaf_result.data,))
+            else:
+                span.set_tag(APPSEC_JSON, _context.get_item("http.request.waf_json", span=span))
+                span.set_tag(APPSEC_JSON, _context.get_item("http.request.waf_json", span=span))
+                span.set_tag("appsec.blocked", True)
```
brettlangdon, line 392:
```suggestion
                span.set_tag_str("appsec.blocked", "true")
```

assuming this is what you want, since below you also have `set_tag_str("appsec.event", "true")` ?

ddtrace/appsec/processor.py
```diff
@@ -293,15 +361,15 @@ def on_span_finish(self, span):
...
+            if not blocked_request:
+                span.set_metric(APPSEC_WAF_DURATION, total_runtime)
+                span.set_metric(APPSEC_WAF_DURATION_EXT, total_overall_runtime)
         except JSONDecodeError:
             log.warning("Error parsing data AppSec In-App WAF metrics report")
         except Exception:
             log.warning("Error executing AppSec In-App WAF metrics report: %s", exc_info=True)
 
-        if ddwaf_result and ddwaf_result.data is not None:
+        if blocked_request or (ddwaf_result and ddwaf_result.data is not None):
```
brettlangdon, line 372: is there a way we can move this block of code earlier?

I think it is a rather jarring block to parse.

e.g.

```python
blocked_request = _context.get_item("http.request.blocked", span=span)
if not blocked_request:
    # run the waf again
    set_metric(APPSEC_WAF_DURATION, ...)
    log.debug(DDAS-011-00])
    set_tag_str(APPSEC_JSON, ddwaf_result.data)
else:
    set_tag(APPSEC_JSON, _context.get_item())
    set_tag("appsec.blocked", True)


try:
    info = self._ddwaf_info
except:
    pass
```

I am not convinced this is the best suggestion to structure it, but I definitely find myself scrolling back and forth a bunch to try and follow the paths since there are a few places where we check both blocked_request and ddwaf_result

ddtrace/appsec/processor.py
```diff
@@ -313,10 +381,18 @@ def on_span_finish(self, span):
 
             if _Addresses.SERVER_RESPONSE_HEADERS_NO_COOKIES in data:
                 _set_headers(span, data[_Addresses.SERVER_RESPONSE_HEADERS_NO_COOKIES], kind="response")
+
+            if not blocked_request and ddwaf_result:
+                # Partial DDAS-011-00
+                log.debug("[DDAS-011-00] AppSec In-App WAF returned: %s", ddwaf_result.data)
+                span.set_tag_str(APPSEC_JSON, '{"triggers":%s}' % (ddwaf_result.data,))
+            else:
+                span.set_tag(APPSEC_JSON, _context.get_item("http.request.waf_json", span=span))
```
brettlangdon, line 390:
```suggestion
                span.set_tag_str(APPSEC_JSON, _context.get_item("http.request.waf_json", span=span))
```

ddtrace/appsec/utils.py
```diff
@@ -13,13 +22,80 @@ def _appsec_rc_features_is_enabled():
...
+
+    if not need_html_template and _JSON_BLOCKED_TEMPLATE_CACHE:
+        return _JSON_BLOCKED_TEMPLATE_CACHE
+
+    if need_html_template:
+        template_path = os.getenv("DD_APPSEC_HTTP_BLOCKED_TEMPLATE_HTML")
+    else:
+        template_path = os.getenv("DD_APPSEC_HTTP_BLOCKED_TEMPLATE_JSON")
+
+    if template_path and os.path.exists(template_path) and os.path.isfile(template_path):
```
brettlangdon, line 82: should we warn if there is a template path given, but the file does not exist?

I wonder if it is worth just removing the exists/isfile check, since `open()` will raise that exception for us?

not sure I love my suggestion, but feels like there is a feedback point missing for when a path is given, but the path does not exist.

ddtrace/contrib/asgi/middleware.py
```diff
@@ -111,104 +118,137 @@ async def __call__(self, scope, receive, send):
...
+                try:
+                    return await send(message)
+                finally:
+                    if message.get("type") == "http.response.body":
+                        span.finish()
 
             try:
-                return await send(message)
+                if _request_blocked(span):
+                    return await self.app(scope, receive, wrapped_blocked_send)
```
brettlangdon, line 241: do we want to call into the app still or just start the response right away?

since this way we'll still execute the web request up until the response is starting to be returned, so a malicious action could still take place?

ddtrace/contrib/django/patch.py
```diff
@@ -331,26 +336,35 @@ def traced_get_response(django, pin, func, instance, args, kwargs):
...
-        try:
-            response = func(*args, **kwargs)
-            return response
-        finally:
-            # DEV: Always set these tags, this is where `span.resource` is set
-            utils._after_request_tags(pin, span, request, response)
+            try:
+                if config._appsec_enabled and _context.get_item("http.request.blocked", span=span):
+                    response = HttpResponseForbidden(
+                        appsec_utils._get_blocked_template(request_headers.get("Accept")), **kwargs
```
brettlangdon, line 361: I am not sure `**kwargs` should be here ?

This function wrapper is around [BaseHandler.get_response](https://github.com/django/django/blob/f3c89744cc801cc7d134bca9958c4a74aa76380f/django/core/handlers/base.py#L136) which it's only argument is a `request` object.

The function signature does not match that of an `HttpResponse` class, so I think we should avoid `**kwargs` here otherwise we are likely to have an issue.

```suggestion
                        appsec_utils._get_blocked_template(request_headers.get("Accept")),
```

ddtrace/contrib/trace_utils.py
```diff
@@ -440,6 +440,7 @@ def set_http_meta(
     :param request_path_params: the parameters of the HTTP URL as set by the framework: /posts/<id:int> would give us
          { "id": <int_value> }
     """
+
```
brettlangdon, line 443:
```suggestion
```

ddtrace/contrib/trace_utils.py
```diff
@@ -479,10 +480,15 @@ def set_http_meta(
...
-                span.set_tag_str("network.client.ip", ip)
+            # Retrieve the IP if it was calculated on AppSecProcessor.on_span_start
+            request_ip = _context.get_item("http.request.remote_ip", span=span)
+
+            if not request_ip:
+                # Not calculated: framework does not support IP blocking or testing env
+                request_ip = _get_request_header_client_ip(request_headers, peer_ip, headers_are_case_sensitive)
+
+            span.set_tag_str(http.CLIENT_IP, request_ip)
+            span.set_tag_str("network.client.ip", request_ip)
```
brettlangdon, line 491: do we need both of these tags?

seems redundant to set the same value twice under two different tags. (I see it was here before, so maybe not worth getting into here)

ddtrace/internal/remoteconfig/client.py
```diff
@@ -357,7 +358,8 @@ def _process_response(self, data):
         if last_targets_version is None or targets is None:
             log.debug("No targets in configuration payload")
             for callback in self._products.values():
-                callback(None, None)
+                if callback:
```
brettlangdon, line 361: do we want a try/except here? should we call every callback for all products?

I see below if we call with `callback(config, callback_action)` we have a `try/except` with a `continue`. Do we need the same here?

brettlangdon (commented):
I need to review changes to the tests still, but wanted to submit what I have so far.

ddtrace/appsec/processor.py
```diff
@@ -275,14 +329,28 @@ def on_span_finish(self, span):
...
-        except OSError:
-            log.warning("Error executing Appsec In-App WAF: ", exc_info=True)
-        except Exception as e:
-            log.warning("Error executing Appsec In-App WAF: %s", repr(e))
+        total_runtime = 0.0
+        total_overall_runtime = 0.0
+        blocked_request = _context.get_item("http.request.blocked", span=span)
+        if not blocked_request:
+            try:
+                ddwaf_result = self._ddwaf.run(data, self._waf_timeout)  # res is a serialized json
```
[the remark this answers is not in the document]
juanjux: Will commit separately since `_run_ddwaf` doesn't take the second argument.

ddtrace/appsec/processor.py
juanjux: Yup.

ddtrace/appsec/processor.py
```diff
@@ -293,15 +361,15 @@ def on_span_finish(self, span):
 
             span.set_metric(APPSEC_EVENT_RULE_LOADED, info.loaded)
             span.set_metric(APPSEC_EVENT_RULE_ERROR_COUNT, info.failed)
-            if ddwaf_result:
-                span.set_metric(APPSEC_WAF_DURATION, ddwaf_result.runtime)
-                span.set_metric(APPSEC_WAF_DURATION_EXT, ddwaf_result.total_runtime)
+            if not blocked_request:
+                span.set_metric(APPSEC_WAF_DURATION, total_runtime)
+                span.set_metric(APPSEC_WAF_DURATION_EXT, total_overall_runtime)
```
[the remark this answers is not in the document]
juanjux: Removed as a product of the small refactor from other comments.

ddtrace/contrib/trace_utils.py
juanjux: It seems we do since they're used on different specs but I need to speak with some people to see if we're to late to unify stuff. Out of the PR context anyway.

ddtrace/appsec/processor.py
juanjux: Yup, the initial check is only for IP blocking with the early information (the IP), but we could block later once we have more information available which means 1/2 calls to the WAF and could be even more since for example the user blocking feature is a function that the user can call and must call the WAF. My gut sense is that once we've merged all the blocking features we can see some way to optimize all this since we'll know all the potential call points for the WAF.

ddtrace/appsec/_remoteconfiguration.py
```diff
@@ -10,8 +10,15 @@
 if TYPE_CHECKING:  # pragma: no cover
```
Kyle-Verhoog, line 10: The changes made to remote config here don't seem necessary to the IP blocking. Why are they made in this PR?

ddtrace/appsec/_remoteconfiguration.py
```diff
@@ -26,9 +33,46 @@ def enable_appsec_rc():
...
+        if not (APPSEC_ENV not in os.environ and rc_appsec_enabled is True) and (
+            asbool(os.environ.get(APPSEC_ENV)) is False or rc_appsec_enabled is False
+        ):
+            _appsec_enabled = False
+            RemoteConfig.unregister(ASM_DATA_PRODUCT)
+        else:
+            RemoteConfig.register(ASM_DATA_PRODUCT, appsec_rc_reload_features(tracer))
+
+        if tracer._appsec_enabled != _appsec_enabled:
+            tracer.configure(appsec_enabled=_appsec_enabled)
```
Kyle-Verhoog, line 75: This is really making me think that we need an ASM singleton for the application. It really doesn't make sense to be on the tracer at this point.

I don't want this to block this change but the coupling is getting pretty gnarly. @juanjux can we see about doing this as follow up work?

ddtrace/appsec/_remoteconfiguration.py
```diff
@@ -26,9 +33,46 @@ def enable_appsec_rc():
...
+        rc_appsec_enabled = features.get("asm", {}).get("enabled")
+
+    if rc_appsec_enabled is not None:
+        from ddtrace.internal.remoteconfig import RemoteConfig
+        from ddtrace.internal.remoteconfig.constants import ASM_DATA_PRODUCT
+
+        log.debug("Reloading Appsec 1-click: %s", rc_appsec_enabled)
+        _appsec_enabled = True
+
+        if not (APPSEC_ENV not in os.environ and rc_appsec_enabled is True) and (
```
Kyle-Verhoog, line 66: this condition is quite hard to read, can we simplify it a bit since these are booleans

```suggestion
        if not (APPSEC_ENV not in os.environ and rc_appsec_enabled) and (
```

ddtrace/appsec/_remoteconfiguration.py
```diff
@@ -26,9 +33,46 @@ def enable_appsec_rc():
...
+            asbool(os.environ.get(APPSEC_ENV)) is False or rc_appsec_enabled is False
```
Kyle-Verhoog, line 67:
```suggestion
            not asbool(os.environ.get(APPSEC_ENV)) or not rc_appsec_enabled
```

ddtrace/appsec/_remoteconfiguration.py
```diff
@@ -50,16 +94,8 @@ def _reload_features(metadata, features):
...
-
-            if not (APPSEC_ENV not in os.environ and rc_appsec_enabled is True) and (
-                asbool(os.environ.get(APPSEC_ENV)) is False or rc_appsec_enabled is False
-            ):
-                _appsec_enabled = False
-
-            tracer.configure(appsec_enabled=_appsec_enabled)
+            log.debug("Updating Appsec Remote Configuration: %s", features)
+            _appsec_rules_data(tracer, features)
+            _appsec_1click_actication(tracer, features)
```
Kyle-Verhoog, line 99:
```suggestion
            _appsec_1click_activation(tracer, features)
```

ddtrace/appsec/processor.py
```diff
@@ -210,9 +217,51 @@ def __attrs_post_init__(self):
         # we always need the response headers
         self._mark_needed(_Addresses.SERVER_RESPONSE_HEADERS_NO_COOKIES)
 
+    def update_rules(self, new_rules):
```
Kyle-Verhoog, line 220: note that this will be a public API if released, is that intentional?

ddtrace/appsec/processor.py
```diff
@@ -210,9 +217,51 @@ def __attrs_post_init__(self):
...
+        # type: (List[Dict[str, Any]]) -> None
+        try:
+            self._ddwaf.update_rules(new_rules)
+        except TypeError as e:
+            log.debug("Error updating RCM rules: %s", str(e))  # noqa: G200
```
Kyle-Verhoog, line 225:
```suggestion
            log.debug("Error updating ASM rules", exc_info=True)  # noqa: G200
```

is this method specific to RCM updates? if so can we rename the method to `_update_rules_from_rcm`.

ddtrace/appsec/processor.py
```diff
@@ -210,9 +217,51 @@ def __attrs_post_init__(self):
...
+                span=span,
+            )
+
+            if peer_ip is not None:
+                ip = trace_utils._get_request_header_client_ip(headers, peer_ip, headers_case_sensitive)
+                # Save the IP and headers in the context so the retrieval can be skipped later
+                _context.set_item("http.request.remote_ip", ip, span=span)
+                if ip and self._is_needed(_Addresses.HTTP_CLIENT_IP):
+                    data = {_Addresses.HTTP_CLIENT_IP: ip}
+                    log.debug("[DDAS-001-00] Executing AppSec In-App WAF with parameters: %s", data)
```
Kyle-Verhoog, line 249:
```suggestion
                    log.debug("[DDAS-001-00] Executing ASM WAF with parameters: %s", data)
```

this will log IPs in plaintext is this a data sensitivity concern?

ddtrace/appsec/processor.py
```diff
@@ -210,9 +217,51 @@ def __attrs_post_init__(self):
...
+                    ddwaf_result = self._run_ddwaf(data)
+
+                    if ddwaf_result and ddwaf_result.actions and "block" in ddwaf_result.actions and ddwaf_result.data:
+                        res_dict = json.loads(ddwaf_result.data)
+                        log.debug("[DDAS-011-00] AppSec In-App WAF returned: %s", res_dict)
```
Kyle-Verhoog, line 254:
```suggestion
                        log.debug("[DDAS-011-00] ASM WAF returned: %s", res_dict)
```

will there be sensitive data in this return value?

ddtrace/appsec/processor.py
```diff
@@ -275,14 +329,28 @@ def on_span_finish(self, span):
...
+        blocked_request = _context.get_item("http.request.blocked", span=span)
+        if not blocked_request:
+            try:
+                ddwaf_result = self._ddwaf.run(data, self._waf_timeout)  # res is a serialized json
+                total_runtime = ddwaf_result.runtime
+                total_overall_runtime = ddwaf_result.total_runtime
+            except OSError:
+                log.warning("Error executing Appsec In-App WAF: ", exc_info=True)
+            except Exception as e:
+                log.warning("Error executing Appsec In-App WAF: %s", repr(e))
```
Kyle-Verhoog, line 350: no `exc_info` here?

ddtrace/appsec/processor.py
```diff
@@ -275,14 +329,28 @@ def on_span_finish(self, span):
...
-            log.warning("Error executing Appsec In-App WAF: %s", repr(e))
+        total_runtime = 0.0
+        total_overall_runtime = 0.0
+        blocked_request = _context.get_item("http.request.blocked", span=span)
+        if not blocked_request:
+            try:
+                ddwaf_result = self._ddwaf.run(data, self._waf_timeout)  # res is a serialized json
+                total_runtime = ddwaf_result.runtime
+                total_overall_runtime = ddwaf_result.total_runtime
+            except OSError:
```
Kyle-Verhoog, line 347: Why the special case of `OSError`?

ddtrace/appsec/utils.py
```diff
@@ -13,13 +22,80 @@ def _appsec_rc_features_is_enabled():
 
 def _appsec_rc_capabilities():
     # type: () -> str
-    """return the bit of the composed capabilities in base64
+    r"""return the bit representation of the composed capabilities in base64
```
Kyle-Verhoog, line 25: Is this change specific to IP blocking? Can we introduce it separately?

ddtrace/appsec/utils.py
```diff
@@ -1,9 +1,18 @@
...
+from typing import Optional
 
+from ddtrace import constants
 from ddtrace.constants import APPSEC_ENV
+from ddtrace.internal.compat import to_bytes_py2
+from ddtrace.internal.logger import get_logger
 from ddtrace.internal.utils.formats import asbool
 
 
+log = get_logger(__name__)
```
Kyle-Verhoog, line 13: We probably shouldn't be introducing a logger here. In general, utilities should `raise` and leave exception handling to the callers of the utility functions. This gives them the flexibility to re-raise to the user in extreme cases or handle the error in whatever way makes sense in context.

ddtrace/constants.py
```diff
@@ -58,3 +58,110 @@
 AUTO_KEEP = 1
 # Use this to explicitly inform the backend that a trace should be kept and stored.
 USER_KEEP = 2
+
+APPSEC_BLOCKED_RESPONSE_HTML = """
```
Kyle-Verhoog, line 62: should this be minified to reduce overhead in the customer application?

Kyle-Verhoog (changes requested):
@juanjux why did you fill out the reviewer checklist yourself? Can you uncheck it for the reviewers?

I did a pass of the first few files. The context stuff I'm mostly ok with, it will be easy to change to a general context API. The processor looks mostly good too besides nits.

Why are we starting by introducing this in all integrations?
- The risk surface area is much larger by starting with all integrations from the start. We might potentially break multiple of our biggest web frameworks.
- It's more effort to review/iterate on review because something in an integration will likely have to be changed num_integrations times.
- We can focus testing the one integration, gain confidence with the approach and quickly add the rest with the precedent.

There look to be some changes made to RCM which aren't related to IP blocking, if that is the case can we do them separately to separate concerns.


(Note that this is the same approach we used when first introducing ASM and allowed us to iterate when things were discovered in testing and beta usage)

ddtrace/appsec/_remoteconfiguration.py
juanjux: They are needed since it's the first RCM product registered outside of 1-click, thus need to register the feature.

ddtrace/appsec/_remoteconfiguration.py
```diff
@@ -10,8 +10,15 @@
 if TYPE_CHECKING:  # pragma: no cover
     from typing import Any
     from typing import Callable
+
+    try:
+        from typing import Literal
+    except ImportError:
+        # Python < 3.8
+        from typing_extensions import Literal  # type: ignore
```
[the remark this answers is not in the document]
juanjux: To avoid mypy saying that it doesn't exist when run under older Pythons without having to add another if to check the version.

ddtrace/appsec/_remoteconfiguration.py
```diff
@@ -26,9 +33,46 @@ def enable_appsec_rc():
...
+    if features is False:
+        rc_appsec_enabled = False
+    else:
+        rc_appsec_enabled = features.get("asm", {}).get("enabled")
+
+    if rc_appsec_enabled is not None:
+        from ddtrace.internal.remoteconfig import RemoteConfig
+        from ddtrace.internal.remoteconfig.constants import ASM_DATA_PRODUCT
+
+        log.debug("Reloading Appsec 1-click: %s", rc_appsec_enabled)
```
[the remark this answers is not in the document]
juanjux: Yep, 1-click is the name of the feature as was announced. It's not all RCM (which would include the blocking features and other things) but the part about enabling or disabling ASM from the web backend.

ddtrace/appsec/processor.py
juanjux: Probably not. Until I can ask @avara1986 tomorrow I will make it private.

ddtrace/appsec/processor.py
juanjux: Maybe. I will remove it just in case. Thanks!

ddtrace/appsec/processor.py
juanjux: Good question... removed.

ddtrace/appsec/processor.py
juanjux: Added to the remaining exception handler.

ddtrace/appsec/processor.py
```diff
@@ -293,15 +361,15 @@ def on_span_finish(self, span):
 
             span.set_metric(APPSEC_EVENT_RULE_LOADED, info.loaded)
             span.set_metric(APPSEC_EVENT_RULE_ERROR_COUNT, info.failed)
-            if ddwaf_result:
-                span.set_metric(APPSEC_WAF_DURATION, ddwaf_result.runtime)
-                span.set_metric(APPSEC_WAF_DURATION_EXT, ddwaf_result.total_runtime)
+            if not blocked_request:
```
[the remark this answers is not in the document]
juanjux: Not now since they'll be set already on `on_span_start` as you suggested on another comment.

ddtrace/appsec/utils.py
juanjux: Is needed for the feature registration, could be introduced separately however.

ddtrace/internal/remoteconfig/client.py
juanjux: Probably yes (need to ask @avara1986 ) wil preemptively add it.

ddtrace/appsec/_remoteconfiguration.py
```diff
@@ -26,9 +33,46 @@ def enable_appsec_rc():
...
+    # type: (Tracer, Union[Literal[False], Mapping[str, Any]]) -> None
+    if features and tracer._appsec_processor:
+        rules_data = features.get("rules_data", [])
+        if rules_data:
+            log.debug("Reloading Appsec rules data: %s", rules_data)
+            tracer._appsec_processor.update_rules(rules_data)
+
+
+def _appsec_1click_actication(tracer, features):
+    # type: (Tracer, Union[Literal[False], Mapping[str, Any]]) -> None
```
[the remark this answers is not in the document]
juanjux: @avara1986 wdyt?

ddtrace/appsec/_remoteconfiguration.py
juanjux: Yup.

ddtrace/appsec/processor.py
juanjux: I think this was talked some time ago on some Slack ASM channel and it was decided that it was fine if the loglevel was debug but I'll ask again tomorrow just to be sure since slack deleted already that conversation.

Kyle-Verhoog: :+1: thanks and apologies if this was already something long decided and I'm digging it back up 😄

ddtrace/appsec/processor.py
juanjux: Simplified the logic and added a boolean for either first or second ddwaf runs returning true which should make this more readable.

ddtrace/appsec/utils.py
juanjux: Removed the isfile/exists checks. The exception potentially raised by open (could not load custom template...) should be enough hint for the user.

ddtrace/constants.py
juanjux: Good suggestion.

juanjux:
@Kyle-Verhoog I don't understand what do you mean by this beging adding "to all integrations"? It's being adding for Django and Flask which is what will be initially required.

Kyle-Verhoog:
>  @Kyle-Verhoog I don't understand what do you mean by this beging adding "to all integrations"? It's being adding for Django and Flask which is what will be initially required.

Ah yeah, sorry I should have been more specific. I meant why add it to more than one integration to start with, for the reasons I mentioned.

juanjux:
> > @Kyle-Verhoog I don't understand what do you mean by this beging adding "to all integrations"? It's being adding for Django and Flask which is what will be initially required.
> 
> Ah yeah, sorry I should have been more specific. I meant why add it to more than one integration to start with, for the reasons I mentioned.

Django and Flask because it’s required for the RCM beta feature that’ll be publicly announced soon and have clients of both frameworks interested. For WSGI it’s only setting the context, which Flask requires. ASGI was not initially there but it’s needed for the Django snapshot tests that @brettlangdon requested so I added support for it.

ddtrace/appsec/_remoteconfiguration.py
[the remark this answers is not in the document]
avara1986: This is the behavior of RC:
- Agent returns data: callback update/enable product
- Empty agent response: callback doesn't do nothing
- "Special" empty agent response: send False to callback and it disables the product

ddtrace/appsec/utils.py
juanjux: @brettlangdon suggested we logged a message here.

juanjux:
@brettlangdon @Kyle-Verhoog I'll split this PR in two, one for the remote configuration registration part and other for the blocking itself so for now I'm closing this PR.

PR #4614, opened 2022-11-23
chore(w3c): add extract for tracecontext propagation

ZStriker19:
## Description

Adds initial support for extracting the W3C tracecontext and tracestate headers needed for distributed tracing with OpenTelemetry.  The list elements of tracestate with the `dd` vendor key are used for propagating Datadog tags.

## Design 

We follow the [W3C Trace Context specification](https://www.w3.org/TR/trace-context/).

The format for ``traceparent``:
```
      HEXDIGLC        = DIGIT / "a" / "b" / "c" / "d" / "e" / "f"
      value           = version "-" version-format
      version         = 2HEXDIGLC
      version-format  = trace-id "-" parent-id "-" trace-flags
      trace-id        = 32HEXDIGLC
      parent-id       = 16HEXDIGLC
      trace-flags     = 2HEXDIGLC

    Example value of HTTP ``traceparent`` header::
        value = 00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01
        base16(version) = 00
        base16(trace-id) = 4bf92f3577b34da6a3ce929d0e0e4736
        base16(parent-id) = 00f067aa0ba902b7
        base16(trace-flags) = 01  // sampled
```

The tracestate field value is a list of list-members separated by commas (,). A list-member is a key/value pair separated by an equals sign (=). Spaces and horizontal tabs surrounding list-members are ignored. There can be a maximum of 32 list-members in a list.

The tracestate header has one list member added to it, `dd`, which contains values that would be in x-datadog-tags as well as those needed for propagation information with shortened keys to meet length restrictions.

Example:  `dd=s:2;o:rum;t.dm:-4;t.usr.id:baz64`

Other vendors can add their own value to the tracestate header, leading to values looking like: `dd=s:2;o:rum;t.dm:-4;t.usr.id:baz64,congo=t61rcWkgMzE,zach=is_cool:yes;lives:in_NH`

This follows the standard format of adding an extractor for http propagation.  In addition we add traceparent and tracestate to the context meta object since it must be maintained and sent on the outgoing request.

## Testing strategy
Added test cases to the original header test cases.

## Checklist

## Reviewer Checklist
- [x] Title is accurate.


ddtrace/propagation/http.py
```diff
@@ -483,10 +489,201 @@ def _extract(headers):
...
+            result = re.search("dd=(.+?),", ts)
+
+            # grab dd value if dd is only list member
+            if not result:
+                result = re.search("dd=(.*)", ts)
+            try:
+                dd = dict(item.split(":") for item in result.group(1).split(";"))  # type: ignore
+            except AttributeError:
+                log.debug(("no dd list member in tracestate from incoming request: %r "), ts)
+                dd = None
```
mabdinur, line 602: This simplifies the regex and removes a redundant try/except block. If `dd = dict(....)` fails we will log: `received invalid dd header value in tracestate: [ts value]`. 

Would something like this work? 

```suggestion
           # match datadog tracestate (excludes trailing comma)
            dd_ts = re.search("dd=(.+?)(?:,|$)", ts)
            dd = None
            if dd_ts:
                dd = dict(item.split(":") for item in dd_ts.split(";"))  # type: ignore
```

ddtrace/propagation/http.py
```diff
@@ -483,10 +489,201 @@ def _extract(headers):
...
+
+            # parse out values
+            if dd:
+                sampling_priority_ts = dd.get("s")
+                if sampling_priority_ts:
+                    sampling_priority_ts = int(sampling_priority_ts)
+                else:
+                    sampling_priority_ts = None
```
mabdinur, line 610: Not required. The default value of dd.get("s") is `None`
```suggestion
```

ddtrace/propagation/http.py
```diff
@@ -483,10 +489,201 @@ def _extract(headers):
...
+                meta.update(other_propagated_tags)
+
+                return sampling_priority_ts, meta, origin
+            else:
+                # even if there's no dd list member, we still need to propagate the other tracestate values
+                return None, meta, ""
+
+        except (TypeError, ValueError, AttributeError):
+            log.debug(("received invalid dd header value in tracestate: %r "), ts)
+            return None
```
mabdinur, line 628: ?
```suggestion
            return None, meta, ""
```

ddtrace/propagation/http.py
```diff
@@ -483,10 +489,201 @@ def _extract(headers):
...
+
+    @staticmethod
+    def _decicide_sampling_priority(sampling_priority_tp, sampling_priority_ts):
```
mabdinur, line 631:
```suggestion
    def _get_sampling_priority(sampling_priority_tp, sampling_priority_ts):
```

ddtrace/propagation/http.py
```diff
@@ -483,10 +489,201 @@ def _extract(headers):
...
+        # type: (int, int) -> int
+        if sampling_priority_tp == 0 and (not sampling_priority_ts or sampling_priority_ts >= 0):
```
mabdinur, line 633: Can we add a doc string with the expected logic. It will make this these if statements easier to review and maintain.

ddtrace/propagation/http.py
```diff
@@ -483,10 +489,201 @@ def _extract(headers):
...
+        return sampling_priority
+
+    @staticmethod
+    def _extract(headers):
+        # type: (Dict[str, str]) -> Optional[Context]
+        traceparent_values = _TraceContext._get_traceparent_values(headers)
+        if traceparent_values:
+            tp, trace_id, span_id, sampling_priority = traceparent_values
+        else:
+            return None
```
mabdinur, line 654: nit
```suggestion
        if not traceparent_values:
            return None
        tp, trace_id, span_id, sampling_priority = traceparent_values     
```

tests/tracer/test_propagation.py
```diff
@@ -368,6 +373,305 @@ def test_get_wsgi_header(tracer):
...
+                    "_dd.p.dm": "-4",
+                    "_dd.p.usr.id": "baz64",
+                },
+                None,
+            ),
+            None,
+        ),
+    ],
+)
+def test_extract_tracestate(caplog, name, ts_string, expected_tuple, expected_logging):
```
mabdinur, line 595: Can we add a test case with non ascii characters (ex: put emojis in the header) and a test where `dd=......,` has an invalid format (ex: `dd=foo|bar:hi|low$candy;:emoji`).

mabdinur (commented):
Looks good just left a few nits

ddtrace/propagation/http.py
ZStriker19: ah great regex thanks! I just used `dd = dict(item.split(":") for item in dd_ts.group(1).split(";"))` instead and it looks good. Thanks!

ddtrace/propagation/http.py
ZStriker19: Thanks let me add a test to cover this too.

ddtrace/propagation/http.py
```diff
@@ -71,20 +79,18 @@ def _extract_header_value(possible_header_names, headers, default=None):
     return default
 
 
-def _b3_id_to_dd_id(b3_id):
+def _hex_id_to_dd_id(hex_id):
```
Kyle-Verhoog, line 82: A bit of a nit...: not sure it actually makes sense to rename this function. Just because B3 happens to use the same "hex" spec that OTel does, doesn't mean that they will remain the same in the future. By merging them into the same function we might implicitly break support for one when changing the function for the other (eg. if we need to add something for OTel we might break B3). 

My preference would be that we keep the `_b3_id_to_dd_id` and just set it equal to `_hex_id_to_dd_id` for now since they happen to be the same. If the implementations were to diverge then `_b3_id_to_dd_id`/`_otel_id_to_dd_id` could be modified to handle the specific cases.

This is probably a bit overkill at the moment, but maybe something to think about now.

ddtrace/context.py
```diff
@@ -53,6 +54,7 @@ def __init__(
         meta=None,  # type: Optional[_MetaDictType]
         metrics=None,  # type: Optional[_MetricDictType]
         lock=None,  # type: Optional[threading.RLock]
+        traceparent=None,  # type: Optional[str]
```
Kyle-Verhoog, line 57: This has a bit of a smell to it. I don't think we should be having to change the public API to add support for OTel. We do have sufficient support for it already with `_meta`, should we just have an internal API for this state instead? `_set_meta()`?

ddtrace/internal/constants.py
```diff
@@ -1,7 +1,15 @@
 PROPAGATION_STYLE_DATADOG = "datadog"
 PROPAGATION_STYLE_B3 = "b3"
 PROPAGATION_STYLE_B3_SINGLE_HEADER = "b3 single header"
-PROPAGATION_STYLE_ALL = (PROPAGATION_STYLE_DATADOG, PROPAGATION_STYLE_B3, PROPAGATION_STYLE_B3_SINGLE_HEADER)
+PROPAGATION_STYLE_TRACECONTEXT = "tracecontext"
+PROPAGATION_STYLE_ALL = (
+    PROPAGATION_STYLE_TRACECONTEXT,
```
Kyle-Verhoog, line 6:
```suggestion
    PROPAGATION_STYLE_W3C_TRACECONTEXT,
```
?

ddtrace/propagation/http.py
```diff
@@ -483,10 +489,201 @@ def _extract(headers):
...
+        base16(trace-flags) = 01  // sampled
+
+    The format for ``tracestate`` is key value pairs with each entry limited to 256 characters.
+    An example of the ``dd`` list member we would add is::
+    "dd=s:2;o:rum;t.dm:-4;t.usr.id:baz64"
+
+    Implementation details:
+      - Datadog Trace and Span IDs are 64-bit unsigned integers.
+      - The W3C Trace Context Trace ID is a 16-byte hexadecimal string.
+      - If the incoming traceparent was valid we DO NOT update the trace-id.
```
Kyle-Verhoog, line 527:
```suggestion
      - If the incoming traceparent is valid we DO NOT update the trace-id.
```
also, should this be "invalid"?

ddtrace/propagation/http.py
```diff
@@ -483,10 +489,201 @@ def _extract(headers):
...
+      - The tracestate header will have one list member added to it, ``dd``, which contains
+        values that would be in x-datadog-tags as well as those needed for propagation information.
+        The keys to the ``dd`` values have been shortened as follows to save space:
+        ``sampling_priority`` = ``s``
+        ``origin`` = ``o``
+        ``_dd.p.`` prefix = ``t.``
+    """
+
+    @staticmethod
+    def _is_valid_datadog_trace_tag_key(key):
```
Kyle-Verhoog, line 541:
```suggestion
    def _is_valid_datadog_trace_tag_key(key):
        # type: (str) -> bool
```

ddtrace/propagation/http.py
```diff
@@ -483,10 +489,201 @@ def _extract(headers):
...
+        return key.startswith("t.")
+
+    @staticmethod
+    def _get_traceparent_values(headers):
+        # type: (Dict[str, str]) -> Optional[Tuple[str,int,int,int]]
```
Kyle-Verhoog, line 546:
```suggestion
        # type: (Dict[str, str]) -> Optional[Tuple[str, int, int, int]]
```
nit

ddtrace/propagation/http.py
```diff
@@ -483,10 +489,201 @@ def _extract(headers):
...
+        ``origin`` = ``o``
+        ``_dd.p.`` prefix = ``t.``
+    """
+
+    @staticmethod
+    def _is_valid_datadog_trace_tag_key(key):
+        return key.startswith("t.")
+
+    @staticmethod
+    def _get_traceparent_values(headers):
```
Kyle-Verhoog, line 545: docstring might be helpful here

ddtrace/propagation/http.py
```diff
@@ -483,10 +489,201 @@ def _extract(headers):
...
+        if tp is None:
+            return None
+
+        try:
+            version, trace_id_hex, span_id_hex, trace_flags_hex = tp.split("-")
+            # check version is a valid hexadecimal
+            int(version, 16)
+            # currently 00 is the only version format, but if future versions come up we may need to add changes
+            if version != "00":
+                log.warning("unsupported traceparent version:%s . Will still attempt to parse.", (version))
```
Kyle-Verhoog, line 558:
```suggestion
                log.warning("unsupported traceparent version: %r, still attempting to parse", version)
```

ddtrace/propagation/http.py
```diff
@@ -483,10 +489,201 @@ def _extract(headers):
...
+        # type: (Dict[str, str]) -> Optional[Tuple[str,int,int,int]]
+
+        tp = _extract_header_value(_POSSIBLE_HTTP_HEADER_TRACEPARENT, headers)
+        if tp is None:
+            return None
+
+        try:
+            version, trace_id_hex, span_id_hex, trace_flags_hex = tp.split("-")
+            # check version is a valid hexadecimal
+            int(version, 16)
```
Kyle-Verhoog, line 555: Do we need to do this check? Doesn't seem like it gains us much and if we don't have a supported version we still continue anyway (see below)

ddtrace/propagation/http.py
```diff
@@ -483,10 +489,201 @@ def _extract(headers):
...
+
+                # All 0s are invalid values
+                assert trace_id != 0
+                assert span_id != 0
+
+                trace_flags = _hex_id_to_dd_id(trace_flags_hex)
+                # there's currently only one trace flag, which denotes sampling priority
+                # was set to keep "01" or drop "00"
+                # trace flags is a bit field: https://www.w3.org/TR/trace-context/#trace-flags
+                sampling_priority = int(trace_flags & 0x1)
```
Kyle-Verhoog, line 572:
```suggestion
                sampling_priority = trace_flags & 0x1
```

the type of `<int> & <int>` is `<int>` so we shouldn't need an additional `int()` call here.

ddtrace/propagation/http.py
```diff
@@ -483,10 +489,201 @@ def _extract(headers):
...
+
+            else:
+                log.debug("W3C traceparent hex length incorrect: %s", tp)
+                raise ValueError
```
Kyle-Verhoog, line 576:
```suggestion
                raise ValueError("W3C traceparent hex length incorrect: %s" % tp)
```

ddtrace/propagation/http.py
```diff
@@ -483,10 +489,201 @@ def _extract(headers):
...
+                assert span_id != 0
+
+                trace_flags = _hex_id_to_dd_id(trace_flags_hex)
+                # there's currently only one trace flag, which denotes sampling priority
+                # was set to keep "01" or drop "00"
+                # trace flags is a bit field: https://www.w3.org/TR/trace-context/#trace-flags
+                sampling_priority = int(trace_flags & 0x1)
+
+            else:
+                log.debug("W3C traceparent hex length incorrect: %s", tp)
```
Kyle-Verhoog, line 575:
```suggestion
```

ddtrace/propagation/http.py
```diff
@@ -483,10 +489,201 @@ def _extract(headers):
...
+
+    @staticmethod
+    def _get_traceparent_values(headers):
+        # type: (Dict[str, str]) -> Optional[Tuple[str,int,int,int]]
+
+        tp = _extract_header_value(_POSSIBLE_HTTP_HEADER_TRACEPARENT, headers)
+        if tp is None:
+            return None
+
+        try:
```
Kyle-Verhoog, line 552: Can we move this error handling up a level? I don't think it should be the responsibility of this method to also handle/log the errors, it should be the caller of this method.

ddtrace/propagation/http.py
Kyle-Verhoog, line 545: This function is also doing a lot. It's attempting to pull the header of of a dictionary of headers, attempting to parse and do error handling.

I think we could have this method just attempt to parse a `traceparent` value and raise if it fails.

The error handling/logging and dictionary look up could be handled by the caller.

ddtrace/propagation/http.py
```diff
@@ -483,10 +489,201 @@ def _extract(headers):
...
+
+    @staticmethod
+    def _get_tracestate_values(ts):
+        # type: (str) -> Optional[Tuple[Optional[int], Dict[str, str], Optional[str]]]
+
+        # tracestate parsing, example: dd=s:2;o:rum;t.dm:-4;t.usr.id:baz64,congo=t61rcWkgMzE
+        # store ts so we keep other vendor data
+        meta = {_TRACESTATE_KEY: ts}
+
+        try:
```
Kyle-Verhoog, line 592: Similar to the above function, I think the error handling should be done by the caller of the function.

ddtrace/propagation/http.py
```diff
@@ -483,10 +489,201 @@ def _extract(headers):
...
+            result = re.search("dd=(.+?),", ts)
+
+            # grab dd value if dd is only list member
+            if not result:
+                result = re.search("dd=(.*)", ts)
+            try:
+                dd = dict(item.split(":") for item in result.group(1).split(";"))  # type: ignore
+            except AttributeError:
+                log.debug(("no dd list member in tracestate from incoming request: %r "), ts)
```
Kyle-Verhoog, line 601:
```suggestion
                log.debug("no dd list member in tracestate from incoming request: %r ", ts)
```

ddtrace/propagation/http.py
```diff
@@ -483,10 +489,201 @@ def _extract(headers):
...
+    @staticmethod
+    def _get_tracestate_values(ts):
+        # type: (str) -> Optional[Tuple[Optional[int], Dict[str, str], Optional[str]]]
+
+        # tracestate parsing, example: dd=s:2;o:rum;t.dm:-4;t.usr.id:baz64,congo=t61rcWkgMzE
+        # store ts so we keep other vendor data
+        meta = {_TRACESTATE_KEY: ts}
+
+        try:
+            result = re.search("dd=(.+?),", ts)
```
Kyle-Verhoog, line 593: Do we really need to regex search here? We just need to find the first occurrence of `dd=` right? Could we just use `index()`?

ddtrace/propagation/http.py
```diff
@@ -483,10 +489,201 @@ def _extract(headers):
...
+
+            # grab dd value if dd is only list member
+            if not result:
+                result = re.search("dd=(.*)", ts)
+            try:
+                dd = dict(item.split(":") for item in result.group(1).split(";"))  # type: ignore
```
Kyle-Verhoog, line 599: why `type: ignore`?

ddtrace/internal/constants.py
ZStriker19: Sure!

ddtrace/context.py
ZStriker19: I can just put traceparent into the meta object and instantiate the Context with the meta.

ddtrace/propagation/http.py
ZStriker19: I changed the wording to make it more clear.

ddtrace/propagation/http.py
ZStriker19: I'm not sure, I guess in my mind there's a difference between version being a valid hexadecimal that we haven't added support for yet, and version being an invalid combination of characters, in which case we should return `None`. Do you still think we should remove it though?

ddtrace/propagation/http.py
ZStriker19: This logging was removed since I didn't need multiple try except blocks

ddtrace/propagation/http.py
ZStriker19: Is the reason for this that it's not good to log multiple messages for a single error even if that message gives more specific information?

ZStriker19: I changed this to `raise ValueError("W3C traceparent hex length incorrect: %s" % tp)`

ddtrace/propagation/http.py
ZStriker19: Ok, but if it's the caller of the method then how do we log the value of traceparent? Since in the caller it hasn't been extracted yet.

ddtrace/propagation/http.py
Kyle-Verhoog: 🤷‍♂️ I don't feel strongly one way or the other. I do like being louder about invalid data rather than pretending it's a valid version and that it isn't supported. Louder and more explicit would help the user more if something were to go wrong. Maybe just worth updating the comment to say this more explicitly.

ddtrace/propagation/http.py
```diff
@@ -483,10 +492,214 @@ def _extract(headers):
...
+            return None
+
+        trace_id, span_id, sampling_priority = traceparent_values
+        meta = {"traceparent": tp}
+
+        ts = _extract_header_value(_POSSIBLE_HTTP_HEADER_TRACESTATE, headers)
+        if ts:
+            # the value MUST contain only ASCII characters in the
+            # range of 0x20 to 0x7E except comma (,) and equal-sign (=)
+            if re.search(r"[^\x20-\x7E]+", ts):
```
mabdinur, line 665: Question: Would we need a similar check for traceparent or is it guaranteed to be ascii? I guess we'll just raise an exception if we can't process the traceparent. Should we follow the same approach here? Attempt to parse the tracestate and then catch the exception if it fails?

ddtrace/propagation/http.py
```diff
@@ -483,10 +492,214 @@ def _extract(headers):
...
+                log.debug("received invalid tracestate header: %r", ts)
+            else:
+                # store ts so we keep other vendor data, even if dd ends up being invalid
```
Kyle-Verhoog, line 668: this is so we can propagate it later on `inject`?

ddtrace/propagation/http.py
```diff
@@ -483,10 +492,214 @@ def _extract(headers):
...
+                    meta.update(other_propagated_tags)
+
+                    sampling_priority = _TraceContext._get_sampling_priority(sampling_priority, sampling_priority_ts)
+
+                    return Context(
+                        trace_id=trace_id,
+                        span_id=span_id,
+                        sampling_priority=sampling_priority,
+                        dd_origin=origin,
+                        meta=cast(_MetaDictType, meta),
```
Kyle-Verhoog, line 687: I don't think we should have to do a cast here. Can we declare the type of meta above when it's initialized?

ddtrace/propagation/http.py
```diff
@@ -483,10 +492,214 @@ def _extract(headers):
...
+                    )
+
+        return Context(
+            trace_id=trace_id,
+            span_id=span_id,
+            sampling_priority=sampling_priority,
+            meta=cast(_MetaDictType, meta),
+        )
```
mabdinur, line 695: nit: can we combine these return statements into one. Ex: Set `origin=None` on line 660 and remove the return statement on line 682. Returning one context object might be easier to read OR we can create the context object first and then update the fields as parse the tracestate.

tests/tracer/test_propagation.py
```diff
@@ -368,6 +373,328 @@ def test_get_wsgi_header(tracer):
...
+                assert tracetstate_values == expected_tuple
+        else:
+            tracetstate_values = _TraceContext._get_tracestate_values(ts_string)
+            if caplog.text:
+                for expected_log in expected_logging:
+                    assert expected_log in caplog.text
+
+
+@pytest.mark.parametrize(
+    "name,headers,expected_context",
```
mabdinur, line 628: I like the name parameter it makes the test so much easier to follow 👍

mabdinur (commented):
some minor nits. Overall LGTM

ddtrace/propagation/http.py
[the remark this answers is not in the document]
mabdinur: I left a similar comment. In another method we attempt to parse the header and then catch an exception if that fails. We might be able to use a similar approach here.

ddtrace/propagation/http.py
ZStriker19: yes that's correct

ddtrace/propagation/http.py
ZStriker19: Ah ok, I was following what other propagators did, let me give that a try.

ddtrace/internal/constants.py
```diff
@@ -1,7 +1,15 @@
...
+PROPAGATION_STYLE_W3C_TRACECONTEXT = "tracecontext"
+PROPAGATION_STYLE_ALL = (
+    PROPAGATION_STYLE_W3C_TRACECONTEXT,
+    PROPAGATION_STYLE_DATADOG,
+    PROPAGATION_STYLE_B3,
+    PROPAGATION_STYLE_B3_SINGLE_HEADER,
+)
 
+_TRACEPARENT_KEY = "traceparent"
+_TRACESTATE_KEY = "tracestate"
```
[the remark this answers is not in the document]
ZStriker19: I'll just remove them. I think I originally was using them more in various place but they're unnecessary now.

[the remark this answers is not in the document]
ZStriker19: Actually I guess I'll just move them for now and use both of them.

ddtrace/propagation/http.py
```diff
@@ -483,10 +492,214 @@ def _extract(headers):
...
+      - The tracestate header will have one list member added to it, ``dd``, which contains
+        values that would be in x-datadog-tags as well as those needed for propagation information.
+        The keys to the ``dd`` values have been shortened as follows to save space:
+        ``sampling_priority`` = ``s``
+        ``origin`` = ``o``
+        ``_dd.p.`` prefix = ``t.``
+    """
+
+    @staticmethod
+    def _is_valid_datadog_trace_tag_key(key):
```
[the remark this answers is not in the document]
ZStriker19: Sure!

ddtrace/propagation/http.py
```diff
@@ -483,10 +492,214 @@ def _extract(headers):
...
+        for list_mem in ts_l:
+            if list_mem.startswith("dd="):
+                # cut out dd= before turning into dict
+                list_mem = list_mem[3:]
+                dd = dict(item.split(":") for item in list_mem.split(";"))
+
+        # parse out values
+        if dd:
+            sampling_priority_ts = dd.get("s")
+            if sampling_priority_ts:
```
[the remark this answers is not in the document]
ZStriker19: ah thank you, I'll add a test for this as well.

ddtrace/propagation/http.py
```diff
@@ -483,10 +492,214 @@ def _extract(headers):
...
+
+            return sampling_priority_ts, other_propagated_tags, origin
+        else:
+            log.debug("no dd list member in tracestate from incoming request: %r", ts)
+            return None
+
+    @staticmethod
+    def _get_sampling_priority(sampling_priority_tp, sampling_priority_ts):
+        # type: (int, int) -> int
+        """
```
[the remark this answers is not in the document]
ZStriker19: Munir suggested I provide documentation for it for clarity. I think it's clear without, but I also wrote it so could be bias.

ddtrace/propagation/http.py
```diff
@@ -483,10 +492,214 @@ def _extract(headers):
...
+        # parse out values
+        if dd:
+            sampling_priority_ts = dd.get("s")
+            if sampling_priority_ts:
+                sampling_priority_ts = int(sampling_priority_ts)
+
+            origin = dd.get("o")
+            # need to convert from t. to _dd.p.
+            other_propagated_tags = {
+                re.sub("t.", "_dd.p.", k): v
```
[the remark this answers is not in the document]
ZStriker19: Ah yup that makes sense. Changed it. Thank you!

ddtrace/propagation/http.py
```diff
@@ -483,10 +492,214 @@ def _extract(headers):
...
+                for (k, v) in dd.items()
+                if (_TraceContext._is_valid_datadog_trace_tag_key(k))
+            }
+
+            return sampling_priority_ts, other_propagated_tags, origin
+        else:
+            log.debug("no dd list member in tracestate from incoming request: %r", ts)
```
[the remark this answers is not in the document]
ZStriker19: we can

ddtrace/propagation/http.py
```diff
@@ -483,10 +492,214 @@ def _extract(headers):
...
+        not present or less than or equal to 0, update the sampling priority field to 1.
+
+        3. If the traceparent sampled flag == 1, and the tracestate sampling priority is greater than 0,
+        use the extracted sampling priority value.
+
+        4. If the traceparent sampled flag == 0, and the tracestate sampling priority is less than or equal to 0,
+        use the extracted sampling priority value.
+        """
+
+        if sampling_priority_tp == 0 and (not sampling_priority_ts or sampling_priority_ts >= 0):
```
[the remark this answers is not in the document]
ZStriker19: ah this helped me understand a typing issue I was having. My typing for _get_sampling_priority was wrong it's `# type: (Optional(int), int) -> int` rather than `# type: (int, int) -> int`

ddtrace/propagation/http.py
```diff
@@ -483,10 +492,214 @@ def _extract(headers):
...
+        # currently 00 is the only version format, but if future versions come up we may need to add changes
+        if version != "00":
+            log.warning("unsupported traceparent version:%r, still attempting to parse", version)
+
+        if len(trace_id_hex) == 32 and len(span_id_hex) == 16 and len(trace_flags_hex) >= 2:
+            trace_id = _hex_id_to_dd_id(trace_id_hex)
+            span_id = _hex_id_to_dd_id(span_id_hex)
+
+            # All 0s are invalid values
+            assert trace_id != 0
```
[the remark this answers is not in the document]
ZStriker19: Yup, adding tests for this too.

ddtrace/propagation/http.py
ZStriker19: Oh nvm, one does that, the other sets the type when it inits like you're recommending. Changed it!

ddtrace/propagation/http.py
Kyle-Verhoog: can we comment this, otherwise it's not entirely obvious why it's saved

ddtrace/propagation/http.py
```diff
@@ -483,10 +492,214 @@ def _extract(headers):
...
+
+        2. If the traceparent sampled flag == 1, and the tracestate sampling priority is
+        not present or less than or equal to 0, update the sampling priority field to 1.
+
+        3. If the traceparent sampled flag == 1, and the tracestate sampling priority is greater than 0,
+        use the extracted sampling priority value.
+
+        4. If the traceparent sampled flag == 0, and the tracestate sampling priority is less than or equal to 0,
+        use the extracted sampling priority value.
+        """
```
majorgreys, line 629: Could this be simplified and be less a direct translation?

```suggestion
        """
        When the traceparent sampled flag is set, the Datadog sampling priority  is either 1 or a positive value of sampling priority if propagated in tracestate.
        
        When the traceparent sampled flag is not set, the Datadog sampling priority is either 0 or a negative value of sampling priority if propagated in tracestate.
        """
```

tests/tracer/test_propagation.py
```diff
@@ -368,6 +373,328 @@ def test_get_wsgi_header(tracer):
...
+        ("1_tp_0_ts", 1, 2, 2),
+        ("0_tp_None_ts", 0, None, 0),
+        ("1_tp_None_ts", 1, None, 1),
+        ("1_tp_-1_ts", 0, -1, -1),
+        ("1_tp_-1_ts", 1, -1, 1),
+        ("1_tp_-2_ts", 1, -2, 1),
+        ("0_tp_-2_ts", 0, -2, -2),
+    ],
+)
+def test_tracecontext_decide_sampling_priority(
```
majorgreys, line 405:
```suggestion
def test_tracecontext_get_sampling_priority(
```

tests/tracer/test_propagation.py
```diff
@@ -368,6 +373,328 @@ def test_get_wsgi_header(tracer):
...
+}
+
+TRACECONTEXT_HEADERS_VALID = {
+    _HTTP_HEADER_TRACEPARENT: "00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01",
+    _HTTP_HEADER_TRACESTATE: "dd=s:2;o:rum;t.dm:-4;t.usr.id:baz64,congo=t61rcWkgMzE",
+}
+
+
+@pytest.mark.parametrize(
+    "name,sampling_priority_tp,sampling_priority_ts,expected_sampling_priority",
```
majorgreys, line 389: We can use the `id` kwarg to `pytest.mark.parameterize` instead of `name`, however I think we can drop it entirely as I don't think there is an easy human readable gloss for each parameter tuple.

https://docs.pytest.org/en/7.1.x/example/parametrize.html#different-options-for-test-ids

tests/tracer/test_propagation.py
```diff
@@ -368,6 +373,328 @@ def test_get_wsgi_header(tracer):
...
+        else:
+            traceparent_values = _TraceContext._get_traceparent_values(headers)
+            assert traceparent_values == expected_tuple
+            if caplog.text:
+                for expected_log in expected_logging:
+                    assert expected_log in caplog.text
+
+
+@pytest.mark.parametrize(
+    "name,ts_string,expected_tuple,expected_logging,expected_exception",
```
majorgreys, line 482: Same as above though the names here provide a reader useful information. In this case, we could replace them with the `id` kwargs.

ddtrace/propagation/http.py
```diff
@@ -483,10 +492,214 @@ def _extract(headers):
...
+            sampling_priority = 0
+
+        elif sampling_priority_tp == 1 and (not sampling_priority_ts or sampling_priority_ts < 0):
+            sampling_priority = 1
+
+        elif sampling_priority_tp == 1 and sampling_priority_ts > 0:
+            sampling_priority = sampling_priority_ts
+
+        elif sampling_priority_tp == 0 and sampling_priority_ts < 0:
+            sampling_priority = sampling_priority_ts
```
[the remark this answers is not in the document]
ZStriker19: ah that's great, thanks!

ddtrace/propagation/http.py
ZStriker19: sure I'll add to the comment

ddtrace/propagation/http.py
```diff
@@ -483,10 +492,214 @@ def _extract(headers):
...
+        """
+
+        version, trace_id_hex, span_id_hex, trace_flags_hex = tp.split("-")
+        # check version is a valid hexadecimal
+        int(version, 16)
+        # currently 00 is the only version format, but if future versions come up we may need to add changes
+        if version != "00":
+            log.warning("unsupported traceparent version:%r, still attempting to parse", version)
+
+        if len(trace_id_hex) == 32 and len(span_id_hex) == 16 and len(trace_flags_hex) >= 2:
```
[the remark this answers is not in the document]
ZStriker19: Yeah at least two bits are required. You can't have just `0` or '1'

ddtrace/propagation/http.py
```diff
@@ -483,10 +492,214 @@ def _extract(headers):
...
+
+        # parse out values
+        if dd:
+            sampling_priority_ts = dd.get("s")
+            if sampling_priority_ts:
+                sampling_priority_ts = int(sampling_priority_ts)
+
+            origin = dd.get("o")
+            # need to convert from t. to _dd.p.
+            other_propagated_tags = {
```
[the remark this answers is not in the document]
ZStriker19: I think I'll keep inline for now, since it's only used once.

ddtrace/propagation/http.py
ZStriker19: For traceparent I guess it could theoretically be anything, but yeah we'd just raise an exception for it because we're casting and converting the values. For tracestate I'm not sure it's as simple, since there can always be unexpected key and tag values, we won't get an error if for example someone were to add `t.€€€€€:not_ascii` in there.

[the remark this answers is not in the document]
ZStriker19: I answer here why it's necessary to check: https://github.com/DataDog/dd-trace-py/pull/4614#discussion_r1034951498

But I agree, it does seem expensive. I'm just not sure of another way. I'll think about it.

ddtrace/propagation/http.py
```diff
@@ -483,10 +492,214 @@ def _extract(headers):
...
+                re.sub("t.", "_dd.p.", k): v
+                for (k, v) in dd.items()
+                if (_TraceContext._is_valid_datadog_trace_tag_key(k))
+            }
+
+            return sampling_priority_ts, other_propagated_tags, origin
+        else:
+            log.debug("no dd list member in tracestate from incoming request: %r", ts)
+            return None
```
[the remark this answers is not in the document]
ZStriker19: Oh interesting, I didn't know that this was preferred. Alright updating.

ddtrace/propagation/http.py
ZStriker19: Sure!

tests/tracer/test_propagation.py
ZStriker19: Why should we replace them with the id kwargs? In my mind it seems like it'd be easier to tell what a test is testing if the name is included.

ddtrace/propagation/http.py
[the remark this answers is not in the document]
ZStriker19: Munir was able to simplify the logic to this though, which I've added:

```
        if sampling_priority_tp == 0 and (not sampling_priority_ts or sampling_priority_ts >= 0):
            sampling_priority = 0

        elif sampling_priority_tp == 1 and (not sampling_priority_ts or sampling_priority_ts < 0):
            sampling_priority = 1
        else:
            # elif sampling_priority_tp == 1 and sampling_priority_ts > 0:
            # elif sampling_priority_tp == 0 and sampling_priority_ts < 0:
            sampling_priority = sampling_priority_ts
```

ddtrace/propagation/http.py
```diff
@@ -483,10 +494,192 @@ def _extract(headers):
...
+            origin = dd.get("o")
+            # need to convert from t. to _dd.p.
+            other_propagated_tags = {"_dd.p.%s" % k[2:]: v for (k, v) in dd.items() if (k.startswith("t."))}
+
+            return sampling_priority_ts, other_propagated_tags, origin
+        else:
+            return None, {}, None
+
+    @staticmethod
+    def _get_sampling_priority(sampling_priority_tp, sampling_priority_ts):
```
majorgreys, line 609: Non-blocker: using the names specific to each extracted value could help reading this code.

```suggestion
    def _get_sampling_priority(traceparent_sampled, tracestate_sampling_priority):
```

ddtrace/propagation/http.py
```diff
@@ -483,10 +492,214 @@ def _extract(headers):
...
+        ts_l = ts.split(",")
+        for list_mem in ts_l:
+            if list_mem.startswith("dd="):
+                # cut out dd= before turning into dict
+                list_mem = list_mem[3:]
+                dd = dict(item.split(":") for item in list_mem.split(";"))
+
+        # parse out values
+        if dd:
+            sampling_priority_ts = dd.get("s")
```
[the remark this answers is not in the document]
ZStriker19: Did this which works and I guess isn't too messy:
```
        if dd:
            sampling_priority_ts = dd.get("s")
            if sampling_priority_ts is not None:
                sampling_priority_ts_int = int(sampling_priority_ts)
            else:
                sampling_priority_ts_int = None
```

ddtrace/propagation/http.py
```diff
@@ -483,10 +494,195 @@ def _extract(headers):
...
+        0 or a negative value of sampling priority if propagated in tracestate.
+        """
+
+        if traceparent_sampled == 0 and (not tracestate_sampling_priority or tracestate_sampling_priority >= 0):
+            sampling_priority = 0
+
+        elif traceparent_sampled == 1 and (not tracestate_sampling_priority or tracestate_sampling_priority < 0):
+            sampling_priority = 1
+        else:
+            # elif sampling_priority_tp == 1 and sampling_priority_ts > 0:
```
ZStriker19, line 627: remove or reword

tests/tracer/test_propagation.py
ZStriker19: nvm, thanks for explaining in person

ddtrace/propagation/http.py
```diff
@@ -21,6 +23,7 @@
 from ..internal.constants import PROPAGATION_STYLE_B3
 from ..internal.constants import PROPAGATION_STYLE_B3_SINGLE_HEADER
 from ..internal.constants import PROPAGATION_STYLE_DATADOG
+from ..internal.constants import PROPAGATION_STYLE_W3C_TRACECONTEXT
```
mabdinur, line 26: We don't do this for any of the other propagation styles constants in this module. I am fine with keeping the code as is for consistency. But in the future we should avoid exposing constants in our public api. 

```suggestion
from ..internal.constants import PROPAGATION_STYLE_W3C_TRACECONTEXT as _PROPAGATION_STYLE_W3C_TRACECONTEXT
```

mabdinur (approved):
LGTM

ddtrace/propagation/http.py
majorgreys: Good point. I agree this should be internal (as is `_TraceContext`) but this aliasing is cumbersome if we were going to standardize on it. Hopefully this improves with refactoring of this module, i.e., split the file into file per propagator and this module only exposes the public interface for use in manual instrumentation.

majorgreys (changes requested):
`PROPAGATION_STYLE_W3C_TRACECONTEXT` should be internal.

ddtrace/propagation/http.py
ZStriker19: I'll update the comment!

majorgreys (approved):

mabdinur (approved):

PR #4710, opened 2022-12-02
fix(botocore): fix inclusion of lambda params.ClientContext tag

brettlangdon:
## Description
`params.ClientContext` is being included by default for calls to Lambda Invoke command. This is potentially sensitive data and should not be included by default.

This change adds `params.ClientContext` to the exclusion list for `lambda` spans.

## Relevant issue(s)

N/A

## Testing strategy

TODO

## Reviewer Checklist
- [ ] Title is accurate.

### DIFF
--- a/ddtrace/ext/aws.py
+++ b/ddtrace/ext/aws.py
@@ -15,6 +15,7 @@ EXCLUDED_ENDPOINT = frozenset({"kms", "sts", "sns", "kinesis", "events"})
 EXCLUDED_ENDPOINT_TAGS = {
     "firehose": frozenset({"params.Records"}),
     "secretsmanager": frozenset({"params.SecretString", "params.SecretBinary"}),
+    "lambda": frozenset({"params.ClientContext"}),
 }
 
 
--- a/releasenotes/notes/exclude-lambda-client-context-param-30c361694f41d3b6.yaml
+++ b/releasenotes/notes/exclude-lambda-client-context-param-30c361694f41d3b6.yaml
@@ -1,29 +1,4 @@
 ---
-#instructions:
-#    The style guide below provides explanations, instructions, and templates to write your own release note.
-#    Once finished, all irrelevant sections (including this instruction section) should be removed,
-#    and the release note should be committed with the rest of the changes.
-#
-#    The main goal of a release note is to provide a brief overview of a change and provide actionable steps to the user.
-#    The release note should clearly communicate what the change is, why the change was made, and how a user can migrate their code.
-#
-#    The release note should also clearly distinguish between announcements and user instructions. Use:
-#    * Past tense for previous/existing behavior (ex: ``resulted, caused, failed``)
-#    * Third person present tense for the change itself (ex: ``adds, fixes, upgrades``)
-#    * Active present infinitive for user instructions (ex: ``set, use, add``)
-#
-#    Release notes should:
-#    * Use plain language
-#    * Be concise
-#    * Include actionable steps with the necessary code changes
-#    * Include relevant links (bug issues, upstream issues or release notes, documentation pages)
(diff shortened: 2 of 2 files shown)


brettlangdon:
There are going to be some failing tests since the tests for validating context propagation through lambda invoke calls relies on the `params.ClientContext` tag on the produced span to validate the data. This strategy would need to change in order for us to enable this.

crebbo:
Sorry to jump on this, but it's not just ClientContext which can have sensitive data added via calls to botocore.

All function params seem to be traced.

An example call to invoke:


```
client = boto3.client("lambda", region_name="eu-west-2")

   response = client.invoke(
        FunctionName=os.environ["Example"],
        InvocationType="RequestResponse",
        Payload=json.dumps(event),
    )
```
results in:

```
params {
  ClientContext: xxx,
  FunctionName: Example,
  InvocationType: "RequestResponse",
  Payload: {} //potentially sensitive data
}
```

Same for other calls such as to dynamodb via boto library. Am I missing something (e.g: some config to tell it not to do this, I can't see where).

brettlangdon:
@crebbo no, there is no specific configuration for this right now.

Since each customer is likely to have their own desires/needs, I am thinking we need a way to expose this configuration to users to set themselves to have better control over it.

I haven't spent too much time thinking about this yet, but thinking about:

- enable/disable specific botocore APIs for tracing
- add to the per-service exclusion list, or a global exclusion list
- change the per-service exclusion list to a per-service (or global) inclusion list
- enable/disable parameter collecting for all services (not inclusion/exclusion, just "don't ever collect any params")


Do you think having this level of configurability would help your case? (I would probably still keep things as-is by default since most people do want the param metadata by default)

Is there another scenario/option you think would better fit your needs?

rl-michaelbrandeis:
There is a Datadog feature where payloads SHOULD be added
https://www.datadoghq.com/blog/troubleshoot-lambda-function-request-response-payloads/

In all other cases the user should expect Payloads to be excluded

If that requires changes to testing, then the tests should ensure payloads exist when requested and do not exist by default

I think the desire to control which botocore modules are instrumented is useful, and a feature we have requested, but should not overly complicate this issue and the serious security gap that has been created by unexpectedly adding payloads to traces.

brettlangdon:
For anyone following this conversation. A potential short term mitigation could be to manually update the `EXCLUDED_ENDPOINT_TAGS` constant in `ddtrace.ext.aws`.

```python
from ddtrace.ext.aws import EXCLUDED_ENDPOINT_TAGS

EXCLUDED_ENDPOINT_TAGS["lambda"] = frozenset({"params.ClientContext"})
```

majorgreys:
@crebbo The `DD_APM_REPLACE_TAGS` environment variable used in the Datadog Agent for scrubbing sensitive data requires that the `name` be the key of the tag value to replace or `*` for a replacement to be applied to all tag values. As far as I understand, you should not be able to apply a replacement for all tag values where the key is prefixed with `params.*`. Instead, you need to add a replace for each param that needs to be scrubbed.

```
DD_APM_REPLACE_TAGS=[
      {
        "name": "params.ClientContext",
        "pattern": "(?s).*"
      },
      {
        "name": "params.Payload",
        "pattern": "(?s).*"
      },
      ...
]
```

This Agent configuration change is an alternative to what @brettlangdon suggested above and might be a preferable short term mitigation.

@crebbo Can you confirm if that mitigation works for you?

crebbo:
Thanks @majorgreys

I think it's a seperate issue we can't do the following in the agent` {"name": "params.*", "pattern": "(?s).*"}]`, a team would now need to identify every botocore library call and it's params/nested structure to (ahead of time) exclude. This doesn't appear scalable or practical?

majorgreys:
Resolved in #4781

PR #4734, opened 2022-12-03
chore(debugger): compile expression lambda with access locals variables

shatzi:
## Description
Debugger expression lambda should be able to access local variables. so user can generate expression like:
`any(array, { @it.value = #arg1 && @it.id = ^local2 } )`

## Motivation
Align with dd-trace-java debugger features

## Reviewer Checklist
- [x] Title is accurate.


ddtrace/debugging/_expressions.py
```diff
@@ -132,17 +149,14 @@ def _compile_arg_predicate(ast):
...
-            + ca
-            + [
-                Instr("LOAD_CONST", fb),
-                Instr("CALL_FUNCTION", 2),
-            ]
-        )
+        if cb is None:
+            raise ValueError("Invalid lmabda: %r" % b)
+
+        return [Instr("LOAD_CONST", lambda i, c: f(c(_) for _ in i))] + ca + cb + [Instr("CALL_FUNCTION", 2)]
```
P403n1x87, line 159: I think we could probably simplify this slightly by doing something like
```suggestion
        return [Instr("LOAD_CONST", lambda i, c, _locals: f(c(_, _locals) for _ in i))] + ca + cb + [Instr("CALL_FUNCTION", 2)]
```
and the pushing a third argument to the stack, before `CALL_FUNCTION` with
```
Instr("LOAD_FAST", "_locals")
```
Then I think we shouldn't need the new helpers. WDYT?

tests/submod/stuff.py
```diff
@@ -135,3 +135,7 @@ def durationstuff(ns):
 
 def mutator(arg):
     arg.append(42)
+
+
+def ageChecker(people, age, name=None):
```
P403n1x87, line 140: snake case 🙂 
```suggestion
def age_checker(people, age, name=None):
```

tests/debugging/test_debugger.py
```diff
@@ -742,3 +742,31 @@ def test_debugger_function_probe_eval_on_exit():
 
         (snapshot,) = d.test_queue
         assert snapshot, d.test_queue
+
+
+def test_debugger_lamba_fuction_access_locals():
```
P403n1x87, line 747:
```suggestion
def test_debugger_lambda_fuction_access_locals():
```

P403n1x87 (approved):
🙌

mabdinur (approved):

mabdinur (commented):
This appears to be a new feature. Can we add a releasenote? Otherwise LGTM

P403n1x87:
> This appears to be a new feature. Can we add a releasenote? Otherwise LGTM

As with other small enhancements in this area, I think we can get away with no relno for now, and reduce the verbosity of the changelog, until these code paths will actually be exercised by users.

PR #4746, opened 2022-12-07
feat(debugger): add support for dynamic log line probe

shatzi:
## Description
Add debugger support for Dynamic Log Line Probes. Log Probe generate a log with a templated message that can capture method arguments and local variables. LogProbe are replacing SnapshotProbes if LogProbe.take_snapshot is set to true.

## Motivation
Allow users to inject dynamic logs into a running python process

## Design 
Log probe generate the templated message (using local variables and DDExpressions). 
Therefore, the cost of dynamic log is the same as generating a normal log message.
Moreover, LogProbe can have take_snapshot = True which capture the arguments + locals as well (previously known as SnapshotProbe). Capture snapshot is require more resources than a normal log message.
Therefore, the default rate limit for snapshot capture is 1 per sec. Otherwise, the default rate limit is 5000 per second.
Probe condition can be used to direct the LogProbe to be invoke only on specific cases using local variables and DDExpressions

## Testing strategy
Unit tests to very line log probes are hooks executed correctly and message is generated based on the given template.
Also checking when evaluation fails the message was still created with evaluationErrors.

## Reviewer Checklist
- [X] Title is accurate.
- [X] Description motivates each change.
- [X] No unnecessary changes were introduced in this PR.
- [X] Avoid breaking [API](https://ddtrace.readthedocs.io/en/stable/versioning.html#interfaces) changes unless absolutely necessary.
- [X] Tests provided or description of manual testing performed is included in the code or PR.
- [X] Release note has been added and follows the [library release note guidelines](https://ddtrace.readthedocs.io/en/stable/contributing.html#Release-Note-Guidelines), or else `changelog/no-changelog` label added.
- [X] All relevant GitHub issues are correctly linked.
- [X] Backports are identified and tagged with Mergifyio.


ddtrace/debugging/_async.py
```diff
@@ -2,7 +2,7 @@
 from types import CoroutineType
 from typing import List
 
-from ddtrace.debugging._snapshot.collector import SnapshotContext
+from ddtrace.debugging._capture.collector import SnapshotContext
```
P403n1x87, line 5: Does this renaming reflect the new data model?

ddtrace/debugging/_capture/collector.py
```diff
@@ -88,6 +90,7 @@ def __enter__(self):
     def __exit__(self, *exc_info):
         # type: (ExcInfoType) -> None
         try:
+            # BUG: I think @duration, @return are not passed to condition
```
P403n1x87, line 93: That's right, they are not yet passed in 🙁 .

ddtrace/debugging/_capture/model.py
```diff
@@ -0,0 +1,109 @@
...
+            return True
+
+        try:
+            return bool(condition(_locals or self.frame.f_locals))
+        except Exception as e:
+            raise ConditionEvaluationError(e)
+
+
+@attr.s
+class LogMessage(object):
```
P403n1x87, line 65: we should probably identify and extract a common base class to avoid redeclaring all these fields

ddtrace/debugging/_encoding.py
```diff
@@ -431,10 +430,10 @@ def _snapshot_v2(snapshot):
     }
 
 
-def _logger_v2(snapshot):
-    # type: (Snapshot) -> Dict[str, Any]
-    thread = snapshot.thread
-    code = snapshot.frame.f_code
+def _logger_v2(thread, frame):
+    # type: (Thread, FrameType) -> Dict[str, Any]
+    thread = thread
```
P403n1x87, line 435:
```suggestion
```

ddtrace/debugging/_encoding.py
```diff
@@ -431,10 +430,10 @@ def _snapshot_v2(snapshot):
     }
 
 
-def _logger_v2(snapshot):
-    # type: (Snapshot) -> Dict[str, Any]
-    thread = snapshot.thread
-    code = snapshot.frame.f_code
+def _logger_v2(thread, frame):
```
P403n1x87, line 433: I think we can perhaps rename the argument but keep the signature of this function? It looks like both snapshots and logs have the thread and frame subfields.

ddtrace/debugging/_probe/model.py
```diff
@@ -107,3 +108,16 @@ class MetricProbeKind(object):
 class MetricProbe(LineProbe):
     kind = attr.ib(type=Optional[str], default=None)
     name = attr.ib(type=Optional[str], default=None)
+
+
+@attr.s
+class TemplateSegment(object):
```
P403n1x87, line 114: What if we had subclasses for the str and the expr segments, each implementing its own evaluation method, to which we pass the locals?

ddtrace/debugging/_probe/remoteconfig.py
```diff
@@ -34,12 +36,12 @@ def _invalid_condition(_):
     return False
 
 
-INVALID_CONDITION = _invalid_condition
+INVALID_EXPRESSION = _invalid_condition
```
P403n1x87, line 39:
```suggestion
INVALID_EXPRESSION = _invalid_expression
```

tests/debugging/test_debugger.py
```diff
@@ -717,6 +719,11 @@ def test_debugger_condition_eval_then_rate_limit():
         for i in range(100):
             Stuff().instancestuff(i)
 
+        # Not sure this test is valid.
+        # Stuff().instancestuff(42) called only once.
+        # Therefore. it only check that condition executed before rate limit.
+        # If rate limit was not working this doesn't catch it as it only run once.
```
P403n1x87, line 725: The property we're testing here is that the condition is evaluated before rate limit. The rate limiter has its own suite of tests. So I think this should be fine 🙂 .

P403n1x87 (commented):
Some initial thoughts on this draft.

ddtrace/debugging/_capture/model.py
shatzi: refactored both classes into a shared based class called CaptureEvent. which have interface for enter/exit/line. each probe type would a different CaptureEvent that can capture different element from the current frame

ddtrace/debugging/_capture/collector.py
shatzi: refactored

ddtrace/debugging/_async.py
shatzi: new data model and also simplification of collector.

ddtrace/debugging/_capture/collector.py
```diff
@@ -0,0 +1,104 @@
...
+    def push(self, event):
+        # type: (CapturedEvent) -> None
+        if event.state == CaptureState.SKIP_COND:
+            meter.increment("skip", tags={"cause": "cond", "probe_id": event.probe.probe_id})
+        elif event.state == CaptureState.SKIP_COND_ERROR:
+            meter.increment("skip", tags={"cause": "cond_exc", "probe_id": event.probe.probe_id})
+        elif event.state == CaptureState.SKIP_RATE:
+            meter.increment("skip", tags={"cause": "rate", "probe_id": event.probe.probe_id})
+        elif event.state == CaptureState.DONE_AND_COMMIT:
+            self._enqueue(event)
```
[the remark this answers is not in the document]
shatzi: for events been collected you mean?

ddtrace/debugging/_capture/model.py
```diff
@@ -0,0 +1,93 @@
...
+
+@attr.s
+class CapturedEvent(six.with_metaclass(abc.ABCMeta)):
+    """Capture Event baseclass.
+
+    Used to store collected when a probe is triggered.
+    """
+
+    probe = attr.ib(type=Probe)  # type: Probe
+    frame = attr.ib(type=FrameType)  # type: FrameType
```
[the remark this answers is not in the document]
shatzi: My IDE doesn't auto complete if I don't have `# type: Probe` comment on.
Maybe I missing something?

ddtrace/debugging/_probe/model.py
```diff
@@ -104,7 +173,127 @@ class MetricProbeKind(object):
...
+
+
+@attr.s
+class LogProbeDetails(six.with_metaclass(abc.ABCMeta)):
+    template = attr.ib(type=Optional[str])
+    segments = attr.ib(type=Optional[List[TemplateSegment]])
+
+
+@attr.s
+class LogLineProbe(Probe, ProbeConditionDetails, LineLocationDetails, LogProbeDetails, SnapshotProbeDetails):
```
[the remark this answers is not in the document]
shatzi: I must say that is a great idea to reverse them.

ddtrace/debugging/_probe/remoteconfig.py
```diff
@@ -83,47 +101,73 @@ def _wrapper(*args, **kwargs):
     return _wrapper
 
 
+def _create_probe_based_on_location(args, attribs, line_class, function_class):
+    # type: (Dict[str, Any], Dict[str, Any], Type, Type) -> Any
+    if attribs["where"].get("sourceFile", None):
```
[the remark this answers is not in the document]
shatzi: not sure what you mean. the configuration only have "snapshotProbe_" as type, regardless of the where value.

What refactoring did you have in mind?

ddtrace/debugging/_probe/model.py
```diff
@@ -104,7 +173,127 @@ class MetricProbeKind(object):
...
+    expr = attr.ib(type=DslExpression, default=None)  # type: DslExpression
+
+    def eval(self, _locals):
+        # type: (Dict[str,Any]) -> Any
+        return self.expr.eval(_locals)
+
+
+@attr.s
+class SnapshotProbeDetails(six.with_metaclass(abc.ABCMeta)):
+    capture = attr.ib(type=CaptureLimits, eq=False)  # type: CaptureLimits
```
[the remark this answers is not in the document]
shatzi: nice

ddtrace/debugging/_probe/model.py
```diff
@@ -104,7 +173,127 @@ class MetricProbeKind(object):
...
+@attr.s
+class TemplateSegment(six.with_metaclass(abc.ABCMeta)):
+    @abc.abstractmethod
+    def eval(self, _locals):
+        # type: (Dict[str,Any]) -> str
+        pass
+
+
+@attr.s
+class ConstTemplateSegment(TemplateSegment):
```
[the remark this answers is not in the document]
shatzi: much better

ddtrace/debugging/_probe/model.py
```diff
@@ -72,27 +107,61 @@ def deactivate(self):
...
+        kwargs.setdefault("active", True)
+        kwargs.setdefault("rate", 1.0)
+        return f(*args, **kwargs)
+
+    return _wrapper
+
+
 @attr.s
-class ConditionalProbe(Probe):
+class ProbeConditionDetails(six.with_metaclass(abc.ABCMeta)):
```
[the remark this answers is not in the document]
shatzi: renamed

ddtrace/debugging/_capture/safe_getter.py
```diff
@@ -0,0 +1,293 @@
+from inspect import CO_VARARGS
```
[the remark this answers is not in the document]
shatzi: sounds like a good move. refactored

ddtrace/debugging/_capture/collector.py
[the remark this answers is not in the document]
P403n1x87: yes

ddtrace/debugging/_capture/model.py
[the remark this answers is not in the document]
P403n1x87: Ah weird. Perhaps it doesn't fully support the `attrs` package? Does mypy complain if you remove the comments?

ddtrace/debugging/_probe/remoteconfig.py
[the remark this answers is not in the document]
P403n1x87: Ah this was just a consideration. I don't know if there is anything much that we can do with the current model, unless we have a `type` field, or we can rely on some prefix, like `snapshotProbe_` to infer the type instead of looking for fields

ddtrace/debugging/_probe/model.py
```diff
@@ -44,13 +50,42 @@ def _resolve_source_file(path):
...
+@attr.s
+class DslExpression(object):
+    dsl = attr.ib(type=str, default=None)  # type: str
+    callable = attr.ib(type=Callable[[Dict[str, Any]], Any], default=None)  # type: Callable[[Dict[str, Any]], Any]
+
+    def eval(self, _locals):
+        try:
+            return self.callable(_locals)
+        except Exception as e:
+            raise ExpressionEvaluationError(self.dsl, e)
```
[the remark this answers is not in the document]
shatzi: good idea

ddtrace/debugging/_capture/model.py
```diff
@@ -0,0 +1,93 @@
...
+    thread = attr.ib(type=Thread)
+
+    context = attr.ib(type=Optional[Context], default=None)
+    args = attr.ib(type=Optional[List[Tuple[str, Any]]], default=None)
+    state = attr.ib(type=str, default=CaptureState.NONE)
+    errors = attr.ib(type=List[EvaluationError], factory=lambda: list())
+    timestamp = attr.ib(type=float, factory=time.time)
+    event_id = attr.ib(type=str, init=False, factory=lambda: str(uuid4()))
+
+    def _evalCondition(self, _locals=None):
```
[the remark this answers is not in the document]
shatzi: yup.

ddtrace/debugging/_probe/model.py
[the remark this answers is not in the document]
shatzi: merged snapshot and log probe into a single clas

ddtrace/debugging/_probe/model.py
```diff
@@ -227,73 +162,30 @@ def eval(self, _locals):
...
-def snapshot_probe_defaults(f):
-    def _wrapper(*args, **kwargs):
-        kwargs.setdefault("capture", CaptureLimits())
-        return f(*args, **kwargs)
-
-    return _wrapper
+class LogProbeMixin(object):
+    template = attr.ib(type=str)
+    segments = attr.ib(type=List[TemplateSegment])
+    capture_snapshot = attr.ib(type=bool)
```
P403n1x87, line 176: `capture_snapshot` feels a bit like two repeated words. How about `has_snapshot`, or `take_snapshot`, or simply `snapshot`/`capture`?

ddtrace/debugging/_probe/model.py
```diff
@@ -227,73 +162,30 @@ def eval(self, _locals):
...
-    @function_location_defaults
-    @snapshot_probe_defaults
-    def create(cls, **kwargs):
-        return LogFunctionProbe(**kwargs)
-
-
-LineProbes = Union[SnapshotLineProbe, LogLineProbe, MetricLineProbe]
-FunctionProbes = Union[SnapshotFunctionProbe, LogFunctionProbe, MetricFunctionProbe]
+LineProbes = Union[LogLineProbe, MetricLineProbe]
+FunctionProbes = Union[LogFunctionProbe, MetricFunctionProbe]
```
P403n1x87, line 191: I think we might want to use the singular for defining type aliases
```suggestion
LineProbe = Union[LogLineProbe, MetricLineProbe]
FunctionProbe = Union[LogFunctionProbe, MetricFunctionProbe]
```

ddtrace/debugging/_probe/registry.py
```diff
@@ -50,9 +50,9 @@ def set_message(self, message):
...
 def _get_probe_location(probe):
     # type: (Probe) -> Optional[str]
-    if isinstance(probe, LineLocationDetails):
+    if isinstance(probe, LineLocationMixin):
         return probe.source_file
-    elif isinstance(probe, FunctionLocationDetails):
+    elif isinstance(probe, FunctionLocationMixin):
         return probe.module
     else:
         raise ValueError("Unsupported probe type: {}".format(type(probe)))
```
P403n1x87, line 58: Non-blocking for this PR: I wonder if we could define a `location` property on the location mixins that would return a tuple with the location information. Then we could drop this helper and simply do `probe.location[0]` to get either the source file or the module. `probe.location[1]` would then be either the line number or the function name.

ddtrace/debugging/_capture/model.py
```diff
@@ -38,9 +38,9 @@ class CaptureState(object):
...
 @attr.s
 class CapturedEvent(six.with_metaclass(abc.ABCMeta)):
-    """Capture Event baseclass.
+    """Captured event base class.
 
-    Used to store collected when a probe is triggered.
+    Used to store captured data when a probe is triggered.
     """
 
     probe = attr.ib(type=Probe)  # type: Probe
```
P403n1x87, line 46: Should we define
```python
class ConditionalProbe(Probe, ProbeConditionMixin):
    pass
```
in the probe model and use that here? It seems that all the probes that we define there have the `ProbeConditionMixin`, so that can perhaps become the base class for all our conditional probes? Then we wouldn't have to do the casting here.

Perhaps we could also define an actual `LogProbe` as
```python
class LogProbe(ConditionalProbe, LogProbeMixin):
    pass
```
(maybe also rename `LogProbeMixin` into something else?) as I think this is what `CapturedEvent` actually deals with. WDYT?

ddtrace/debugging/_capture/snapshot.py
```diff
@@ -53,68 +60,104 @@ class Snapshot(CapturedEvent):
...
+            )
         self.duration = duration
         self.state = CaptureState.DONE_AND_COMMIT
+        if probe.evaluate_at != ProbeEvaluateTimingForMethod.ENTER:
+            self._eval_message(dict(_args))
 
     def line(self, _locals=None, exc_info=(None, None, None)):
         frame = self.frame
-        probe = cast(SnapshotLineProbe, self.probe)
+        probe = cast(LogLineProbe, self.probe)
```
P403n1x87, line 145: I think we should turn this `cast` into an `isinstance` check in this case.

ddtrace/debugging/_debugger.py
```diff
@@ -282,21 +277,10 @@ def _dd_debugger_hook(self, probe):
...
-                )
-                logMessage.line(exc_info=sys.exc_info())
-                self._collector.push(logMessage)
-                return
-
-            if isinstance(probe, SnapshotLineProbe):
-                # TODO: Global limit evaluated before probe conditions
-                if self._global_rate_limiter.limit() is RateLimitExceeded:
-                    return
+                if cast(LogLineProbe, probe).capture_snapshot:
```
P403n1x87, line 280:
```suggestion
                if probe.capture_snapshot:
```

P403n1x87 (commented):
This is shaping up nicely! 🙌 

Do you reckon we could maybe split this PR into a refactor + new feature? Given the current size, it might make it easier for others to do a review, if we can 🙏 .

ddtrace/debugging/_capture/model.py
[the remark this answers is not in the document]
shatzi: removed

ddtrace/debugging/_probe/remoteconfig.py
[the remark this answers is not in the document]
shatzi: I can't think of a way to remove this if (only move it to a different location

ddtrace/debugging/_probe/model.py
shatzi: renamed to take_snapshot

ddtrace/debugging/_probe/model.py
shatzi: renamed

ddtrace/debugging/_capture/model.py
shatzi: while I loved the idea. ConditionalProbe have the condition optional field - which require that mixing to always be the last in the chain.

ddtrace/debugging/_capture/snapshot.py
shatzi: refactored

ddtrace/debugging/_capture/collector.py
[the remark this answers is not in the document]
shatzi: added

ddtrace/debugging/_probe/registry.py
shatzi: refactored - not sure if that what you wanted to see

ddtrace/debugging/_probe/model.py
```diff
@@ -73,26 +93,62 @@ def deactivate(self):
...
 @attr.s
-class FunctionProbe(ConditionalProbe):
-    module = attr.ib(type=Optional[str], default=None)
-    func_qname = attr.ib(type=Optional[str], default=None)
+class LineLocationMixin(ProbeLocationMixin):
+    source_file = attr.ib(type=str, converter=_resolve_source_file)  # type: ignore[misc]
+    line = attr.ib(type=int)
+
+    def location(self):
+        return self.source_file
```
P403n1x87, line 134:
```suggestion
        return (self.source_file, self.line)
```

ddtrace/debugging/_probe/model.py
```diff
@@ -73,26 +93,62 @@ def deactivate(self):
...
+
+
+@attr.s
+class FunctionLocationMixin(ProbeLocationMixin):
+    module = attr.ib(type=str)
+    func_qname = attr.ib(type=str)
+    evaluate_at = attr.ib(type=ProbeEvaluateTimingForMethod)
+
+    def location(self):
+        return self.module
```
P403n1x87, line 151:
```suggestion
        return (self.module, self.func_qname)
```

ddtrace/debugging/_debugger.py
```diff
@@ -307,32 +306,39 @@ def _dd_debugger_wrapper(self, wrappers):
...
             # type: (FunctionType, Tuple[Any], Dict[str,Any]) -> Any
-            if not any(probe.active for probe in wrappers.values()):
+            active_probes = [probe for probe in wrappers.values() if probe.active]
+
+            if not active_probes:
                 return wrapped(*args, **kwargs)
 
             argnames = wrapped.__code__.co_varnames
-            frame = currentframe()
+            actual_frame = currentframe().f_back.f_back  # type: ignore[union-attr]
```
P403n1x87, line 315:
```suggestion
            actual_frame = sys._getframe(2)  # type: ignore[union-attr]
```

ddtrace/debugging/_probe/model.py
```diff
@@ -44,16 +55,25 @@ def _resolve_source_file(path):
...
-
-    def __attrs_post_init__(self):
-        self.limiter = RateLimiter(
+    tags = attr.ib(type=dict, eq=False)
+    active = attr.ib(type=bool, eq=False)
+    rate = attr.ib(type=float, eq=False)
+    limiter = attr.ib(type=RateLimiter, init=False, repr=False, eq=False)  # type: RateLimiter
+
+    @limiter.default
+    def _create_limiter(self):
```
P403n1x87, line 75: I think we could do
```suggestion
    def _(self):
```

ddtrace/debugging/_probe/model.py
```diff
@@ -73,26 +93,62 @@ def deactivate(self):
...
     probe.
     """
 
-    condition = attr.ib(type=Optional[Callable[[Dict[str, Any]], Any]], default=None)
+    condition = attr.ib(type=Optional[DDExpression])  # type: Optional[DDExpression]
+    condition_error_rate = attr.ib(type=float, eq=False)
+    condition_error_limiter = attr.ib(type=RateLimiter, init=False, repr=False, eq=False)  # type: RateLimiter
+
+    @condition_error_limiter.default
+    def _create_condition_error_limiter(self):
```
P403n1x87, line 108:
```suggestion
    def _(self):
```
?

ddtrace/debugging/_probe/registry.py
```diff
@@ -50,10 +49,8 @@ def set_message(self, message):
 
 def _get_probe_location(probe):
     # type: (Probe) -> Optional[str]
-    if isinstance(probe, LineProbe):
-        return probe.source_file
-    elif isinstance(probe, FunctionProbe):
-        return probe.module
+    if isinstance(probe, ProbeLocationMixin):
+        return probe.location()
```
P403n1x87, line 53:
```suggestion
        return probe.location()[0]
```

ddtrace/debugging/_capture/collector.py
```diff
@@ -0,0 +1,104 @@
...
+class CapturedEventWithContext(object):
+    """Snapshot context manager.
+
+    This is used to capture snapshot data for function invocation, whose return
+    value needs to be captured as well.
+    """
+
+    def __init__(
+        self,
+        collector,  # type: CapturedEventCollector
```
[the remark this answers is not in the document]
shatzi: fixed.

shatzi:
> This is shaping up nicely! 🙌
> 
> Do you reckon we could maybe split this PR into a refactor + new feature? Given the current size, it might make it easier for others to do a review, if we can 🙏 .

I think we can. it would make those two changes smaller. I was thinking doing model refactor + remote-config client as one PR and the second to replace snapshot probe with log-probe. I will give it a try.

tests/debugging/py35/test_async.py
```diff
@@ -3,7 +3,7 @@
 from ddtrace.debugging._async import dd_coroutine_wrapper
 
 
-class MockSnapshotContext:
+class MockCapturedEventWithcontext:
```
P403n1x87, line 6:
```suggestion
class MockCapturedEventWithContext:
```

ddtrace/debugging/_debugger.py
```diff
@@ -308,31 +306,39 @@ def _dd_debugger_wrapper(self, wrappers):
 
         def _(wrapped, args, kwargs):
             # type: (FunctionType, Tuple[Any], Dict[str,Any]) -> Any
-            if not any(probe.active for probe in wrappers.values()):
+            active_probes = [probe for probe in wrappers.values() if probe.active]
+
+            if not active_probes:
```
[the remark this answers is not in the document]
shatzi: I thought it would be more efficient as in the old implementation we go over the list twice. I would say maybe we should avoid the "disabled" stuff all together? I'm not sure this been used anymore.

ddtrace/debugging/_capture/collector.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+NO_RETURN_VALUE = object()
+
+
+class CapturedEventWithContext(object):
+    """CapturedEvent with context.
+
+    This is used to capture snapshot data for function invocation, whose return
+    value needs to be captured as well.
```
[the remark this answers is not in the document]
shatzi: updated comments. good enough?

P403n1x87 (dismissed):

tests/debugging/test_uploader.py
```diff
@@ -44,16 +44,13 @@ def test_uploader_full_buffer():
     size = 1 << 8
     with ActiveBatchJsonEncoder(size=size, interval=0.5) as uploader:
```
P403n1x87, line 45:
```suggestion
    with ActiveBatchJsonEncoder(size=size, interval=1) as uploader:
```
We see this test being flaky so we might find it beneficial to increase this interval

tests/debugging/test_uploader.py
```diff
@@ -44,16 +44,13 @@ def test_uploader_full_buffer():
...
         item = "hello" * 10
-        n = size // len(item)
+        item_len = len(item)
+        n = (size + item_len - 1) // item_len
```
P403n1x87, line 48: I would keep this as `n = size // len(item)` and then do a loop `for _ in range(2 * n)` so that we are sure enough, but without doing an infinite loop. WDYT?

shatzi: reverted

tests/debugging/test_uploader.py
shatzi: increased and also changed the second wait with awake()

P403n1x87 (approved):

ddtrace/debugging/_capture/collector.py
```diff
@@ -0,0 +1,110 @@
+from typing import Any
+from typing import Callable
+from typing import List
+from typing import Optional
+from typing import Tuple
+
+from ddtrace.debugging._capture.model import CaptureState
+from ddtrace.debugging._capture.model import CapturedEvent
+from ddtrace.debugging._encoding import BufferedEncoder
+from ddtrace.debugging._metrics import metrics
```
juanjux, line 10: nit (just complete when read, not preventing approval): if these are used only for typing (can't find other uses in the PR) they could go inside a:

```
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from ddtrace.debugging._encoding import BufferedEncoder
```

This way you avoid some imports at runtime.

majorgreys (dismissed):

P403n1x87 (dismissed):

majorgreys (approved):

P403n1x87 (approved):

PR #4748, opened 2022-12-07
fix(rate_limiter): reset last_update_ns every time _replenish is called

ZStriker19:
## Description
A bug was found in the Python tracer while using system-tests for span sampling. Essentially the issue was that upon instantiating the rule, the time since last update was set: `self.last_update_ns = compat.monotonic_ns()`. Then upon the first calling of `_replenish` if the max number of tokens was already hit, we would not reset the time:
```
        # If we are at the max, we do not need to add any more
        if self.tokens == self.max_tokens:
            return
```
Therefore, as long as the second hit of `_replenish`, when using the calculation `elapsed = (timestamp_ns - self.last_update_ns) / 1e9` it would often come out to greater than 1, since `self.last_update_ns` would be the amount of time since instantiation, not since last hit (which is what it should be). 

To fix this I simply added an update to `self.last_update_ns` even if we are at max tokens:

```
        # If we are at the max, we do not need to add any more
        if self.tokens == self.max_tokens:
            self.last_update_ns = timestamp_ns
            return
```

## Relevant issue(s)
 https://github.com/DataDog/system-tests/pull/690

## Testing strategy
Tests were added both for span sampling and for the rate limiter specifically.

## Checklist

## Reviewer Checklist
- [ ] Title is accurate.


ddtrace/internal/rate_limiter.py
```diff
@@ -113,6 +113,7 @@ def _replenish(self, timestamp_ns):
         # type: (int) -> None
         # If we are at the max, we do not need to add any more
         if self.tokens == self.max_tokens:
+            self.last_update_ns = timestamp_ns
```
P403n1x87, line 116: perhaps we can move this at the function entry and remove line 122? Or do a `try...finally` and update `last_update_ns` in the finally block?

ZStriker19: If we move it to the function entry though our calculation for `elapsed = (timestamp_ns - self.last_update_ns) / 1e9` will be wrong though, won't it?  I'll use a `try finally` block.

ZStriker19: Thanks for the recommendation! I couldn't think of the best way to do it.

P403n1x87: > If we move it to the function entry though our calculation for `elapsed = (timestamp_ns - self.last_update_ns) / 1e9` will be wrong though, won't it? I'll use a `try finally` block.

ah right 🤦 ! I didn't look carefully enough, sorry about that!

tests/tracer/test_rate_limiter.py
```diff
@@ -177,6 +177,17 @@ def test_rate_limiter_effective_rate_starting_rate():
     assert limiter.prev_window_rate == 0.5
 
 
+def test_rate_limiter_3():
+    limiter = RateLimiter(rate_limit=2)
+    for i in range(3):
+        decision = limiter.is_allowed(compat.monotonic_ns())
+        # the first two should be allowed, the third should not
+        if i in range(2):
```
brettlangdon, line 185:
```suggestion
        if i < 2:
```

ddtrace/tracer.py
```diff
@@ -704,7 +704,6 @@ def _start_span(
 
         if config.env:
             span.set_tag_str(ENV_KEY, config.env)
-
```
brettlangdon, line 707: unnecessary change?

tests/tracer/test_single_span_sampling_rule.py
```diff
@@ -170,3 +170,14 @@ def test_match_max_per_sec():
...
 
     assert_sampling_decision_tags(rate_limited_span, sample_rate=None, mechanism=None, limit=None)
+
+
+def test_max_per_sec_3():
+    rule = SpanSamplingRule(service="test_service", name="test_name", sample_rate=1.0, max_per_second=2)
+    for i in range(3):
+        span = traced_function(rule)
+        # the first two spans should be sampled, the third should not
+        if i in range(2):
```
brettlangdon, line 180:
```suggestion
        if i < 2:
```

ddtrace/internal/rate_limiter.py
brettlangdon: can you add a comment for that here? I almost made the same suggestion as Gab here.

brettlangdon (approved):

P403n1x87 (approved):

mabdinur (approved):

juanjux (approved):

PR #4756, opened 2022-12-08
chore(W3C): cover tracecontext header value edge cases

ZStriker19:
## Description
There are a few edge cases for W3C tracecontext that we need to cover:
1. hexadecimal values must be lowercase to be considered valid for the traceparent fields: https://www.w3.org/TR/trace-context/#traceparent-header-field-values
2. The incoming traceparent header value name and tracestate header name may be in any case (upper, lower, mixed) (actually already covered by our `HTTPPropagator.extract()` method since it does .lower(): https://www.w3.org/TR/trace-context/#header-name
3. The incoming traceparent header value may have leading or trailing whitespace characters, which may be the characters 0x09 (horizontal tab \t) and 0x20 (space): https://github.com/w3c/trace-context/blob/main/test/test.py#L407
4. The incoming tracestate header may have leading or trailing whitespace characters, which may be the characters 0x09: https://www.w3.org/TR/trace-context/#tracestate-header-field-values
5. for a traceparent whose version is "ff" it is invalid: https://www.w3.org/TR/trace-context/#version

## Testing Strategy
The testing for this is being covered by system-tests: https://github.com/DataDog/system-tests/pull/691

## Checklist

## Reviewer Checklist
- [ ] Title is accurate.


ddtrace/propagation/http.py
```diff
@@ -543,6 +543,16 @@ class _TraceContext:
         ``_dd.p.`` prefix = ``t.``
     """
 
+    # incoming traceparent header name and tracestate header name may be in any case (upper, lower, mixed)
+    # they may also have leading or trailing whitespace
+    @staticmethod
+    def _extract_header_value_tracecontext(possible_header_names, headers, default=None):
+        # type: (FrozenSet[str], Dict[str, str], Optional[str]) -> Optional[str]
+        headers_lower = {k.lower().strip(): v for k, v in headers.items()}
```
brettlangdon, line 551: don't these get normalized already below in `HTTPPropagator.extract`?

brettlangdon: if we do this again here, the performance is going to be **bad**

ddtrace/propagation/http.py
```diff
@@ -637,18 +650,21 @@ def _extract(headers):
         # type: (Dict[str, str]) -> Optional[Context]
 
         try:
-            tp = _extract_header_value(_POSSIBLE_HTTP_HEADER_TRACEPARENT, headers)
+            tp = _TraceContext._extract_header_value_tracecontext(_POSSIBLE_HTTP_HEADER_TRACEPARENT, headers)
             if tp is None:
                 log.debug("no traceparent header")
                 return None
+            # uppercase char in tp makes it invalid
+            if tp.isupper():
```
brettlangdon, line 658: this will only be true if all characters are uppercase, not if a single character is uppercase.

there has to be a better way, but `if any(c.isupper() for c in tp):` would work.

(we might want a generator expression here (what I have) instead of a list comprehension to ensure we don't process any more characters than needed to find the first uppercase)

brettlangdon: can we just `tp = tp.lower()`? does it matter?

brettlangdon (changes requested):
do we have test cases for these scenarios?

ZStriker19:
> do we have test cases for these scenarios?
https://github.com/DataDog/dd-trace-py/pull/4756#issue-1485092762

ddtrace/propagation/http.py
ZStriker19: In that case, should I just change the headers in HTTPPropagator.extract to be lowercase and have no white space? The issue I see with that, is that `b3 single header` is supposed to have whitespace.

ddtrace/propagation/http.py
```diff
@@ -554,6 +564,9 @@ def _get_traceparent_values(tp):
         version, trace_id_hex, span_id_hex, trace_flags_hex = tp.split("-")
         # check version is a valid hexadecimal, if not it's invalid we will move on to the next prop method
         int(version, 16)
+        # https://www.w3.org/TR/trace-context/#version
+        if version == "ff":
+            raise ValueError("'ff' is an invalid traceparent version")
```
mabdinur, line 569: Do we catch this exception in our library? If not, we should log a error and stop processing instead

ddtrace/propagation/http.py
ZStriker19: Would `if not tp.islower()` be better? "A string is lowercase if all cased characters in the string are lowercase and there is at least one cased character in the string"

ZStriker19: We can't accept it if there is an uppercase value. I asked Zach Montoya the same question about just lowering it and moving on.

ddtrace/propagation/http.py
ZStriker19: It's caught by the parent

ddtrace/propagation/http.py
brettlangdon: yeah, just add `.strip()` there

brettlangdon (approved):

mabdinur (approved):
Awesome work 😁

PR #4782, opened 2022-12-13
fix(tracing): fix usage of mapped_service when creating a new span

brettlangdon:
## Description

After getting the remapped service name we were not using it to determine if the `DD_VERSION` tag should be set, or when keeping track of all the service names seen by the tracer

## Testing strategy

Added two tests to validate that when setting `DD_SERVICE_MAPPING` we get the expected results for `version` tag and the collected list of services seen by the tracer.

## Reviewer Checklist
- [ ] Title is accurate.


ddtrace/tracer.py
```diff
@@ -632,6 +632,7 @@ def _start_span(
             else:
                 service = config.service
 
+        # DEV: We *MUST* use `mapped_service` from here on
```
P403n1x87, line 635: If we don't need the original service name beyond this point, can we just do
~~~python
service= config.service_mapping.get(service, service)
~~~
?

brettlangdon: I did think about this too, I was hoping someone would call it out.

This is probably the more clear change. Thanks!

mabdinur (approved):

Yun-Kim (approved):

PR #4791, opened 2022-12-14
fix(w3c): fix traceparent validation

mabdinur:
## Description

Running the W3C trace context parametric tests (introduced [here](https://github.com/DataDog/system-tests/pull/691)) surfaced four issues with how the python tracer propagates traceparent and tracestate headers. These issues are described below and are fixed in this PR.

### Edge Case 1 (most important)
Currently `trace_parent.islower()` is used to check if a traceparent contains lower case letters and digits (uppercase letters are invalid). Unfortunately `trace_parent.islower()` also evaluates to False if there are no lowercase letters are present in the traceparent (ex: traceparent=00-12312312003-1232131321 would be marked as invalid and dropped). This is incorrect. Fix: ddb9dc565aa6916728c1e105d07ab06a7cd214e9.

Since the traceparent should contain 48 randomly generated hex values it is unlikely to only contain digits. This edge case is very unlikely to occur.  

Current behavior:
- `00-123123123-23123213AAA` ->   uppercase letter detected traceparent is ignored
-  `00-123123123-23123213132` -> no lowercase letter detected traceparent is ignored [WRONG]
-  `00-123123123-2312322aaaa` -> lowercase letter detected traceparent is propagated 

Expected behavior:
- `00-123123123-23123213132AAAA` ->   uppercase detected traceparent is ignored
-  `00-123123123-23123213132` -> no uppercase detected traceparent is propagated 
-  `00-123123123-23123213132aaaa` -> no uppercase detected traceparent is propagated

### Edge Case 2

If traceparent contains an invalid characters (ex: period) do not propagate the traceparent. Start a new trace. Currently we can propagate invalid trace_ids (this is because we truncate the trace_id and only convert the last 16 digits to hex, we ignore all the characters in the first 16 digits). Fix: ae1e73e36d933af511b863ba4d4a83a768173e25

### Edge Case 3

Ensure that traceflags contain ONLY two digits (`00` or `01`). `000` and `001` are examples of invalid values that should not be propagated. Fix: adc4f80bf027361e670296c6269c19a84c94ed09

### Edge Case 4

Ensure traceparent version contain ONLY two digits (`00` or `01`). `000` and `001` are examples of invalid values that should not be propagated. Fix: 2e416c5abb5e6fd5438a2759682ee19b4ceb4650

### Edge Case 5

The W3C specification is additive. Although only traceparents with the version `00` are supported we should attempt to parse traceparent with different formats. 

Example: "01-12345678901234567890123456789012-1234567890123456-01-what-the-future-will-be-like"

In the traceparent example above we should parse the version, trace id, span id, and sample flag and ignore the trailing values.   Fix: 4cebe7514682787f6068754037fba569f4af3d60

### Edge Case 6

This edge case is not addressed in this PR. I am including it here for completeness. In the W3C tracecontext specification a tracer SHOULD set two http headers, one header should set the tracestate and the other should set the traceparent. However, if duplicate traceparent and tracestate headers are received, the tracer must processes and reconcile these headers (logic: 1) receive duplicate traceparent headers with different values -> drop these values and start a new trace.  2) receive duplicate tracestates with different tags -> combine the tracestates and propagate it). 

In the ddtrace library http headers are added to a dictionary, so duplicate http header values are overwritten. To address this edge case the tracer must be able to store detect and store all key value pairs in http headers. Since this edge case requires significant changes to distributed tracing and does not resolve a critical issue this work can be deferred.

## Testing 

Testing will covered by system tests.

## Reviewer Checklist
- [ ] Title is accurate.


ddtrace/propagation/http.py
```diff
@@ -646,7 +648,7 @@ def _extract(headers):
                 return None
             # uppercase char in tp makes it invalid:
             # https://www.w3.org/TR/trace-context/#traceparent-header-field-values
-            if not tp.islower():
+            if any(x.isupper() for x in tp):
```
mabdinur, line 651: edge case #1

ddtrace/propagation/http.py
```diff
@@ -552,16 +552,18 @@ def _get_traceparent_values(tp):
...
+            int(val, 16)
         # https://www.w3.org/TR/trace-context/#version
         if version == "ff":
             raise ValueError("'ff' is an invalid traceparent version")
         # currently 00 is the only version format, but if future versions come up we may need to add changes
         if version != "00":
             log.warning("unsupported traceparent version:%r, still attempting to parse", version)
 
-        if len(trace_id_hex) == 32 and len(span_id_hex) == 16 and len(trace_flags_hex) >= 2:
+        if len(version) == 2 and len(trace_id_hex) == 32 and len(span_id_hex) == 16 and len(trace_flags_hex) == 2:
```
mabdinur, line 566: edge cases 3 and 4

Yun-Kim (approved):

tests/tracer/test_propagation.py
```diff
@@ -448,17 +448,21 @@ def test_tracecontext_get_sampling_priority(sampling_priority_tp, sampling_prior
             "0-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01",
             # tp, trace_id, span_id, sampling_priority
             (11803532876627986230, 67667974448284343, 1),
-            ["unsupported traceparent version:'0', still attempting to parse"],
-            None,
+            [],
+            ValueError,
+        ),
+        (
+            "ff-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01",
```
brettlangdon, line 455: can we add some test cases for when the lengths are wrong, contain non-hex, have uppercase, are missing sections, etc ?

tests/tracer/test_propagation.py
```diff
@@ -448,17 +448,21 @@ def test_tracecontext_get_sampling_priority(sampling_priority_tp, sampling_prior
...
+            # tp, trace_id, span_id, sampling_priority
+            (11803532876627986230, 67667974448284343, 1),
```
brettlangdon, line 457: If we get a `ValueError` should this be `None` ?

ddtrace/propagation/http.py
```diff
@@ -73,6 +73,10 @@ def _possible_header(header):
 _POSSIBLE_HTTP_HEADER_TRACESTATE = _possible_header(_HTTP_HEADER_TRACESTATE)
 
 
+# https://www.w3.org/TR/trace-context/#traceparent-header-field-values
+_TRACEPARENT_HEX_REGEX = "^[a-f0-9]{2}-[a-f0-9]{32}-[a-f0-9]{16}-[a-f0-9]{2}$"
```
brettlangdon, line 77: we should pre-compile this, it'll make a _big_ difference.

```suggestion
_TRACEPARENT_HEX_REGEX = re.compile("^[a-f0-9]{2}-[a-f0-9]{32}-[a-f0-9]{16}-[a-f0-9]{2}$")
```

and then below you can do:

```python
if not _TRACEPARENT_HEX_REGEX.match(tp) or ... 
```

ddtrace/propagation/http.py
```diff
@@ -550,35 +554,29 @@ def _get_traceparent_values(tp):
...
+            raise ValueError("W3C traceparent hex is invalid: %s" % tp)
 
         version, trace_id_hex, span_id_hex, trace_flags_hex = tp.strip().split("-")
-        # check version is a valid hexadecimal, if not it's invalid we will move on to the next prop method
-        int(version, 16)
-        # https://www.w3.org/TR/trace-context/#version
-        if version == "ff":
-            raise ValueError("'ff' is an invalid traceparent version")
-        # currently 00 is the only version format, but if future versions come up we may need to add changes
         if version != "00":
```
brettlangdon, line 563: Removing the comment? it seems good to assert that this should be `00`, but we don't error if it isn't.

brettlangdon (commented):
some minor nits and suggestions for additional tests, but otherwise lgtm

ddtrace/propagation/http.py
```diff
@@ -550,35 +554,29 @@ def _get_traceparent_values(tp):
         Otherwise we extract the trace-id, span-id, and sampling priority from the
         traceparent header.
         """
+        tp = tp.strip()
+        if not re.match(_TRACEPARENT_HEX_REGEX, tp) or tp.startswith("ff"):
```
majorgreys, line 558: What's the benefit of adding this regex?

ddtrace/propagation/http.py
```diff
@@ -550,35 +554,29 @@ def _get_traceparent_values(tp):
...
+            # ff is an invalid traceparent version: https://www.w3.org/TR/trace-context/#version
+            raise ValueError("W3C traceparent hex is invalid: %s" % tp)
```
majorgreys, line 560:
```suggestion
            raise ValueError("Invalid traceparent version: %s" % tp)
```

tests/tracer/test_propagation.py
mabdinur: Yes, I made this fix

ddtrace/propagation/http.py
mabdinur: The comment was move to the function that validates the regex

ddtrace/propagation/http.py
mabdinur: The regex allows us to combine multiple checks into one. This regex validates that the traceparent does not have uppercase letters and that all the components of the traceparent have the correct length.

mabdinur: It's an optimization to the bug fixes that were introduced earlier.

ddtrace/propagation/http.py
```diff
@@ -73,6 +73,12 @@ def _possible_header(header):
 _POSSIBLE_HTTP_HEADER_TRACESTATE = _possible_header(_HTTP_HEADER_TRACESTATE)
 
 
+# https://www.w3.org/TR/trace-context/#traceparent-header-field-values
+# Future proofing: The traceparent spec is additive, future traceparent versions may contain more than 4 values
+# The regex below matches the version, trace id, span id, sample flag, and end-string/future values (if version>00)
+_TRACEPARENT_HEX_REGEX = re.compile("^([a-f0-9]{2})-([a-f0-9]{32})-([a-f0-9]{16})-([a-f0-9]{2})($|-.+)")
```
brettlangdon, line 79: Fun fact. (you don't have to make this change) You can add comments to regular expressions.


```suggestion
_TRACEPARENT_HEX_REGEX = re.compile(r"""
     ^                  # Start of string
     ([a-f0-9]{2})-     # 2 character hex version
     ([a-f0-9]{32})-    # 32 character hex trace id
     ([a-f0-9]{16})-    # 16 character hex span id
     ([a-f0-9]{2})      # 2 character hex sample flag
     ($|-.+)")          # end of string or start of an additional value
     """, re.VERBOSE)
```

brettlangdon: This could ("should") also be:

`"^([a-f0-9]{2})-([a-f0-9]{32})-([a-f0-9]{16})-([a-f0-9]{2})(-.+)?$"`

tests/tracer/test_propagation.py
```diff
@@ -444,7 +444,7 @@ def test_tracecontext_get_sampling_priority(sampling_priority_tp, sampling_prior
             None,
         ),
         (
-            "01-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01",
+            "01-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01-what-the-future-looks-like",
```
brettlangdon, line 447: Can we add a comment that `01` is required for this to work?

Also, we might want to add a test case for when we have version `00` and extra at the end.

Yun-Kim (dismissed):

ddtrace/propagation/http.py
```diff
@@ -73,6 +73,22 @@ def _possible_header(header):
...
+# Future proofing: The traceparent spec is additive, future traceparent versions may contain more than 4 values
+# The regex below matches the version, trace id, span id, sample flag, and end-string/future values (if version>00)
+_TRACEPARENT_HEX_REGEX = re.compile(
+    r"""
+     ^                  # Start of string
+     ([a-f0-9]{2})-     # 2 character hex version
+     ([a-f0-9]{32})-    # 32 character hex trace id
+     ([a-f0-9]{16})-    # 16 character hex span id
+     ([a-f0-9]{2})      # 2 character hex sample flag
+     ($|-.+)            # end of string or start of an additional value
```
brettlangdon, line 86:
```suggestion
     (-.+)?             # optional, start of any additional values
     $                  # end of string
```

brettlangdon: I cannot tell if I have the comments lined up correctly.

ddtrace/propagation/http.py
```diff
@@ -550,35 +566,35 @@ def _get_traceparent_values(tp):
...
         if version == "ff":
-            raise ValueError("'ff' is an invalid traceparent version")
-        # currently 00 is the only version format, but if future versions come up we may need to add changes
-        if version != "00":
+            # https://www.w3.org/TR/trace-context/#version
+            raise ValueError("ff is an invalid traceparent version: %s" % tp)
+        elif version != "00":
+            # currently 00 is the only version format, but if future versions come up we may need to add changes
             log.warning("unsupported traceparent version:%r, still attempting to parse", version)
+        elif version == "00" and future_vals != "":
```
mabdinur, line 581:
```suggestion
        elif version == "00" and future_vals is not None:
```

ddtrace/propagation/http.py
```diff
@@ -550,35 +567,35 @@ def _get_traceparent_values(tp):
         Otherwise we extract the trace-id, span-id, and sampling priority from the
         traceparent header.
         """
+        valid_tp_values = _TRACEPARENT_HEX_REGEX.match(tp.strip())
+        if valid_tp_values is None:
+            raise ValueError("Invalid traceparent version: %s" % tp)
+
+        version, trace_id_hex, span_id_hex, trace_flags_hex, future_vals = valid_tp_values.groups()
```
brettlangdon, line 574: Adding type hinting here would be helpful.

brettlangdon (approved):

Yun-Kim (approved):

PR #4814, opened 2022-12-21
fix(deps): Python 3.6 is now deprecated in attrs==22.2.0

brettlangdon:
## Description
`attrs==22.2.0` removed support for Python 3.5, and then made Python 3.6 deprecated.

https://www.attrs.org/en/stable/changelog.html#backwards-incompatible-changes

We are having an issue with our testing since these deprecation warnings are causing errors.

This change ensures we do not try to install `attrs>=22.2.0` if we have Python 3.6

The downside to this approach is that Python 3.6 support is only deprecated and not fully removed by `attrs==22.2.0` which means we might have a customer who _wants_ to use this version of attrs on Python 3.6. This change would mean they cannot make this change and update to this version.

## Reviewer Checklist
- [ ] Title is accurate.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -261,6 +261,7 @@ setup(
         "tenacity>=5",
         "attrs>=20; python_version>'2.7'",
         "attrs>=20,<22; python_version=='2.7'",
+        "attrs>=22,<22.2; python_version=='3.6'",
         "contextlib2<1.0; python_version=='2.7'",
         "cattrs",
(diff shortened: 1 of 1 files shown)


brettlangdon:
Alternative. If `attrs` is causing failures only on a select few test suites, we can pin `attrs<=22.2.0` when testing Python 3.6 for only those test suites.

PR #4822, opened 2022-12-21
chore(w3c): replace ~ with _ and = with ~

ZStriker19:
## Description
Upon review we found that `:` should be allowed in the tags value: https://docs.datadoghq.com/getting_started/tagging/#define-tags  

We were using `:` as a way to represent an "encoded" = in the propagated trace tag values, so we decided to change the "encoded" value of the `=` to `~`.

Therefore, we should allow `=` in tag values. In order to do this in tracestate, we need to encode it. Therefore we encode `~`s in the value as `_` first, and then encode `=` as `~`.

## Checklist

## Reviewer Checklist
- [ ] Title is accurate.

### DIFF
--- a/ddtrace/internal/utils/http.py
+++ b/ddtrace/internal/utils/http.py
@@ -158,11 +158,9 @@ def w3c_get_dd_list_member(context):
         tags.append("{}:{}".format(W3C_TRACESTATE_SAMPLING_PRIORITY_KEY, context.sampling_priority))
     if context.dd_origin:
         # the origin value has specific values that are allowed.
-        tags.append(
-            "{}:{}".format(
-                W3C_TRACESTATE_ORIGIN_KEY, re.sub(r",|;|~|[^\x20-\x7E]+", "_", context.dd_origin).replace("=", "~")
-            )
-        )
+        tag_val = re.sub(r",|;|~|[^\x20-\x7E]+", "_", context.dd_origin)
+        tag_val = w3c_encode_equals(tag_val)
+        tags.append("{}:{}".format(W3C_TRACESTATE_ORIGIN_KEY, tag_val))
 
     sampling_decision = context._meta.get(SAMPLING_DECISION_TRACE_TAG_KEY)
     if sampling_decision:
@@ -183,10 +181,9 @@ def w3c_get_dd_list_member(context):
             # for key replace ",", "=", and characters outside the ASCII range 0x20 to 0x7E
             # for value replace ",", ";", "~" and characters outside the ASCII range 0x20 to 0x7E
             # for value also replace = with ~
-            next_tag = "{}:{}".format(
-                re.sub("_dd.p.", "t.", re.sub(r",| |=|[^\x20-\x7E]+", "_", k)),
-                re.sub(r",|;|~|[^\x20-\x7E]+", "_", v).replace("=", "~"),
-            )
+            tag_val = re.sub(r",|;|~|[^\x20-\x7E]+", "_", v)
+            tag_val = w3c_encode_equals(tag_val)
+            next_tag = "{}:{}".format(re.sub("_dd.p.", "t.", re.sub(r",| |=|[^\x20-\x7E]+", "_", k)), tag_val)
             # we need to keep the total length under 256 char
             potential_current_tags_len = current_tags_len + len(next_tag)
             if not potential_current_tags_len > 256:
@@ -196,3 +193,8 @@ def w3c_get_dd_list_member(context):
                 log.debug("tracestate would exceed 256 char limit with tag: %s. Tag will not be added.", next_tag)
 
     return ";".join(tags)
+
+
+def w3c_encode_equals(tag_value):
+    # type: (str) -> str
+    return tag_value.replace("=", "~")
--- a/ddtrace/internal/utils/http.py
+++ b/ddtrace/internal/utils/http.py
@@ -19,7 +19,7 @@ from ddtrace.internal.sampling import SAMPLING_DECISION_TRACE_TAG_KEY
 from ddtrace.internal.utils.cache import cached
 
 
-_W3C_TRACESTATE_INVALID_CHARS_REGEX = r",|;|:|[^\x20-\x7E]+"
+_W3C_TRACESTATE_INVALID_CHARS_REGEX = r",|;|[^\x20-\x7E]+"
 
 
 Connector = Callable[[], ContextManager[compat.httplib.HTTPConnection]]
@@ -158,7 +158,12 @@ def w3c_get_dd_list_member(context):
         tags.append("{}:{}".format(W3C_TRACESTATE_SAMPLING_PRIORITY_KEY, context.sampling_priority))
     if context.dd_origin:
         # the origin value has specific values that are allowed.
-        tags.append("{}:{}".format(W3C_TRACESTATE_ORIGIN_KEY, re.sub(r",|;|=|[^\x20-\x7E]+", "_", context.dd_origin)))
+        tags.append(
+            "{}:{}".format(
+                W3C_TRACESTATE_ORIGIN_KEY, re.sub(r",|;|~|[^\x20-\x7E]+", "_", context.dd_origin).replace("=", "~")
+            )
+        )
+
     sampling_decision = context._meta.get(SAMPLING_DECISION_TRACE_TAG_KEY)
     if sampling_decision:
         tags.append("t.dm:{}".format(re.sub(_W3C_TRACESTATE_INVALID_CHARS_REGEX, "_", sampling_decision)))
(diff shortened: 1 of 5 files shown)


ddtrace/internal/utils/http.py
```diff
@@ -19,7 +19,7 @@
 from ddtrace.internal.utils.cache import cached
 
 
-_W3C_TRACESTATE_INVALID_CHARS_REGEX = r",|;|:|[^\x20-\x7E]+"
+_W3C_TRACESTATE_INVALID_CHARS_REGEX = r",|;|[^\x20-\x7E]+"
```
Yun-Kim, line 22: Is this related to the `~/_/=/~` replacement changes? If it is, can you add it to the PR description? Thanks!

ddtrace/internal/utils/http.py
```diff
@@ -158,7 +158,12 @@ def w3c_get_dd_list_member(context):
         tags.append("{}:{}".format(W3C_TRACESTATE_SAMPLING_PRIORITY_KEY, context.sampling_priority))
     if context.dd_origin:
         # the origin value has specific values that are allowed.
-        tags.append("{}:{}".format(W3C_TRACESTATE_ORIGIN_KEY, re.sub(r",|;|=|[^\x20-\x7E]+", "_", context.dd_origin)))
+        tags.append(
+            "{}:{}".format(
+                W3C_TRACESTATE_ORIGIN_KEY, re.sub(r",|;|~|[^\x20-\x7E]+", "_", context.dd_origin).replace("=", "~")
```
Yun-Kim, line 163: I wonder if we can just add `~_=~` into the regex itself instead of doing a replacement outside the regex operation? Feels like it would be cleaner this way unless these replacements have nothing to do with the regular expression.

ddtrace/internal/utils/http.py
```diff
@@ -158,7 +158,12 @@ def w3c_get_dd_list_member(context):
...
+            )
```
mabdinur, line 164: something like this might be more readable
```suggestion
        # replace unsupported characters with `_`
        tag_val = re.sub(r",|;|~|[^\x20-\x7E]+", "_", context.dd_origin)
        # Since `=` is an allowed character in datadog tags but not tracestate encode `=` as `~`
        tag_val = tag_val.replace("=", "~")
        tags.append("{}:{}".format(W3C_TRACESTATE_ORIGIN_KEY, tag_val)
```

mabdinur (commented):
Should we update https://github.com/DataDog/dd-trace-py/blob/v1.7.0rc2/ddtrace/propagation/http.py#L631 to decode `~` as `=`?

ddtrace/internal/utils/http.py
ZStriker19: Yeah I couldn't find a way to specify multiple replacements without a lambda for re.sub without using lambda which I don't think is supported in all versions.  I'll break it out into a separate function and make it multi-step for clarity.

ddtrace/internal/utils/http.py
ZStriker19: Yup just added, thanks!

ddtrace/internal/utils/http.py
ZStriker19: I couldn't figure out how to do this without using lambda, which I don't think is supported in all Python versions we support? However, if you know there's a way, I agree, it would be cleaner.

ZStriker19:
> Should we update https://github.com/DataDog/dd-trace-py/blob/v1.7.0rc2/ddtrace/propagation/http.py#L631 to decode `~` as `=`?

I think we're already doing this, no?

ddtrace/propagation/http.py
```diff
@@ -616,7 +616,8 @@ def _get_tracestate_values(ts):
             if list_mem.startswith("dd="):
                 # cut out dd= before turning into dict
                 list_mem = list_mem[3:]
-                dd = dict(item.split(":") for item in list_mem.split(";"))
+                # since tags can have a value with a :, we need to only split on the first instance of :
+                dd = dict(item.split(":", 1) for item in list_mem.split(";"))
```
mabdinur, line 620: Are we making this change to address the following edge case: `s::2` -> {`s` : `:2`} ? 

Do we need to handle the case where `:` is in the key? ex: `s:_h:2` -> {`s:_h` : `:2`} ? Is this possible?

ddtrace/propagation/http.py
```diff
@@ -627,8 +628,13 @@ def _get_tracestate_values(ts):
                 sampling_priority_ts_int = None
 
             origin = dd.get("o")
```
mabdinur, line 630: Can we replace `~` to `=` once?
```suggestion
            # we encode "=" as "~" in tracestate so need to decode here
            dd = dd.replace("~", "=")
            origin = dd.get("o")
```

mabdinur (commented):
Left some small nits but overall LGTM. Can we run the tracestate system tests with the latest commit in this branch: [parametric/test_headers_tracestate_dd.py: test_headers_tracestate_dd_propagate_propagatedtags](https://github.com/DataDog/system-tests/pull/691/files#diff-fcfebc861ddd73fc69d050c96029009426162e72b61faf5198c251fb49b67697)?

ddtrace/propagation/http.py
ZStriker19: Yes to the first question. 

To the second I'm almost positive the answer is no. Otherwise it would be impossible to determine what was the key and what was the value. E.g. [in the docs](https://docs.datadoghq.com/getting_started/tagging/#define-tags) we can see `env:staging:east` key is `env`, value is `staging:east`

ddtrace/propagation/http.py
ZStriker19: I don't think `replace` is a dict command, but we could do a dict comprehension if you think that'd be better? However then we'd be doing two dictionary comprehensions. One for the the char replacement, and one for creating `other_propagated_tags` below.

ddtrace/internal/utils/http.py
```diff
@@ -190,3 +193,8 @@ def w3c_get_dd_list_member(context):
                 log.debug("tracestate would exceed 256 char limit with tag: %s. Tag will not be added.", next_tag)
 
     return ";".join(tags)
+
+
+def w3c_encode_equals(tag_value):
```
mabdinur, line 198: Would it make sense to create helper that handles all the encoding (so regex sub + converting the `=` character)? Would this simplify `w3c_get_dd_list_member` function? 
We can maybe do the same with the the decode tag logic too.
 
 
```suggestion
def w3c_encode_tag(tag_value):
```

mabdinur, line 198: Would it make sense to create helper that handles all the encoding (so regex sub + converting the `=` character)? Would this simplify `w3c_get_dd_list_member` function? 
We can maybe do the same with the the decode tag logic too.
 
 
```suggestion
def w3c_encode_tag(tag_value):
```

mabdinur, line 198: Would it make sense to create helper that handles all the encoding (so regex sub + converting the `=` character)? Would this simplify `w3c_get_dd_list_member` function? 
We can maybe do the same with the the decode tag logic too.
 
 
```suggestion
def w3c_encode_tag(tag_value):
```

ddtrace/internal/utils/http.py
```diff
@@ -176,10 +179,10 @@ def w3c_get_dd_list_member(context):
             and k not in [SAMPLING_DECISION_TRACE_TAG_KEY, USER_ID_KEY]
         ):
             # for key replace ",", "=", and characters outside the ASCII range 0x20 to 0x7E
-            # for value replace ",", ";", ":" and characters outside the ASCII range 0x20 to 0x7E
+            # for value replace ",", ";", "~" and characters outside the ASCII range 0x20 to 0x7E
             next_tag = "{}:{}".format(
-                re.sub("_dd.p.", "t.", re.sub(r",| |=|[^\x20-\x7E]+", "_", k)),
-                re.sub(_W3C_TRACESTATE_INVALID_CHARS_REGEX, "_", v),
+                w3c_encode_tag("_dd.p.", "t.", w3c_encode_tag(r",| |=|[^\x20-\x7E]+", "_", k)),
```
ZStriker19, line 184: rather than w3c_encode_tag twice on the key here, should I change the method to take a dict like {regex: replacement} and loop through it? This is the only instance of two replacements being done on a single str so wasn't sure what was cleaner.

ddtrace/propagation/http.py
mabdinur: Makes sense. If we need to handle the second case then we should raise this issue to the team.

ddtrace/propagation/http.py
```diff
@@ -627,8 +633,13 @@ def _get_tracestate_values(ts):
...
+            if origin:
+                # we encode "=" as "~" in tracestate so need to decode here
+                origin = _TraceContext.decode_tag_val(origin)
             # need to convert from t. to _dd.p.
-            other_propagated_tags = {"_dd.p.%s" % k[2:]: v for (k, v) in dd.items() if (k.startswith("t."))}
+            other_propagated_tags = {
+                "_dd.p.%s" % k[2:]: _TraceContext.decode_tag_val(v) for (k, v) in dd.items() if (k.startswith("t."))
```
mabdinur, line 641: nit: you don't the brackets
```suggestion
                "_dd.p.%s" % k[2:]: _TraceContext.decode_tag_val(v) for (k, v) in dd.items() if k.startswith("t.")
```

ddtrace/internal/utils/http.py
ZStriker19: I think breaking this out into a separate function makes it cleaner.

mabdinur (approved):

P403n1x87 (approved):

Yun-Kim (approved):

PR #4884, opened 2023-01-11
feat: change log level and remove validation in Remote Configuration client

avara1986:
## Description
- Change log level to debug in Remote Configuration client
- Remove duplicate validation in Remote Configuration client (The agent has this validation)


gnufede (approved):

Kyle-Verhoog (commented):
Why should these be `debug` rather than `warning`?

juanjux (approved):

avara1986:
> Why should these be `debug` rather than `warning`?

Remote Configuration make a request each 5s by default, if there is an error, it could create a lot of noise in logs (e.g. the error with gevent + threads reports timeouts most of the time)

brettlangdon:
If we use the internal interface to create loggers, those loggers are rate limited by default to only once every 60 seconds.

PR #4888, opened 2023-01-11
chore(debugger): use new reference operators in JSON DSL

shatzi:
## Description

Update expression parser to match with the updated debugger backend expression json format. This change replaces the old syntax for field references with the new reference operators `ref`, `getmember` and `index`.

## Checklist

### DIFF
--- a/ddtrace/debugging/_expressions.py
+++ b/ddtrace/debugging/_expressions.py
@@ -9,12 +9,9 @@ Full grammar:
     predicate               =>  <direct_predicate> | <arg_predicate> | <value_source>
     direct_predicate        =>  {"<direct_predicate_type>": <predicate>}
     direct_predicate_type   =>  not | isEmpty | isUndefined
-    value_source            =>  <literal> | <value_reference> | <operation>
+    value_source            =>  <literal> | <operation>
     literal                 =>  <number> | true | false | "string"
     number                  =>  0 | ([1-9][0-9]*\.[0-9]+)
-    value_reference         =>  "<reference_prefix><reference_path>"
-    reference_prefix        =>  @ | ^ | # | .
-    reference_path          =>  identifier(.identifier)*
     identifier              =>  [a-zA-Z][a-zA-Z0-9_]*
     arg_predicate           =>  {"<arg_predicate_type>": [<argument_list>]}
     arg_predicate_type      =>  eq | ne | gt | ge | lt | le | any | all | and | or
@@ -22,10 +19,11 @@ Full grammar:
     argument_list           =>  <predicate>(,<predicate>)+
     operation               =>  <direct_operation> | <arg_operation>
     direct_opearation       =>  {"<direct_op_type>": <value_source>}
-    direct_op_type          =>  len | count
+    direct_op_type          =>  len | count | ref
     arg_operation           =>  {"<arg_op_type>": [<argument_list>]}
-    arg_op_type             =>  filter | substring
+    arg_op_type             =>  filter | substring | getmember | index
 """  # noqa
+from itertools import chain
 import re
 from types import FunctionType
 from typing import Any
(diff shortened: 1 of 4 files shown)


ddtrace/debugging/_expressions.py
```diff
@@ -183,9 +183,31 @@ def _compile_direct_operation(ast):
             raise ValueError("Invalid argument: %r" % arg)
         return [Instr("LOAD_CONST", len)] + value + [Instr("CALL_FUNCTION", 1)]
 
+    if _type == "ref":
+        if not isinstance(arg, str):
+            return None
```
P403n1x87, line 188: Whilst it makes sense for now to support just `str`, perhaps we could just try to compile `arg` as a `value_source`, in case in the future we'd want to offer some reflection-like capabilities. WDYT?

P403n1x87 (commented):
🙌  neato! I've left a question; also can we update the grammar at the top of this file as well? Otherwise LGTM!

ddtrace/debugging/_expressions.py
shatzi: LOVE THE IDEA.

Two thoughts, I don't see why ref op would need it -> do you have an example in mind?
Index op already doing it.
Maybe _compile_identifier could have this feature. so we can do dynamic GET_ATTR based on value source.
However, not sure how the DI syntax would look like, as "[]" is for index. so obj[some_var_as_attribute] will be translated to BIN_SUBSCR. 

So, I think leave it be for now. WDYT?

shatzi:
updated comments at top of FILE

ddtrace/debugging/_expressions.py
P403n1x87: > So, I think leave it be for now. WDYT?

👍  I don't think this is a blocker. This would just be a slight deviation from the stated syntax, since the argument is supposed to be a generic value and not just a literal string. The use case is if someone wanted to access a field dynamically, like it's done in Python with `getattr(obj, <dynamically computed attribute name>)`, but we can always add support for this later if we need to.

ddtrace/debugging/_expressions.py
```diff
@@ -321,7 +287,7 @@ def _compile_literal(ast):
 def _compile_value_source(ast):
     # type: (DDASTType) -> Optional[List[Instr]]
     # value_source  =>  <literal> | <value_reference> | <operation>
```
P403n1x87, line 289:
```suggestion
    # value_source  =>  <literal> | <operation>
```

ddtrace/debugging/_expressions.py
shatzi: yup. I think this also a good syntax: getattr(obj, var) - so for python developer it would be natural. then maybe the UI can even change it to `{"getmember":[{ref:"obj"},{"ref:":"var"}]}`. or even make it generic and have getmember(obj,field) as a method in EL.

P403n1x87 (approved):

ddtrace/debugging/_expressions.py
```diff
@@ -183,9 +180,31 @@ def _compile_direct_operation(ast):
...
     return None
 
 
+def _compile_identifier(ast):
+    # type: (DDASTType) -> Optional[List[Instr]]
+    # identifier  =>  [a-zA-Z][a-zA-Z0-9_]*
+    if not isinstance(ast, str) or not IDENT_RE.match(ast):
+        return None
+
+    return [Instr("LOAD_ATTR", ast)]
```
P403n1x87, line 205: Just a note for follow-up work: it's probably best to move the `LOAD_ATTR` outside of this rule, as strictly speaking this is not how the identifier rule should work. This is fine for now because this rule is only used by the `getmember` operation.

shatzi: refactored

P403n1x87 (approved):

P403n1x87 (approved):

brettlangdon (dismissed):

PR #4895, opened 2023-01-12
fix(profiler): indirectly access frame objects for tracebacks for Python 3.11

Yun-Kim:
## Description
#4511 added support for Python 3.11 for the profiler, but missed some `PyFrameObject`-related upgrades required to be compatible with Python 3.11, most notably in the stack collector's traceback module. Specifically, `PyFrameObjects` and its member values are now lazily created/computed, which means we can no longer directly access its member values. 
According to https://docs.python.org/3/whatsnew/3.11.html#pyframeobject-3-11-hiding, we need to use get functions to access its member values.

## Checklist

## Reviewer Checklist
- [ ] Title is accurate.


ddtrace/profiling/collector/_traceback.pyx
```diff
@@ -33,8 +44,14 @@ cpdef traceback_to_frames(traceback, max_nframes):
...
-            code = frame.f_code
-            lineno = 0 if frame.f_lineno is None else frame.f_lineno
+            # Python 3.11 moved PyFrameObject to internal C API and cannot be directly accessed from tstate
+            IF PY_MAJOR_VERSION >= 3 and PY_MINOR_VERSION >= 11:
+                code = PyFrame_GetCode(frame)
+                lineno = PyFrame_GetLineNumber(frame)
+                lineno = 0 if lineno is None else lineno
+            ELSE:
+                code = frame.f_code
+                lineno = 0 if frame.f_lineno is None else frame.f_lineno
```
Yun-Kim, line 54: @P403n1x87 I was unsure if this is also required to be changed, since this is taking a frame directly from `PyException_GetTraceback()`. I couldn't find in documentation if there's an accessor function to get a frame from a traceback - what do you think?

Yun-Kim (commented):
@P403n1x87 At first glance it looks like it's only the `_traceback.pyx` module that works with PyFrameObjects, but to be fair this is the only obvious fix relating to Python 3.11. Would love your feedback/second look 🙇

ddtrace/profiling/collector/_traceback.pyx
```diff
@@ -1,22 +1,45 @@
+from cpython.object cimport PyObject
+
+
+IF PY_MAJOR_VERSION >= 3 and PY_MINOR_VERSION >= 9:
+    cdef extern from "<Python.h>":
+        PyObject* PyFrame_GetCode(PyObject* frame)
+        PyObject* PyFrame_GetBack(PyObject* frame)
+        IF PY_MINOR_VERSION >= 11:
+            PyObject* PyFrame_GetLocals(PyObject* frame)
+            int PyFrame_GetLineNumber(PyObject* frame)
```
Yun-Kim, line 10: This isn't building on Python 3.11 properly with compilation errors like:
```
Object of type 'PyObject' has no attribute 'co_varnames'
```
Looks like I need to either import `PyFrameObject` directly or change around these types.

ddtrace/profiling/collector/_traceback.pyx
P403n1x87: If there is no documented API to do this then I guess this is the best we can do. Frame objects are guaranteed to be there with Python < 3.11, so we should be safe.

ddtrace/profiling/collector/_traceback.pyx
```diff
@@ -1,22 +1,46 @@
...
-        try:
-            if argname == "self":
-                return object.__getattribute__(type(value), "__name__")  # use type() and object.__getattribute__ to avoid side-effects
-            if argname == "cls":
-                return object.__getattribute__(value, "__name__")
-        except AttributeError:
-            return ""
+    # Python 3.11 moved PyFrameObject to internal C API and cannot be directly accessed from tstate
+    IF PY_MAJOR_VERSION >= 3 and PY_MINOR_VERSION >= 11:
+        co_varnames = PyFrame_GetCode(frame).co_varnames
```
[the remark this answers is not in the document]
Yun-Kim: This makes sense. I didn't catch that `PyCodeObject` member values also cannot be accessed directly anymore: https://docs.python.org/3/whatsnew/3.11.html#whatsnew311-c-api-porting

ddtrace/profiling/collector/_traceback.pyx
```diff
@@ -1,22 +1,46 @@
+from cpython.object cimport PyObject
+
+
+IF PY_MAJOR_VERSION >= 3 and PY_MINOR_VERSION >= 9:
```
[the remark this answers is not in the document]
brettlangdon: any reason this was resolved? seems like a decent suggestion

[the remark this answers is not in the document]
Yun-Kim: Looks like I forgot to push the changes after I resolved this 😓

ddtrace/profiling/collector/_traceback.pyx
```diff
@@ -1,22 +1,57 @@
...
-            return ""
+    # Python 3.11 moved PyFrameObject to internal C API and cannot be directly accessed from tstate
+    IF PY_MAJOR_VERSION > 3 or (PY_MAJOR_VERSION == 3 and PY_MINOR_VERSION >= 11):
+        code = PyFrame_GetCode(<PyFrameObject*> frame)
+        co_varnames = PyCode_GetVarnames(code)
+        if co_varnames:
+            argname = PyTuple_GetItem(co_varnames, 0)
+            try:
+                f_locals = PyFrame_GetLocals(<PyFrameObject*>frame)
+                value = <object>PyDict_GetItem(f_locals, argname)
```
Yun-Kim, line 34: According to https://docs.python.org/3/c-api/dict.html#c.PyDict_GetItem, this returns a borrowed reference. Am I correct in not needing to call `Py_XDECREF(value)` later?

ddtrace/profiling/collector/_traceback.pyx
```diff
@@ -1,22 +1,57 @@
+IF PY_MAJOR_VERSION > 3 or (PY_MAJOR_VERSION == 3 and PY_MINOR_VERSION >= 9):
+    from cpython cimport PyFrameObject
+    from cpython.object cimport PyObject
+    from cpython.ref cimport Py_XDECREF
+
+    cdef extern from "<Python.h>":
+        ctypedef struct PyCodeObject:
+            pass
+        PyObject* PyTuple_GetItem(PyObject *p, Py_ssize_t pos)
```
brettlangdon, line 9: why not use `from cpython.tuple import PyTuple_GetItem` ?

ddtrace/profiling/collector/_traceback.pyx
```diff
@@ -1,22 +1,57 @@
...
+        PyObject* PyDict_GetItem(PyObject *p, PyObject *key)
```
brettlangdon, line 10: same here, we should have `from cpython.dict import PyDict_GetItem` ?

ddtrace/profiling/collector/_traceback.pyx
```diff
@@ -1,22 +1,57 @@
...
+        code = PyFrame_GetCode(<PyFrameObject*> frame)
+        co_varnames = PyCode_GetVarnames(code)
+        if co_varnames:
+            argname = PyTuple_GetItem(co_varnames, 0)
+            try:
+                f_locals = PyFrame_GetLocals(<PyFrameObject*>frame)
+                value = <object>PyDict_GetItem(f_locals, argname)
+                Py_XDECREF(f_locals)
+                Py_XDECREF(argname)
+                Py_XDECREF(co_varnames)
```
brettlangdon, line 37: these should be handled by Cython for us ?

If you do want to manage these yourself, then you need to make sure they are called in ALL scenarios. e.g. if there is an exception raised then no XDECREF will get called. To handle better, you can predefine all of the `PyObject*` vars earlier and do a `try/finally` or make sure all branches will decref them.

ddtrace/profiling/collector/_traceback.pyx
```diff
@@ -51,9 +92,18 @@ cpdef pyframe_to_frames(frame, max_nframes):
...
         nframes += 1
-        if len(frames) < max_nframes:
-            code = frame.f_code
-            lineno = 0 if frame.f_lineno is None else frame.f_lineno
-            frames.append((code.co_filename, lineno, code.co_name, _extract_class_name(frame)))
-        frame = frame.f_back
+        # Python 3.11 moved PyFrameObject to internal C API and cannot be directly accessed from tstate
+        IF PY_MAJOR_VERSION > 3 or (PY_MAJOR_VERSION == 3 and PY_MINOR_VERSION >= 11):
+            if len(frames) < max_nframes:
+                code = PyFrame_GetCode(<PyFrameObject*> frame)
```
brettlangdon, line 98:
```suggestion
                code = <object>PyFrame_GetCode(<PyFrameObject*> frame)
```

or maybe `PyCodeObject*`, and then define the properties on the cdef extern above so it knows about `co_name`/`co_filename` ?

ddtrace/profiling/collector/_traceback.pyx
brettlangdon: Cython is probably INCREF and DECREFing it anyways? I think when we looked at the annotation it was doing that.

ddtrace/profiling/collector/_traceback.pyx
```diff
@@ -51,9 +92,18 @@ cpdef pyframe_to_frames(frame, max_nframes):
...
+                lineno = PyFrame_GetLineNumber(<PyFrameObject*> frame)
+                lineno = 0 if lineno is None else lineno
+                frames.append(((<object>code).co_filename, lineno, (<object>code).co_name, _extract_class_name(frame)))
+            frame = <object>PyFrame_GetBack(<PyFrameObject*> frame)
```
Yun-Kim, line 102: `PyFrame_GetBack()` returns a strong reference. Should I be calling `Py_XDECREF(frame)` even though we are iterating through these references?

ddtrace/profiling/collector/_traceback.pyx
```diff
@@ -1,22 +1,57 @@
...
-            if argname == "self":
-                return object.__getattribute__(type(value), "__name__")  # use type() and object.__getattribute__ to avoid side-effects
-            if argname == "cls":
-                return object.__getattribute__(value, "__name__")
-        except AttributeError:
-            return ""
+    # Python 3.11 moved PyFrameObject to internal C API and cannot be directly accessed from tstate
+    IF PY_MAJOR_VERSION > 3 or (PY_MAJOR_VERSION == 3 and PY_MINOR_VERSION >= 11):
+        code = PyFrame_GetCode(<PyFrameObject*> frame)
+        co_varnames = PyCode_GetVarnames(code)
```
Yun-Kim, line 29: `PyCode_GetVarnames()` returns a new reference. Should I be calling `Py_XDECREF(co_varnames)` as well here? Or every exit point in this function?

ddtrace/profiling/collector/_traceback.pyx
Yun-Kim: I'm hesitant to let Cython handle this for us, considering the source of the memory leak was due to a strong reference not being XDECREF'd. 
I'll go with your latter idea.

ddtrace/profiling/collector/_traceback.pyx
brettlangdon: the frame isn't actually needed anymore after L114 right? since we are creating/returning a list of tuples?

does the frame getting GC'd cause co_filename/co_name/etc to also get GC'd? are we appropriately incref/decref-ing those references?

ddtrace/profiling/collector/_traceback.pyx
brettlangdon: does Cython do that for you? probably look for something like `__py_x_var_co_varnames` (or w/e cython does for naming the variables in the C code).

there should be exit/jump points in the function definitions for errors or exiting and should show decrefs on the variables.

releasenotes/notes/profiler-3-11-traceback-fix-4c98c3ce96833df6.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    profiling: This fix resolves an issue in where the stack collector did not properly access PyFrameObject
+    member values, as PyFrameObject is now created and computed lazily in Python 3.11.
```
mabdinur, line 5: Reading this release note it's unclear how this issue manifests. Are we fixing segfaults, memory leaks, improving the accuracy of the profiler, etc....?

```suggestion
    profiling: This fix resolves an issue in where the stack collector did not properly access PyFrameObject
    member values, as PyFrameObject is now created and computed lazily in Python 3.11.

```

ddtrace/profiling/collector/_traceback.pyx
```diff
@@ -49,11 +102,25 @@ cpdef pyframe_to_frames(frame, max_nframes):
...
+        while pyframe != NULL:
+            nframes += 1
+            if len(frames) < max_nframes:
+                code = PyFrame_GetCode(pyframe)
+                lineno = PyFrame_GetLineNumber(pyframe)
+                lineno = 0 if lineno is None else lineno
+                frames.append(((<object>code).co_filename, lineno, (<object>code).co_name, _extract_class_name(<object>pyframe)))
+                Py_XDECREF(<PyObject*>code)
+            pyframe = PyFrame_GetBack(pyframe)
+            # FIXME: Where to Py_XDECREF(cframe)?
```
brettlangdon, line 117: `s/cframe/pyframe/` ?

ddtrace/profiling/collector/_traceback.pyx
```diff
@@ -49,11 +102,25 @@ cpdef pyframe_to_frames(frame, max_nframes):
...
-        nframes += 1
-        if len(frames) < max_nframes:
-            code = frame.f_code
-            lineno = 0 if frame.f_lineno is None else frame.f_lineno
-            frames.append((code.co_filename, lineno, code.co_name, _extract_class_name(frame)))
-        frame = frame.f_back
+    # Python 3.11 moved PyFrameObject members to internal C API and cannot be directly accessed
+    IF PY_MAJOR_VERSION > 3 or (PY_MAJOR_VERSION == 3 and PY_MINOR_VERSION >= 11):
+        cdef PyFrameObject* pyframe = <PyFrameObject*> frame
+        while pyframe != NULL:
```
brettlangdon, line 108: 👍🏻

ddtrace/profiling/collector/_traceback.pyx
```diff
@@ -49,11 +102,25 @@ cpdef pyframe_to_frames(frame, max_nframes):
...
+            nframes += 1
+            if len(frames) < max_nframes:
+                code = PyFrame_GetCode(pyframe)
+                lineno = PyFrame_GetLineNumber(pyframe)
```
brettlangdon, line 112: do you get a `PyObject*` here? does Cython handle `lineno is None` ok here, it didn't like it for `PyFrame_GetBack` though, right?

ddtrace/profiling/collector/_traceback.pyx
```diff
@@ -33,9 +78,17 @@ cpdef traceback_to_frames(traceback, max_nframes):
     while tb is not None:
         if nframes < max_nframes:
             frame = tb.tb_frame
-            code = frame.f_code
-            lineno = 0 if frame.f_lineno is None else frame.f_lineno
-            frames.insert(0, (code.co_filename, lineno, code.co_name, _extract_class_name(frame)))
+            # Python 3.11 moved PyFrameObject members to internal C API and cannot be directly accessed
+            IF PY_MAJOR_VERSION > 3 or (PY_MAJOR_VERSION == 3 and PY_MINOR_VERSION >= 11):
+                code = PyFrame_GetCode(<PyFrameObject*> frame)
```
brettlangdon, line 83: could frame ever be `NULL` here? does `PyFrame_GetCode` care?

should you do a similar cast as in the other function to a `PyFrameObject*` ?

ddtrace/profiling/collector/_traceback.pyx
```diff
@@ -33,9 +78,17 @@ cpdef traceback_to_frames(traceback, max_nframes):
...
+                lineno = PyFrame_GetLineNumber(<PyFrameObject*> frame)
+                lineno = 0 if lineno is None else lineno
+                frames.insert(0, ((<object>code).co_filename, lineno, (<object>code).co_name, _extract_class_name(frame)))
```
brettlangdon, line 86: should you cast `code` to `object` once? does it matter?

could `code` ever be `NULL` ?

ddtrace/profiling/collector/_traceback.pyx
```diff
@@ -1,23 +1,68 @@
...
+        co_varnames = PyCode_GetVarnames(code)
+        if PyTuple_Size(co_varnames) != 0:
+            argname = PyTuple_GetItem(co_varnames, 0)
+            try:
+                f_locals = PyFrame_GetLocals(pyframe)
+                value = PyDict_GetItem(f_locals, argname)
+            except KeyError:
+                return ""
+            try:
+                if <str>argname == "self":
```
brettlangdon, line 43: we cannot combine the below code and this?

e.g.

```python
cdef str argname
cdef object value

IF 3.11
    argname_pointer = PyTuple_GetItem(...)
    argname = <str>argname_pointer
    
    ...
ELSE
    argname = frame.f_code.co_varnames[0]
    ...

if argname == "self":
    return .... 
```

wdyt?

ddtrace/profiling/collector/_traceback.pyx
Yun-Kim: From what I understand, a traceback is always associated with a frame. @P403n1x87 can you confirm? 
I think we don't necessarily need to explicitly cast to a new `PyFrameObject*` variable here like in the other functions since we don't need to iterate through the object itself.

ddtrace/profiling/collector/_traceback.pyx
Yun-Kim: 💢

ddtrace/profiling/collector/_traceback.pyx
Yun-Kim: `PyFrame_GetLineNumber` returns an int, so I think we're okay using a `is None` check. `PyFrame_GetBack` returns a pointer `PyFrameObject*` so that's why we couldn't do a `is None` check I believe.

ddtrace/profiling/collector/_traceback.pyx
Yun-Kim: Thanks for the catch!

I am looking for some suggestions on where to call `Py_XDECREF(pyframe)` though. Since `PyFrame_GetBack(pyframe)` returns a strong reference of the `PyFrameObject*`, should I be decrementing the refcount for every time we call `PyFrame_GetBack(pyframe)`? But since we're using the result of the `PyFrame_GetBack()` in the while loop, I don't think I can decrement the refcount without interfering with the rest of the function.

releasenotes/notes/profiler-3-11-traceback-fix-4c98c3ce96833df6.yaml
Yun-Kim: This is a compatibility issue with Python 3.11.

ddtrace/profiling/collector/_traceback.pyx
P403n1x87: I believe "new reference" is synonym with "strong reference", so since `code` is `PyCodeObject*` we should xdecref here

ddtrace/profiling/collector/_traceback.pyx
P403n1x87: This looks fine to me. I also would expect `tb_frame != NULL` if `tb` is a valid traceback object.

ddtrace/profiling/collector/_traceback.pyx
P403n1x87: What happens if we cast the return value of `PyFrame_GetCode` to `<object>` instead?

ddtrace/profiling/collector/_traceback.pyx
P403n1x87: Note that `int` is a C `int` here. So it can't be a reference to `None`. My guess is that it would return either 0 or a negative number to signal an error

ddtrace/profiling/collector/_traceback.pyx
P403n1x87: `Py_XDECREF(pyframe)` has to be called every time you are done with that particular `pyframe` reference, so within the loop. The cast to `<object>` is taking care of ref counting when passing values around in Cython, so as long as our ref counting for when we invoke the C API directly is correct we should be good.

ddtrace/profiling/collector/_traceback.pyx
brettlangdon: let's make sure we have comments for any calls and the type of reference they return

ddtrace/profiling/collector/_traceback.pyx
```diff
@@ -33,9 +69,17 @@ cpdef traceback_to_frames(traceback, max_nframes):
...
-            code = frame.f_code
-            lineno = 0 if frame.f_lineno is None else frame.f_lineno
-            frames.insert(0, (code.co_filename, lineno, code.co_name, _extract_class_name(frame)))
+            # Python 3.11 moved PyFrameObject members to internal C API and cannot be directly accessed
+            IF PY_MAJOR_VERSION > 3 or (PY_MAJOR_VERSION == 3 and PY_MINOR_VERSION >= 11):
+                code = PyFrame_GetCode(<PyFrameObject*> frame)
+                lineno = PyFrame_GetLineNumber(<PyFrameObject*> frame)
+                lineno = 0 if lineno is None else lineno
+                frames.insert(0, ((<object>code).co_filename, lineno, (<object>code).co_name, _extract_class_name(frame)))
+                Py_XDECREF(<PyObject*>code)
```
brettlangdon, line 78: we might want a try/finally ? what happens if we get an error, we don't decref this.

ddtrace/profiling/collector/_traceback.pyx
```diff
@@ -49,11 +93,25 @@ cpdef pyframe_to_frames(frame, max_nframes):
...
+        while pyframe != NULL:
+            nframes += 1
+            if len(frames) < max_nframes:
+                code = PyFrame_GetCode(pyframe)
+                lineno = PyFrame_GetLineNumber(pyframe)
+                lineno = 0 if lineno is None else lineno
+                frames.append(((<object>code).co_filename, lineno, (<object>code).co_name, _extract_class_name(<object>pyframe)))
+                Py_XDECREF(<PyObject*>code)
+            pyframe = PyFrame_GetBack(pyframe)
+            # FIXME: Where to Py_XDECREF(pyframe)?
```
brettlangdon, line 108: you might need to do something like:

```python

cdef PyFrameObject* cur_frame = pyframe
pyframe = PyFrame_GetBack(cur_frame)
Py_XDECREF(cur_frame)
```

? and then a final `Py_XDECREF(pyframe)` after the while ?

Yun-Kim: I think this makes sense. Trying that now.

ddtrace/profiling/collector/_traceback.pyx
P403n1x87: I think we can adopt the convention that `lineno <= 0` means we failed to retrieve the line number, so we don't need to do further checks on the return value.

ddtrace/profiling/collector/_traceback.pyx
P403n1x87: Yep, the only "weak" line that I can see here in this regard is the "pure Python" `frames.insert` statement, as this could cause exceptions, but with the guarantees around `code` I think it is unlikely it will throw an exception.

ddtrace/profiling/collector/_traceback.pyx
P403n1x87: I would have thought that
~~~
Py_XDECREF(pyframe)
pyframe = PyFrame_GetBack(cur_frame)
~~~
is enough, but @brettlangdon's suggestion works too.

ddtrace/profiling/collector/_traceback.pyx
```diff
@@ -1,22 +1,62 @@
+IF PY_MAJOR_VERSION > 3 or (PY_MAJOR_VERSION == 3 and PY_MINOR_VERSION >= 9):
```
mabdinur, line 1: ?
```suggestion
IF PY_MAJOR_VERSION > 3 or (PY_MAJOR_VERSION == 3 and PY_MINOR_VERSION >= 11):
```

Yun-Kim:
Currently running into a segmentation fault after adding the `Py_XDECREF(pyframe)` in `pyframe_to_frames()`. It appears that this segfault appears after `pyframe_to_frames()` is run with asyncio task frames as the argument:
https://github.com/Yun-Kim/dd-trace-py/blob/0cf1580cfa638155bb7055a1767ca9083f53c24f/ddtrace/profiling/collector/stack.pyx#L338
				- https://github.com/Yun-Kim/dd-trace-py/blob/0cf1580cfa638155bb7055a1767ca9083f53c24f/ddtrace/profiling/collector/stack.pyx#L326
				- https://github.com/Yun-Kim/dd-trace-py/blob/0cf1580cfa638155bb7055a1767ca9083f53c24f/ddtrace/profiling/collector/_task.pyx#L125
				- https://github.com/Yun-Kim/dd-trace-py/blob/0cf1580cfa638155bb7055a1767ca9083f53c24f/ddtrace/profiling/collector/_task.pyx#L52

Yun-Kim:
@P403n1x87 Closing this PR until we can confirm that the issue persists after #5019, in which case I'll reopen this.

PR #4898, opened 2023-01-13
chore(internal): remove single-span-sampling tags from spans in sampled traces

emmettbutler:
This PR removes single-span-sampling tags from span contexts in sampled traces. These tags are redundant in sampled traces because from the perspective of an individual span they have the same effect as trace-level sampling tags; namely, marking the span as having been sampled. 

These are the specific tags/metrics removed:

* `_dd.span_sampling.mechanism`
* `_dd.span_sampling.rule_rate`
* `_dd.span_sampling.max_per_second`

These tags are removed from a given span when that span contains a well-formed `_sampling_priority_v1` tag indicating trace-level sampling.

This behavior is actually not new - it's validated in [this existing test](https://github.com/DataDog/dd-trace-py/blob/a01c18f20de2348ed34bde3ac2fe7a1e010a2d38/tests/tracer/test_processors.py#L422-L425). Thus this change is a defensive measure against these redundant tags still being sent even though we think that "should be" impossible.

Performance environment test doesn't show anything that looks concerning: https://ddstaging.datadoghq.com/dashboard/h8c-888-v2e/python-reliability-env-dashboard?from_ts=1673622354286&to_ts=1673623074286&live=false


releasenotes/notes/single-span-sampling-tags-fix-235b289006f2c635.yaml
```diff
@@ -0,0 +1,3 @@
+fixes:
+  - |
+    tracing: This fix removes single-span-sampling tags from span contexts in sampled traces, in which case they are redundant
```
brettlangdon, line 3: I'll let someone else comment to confirm, but we probably don't need a release note here since this is an internal implementation detail, so nothing should change for users.

meaning we can likely change the PR type from `fix:` to `chore(internal):`

emmettbutler: How do I stop this check from failing? https://github.com/DataDog/dd-trace-py/actions/runs/3911832547/jobs/6685680025

tests/snapshots/tests.integration.test_propagation.test_single_span_sampling_tags_are_removed_when_entire_trace_is_sampled.json
```diff
@@ -0,0 +1,23 @@
+[[
```
ZStriker19, line 1: Wouldn't this span be dropped by the Datadog agent, when it should be kept, since the single span sampling metrics are removed?

ddtrace/internal/processor/trace.py
```diff
@@ -79,9 +82,18 @@ def process_trace(self, trace):
...
                     return single_spans or None
 
+            at_least_one_sampled = False
             for span in trace:
+                # DEV: if the whole trace is being returned, single-span-sampling metrics don't matter anymore
+                if is_single_span_sampled(span):
+                    span._remove_metric(_SINGLE_SPAN_SAMPLING_MECHANISM)
+                    span._remove_metric(_SINGLE_SPAN_SAMPLING_RATE)
+                    span._remove_metric(_SINGLE_SPAN_SAMPLING_MAX_PER_SEC)
                 if span.sampled:
```
brettlangdon, line 92: this is a digression, I wonder if we can safely say... if at least one span is *not* sampled, then the whole trace isn't sampled?

vs the way we have it now is if at least one span is sampled then the trace should be sampled?

I am wondering if we can add a short circuit in here for a case when the trace wouldn't be kept?

(I don't think so, but asking the question anyways)

tests/integration/test_propagation.py
```diff
@@ -68,16 +69,36 @@ def downstream_tracer():
...
+    context = HTTPPropagator.extract(CONTEXT_HEADERS)
     downstream_tracer.context_provider.activate(context)
 
     with downstream_tracer.trace("p", service="downstream") as span:
         span.set_tag(MANUAL_DROP_KEY)
+
+
+@pytest.fixture
+def tracer_with_single_span_sampling_enabled():
+    with override_env(dict(DD_TRACE_SAMPLING_RULES='[{"sample_rate":1.0}]')):
```
brettlangdon, line 91: I think you want this configuration instead: https://docs.datadoghq.com/tracing/trace_pipeline/ingestion_mechanisms/?tab=python&code-lang=python#single-spans

E_TOO_MANY_SAMPLING_OPTIONS

ddtrace/internal/processor/trace.py
```diff
@@ -79,9 +82,18 @@ def process_trace(self, trace):
 
                     return single_spans or None
 
+            at_least_one_sampled = False
             for span in trace:
+                # DEV: if the whole trace is being returned, single-span-sampling metrics don't matter anymore
```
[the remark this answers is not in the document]
emmettbutler: @ZStriker19 is there something here that you think I should change? This comment is my current understanding of the issue that this PR is fixing.

tests/snapshots/tests.integration.test_propagation.test_single_span_sampling_tags_are_removed_when_entire_trace_is_sampled.json
ZStriker19: Oh, I didn't know that the agent cared about `_dd.p.dm`. Did we confirm that?

ddtrace/internal/processor/trace.py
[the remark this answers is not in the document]
ZStriker19: No you're good! I'll resolve!

ddtrace/internal/processor/trace.py
emmettbutler: I think this would work when the only sampler in use is the default sampler with its default `SamplingRule`, because it uses the [trace ID](https://github.com/DataDog/dd-trace-py/blob/5388335cfbe3c6d6f4a9ec914eb9422117bca5fc/ddtrace/sampler.py#L504) to make the sampling decision, meaning that `span.sampled` will be the same for all spans in the trace. However, I'm not sure we can maintain that guarantee when other `SamplingRules` are in use.

This question also doesn't seem to affect the performance of the change in this PR, because the single-span-sampling tag check needs to happen on all spans regardless of the state of `span.sampled`, meaning that a full iteration is still necessary.

As I see this currently there's no change necessary to this diff block. What am I missing? Thanks for your input.

brettlangdon: You are right that it should be the same for all spans, but because it is a mutable property on spans, they could be different.


```python
>>> import ddtrace
>>> with ddtrace.tracer.trace("root") as root:
...     with ddtrace.tracer.trace("child") as child:
...             pass
...
>>> root.sampled
True
>>> child.sampled
True
>>> child.sampled = False
>>> root.sampled
True
>>> child.sampled
False
```

they _should_ be the same, but there is no guarantee :/

brettlangdon: maybe we don't need to make a change, maybe more "FYI, should we care?"

tests/snapshots/tests.integration.test_propagation.test_single_span_sampling_tags_are_removed_when_entire_trace_is_sampled.json
emmettbutler: I haven't - I took it as a given based on some explanation I got from Brett. I'll see if I can figure out how to add that piece of knowledge to a test.

emmettbutler: @ZStriker19 Your comment here prompted me to take another look at the various interactions of types of sampling, and I fixed the tests in this pull request to be much clearer about the expected behavior.

emmettbutler:
This is much clearer than it was the first time around. It's ready for another review!

mabdinur:
We will need to update following shared tests in [system-tests](https://github.com/DataDog/system-tests/blob/main/parametric/test_span_sampling.py) repo:
```
test_span_sampling.py::test_single_rule_match_span_sampling_sss001
test_span_sampling.py::test_special_glob_characters_span_sampling_sss002
test_span_sampling.py::test_single_rule_only_service_pattern_match_span_sampling_sss004
test_span_sampling.py::test_multi_rule_keep_drop_span_sampling_sss006
test_span_sampling.py::test_sampling_rate_not_absolute_value_sss009
test_span_sampling.py::test_single_rule_always_keep_span_sampling_sss011
```

PR #4904, opened 2023-01-13
fix: remove url parts from http.url tag

majorgreys:
This fix removes unintended url parts in the `http.url` tag.

### DIFF
--- a/tests/tracer/test_trace_utils.py
+++ b/tests/tracer/test_trace_utils.py
@@ -342,6 +342,9 @@ def test_ext_service(int_config, pin, config_val, default, expected):
             {"id": "val", "name": "vlad"},
             None,
         ),
+        ("GET", "http://user:pass@localhost/", 0, None, None, None, None, None, None, None),
+        ("GET", "http://user@localhost/", 0, None, None, None, None, None, None, None),
+        ("GET", "http://user:pass@localhost/api?q=test", 0, None, None, None, None, None, None, None),
     ],
 )
 def test_set_http_meta(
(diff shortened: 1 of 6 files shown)


brettlangdon (approved):

mabdinur (approved):

tests/tracer/test_trace_utils.py
```diff
@@ -382,10 +385,15 @@ def test_set_http_meta(
         assert http.METHOD not in span.get_tags()
 
     if url is not None:
+        if url.startswith("http://user"):
+            expected_url = url[: url.index(":")] + "://" + url[url.index("@") + 1 :]
```
ZStriker19, line 389: I think a comment here might be helpful.

ZStriker19 (approved):
nit for comment but looks good

tests/tracer/test_trace_utils.py
brettlangdon:
```suggestion
            # Remove any userinfo that may be in the original url
            expected_url = url[: url.index(":")] + "://" + url[url.index("@") + 1 :]
```

brettlangdon (approved):

ZStriker19 (approved):

mabdinur (approved):

tests/tracer/test_trace_utils.py
Yun-Kim: This feels a little too engineered (i.e. we test our code against exactly what our code does). It would probably be better if we make the expected url part of the parameters instead?

Yun-Kim (commented):
Implementation looks good, but not the biggest fan of how we're testing for it.

PR #4926, opened 2023-01-17
refactor(profiling): eager code provenance file-package mapping

P403n1x87:
## Description

This refactor ensures that the file-to-package mapping is created as early as possible in the parent process, when code provenance is enabled, to avoid repeating the work in child worker processes.

## Checklist

## Reviewer Checklist
- [x] Title is accurate.


mabdinur (approved):
Can we benchmark the overhead of enabling `DD_PROFILING_ENABLE_CODE_PROVENANCE`? 

Since `DD_PROFILING_ENABLE_CODE_PROVENANCE` is disabled by default this is a low impact change and I am fine with merging this change as is. If/when we enable this feature by default this change has the potential to increase startup time for all ddtrace customers. Currently this module is only imported when profiling is enabled but I can see this changing unexpectedly.

P403n1x87:
> Can we benchmark the overhead of enabling `DD_PROFILING_ENABLE_CODE_PROVENANCE`?

Unfortunately I don't think we can make craft a sensible scenario for this. The current implementation relies on `importlib.metadata` which does quite a fair bit of I/O. The larger the number of package distributions in the environment, the longer the execution time in theory.

> If/when we enable this feature by default this change has the potential to increase startup time for all ddtrace customers. Currently this module is only imported when profiling is enabled but I can see this changing unexpectedly.

Indeed this will affect all users of the ddtrace profiler. The cost of initialising the mapping will have to be paid at some point and IMO it's best to pay this upfront _once_ than lazily and repeatedly in child processes.

ddtrace/profiling/exporter/_packages.py
```diff
@@ -90,30 +91,28 @@ def _build_package_file_mapping():
...
-
-_FILE_PACKAGE_MAPPING = None  # type: typing.Optional[typing.Union[typing.Dict[str, Distribution], _load_status]]
+_FILE_PACKAGE_MAPPING = None  # type: typing.Optional[typing.Dict[str, Distribution]]
+if asbool(os.getenv("DD_PROFILING_ENABLE_CODE_PROVENANCE", False)):
+    # DEV: If code provenance is enabled, create the mapping as soon as this
+    # module is imported. This should happen in the parent process to
+    # guarantee that the work is not repeated in child worker processes.
+    try:
+        _FILE_PACKAGE_MAPPING = _build_package_file_mapping()
+    except Exception:
```
ZStriker19, line 101: Is `Exception` used because there's a really wide range of exceptions that could occur here?

P403n1x87: Yes, I suppose we don't really know what could go wrong here, so we catch any errors and report them.

PR #4937, opened 2023-01-18
chore(profiler): fix regression test for frame object garbage collection in Python 3.11

Yun-Kim:
## Description
#4901 added a fix to properly decrement frametype objects in Python 3.11 in our stack collector, with a regression test. We found that this regression test does not properly assert that frametype objects were all garbage collected, which this test fixes.

Specifically, the previous change was looking for only frame objects related to a function call `_foo()`. However, after reverting the fix from #4901 this test still passed, because all frametype objects have been garbage collected. This change makes a broader assertion instead that all frametype objects have been garbage collected, which does expectedly fail if the fix from #4901 is reverted.

## Checklist

## Reviewer Checklist
- [ ] Title is accurate.
- [ ] Telemetry is meaningful, actionable and does not have the potential to leak sensitive data.


tests/profiling/collector/test_stack.py
```diff
@@ -761,7 +761,4 @@ def _foo():
             _foo()
 
     gc.collect()  # Make sure we don't race with gc when we check frame objects
-    frametype_objs = [obj for obj in gc.get_objects() if isinstance(obj, FrameType)]
-    for obj in frametype_objs:
-        # Ensure that all frames relating to _foo() have been gc'd already
-        assert obj.f_code.co_name != "_foo"
+    assert sum(isinstance(_, FrameType) for _ in gc.get_objects()) == 0
```
brettlangdon, line 764: wouldn't we potentially have frame objects from the current stack here?

what determines when a `FrameType` object is created? is that not the same as the `PyFrame_Object`(sp?) ?

Yun-Kim: Yes, `FrameType` is referencing `PyFrame_Object`. This is created in the stack collector https://github.com/DataDog/dd-trace-py/blob/a01c18f20de2348ed34bde3ac2fe7a1e010a2d38/ddtrace/profiling/collector/stack.pyx#L248. The fix from #4901 sought to properly decrement its reference counts so that they can be garbage collected instead of leaking strong references.

We do `gc.collect()` the line above to make sure any frame objects that are being tracked are collected before we make this assertion.

P403n1x87 (approved):

mabdinur (approved):

PR #4938, opened 2023-01-18
chore(tracing): add language tag to all chunk root spans

wconti27:
## Description
Add the language tag to every first span within a chunk of spans. 

## Motivation
Implement new Datadog span specification.

## Testing strategy
Snapshots updated.

## Reviewer Checklist
- [x] Title is accurate.


ddtrace/internal/processor/trace.py
```diff
@@ -126,6 +126,7 @@ def process_trace(self, trace):
             return trace
 
         ctx._update_tags(chunk_root)
+        chunk_root.set_tag_str("language", "python")
```
mabdinur, line 129: Is `language` a trace level tag? If so we should reorder these lines:

```suggestion
        chunk_root.set_tag_str("language", "python")
        ctx._update_tags(chunk_root)
```

brettlangdon: not sure we want this to be over-rideable?

tests/tracer/runtime/test_runtime_metrics.py
```diff
@@ -131,12 +131,12 @@ def test_only_root_span_runtime_internal_span_types(self):
                 assert root.get_tag("language") == "python"
                 assert child.get_tag("language") is None
 
-    def test_span_no_runtime_tags(self):
+    def test_span_always_runtime_tags(self):
```
brettlangdon, line 134: I am not sure we even need this test case anymore, since we always have the language tag on spans, we could remove this test and instead add one for the trace tags processor to ensure it always sets it on the chunk root span.

brettlangdon (commented):
suggestion for tests

ddtrace/internal/processor/trace.py
mabdinur: On second thought is probably a bad idea. A trace can contain chunks/spans from different languages. Depending on how the agent sets trace level tags this could overwrite values set by other libraries. We should probably add comment stating this.

brettlangdon: I am not sure I follow your line of thinking.

emmettbutler: @mabdinur how can a trace contain spans from multiple languages?

ZStriker19: I think the order looks correct, and I'm not aware of any special agent/backend trace level tags magic that occurs regarding whatever tags are set on the local root span, which in this case I'm pretty sure will only ever be a Python span.

brettlangdon: > trace level tags magic

👍🏻 correct, they'd only be for the portion of the trace from this process. we don't have any way to set a trace "global" (?) tag.

ZStriker19 (approved):

tests/tracer/runtime/test_runtime_metrics.py
wconti27: Did this.

ddtrace/internal/processor/trace.py
mabdinur: Yup, I was mistaken. I thought we submitted "tracer level tags" to the agent but this is not the case. I'll resolve this thread and my suggestion

mabdinur:
rq and integration_testagent snapshot tests need to be updated: https://app.circleci.com/pipelines/github/DataDog/dd-trace-py/27138/workflows/7558c62a-a800-4c60-a88c-78642e9406ef/jobs/1854933, https://app.circleci.com/pipelines/github/DataDog/dd-trace-py/27138/workflows/7558c62a-a800-4c60-a88c-78642e9406ef/jobs/1854933.

 Aside from that LGTM

mabdinur (dismissed):

ZStriker19 (dismissed):

brettlangdon (approved):

mabdinur (approved):

PR #4943, opened 2023-01-19
feat: telemetry metrics

avara1986:
## Description
Telemetry metrics, draft PR to discuss


ddtrace/internal/writer.py
```diff
@@ -599,15 +600,13 @@ def flush_queue(self, raise_exc=False):
...
-                    # This really isn't ideal as now we're going to do a ton of socket calls.
-                    self.dogstatsd.distribution("datadog.tracer.http.sent.bytes", len(encoded))
-                    self.dogstatsd.distribution("datadog.tracer.http.sent.traces", n_traces)
+                if self._report_metrics:
+                    telemetry_writer.add_count_metric(TELEMETRY_TRACER, "http.sent.bytes", len(encoded), {})
+                    telemetry_writer.add_count_metric(TELEMETRY_TRACER, "http.sent.traces", n_traces, {})
                     for name, metric in self._metrics.items():
-                        self.dogstatsd.distribution("datadog.tracer.%s" % name, metric["count"], tags=metric["tags"])
+                        telemetry_writer.add_count_metric(
+                            TELEMETRY_TRACER, "%s" % name, metric["count"], metric["tags"]
```
mabdinur, line 608:
```suggestion
                            TELEMETRY_TRACER, name, metric["count"], metric["tags"]
```

tests/telemetry/test_writer.py
```diff
@@ -189,6 +193,114 @@ def test_telemetry_graceful_shutdown(telemetry_writer, test_agent_session):
     assert events[1]["request_type"] == "app-started"
 
 
+def test_send_tracers_count_metric(mock_time, telemetry_writer, test_agent_session):
+    telemetry_writer.add_count_metric(TELEMETRY_TRACER, "test-metric", 1, {"hi": "HELLO", "NAME": "CANDY"})
+    telemetry_writer.add_count_metric(TELEMETRY_TRACER, "test-metric", 1, {})
```
mabdinur, line 198: These two data points should not be combined. This was a bug in my initial implementation. Tags should be associated with individual data points not metric names. 

We should only combine datapoints into one series object if they contain the same tags. For now we should avoid combing metrics and this optimization in a follow up PR

ddtrace/internal/telemetry/constants.py
```diff
@@ -0,0 +1,2 @@
+TELEMETRY_TRACER = "tracers"
+TELEMETRY_APPSEC = "appsec"
```
mabdinur, line 2: We can move these constants to  `ddtrace/internal/constants.py`

mabdinur (commented):
Something to note:

When a process forks we set `TelemetryWriter._forked = True` and prevent [appstarted](https://github.com/DataDog/dd-trace-py/blob/v1.7.2/ddtrace/internal/telemetry/writer.py#L203), [appclosed](https://github.com/DataDog/dd-trace-py/blob/v1.7.2/ddtrace/internal/telemetry/writer.py#L228) and [heartbeat](https://github.com/DataDog/dd-trace-py/blob/v1.7.2/ddtrace/internal/telemetry/writer.py#L215) events from being queued ([link](https://github.com/DataDog/dd-trace-py/blob/v1.7.2/ddtrace/internal/telemetry/writer.py#L57)). These events should only be sent on the "application" level not on individual processes. 

I am not sure if this impacts the metrics event or if we even need to track metrics sent in different processes in the same application. This is just something to keep in mind. @mehulsonowal any thoughts on how we should approach this?

tests/telemetry/test_writer.py
avara1986: Thanks @mabdinur! PR updated with this comment :wink:

ddtrace/internal/telemetry/writer.py
```diff
@@ -158,8 +184,49 @@ def _stop_service(self, *args, **kwargs):
...
+        Queues count metric
+        """
+        self._add_metric("count", namespace, name, value, tags)
+
+    def _add_metric(self, metric_type, namespace, name, value=1.0, tags={}):
+        # type: (MetricType, str,str, float, MetricTagType) -> None
+        """
+        Queues metric
+        """
+        with self._lock:
```
mabdinur, line 213: This lock scares me. I can see a scenario where we can inadvertently introduce deadlocks if decide to collect telemetry metrics in certain parts of our codebase. Not sure what we can do here. Just wanted to call it out.

mabdinur (commented):
Left some general thoughts. Overall I like the direction of this PR. I'll give it a more through review when it's no longer in a draft state. 

Thanks again for pushing this forward

ddtrace/internal/remoteconfig/client.py
```diff
@@ -187,33 +188,6 @@ class RemoteConfigClient(object):
     and dispatches configurations to registered products.
     """
 
-    @staticmethod
```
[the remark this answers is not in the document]
avara1986: Yes, it was. We need this "special" parsed version in all agent payload requests. we need this function in RemoteConfigClient and TelemetryWorker

ddtrace/internal/telemetry/writer.py
```diff
@@ -52,15 +60,20 @@ def __init__(self, agent_url=None):
...
         self._lock = forksafe.Lock()  # type: forksafe.ResetObject
         self._forked = False  # type: bool
         forksafe.register(self._fork_writer)
-
         self._headers = {
             "Content-type": "application/json",
             "DD-Telemetry-API-Version": "v1",
+            "DD-Client-Library-Language": "python",
+            "DD-Client-Library-Version": _get_version_agent_format(),
+            "DD-Agent-Env": "config.env",  # TODO
```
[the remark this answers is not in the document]
avara1986: the `ddtrace.config.env` but I get a circular import. I'm looking for a solution :grimacing:

ddtrace/appsec/processor.py
```diff
@@ -289,13 +291,21 @@ def on_span_finish(self, span):
             if info.errors:
                 span.set_tag_str(APPSEC_EVENT_RULE_ERRORS, json.dumps(info.errors))
             span.set_tag_str(APPSEC_EVENT_RULE_VERSION, info.version)
+            # telemetry_writer.add_gauge_metric(TELEMETRY_APPSEC, "event_rules.version", info.version, {})
             span.set_tag_str(APPSEC_WAF_VERSION, version())
+            # telemetry_writer.add_gauge_metric(TELEMETRY_APPSEC, "waf.version", version(), {})
 
             span.set_metric(APPSEC_EVENT_RULE_LOADED, info.loaded)
+            telemetry_writer.add_count_metric(TELEMETRY_APPSEC, "event_rules.loaded", float(info.loaded), {})
```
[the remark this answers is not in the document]
avara1986: `add_count_metric` doesn't send metrics to the agent, this function adds the metric information in a queue attribute of telemetry_writer and then, each 60s, the worker sends all metrics to the agent and then flush the queue :upside_down_face:  

About the memory usage, you're right, we should measure the memory :+1:

ddtrace/internal/remoteconfig/client.py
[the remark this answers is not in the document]
avara1986: Ready: https://github.com/DataDog/dd-trace-py/pull/5017

ddtrace/internal/telemetry/writer.py
```diff
@@ -52,15 +60,20 @@ def __init__(self, agent_url=None):
...
+            "DD-Agent-Hostname": "",  # TODO
         }  # type: Dict[str, str]
         additional_header_str = os.environ.get("_DD_TELEMETRY_WRITER_ADDITIONAL_HEADERS")
+        self._debug = asbool(os.environ.get("DD_TELEMETRY_DEGUG", "false"))
```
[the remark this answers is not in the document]
avara1986: Done! https://github.com/DataDog/dd-trace-py/pull/5020

ddtrace/internal/telemetry/metrics.py
```diff
@@ -0,0 +1,129 @@
...
+from typing_extensions import Literal
+
+from ..hostname import get_hostname
+
+
+MetricType = Literal["count", "gauge", "rate"]
+MetricTagType = Dict[str, str]
+
+
+class Metric(six.with_metaclass(abc.ABCMeta)):
```
P403n1x87, line 19: Is there anything in https://github.com/DataDog/dd-trace-py/blob/1.x/ddtrace/internal/metrics.py that we can reuse here?

mabdinur:
I opened this PR to clean up the telemetry writer. This PR should not change any new functionality but it may conflict with this PR: https://github.com/DataDog/dd-trace-py/pull/5038

ddtrace/internal/telemetry/metrics.py
avara1986: mmm, I don't think so, they have the same name but they're totally different

PR #4957, opened 2023-01-24
chore(tracing): align destination port tag across integrations

jbertran:
## Description

Align all relevant contrib modules from `out.port` to `network.destination.port` for outgoing port information.

## Motivation
Match new Datadog Span specs - for client processes, outgoing client port tag moves to `network.destination.port`. Previously, databases and caches used `out.port`, and grpc clients used `grpc.port`.

## Reviewer Checklist
- [x] Title is accurate.


ddtrace/contrib/aiomysql/patch.py
```diff
@@ -21,7 +21,7 @@
 
 CONN_ATTR_BY_TAG = {
     net.TARGET_HOST: "host",
-    net.TARGET_PORT: "port",
```
wconti27, line 24: Looks good to me! To fix the commit message, we use conventional commits. See [here](https://www.conventionalcommits.org/en/v1.0.0/). For merging, we may want to reach out to Python core team and see if they're okay with switching just the port tag for now (leaving `out.host`). If they are, we can just switch the target branch to 1.x instead. The rest of networking tags are blocked by security changes to come.

brettlangdon: > we may want to reach out to Python core team and see if they're okay with switching just the port tag for now (leaving out.host)

👍🏻 to lowering scope to make PR smaller/easier to merge. Especially if it can be merged as-is without any other dependencies with other changes.

jbertran: Rebased and targeted `1.x`

ddtrace/contrib/aiomysql/patch.py
```diff
@@ -21,7 +21,7 @@
...
+    "network.destination.port": "port",
```
mabdinur, line 24: Since we set `network.destination.port` in a few places we should create a constant for it. For consistency we can define it in https://github.com/DataDog/dd-trace-py/blob/1.x/ddtrace/ext/db.py

mabdinur (commented):
Should we also update the following database integrations: asyncpg, cassandra, snowflake, and postgres (psycopg)?

ddtrace/contrib/aiomysql/patch.py
jbertran: This task was my "get your legs under you" task, which I noticed a bit late was intentionally underscoped (otherwise this would be a draft 🙁 ).

The idea of literals instead of a separate constant was to avoid having a "temporary" constant that holds the new name until all integrations are switched to the new tag.

Since my legs are now _somewhat_ under me, I have a follow-up task set aside to catch up for the rest of the integrations (including the missing ~SQL modules), which I can handle in the same PR just by switching the value of the existing constant at the end of switching everything to the literal (this way the last commit is just a big `sed` for me, but I can still run tests locally along the way).

jbertran: `out.port` is now totally replaced in all contrib modules using it

jbertran:
> Should we also update the following database integrations: asyncpg, cassandra, snowflake, and postgres (psycopg)?

`out.port` is now totally replaced in all contrib modules using it (see https://github.com/DataDog/dd-trace-py/pull/4957#discussion_r1086790796)

ddtrace/contrib/grpc/constants.py
```diff
@@ -12,7 +12,6 @@
 GRPC_REQUEST_METADATA_PREFIX_KEY = "grpc.request.metadata."
 GRPC_RESPONSE_METADATA_PREFIX_KEY = "grpc.response.metadata."
 GRPC_HOST_KEY = "grpc.host"
-GRPC_PORT_KEY = "grpc.port"
```
wconti27, line 15: Confirmed on Metabase that this tag name is not in use by any clients. Good to switch to new tag name.

ddtrace/contrib/grpc/constants.py
```diff
@@ -12,7 +12,6 @@
 GRPC_REQUEST_METADATA_PREFIX_KEY = "grpc.request.metadata."
```
wconti27, line 12: Everything looks good to me. @brettlangdon Do we need a changelog for this if we're switching the constant value?

ddtrace/contrib/grpc/constants.py
jbertran: Sounds like I need a more complete view of the complete lifecycle of tags. This one was the last remaining `.port` tag generated by a client process. Thanks for confirming!

ddtrace/contrib/grpc/constants.py
mabdinur: Removing this constant is a breaking change. Instead we should deprecate it and add a release note (we should set the removal version to v2.0.0)

jbertran: Sure! Is this just about setting a release note, or is there any source to change (besides reestablishing that constant in this PR)?

mabdinur: We'll need a releasenote and a code change. Here an example of how we deprecated constants in the past: https://github.com/DataDog/dd-trace-py/commit/be529ca80a2f1d32ec94e3f3a53e0e9830323f59. You'll need a code change like this:

```
from ddtrace.vendor.debtcollector import deprecate

def __getattr__(name):
    if name == "GRPC_PORT_KEY":
       deprecate(
            ("%s.%s is deprecated" % (__name__, name)),
            removal_version="2.0.0",
        )
        return "grpc.port"

    if name in globals():
        return globals()[name]

    raise AttributeError("%s has no attribute %s", __name__, name)
```

releasenotes/notes/deprecate-constant-9bc432d6647766f3.yaml
```diff
@@ -0,0 +1,4 @@
+---
+deprecations:
+  - |
+    ``ddtrace.contrib.grpc.constants.GRPC_PORT_KEY`` is deprecated. Use ``ddtrace.ext.net.TARGET_PORT`` instead.
```
mabdinur, line 4: We should use third person present tenses to describe changes in release notes [guidelines](https://ddtrace.readthedocs.io/en/stable/contributing.html#release-note-style-guidelines).

```suggestion
    grpc: deprecates ``ddtrace.contrib.grpc.constants.GRPC_PORT_KEY``. Use ``ddtrace.ext.net.TARGET_PORT`` instead.
```

ddtrace/contrib/grpc/constants.py
mabdinur: Also can we add a test case for this deprecation warning. Here's an example of how we did this in the past: https://github.com/DataDog/dd-trace-py/blob/0.61/tests/tracer/test_constants.py#L20

releasenotes/notes/deprecate-constant-9bc432d6647766f3.yaml
jbertran: Teachable moment about cargo-culting the wording instead of skimming over that part of release note docs 😅

releasenotes/notes/deprecate-constant-9bc432d6647766f3.yaml
```diff
@@ -0,0 +1,4 @@
+---
+deprecations:
+  - |
+    grpc: deprecates ``ddtrace.contrib.grpc.constants.GRPC_PORT_KEY``ß. Use ``ddtrace.ext.net.TARGET_PORT`` instead.
```
mabdinur, line 4:
```suggestion
    grpc: deprecates ``ddtrace.contrib.grpc.constants.GRPC_PORT_KEY``. Use ``ddtrace.ext.net.TARGET_PORT`` instead.
```

mabdinur (commented):
left a suggestion but I am fine with merging this PR as is. Thanks for pushing this forward.

releasenotes/notes/deprecate-constant-9bc432d6647766f3.yaml
jbertran: Not used to mac keyboards yet 😅

ddtrace/internal/compat.py
```diff
@@ -268,6 +268,22 @@ def get_connection_response(
...
+            Pep562(module_name)
+
+
+except ImportError:
+
+    def ensure_pep562(module_name):
+        # type: (str) -> None
+        pass
+
+
```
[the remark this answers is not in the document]
jbertran: Let's merge now if we can - those are the only changes in `ddtrace.internal` so they'll be easy to pick out.

majorgreys:
While this isn't a breaking change, this change could still be categorized as a fix and benefit by the addition of a release note. Otherwise, a user upgrading would not know to expect the renamed span tags.

mabdinur:
> While this isn't a breaking change, this change could still be categorized as a fix and benefit by the addition of a release note. Otherwise, a user upgrading would not know to expect the renamed span tags.

I don't think users will see any changes to their traces. I believe the backend already converts `out.port`  to `network.destination.port`

releasenotes/notes/deprecate-constant-9bc432d6647766f3.yaml
```diff
@@ -0,0 +1,4 @@
+---
+deprecations:
+  - |
+    grpc: deprecates ``ddtrace.contrib.grpc.constants.GRPC_PORT_KEY``. Use ``ddtrace.ext.net.TARGET_PORT`` instead.
```
mabdinur, line 4:
```suggestion
deprecations:
  - |
    grpc: deprecates ``ddtrace.contrib.grpc.constants.GRPC_PORT_KEY``. Use ``ddtrace.ext.net.TARGET_PORT`` instead.
other:
  - |
     tracing: renames database connection tag from ``out.port`` to ``network.destination.port``.



```

mabdinur, line 4:
```suggestion
    grpc: Deprecates ``ddtrace.contrib.grpc.constants.GRPC_PORT_KEY``. Use ``ddtrace.ext.net.TARGET_PORT`` instead.
```

mabdinur (approved):

gnufede (approved):

PR #4962, opened 2023-01-25
chore(gunicorn): test PeriodicService under various gunicorn+ddtracepy configs

emmettbutler:
## Description

This pull request tries to answer two questions: 

* Is the [public documentation about gunicorn](https://ddtrace.readthedocs.io/en/stable/advanced_usage.html#gunicorn) accurate and complete?
* What problematic ddtracepy behavior was [this commit](https://github.com/DataDog/dd-trace-py/commit/e1ab88c78050f0fbaf1232a7cc8431b097b4f048) fixing?

It does so by implementing a series of integration tests against a real gunicorn server. Each test configures the server and ddtracepy in a different way. These tests check each configuration against known failure modes. The set of configurations that avoid all known failure modes is the one that the updated documentation recommends.

There were a lot more tests exercising various buggy configurations, but I removed them to make CI run faster. Let me know if you think some of those other configurations need to be included in automated testing.

By "known failure modes", I mean degraded performance, duplicated work, or crashes that I found through experimentation based on following logical threads from the relevant issues below. The big comment in the diff explains it in detail.

## Checklist
- [x] [Library documentation](https://github.com/DataDog/dd-trace-py/tree/1.x/docs)
- [x] update these https://docs.datadoghq.com/tracing/trace_collection/compatibility/python/

Main DD Documentation PR: https://github.com/DataDog/documentation/pull/16675

## Relevant issue(s)

https://github.com/DataDog/dd-trace-py/pull/2894
https://github.com/DataDog/dd-trace-py/pull/4810
https://github.com/DataDog/dd-trace-py/pull/4070
https://github.com/DataDog/dd-trace-py/pull/1799

## Testing strategy

After the experimentation process that informed the integration test writing, the testing strategy is the tests themselves.

## Reviewer Checklist
- [x] Title is accurate.


tests/contrib/gunicorn/wsgi_mw_app.py
```diff
@@ -1,20 +1,64 @@
...
+from ddtrace.internal.remoteconfig import RemoteConfig
+from ddtrace.profiling.collector.memalloc import MemoryCollector
+from ddtrace.profiling.profiler import Profiler
 from tests.webclient import PingFilter
 
 
+# starting the profiler here allows test cases to exercise the code paths that
+# restart threads preexisting a forked process
+Profiler._profiler = Profiler()
+Profiler._profiler.start()
```
brettlangdon, line 25: Why do we enable profiling in this way? this isn't normally how a user would enable.

```python
profiler = Profiler()
profiler.start()
```

?

or can do:

```python
import ddtrace.profiling.auto
```

It also looks like `Profiler._profiler` is meant to be a `_ProfilerInstance`, right?

brettlangdon (commented):
initial quick glance with some thoughts

tests/contrib/gunicorn/wsgi_mw_app.py
```diff
@@ -10,11 +29,28 @@
 )
 
 
+def aggressive_shutdown():
+    tracer.shutdown(timeout=1)
+    RemoteConfig.disable()
+    DynamicInstrumentation.disable()
+    bootstrap.profiler._scheduler.stop()
```
brettlangdon, line 36: this should get handled by `profiler.stop()`


https://github.com/DataDog/dd-trace-py/blob/6bf92f8fda1b7e5404a22320b96def4b536ee89c/ddtrace/profiling/profiler.py#L78

->

https://github.com/DataDog/dd-trace-py/blob/6bf92f8fda1b7e5404a22320b96def4b536ee89c/ddtrace/profiling/profiler.py#L286-L293

emmettbutler: For some reason this "aggressive shutdown" approach appears to be necessary under the test conditions I'm creating. Without it, some of the tests hang forever.

tests/contrib/gunicorn/test_gunicorn.py
```diff
@@ -52,79 +97,107 @@ def _gunicorn_settings_factory(
...
-@pytest.mark.parametrize(
-    "gunicorn_server_settings", [_gunicorn_settings_factory(app_path="tests.contrib.gunicorn.wsgi_mw_app:app")]
-)
-def test_traced_basic(gunicorn_server_settings, gunicorn_server):
-    # meta.result_class is listiterator vs list_iterator in PY2 vs PY3.
-    # Ignore this field to avoid having to create mostly duplicate snapshots in Python 2 and 3.
-    r = gunicorn_server.get("/")
-    assert r.status_code == 200
-    assert r.content == b"Hello, World!\n"
+if PY3:
```
P403n1x87, line 189: can't we use a `skipif` mark instead?

tests/contrib/gunicorn/test_gunicorn.py
```diff
@@ -1,16 +1,28 @@
...
 import pytest
 import tenacity
 
+from ddtrace.internal.compat import PY2
+from ddtrace.internal.compat import PY3
 from ddtrace.internal.compat import stringify
 from tests.webclient import Client
 
 
+PYTHON_VERSION = tuple(int(v) for v in platform.python_version_tuple())
```
P403n1x87, line 20:
```suggestion
PYTHON_VERSION = sys.version_info
```
?

P403n1x87 (commented):
This is 💪! A couple of nits, otherwise LGTM!

releasenotes/notes/gunicorn-issue-09308901fa00e76c.yaml
```diff
@@ -0,0 +1,3 @@
+---
+prelude: >
+  gunicorn: known issue: ddtrace-run does not work with gunicorn
```
mabdinur, line 3: Is this release note required? This issue has been documented for a few years now. Calling it out in a new release note might cause confusion

riotfile.py
```diff
@@ -2595,13 +2595,11 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
         Venv(
             name="gunicorn",
             command="pytest {cmdargs} tests/contrib/gunicorn",
-            pkgs={"requests": latest},
+            pkgs={"requests": latest, "gevent": latest},
+            env={
+                "AGENT_VERSION": "latest",
+            },
```
mabdinur, line 2601:
```suggestion
```

This environment variable is  used in the following test suite: [integration](https://github.com/DataDog/dd-trace-py/blob/1.x/tests/integration/test_integration.py#L25). It's not used in gunicorn tests

riotfile.py
```diff
@@ -2595,13 +2595,11 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
...
             venvs=[
-                Venv(
```
mabdinur, line 2600: Gunicorn dropped support for python 2.x in [v20.0](https://docs.gunicorn.org/en/stable/2019-news.html#id6) not 19.10. 

If we want to drop py2.7 support for gunicorn we should we update our [minimum support version](https://ddtrace.readthedocs.io/en/latest/index.html#supported-libraries) to 20.0.

tests/contrib/gunicorn/post_fork.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+# take some notes about the relative ordering of thread creation and
+# monkeypatching
+monkeypatch_happened = gevent.monkey.is_module_patched("threading")
+
+# enabling DI here allows test cases to exercise the code paths that handle
+# gevent monkeypatching of running threads
+# post_fork is called before gevent.monkey.patch_all()
+if sys.version_info[1] < 11:
```
mabdinur, line 17: We can future proof this check:

```suggestion
if sys.version_info < (3, 11):
```

mabdinur: We should remove this check after DynamicInstrumentation add support for python 3.11. Maybe we should leave comment here?

tests/contrib/gunicorn/wsgi_mw_app.py
```diff
@@ -1,20 +1,59 @@
...
     }
 )
 
+if sys.version_info[1] < 11:
+    from ddtrace.debugging import DynamicInstrumentation
+
+
+def aggressive_shutdown():
+    RemoteConfig.disable()
+    if sys.version_info[1] < 11:
```
mabdinur, line 33: nit:
```suggestion
    if sys.version_info < (3, 11):
```

tests/contrib/gunicorn/wsgi_mw_app.py
```diff
@@ -1,20 +1,59 @@
...
+        DynamicInstrumentation.disable()
+    if PY3:
```
mabdinur, line 35: I don't think we're running these tests with python2. Is this PY3 check required?

releasenotes/notes/gunicorn-issue-09308901fa00e76c.yaml
emmettbutler: 🤷 I don't really know the norms, but including it in a release note seems reasonable to me. @brettlangdon suggested that I should add a release note.

tests/contrib/gunicorn/post_fork.py
emmettbutler: I think a recursive grep for `(3, 11)` will be enough to track this down once the DI change is merged.

releasenotes/notes/gunicorn-issue-09308901fa00e76c.yaml
brettlangdon: I think it makes sense to include as a known issue since the previous suggestion was that yes you could use `ddtrace-run` with gunciorn and gevent, but now you cannot use them together.

brettlangdon, line 3:
```suggestion
issues:
  - |
    gunicorn: ddtrace-run does not work with gunicorn
```

If we list under `issues:` it'll show up as a "Known Issue" in the release notes.

But, we should improve this message as well to include the new suggested steps, and link to the new gunicorn docs.

emmettbutler:
@brettlangdon do you think [this](https://docs.datadoghq.com/integrations/gunicorn/#pagetitle) is the right page on docs.datadog.com to add the ddtracepy+gunicorn integration notes?

emmettbutler:
@brettlangdon @Yun-Kim @Kyle-Verhoog do you think [this](https://docs.datadoghq.com/integrations/gunicorn/#pagetitle) is the right page on docs.datadog.com to add the ddtracepy+gunicorn integration notes? Or maybe [here](https://docs.datadoghq.com/tracing/trace_collection/dd_libraries/python/?tab=containers)? Both might be a bit general to include framework-specific gotchas.

P403n1x87 (approved):

mabdinur (approved):

tests/contrib/gunicorn/test_gunicorn.py
```diff
@@ -52,79 +94,103 @@ def _gunicorn_settings_factory(
...
-            client.wait(max_tries=20, delay=0.5)
+            client.wait(max_tries=100, delay=0.1)
         except tenacity.RetryError:
-            # if proc.returncode is not None:
-            # process failed
             raise TimeoutError("Server failed to start, see stdout and stderr logs")
-
-        yield client
-
+        time.sleep(SERVICE_INTERVAL)
```
Yun-Kim, line 157: Out of curiousity, what does waiting here before yielding accomplish?

tests/contrib/gunicorn/test_gunicorn.py
```diff
@@ -52,79 +94,103 @@ def _gunicorn_settings_factory(
...
+)
 
 
-@pytest.mark.snapshot(ignores=["meta.result_class"])
 @pytest.mark.parametrize(
-    "gunicorn_server_settings", [_gunicorn_settings_factory(app_path="tests.contrib.gunicorn.wsgi_mw_app:app")]
+    "gunicorn_server_settings",
+    [
+        SETTINGS_GEVENT_APPIMPORT_PATCH_POSTWORKERSERVICE,
+        SETTINGS_GEVENT_POSTWORKERIMPORT_PATCH_POSTWORKERSERVICE,
```
Yun-Kim, line 188: Are we adding any scenarios that end in failure modes? Or are we just testing the two scenarios which don't result in failure modes? Should we include any expected failure test scenarios (not too many but at least one or two)? I worry once we have a "solution" for this gevent support these two scenarios alone won't make it obvious that the "solution" is indeed a solution.

tests/contrib/gunicorn/test_gunicorn.py
```diff
@@ -52,79 +94,103 @@ def _gunicorn_settings_factory(
...
 
 
-def test_basic(gunicorn_server):
-    r = gunicorn_server.get("/")
-    assert r.status_code == 200
-    assert r.content == b"Hello, World!\n"
+SETTINGS_GEVENT_APPIMPORT_PATCH_POSTWORKERSERVICE = _gunicorn_settings_factory(
+    worker_class="gevent",
+    use_ddtracerun=False,
+    import_sitecustomize_in_app=True,
```
Yun-Kim, line 171: Was there any specific reason `import_sitecustomize_in_app` wasn't included in the actual `GunicornServerSettings` namedTuple while `import_sitecustomize_in_postworkerinit` was?

tests/contrib/gunicorn/post_fork.py
```diff
@@ -0,0 +1,21 @@
...
+monkeypatch_happened = gevent.monkey.is_module_patched("threading")
+
+# enabling DI here allows test cases to exercise the code paths that handle
+# gevent monkeypatching of running threads
+# post_fork is called before gevent.monkey.patch_all()
+if sys.version_info < (3, 11):
+    DynamicInstrumentation.enable()
+
+RemoteConfig._was_enabled_after_gevent_monkeypatch = monkeypatch_happened
+Profiler._was_enabled_after_gevent_monkeypatch = monkeypatch_happened
```
Yun-Kim, line 21: Do we need this?

Yun-Kim (commented):
Just a few clarifying questions.

tests/contrib/gunicorn/post_fork.py
emmettbutler: Nope, good catch.

tests/contrib/gunicorn/test_gunicorn.py
emmettbutler: `import_sitecustomize_in_app` can be checked via `GunicornServerSettings().env.get("_DD_TEST_IMPORT_SITECUSTOMIZE")`, and I left it out of the `GunicornServerSettings` top-level attributes to avoid duplicating it.

tests/contrib/gunicorn/test_gunicorn.py
emmettbutler: It gives the `PeriodicService` time to wake up and try to run. It's only when that happens that the service's zombie threads self-destruct. This is definitely worth adding a comment about.

tests/contrib/gunicorn/test_gunicorn.py
emmettbutler: I was wondering this as well. I wrote a bunch of other tests that I then deleted to save time in CI. Which of those do you think are the most important to add back? https://github.com/DataDog/dd-trace-py/pull/4962/commits/564037ac3868b140abb181476b810d4790ba4025

tests/contrib/gunicorn/test_gunicorn.py
```diff
@@ -21,29 +30,62 @@
...
+
+def parse_payload(data):
+    decoded = data
+    if sys.version_info[1] == 5:
+        decoded = data.decode("utf-8")
+    return json.loads(decoded)
+
+
+def assert_remoteconfig_started_successfully(response, check_patch=True):
+    if sys.version_info[1] in (5, 11):
```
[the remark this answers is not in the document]
emmettbutler: Under the configs I've set up in this test suite, it does not. It's such an edge case that I decided not to mention it in the documentation change. What do you think?

tests/contrib/gunicorn/test_gunicorn.py
```diff
@@ -21,29 +30,62 @@
         ("worker_class", str),
         ("bind", str),
         ("use_ddtracerun", bool),
-        ("post_worker_init", str),
+        ("import_sitecustomize_in_postworkerinit", bool),
+        ("start_service_in_hook_named", str),
```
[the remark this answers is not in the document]
emmettbutler: Really the two options are `post_worker_init` and "every other string" - there's default behavior if it's not `post_worker_init`.

ddtrace/internal/periodic.py
```diff
@@ -54,15 +54,46 @@ def stop(self):
...
+
+        ...or is it?
+
+        As it turns out, gevent has some of its own post-fork logic that complicates things. All of the above is
+        accurate, except for the bit about the child process' Profiler being the only one alive. In fact, gunicorn's
+        "gevent worker" notices that there was a Profiler (PeriodicThread) running before the fork, and attempts to
+        bring it back to life after the fork.
+
+        Outside of ddtrace-run, the only apparent problem this causes is duplicated work and decreased performance.
+        Under ddtrace-run, because it triggers an additional fork to which gevent's post-fork logic responds, the
```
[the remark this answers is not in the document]
emmettbutler: Here: https://github.com/DataDog/dd-trace-py/blob/67039008e64fa560373d677372179c75f3059bd0/ddtrace/commands/ddtrace_run.py#L143
I'm not 100% sure from the `execl` documentation whether it actually triggers a fork, but it certainly appears to based on the behavior I've observed and tried to capture in the tests.

tests/contrib/gunicorn/test_gunicorn.py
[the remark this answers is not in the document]
Yun-Kim: I think it might be worth adding as a comment just for anyone like me in the future that might be confused 😄

tests/contrib/gunicorn/test_gunicorn.py
Yun-Kim: I see, a comment here definitely sounds helpful!

tests/contrib/gunicorn/test_gunicorn.py
Yun-Kim: I think in an ideal scenario where we don't have to worry about flakiness of starting up a new gunicorn server for each test, it would be great to keep all of them. However in our case, `SETTINGS_GEVENT_DDTRACERUN_PATCH` might be the most important to add back just to illustrate that scenarios where we use `ddtrace-run` with gevent workers results in profiler issues.

lattwood:
@emmettbutler Does this mean the only supported worker type is `gevent` under Gunicorn, or that if you want to use the `gevent` worker type you have to jump through hoops, and the other worker types are supported?

Docs are unclear, hence asking :\

PR #4968, opened 2023-01-26
treat frame as a generic Python object

P403n1x87:
## Checklist

## Reviewer Checklist
- [ ] Title is accurate.
- [ ] Telemetry is meaningful, actionable and does not have the potential to leak sensitive data.

### DIFF
--- a/ddtrace/profiling/collector/_traceback.pyx
+++ b/ddtrace/profiling/collector/_traceback.pyx
@@ -1,61 +1,22 @@
-IF PY_MAJOR_VERSION > 3 or (PY_MAJOR_VERSION == 3 and PY_MINOR_VERSION >= 11):
-    from cpython.object cimport PyObject
-    from cpython.ref cimport Py_XDECREF
-
-    cdef extern from "<Python.h>":
-        ctypedef struct PyCodeObject:
-            pass
-        ctypedef struct PyFrameObject:
-            pass
-        PyObject* PyDict_GetItem(PyObject* p, PyObject* key)
-        PyObject* PyTuple_GetItem(PyObject* p, Py_ssize_t pos)
-        Py_ssize_t PyTuple_Size(PyObject* p)
-
-        PyCodeObject* PyFrame_GetCode(PyFrameObject* frame)
-        PyFrameObject* PyFrame_GetBack(PyFrameObject* frame)
-        PyObject* PyFrame_GetLocals(PyFrameObject* frame)
-        int PyFrame_GetLineNumber(PyFrameObject* frame)
-        PyObject* PyCode_GetVarnames(PyCodeObject* co)
-
-
-cpdef _extract_class_name(frame):
+cdef str _extract_class_name(object frame):
     # type: (...) -> str
     """Extract class name from a frame, if possible.
 
     :param frame: The frame object.
     """
-    # Python 3.11 moved PyFrameObject members to internal C API and cannot be directly accessed
-    IF PY_MAJOR_VERSION > 3 or (PY_MAJOR_VERSION == 3 and PY_MINOR_VERSION >= 11):
-        code = PyFrame_GetCode(<PyFrameObject*>frame)  # Returns strong reference
-        co_varnames = PyCode_GetVarnames(code)  # Returns new reference
-        if co_varnames == NULL or PyTuple_Size(co_varnames) == 0:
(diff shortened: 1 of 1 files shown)


ddtrace/profiling/collector/_traceback.pyx
```diff
@@ -1,61 +1,22 @@
...
-        code = PyFrame_GetCode(<PyFrameObject*>frame)  # Returns strong reference
-        co_varnames = PyCode_GetVarnames(code)  # Returns new reference
-        if co_varnames == NULL or PyTuple_Size(co_varnames) == 0:
-            Py_XDECREF(<PyObject*>code)
-            return ""
-        argname = PyTuple_GetItem(co_varnames, 0)
-        Py_XDECREF(<PyObject*>code)
-        Py_XDECREF(co_varnames)
+    if frame.f_code.co_varnames:
+        argname = frame.f_code.co_varnames[0]
```
brettlangdon, line 8: are we sure we'll have at least 1?

ddtrace/profiling/collector/_traceback.pyx
```diff
@@ -67,22 +28,16 @@ cpdef traceback_to_frames(traceback, max_nframes):
     :return: The serialized frames and the number of frames present in the original traceback.
     """
     tb = traceback
-    frames = []
-    nframes = 0
+
+    cdef list frames = []
```
brettlangdon, line 32: `cdef tb` too ? 🤷🏻

ddtrace/profiling/collector/_traceback.pyx
```diff
@@ -67,22 +28,16 @@ cpdef traceback_to_frames(traceback, max_nframes):
...
+    cdef int nframes = 0
+    
     while tb is not None:
         if nframes < max_nframes:
```
brettlangdon, line 36: this is weird, should this be something more like:

```python
if nframes >= max_nframes:
    break
```

it seems the way this is written we will traverse the entire traceback every time, no matter what ?

brettlangdon: (I think Yun saw this behavior when they were debugging too)

ddtrace/profiling/collector/_traceback.pyx
```diff
@@ -94,26 +49,17 @@ cpdef pyframe_to_frames(frame, max_nframes):
...
-                frames.append((code.co_filename, lineno, code.co_name, _extract_class_name(frame)))
-            frame = frame.f_back
+    if frame is None:
+        return [], 0
+
+    cdef list frames = []
+    cdef int nframes = 0
+
+    while frame is not None:
+        if nframes < max_nframes:
```
brettlangdon, line 59: same here, `if nframes >= max_nframes: break` ?

brettlangdon: nframes vs `len(frames)` ? probably splitting hairs on which would be faster.

ddtrace/profiling/collector/_traceback.pyx
P403n1x87: then `frame.f_code.co_varnames` would evaluate to `False`, no? This is the current coding btw

ddtrace/profiling/collector/_traceback.pyx
P403n1x87: Yep we could do, but I wouldn't bother too much as this is not a mile off from what we currently have implemented, so we might as well just close this PR entirely as I don't think we gain much from these little tweaks for now

ddtrace/profiling/collector/_traceback.pyx
P403n1x87: Yes I think the point is that we still want to count the number of frames, even though we then truncate the frame stack. Thats why we return `nframes` instead of just returning `frames` and deriving `nframes` from `len(frames)`.

ddtrace/profiling/collector/_traceback.pyx
brettlangdon: this was pre-coffee 🤦🏻

ddtrace/profiling/collector/_traceback.pyx
brettlangdon: oh, you are right, we still want to return them all. ok I didn't quite notice that

PR #4974, opened 2023-01-26
feat(rcm): add more validations to Remote Config response from agent [backport #4970 to 1.8]

avara1986:
Backport of #4970

Add more validations to Remote Config response from agent


juanjux (approved):

gnufede (dismissed):

brettlangdon (changes requested):
is this required for 1.8? since this is a new feature and it did not make the cut off, it should be for 1.9 instead.

ddtrace/internal/remoteconfig/client.py
```diff
@@ -341,7 +347,7 @@ def _process_response(self, data):
         paths = {_.path for _ in payload.target_files}
         paths = paths.union({_["path"] for _ in self.cached_target_files})
 
-        if not set(payload.client_configs) <= paths:
+        if set(payload.client_configs) > paths:
```
P403n1x87, line 350: `!(A is a subset of B)` is not equivalent to `B is a proper subset of A`, so this is now a different check. Was the original one wrong?

avara1986: My bad, I didn't know `set_a() < set_b()`  == `set_a.issubset(set_b)` fixed in https://github.com/DataDog/dd-trace-py/pull/4988

brettlangdon (approved):

brettlangdon (approved):

juanjux (approved):

PR #4975, opened 2023-01-26
chore(asm): add more error/type checking to ddwaf

gnufede:
## Description

- Avoid any errors or exceptions raised while running the WAF to interrupt the request
- Add test coverage and regression tests

## Checklist

## Reviewer Checklist
- [x] Title is accurate.


christophe-papazian (dismissed):

ddtrace/appsec/ddwaf/__init__.py
```diff
@@ -109,19 +110,24 @@ def run(
...
+            error = ddwaf_run(ctx, wrapper, ctypes.byref(result), timeout_ms * 1000)
+            if error:
+                LOGGER.warning("DDWAF error: %d", error)
+
+            libddwaf_result = DDWaf_result(
+                result.data.decode("UTF-8", errors="ignore") if hasattr(result, "data") and result.data else None,
+                [result.actions.array[i].decode("UTF-8", errors="ignore") for i in range(result.actions.size)],
+                result.total_runtime / 1e3,
+                (time.time() - start) * 1e6,
+            )
```
brettlangdon, line 127: everything in `result` gets free'd right? so the only reason this does not cause a bad pointer/reference is because everything is transformed/copied right?

if you didn't call `.decode()` on `result.data`, then we'd have issues?


if so, maybe a comment or helper would be better, I could see an easy mistake to end up with an invalid pointer to something here.

if not, feel free to resolve with no action!

brettlangdon: I wonder if we should still have `try/finally` here, since if we end up with an error in any of the transformations when we could end up leaking memory.

The only issues I potentially see are data that aren't actually bytes, or is `size` is wrong.

```python
try:
    return DDWaf_result(...)
finally:
    ddwaf_result_free(ctypes.byref(result))
    ddwaf_context_destroy(ctx)
```


In fact, `ctx` could leak as well, if there is an error anywhere between L116 and the end of the function.

maybe better as:


```python
ctx = ddwaf_context_init(self._handle)
try:
    result = ddwaf_result()
    try:
        return DDWaf_result(...)
    finally:
        ddwaf_result_free(ctypes.byref(result))
finally:
    ddwaf_context_destroy(ctx)
```

I know it isn't very pretty, but we really want to be sure we don't end up with a memory leak.

ddtrace/appsec/ddwaf/ddwaf_types.py
```diff
@@ -130,15 +128,18 @@ def __init__(self, struct=None):
...
                 res_key = key.encode("UTF-8", errors="ignore") if isinstance(key, unicode) else key
                 obj = ddwaf_object(val)
                 if obj.type:  # discards invalid objects
                     assert ddwaf_object_map_add(map_o, res_key, obj)
         else:
-            raise TypeError("ddwaf_object : unknown type in structure. " + repr(type(struct)))
+            if struct is not None:
+                log.warning("DDWAF object init called with unknown data structure")
+
+            ddwaf_object_invalid(self)
```
brettlangdon, line 142: is this safe? will everything else operate ok if a value gets converted to `None`?

ddtrace/appsec/processor.py
```diff
@@ -270,7 +276,11 @@ def on_span_finish(self, span):
                 data[_Addresses.SERVER_REQUEST_BODY] = body
 
         log.debug("[DDAS-001-00] Executing AppSec In-App WAF with parameters: %s", data)
-        ddwaf_result = self._ddwaf.run(data, self._waf_timeout)  # res is a serialized json
+        ddwaf_result = None
+        try:
+            ddwaf_result = self._ddwaf.run(data, self._waf_timeout)  # res is a serialized json
+        except Exception:
+            log.warning("Error executing Appsec In-App WAF")
```
brettlangdon, line 283: do we want `exc_info=True` here?

do we think there would be concerns of leaking data via this warning log if we include the traceback? would just the exception type be at least enough to make it more useful? (or do the other logs help us out here?)

ddtrace/appsec/ddwaf/ddwaf_types.py
```diff
@@ -151,17 +152,18 @@ def struct(self):
...
         if self.type == DDWAF_OBJ_TYPE.DDWAF_OBJ_MAP:
             return {
-                self.value.array[i].parameterName.decode("UTF-8"): self.value.array[i].struct
+                self.value.array[i].parameterName.decode("UTF-8", errors="ignore"): self.value.array[i].struct
                 for i in range(self.nbEntries)
             }
         if self.type == DDWAF_OBJ_TYPE.DDWAF_OBJ_BOOL:
             return self.value.boolean
-        raise ValueError("ddwaf_object: unknown object")
+        log.warning("ddwaf_object struct: unknown object type")
```
brettlangdon, line 165: can we include the unknown object type? (just the type, not the value)

ddtrace/appsec/ddwaf/ddwaf_types.py
```diff
@@ -130,15 +128,18 @@ def __init__(self, struct=None):
...
                 if not isinstance(key, (bytes, unicode)):  # discards non string keys
                     continue
                 res_key = key.encode("UTF-8", errors="ignore") if isinstance(key, unicode) else key
                 obj = ddwaf_object(val)
                 if obj.type:  # discards invalid objects
                     assert ddwaf_object_map_add(map_o, res_key, obj)
         else:
-            raise TypeError("ddwaf_object : unknown type in structure. " + repr(type(struct)))
+            if struct is not None:
+                log.warning("DDWAF object init called with unknown data structure")
```
brettlangdon, line 140: include the unknown type here, I think the `repr(type(struct))` above was a good idea ?

tests/appsec/test_processor.py
```diff
@@ -423,3 +431,112 @@ def test_ddwaf_info_with_3_errors():
...
+                    "x-ratelimit-name": "role_api",
+                    "x-ratelimit-limit": "500",
+                    "x-ratelimit-period": "60",
+                    "content-type": "application/json",
+                    "x-ratelimit-reset": "16",
+                },
+                request_body={"_authentication_token": u"2b0297348221f294de3a047e2ecf1235abb866b6"},
+            )
+
+    assert "Error parsing data AppSec In-App WAF metrics report" in caplog.text
```
brettlangdon, line 469: do/should we need to assert anything about the span? I am guessing this test would fail with a raised exception before the changes, so maybe we don't 🤔

tests/appsec/test_processor.py
```diff
@@ -423,3 +431,112 @@ def test_ddwaf_info_with_3_errors():
...
+
+
+def test_ddwaf_run_reggression_test(tracer_appsec):
```
brettlangdon, line 472: what is this a regression for? what was the behavior before?

brettlangdon (commented):
nit/suggestions, but otherwise looks good

tests/appsec/test_processor.py
gnufede: This was the data that raised the UnboundLocalError while running the waf, but I couldn't reproduce it locally

tests/appsec/test_processor.py
gnufede: We could assert that the span is finished (so that the Exception is captured), but not much more, as the appsec part will not be included in it, what do you think?

brettlangdon: I think the function not raising a exception is proof enough that all the `on_span_finish`s got called without errors.

I wanted to suggest making it a snapshot test to confirm it does make it's way through to the writer/agent without issues, but I am not sure, that is probably overkill.

this might be fine as-is. as long as you stopped to ponder some options, feel free to resolve with no action if you don't think any are necessary.

ddtrace/appsec/ddwaf/ddwaf_types.py
christophe-papazian: It should not happen, but if for any reason, an arbitratry object is used to feed the waf, it should count as invalid. This is an expected behaviour.

tests/appsec/test_processor.py
brettlangdon: mind adding a comment, the test name alone isn't clear enough (IMHO)

ddtrace/appsec/ddwaf/ddwaf_types.py
brettlangdon: cool, awesome. thanks for confirming. I'd always welcome a comment to explain that when I try to read this code again in the future 😆

ddtrace/appsec/processor.py
gnufede: We'll add two checks, one for OSError that should be safe regarding the data leaked, so adding `exc_info=True` in that case, and a second `except Exception` logging only the type of exception.

tests/appsec/test_processor.py
gnufede: Maybe we can remove it altogether given that it is not enough information to reproduce the problem

ddtrace/appsec/ddwaf/__init__.py
```diff
@@ -109,18 +110,29 @@ def run(
...
             if ctx == 0:
-                raise RuntimeError
+                LOGGER.error("DDWaf failure to create the context")
+                return DDWaf_result(None, [], 0, (time.time() - start) * 1e6)
+
+            result = ddwaf_result()
+            wrapper = ddwaf_object(data)
+            error = ddwaf_run(ctx, wrapper, ctypes.byref(result), timeout_ms * 1000)
+            if error:
+                LOGGER.warning("DDWAF error: %d", error)
```
brettlangdon, line 120: this all needs to be in the first `try` right?

gnufede: good catch, thanks!

christophe-papazian (dismissed):

brettlangdon (approved):

christophe-papazian (approved):

ddtrace/appsec/ddwaf/__init__.py
```diff
@@ -24,11 +24,12 @@
     from .ddwaf_types import ddwaf_get_version
     from .ddwaf_types import ddwaf_init
     from .ddwaf_types import ddwaf_object
+    from .ddwaf_types import ddwaf_result
     from .ddwaf_types import ddwaf_result_free
     from .ddwaf_types import ddwaf_ruleset_info
+    from .ddwaf_types import ddwaf_run
     from .ddwaf_types import ddwaf_update_rule_data
     from .ddwaf_types import py_ddwaf_required_addresses
-    from .ddwaf_types import py_ddwaf_run
```
mabdinur, line 31: Since this is a public module removing this import is a "breaking change". The impact should be small so I think it should be fine.

ddtrace/appsec/processor.py
```diff
@@ -34,6 +34,12 @@
 from ddtrace.internal.rate_limiter import RateLimiter
 
 
+try:
+    from json.decoder import JSONDecodeError
+except ImportError:
+    # handling python 2.X import error
+    JSONDecodeError = ValueError  # type: ignore
```
mabdinur, line 41: We have this logic in a few places. We should define this type in a common module.

mabdinur (commented):
left some small nits

mabdinur:
oh well 😭. Looks Okay to Me!

PR #5029, opened 2023-02-03
fix(apm): fix unicode decode errors in pylons

christophe-papazian:
## Checklist

This PR is related to a critical exception detected two days ago on https://app.datadoghq.com/logs

Pylons Unicode Error

Silently drop non translatable UTF-8 characters instead of raising critical exception in Pylons framework.

Added regression tests to make sure that the UnicodeDecodeError is not logged as it was before

## Reviewer Checklist

- [x] Title is accurate.


P403n1x87:
@christophe-papazian I think it might be best to have a screenshot/description of the issue than a link to some data that might eventually go away. Just in case we'd have to look back at this in the future and we find that we lost some of the context. 🙏

gnufede:
Maybe add a regression test?

christophe-papazian:
> 



> Maybe add a regression test?

The actual situation is that uncatched unicode errors will crash the thread. This fix only suppresses them by discarding problematic bytes, but it doesn't change the previous behaviour on well formed byte input.

brettlangdon (changes requested):
👍🏻 to adding a regression test.

christophe-papazian:
> 👍🏻 to adding a regression test.

@brettlangdon The fix is urgent, the test is not. And I have not time right now to understand how pylons tests work.

ddtrace/contrib/pylons/middleware.py
```diff
@@ -107,11 +107,11 @@ def __call__(self, environ, start_response):
                         if hasattr(request, "json"):
                             req_body = request.json
                         else:
-                            req_body = json.loads(request.body.decode("UTF-8"))
+                            req_body = json.loads(request.body.decode("UTF-8", "ignore"))
```
mabdinur, line 110: Question: `xmltodict.parse` uses uft8 by default but it can [handle different encodings](https://omz-software.com/pythonista/docs/ios/xmltodict.html). Instead of ignoring errors can we use [request.charset](https://github.com/Pylons/webob/blob/main/src/webob/request.py#L131) to get the correct encoding?

mabdinur (commented):
Can we add a regression test. Left a question but aside from that this change looks good to me.

releasenotes/notes/fix-unicode-decode-errors-pylons-52d28b22f82e184a.yaml
```diff
@@ -0,0 +1,4 @@
+---
+fixes:
+  - |
+    pylons: This fix resolves an issue where str.decode caused critical unicode decode errors.
```
brettlangdon, line 4:
```suggestion
    pylons: This fix resolves an issue where ``str.decode`` caused critical unicode decode errors.
```

the `build_docs` job was failing because of spellchecker. if we make this an inline code block the spellchecker will ignore it.

the alternative is to add `str` to our spelling word list. (inline code block seemed to make sense for this specific instance)

brettlangdon, line 4: This is only for when appsec is enabled right? so should we say this is an appsec bug?

```suggestion
    pylons: This fix resolves an issue where ``str.decode`` could cause critical unicode decode errors when ASM is enabled. ASM is disabled by default.
```

wdyt?

tests/contrib/pylons/test_pylons.py
```diff
@@ -674,11 +675,36 @@ def test_pylons_body_xml(self):
...
+                spans = self.pop_spans()
+                assert spans
+
+                root_span = spans[0]
+                appsec_json = root_span.get_tag("_dd.appsec.json")
+                assert appsec_json is None
+
+                assert (
+                    "UnicodeDecodeError: 'utf8' codec can't decode byte 0x80 in position 0: invalid start byte"
+                    not in self._caplog.text
```
brettlangdon, line 699:
```suggestion
                    "UnicodeDecodeError" not in self._caplog.text
```

Should we keep it simple, just in case the error message format were to ever change?


this is a thought, doesn't need to be changed. feel free to resolve without accepting.

tests/contrib/pylons/test_pylons.py
```diff
@@ -690,6 +716,11 @@ def test_pylons_body_xml_attack(self):
 
             root_span = spans[0]
             assert root_span
+
+            assert (
+                "UnicodeDecodeError: 'utf8' codec can't decode byte 0x80 in position 0: invalid start byte"
+                not in self._caplog.text
```
brettlangdon, line 722:
```suggestion
                "UnicodeDecodeError" not in self._caplog.text
```

same here too, optional, just suggesting to make it easier to update both if you decided to.

tests/contrib/pylons/test_pylons.py
```diff
@@ -659,7 +660,7 @@ def test_pylons_body_xml(self):
             response = self.app.post(
                 url_for(controller="root", action="body"),
                 params=payload,
-                extra_environ={"CONTENT_TYPE": "application/xml"},
+                extra_environ={"CONTENT_TYPE": "application/xml; charset=euc-jp"},
```
brettlangdon, line 663: I am not sure I love that we are tossing these into other tests randomly rather than having dedicated tests for this, but it gets the job done for now.

brettlangdon (approved):
a few thoughts/alternative ideas, but overall lgtm

tests/contrib/pylons/test_pylons.py
gnufede: Sorry, these are the start of other interesting corner cases to try, but they require some more time to do them properly. They are harmless, but I can remove them if they could be misleading.

brettlangdon (approved):
Thanks for removing those charset test changes. I think you are right, they were probably a little misleading.

avara1986 (approved):

tests/contrib/pylons/test_pylons.py
```diff
@@ -674,11 +676,60 @@ def test_pylons_body_xml(self):
             assert span
             assert span["mytestingbody_key"] == "mytestingbody_value"
 
-    def test_pylons_body_xml_attack(self):
+    def test_pylons_body_json_unicode_decode_error_charset(self):
+        # Regression test, if request.charset returns None, a TypeError is raised
```
emmettbutler, line 680: I don't see where in the test code this `TypeError` is checked for. What am I missing?

emmettbutler (changes requested):
This all makes sense to me. I added one question about a test that appears to be redundant. You might consider changing that test to increase the coverage of different code paths.

tests/contrib/pylons/test_pylons.py
gnufede: You are right, it's not checked for: if a regression happens, the TypeError will be raised and the test will fail.

PR #5032, opened 2023-02-03
feat(asm): add blocking scaffolding

juanjux:
# Description

Adds the base features and utilities to allow for IP blocking without adding any framework (this will come in the next 2 PRs).

# Motivation

Having IP Blocking with is an ASM OKR.

# Design

At the very start of Django or Flask requests, this will store the headers and IP, if available in the new context manager. Those will then be checked in the AppsecSpanProcessor.on_span_start which will do an early call of the WAF. If it triggers a result, some tags will be set and the _context will store a http.request.blocked field. This will be checked inside the request by the frameworks to interrupt the request. on_span_finish will add more tags if needed but not call the WAF a second time (however, it could be called twice if the request was not blocked since the second call could use additional information from the request like the user id; we'll work on improve this once we incorporate user id blocking and suspicious request blocking and have a good picture of all the data needed for the WAF).

# Testing strategy

- This includes some unittests. The next PRs adding support for Django and Flask (and indirectly ASGI) will add more focused tests. 
- Has gone trough two interactive ASM peer testing sessions.
- On stagging, manual testing would be producing an event with, for example:
`curl -v -s -k 'GET' http://stagging_test_url/ --header "x-cluster-client-ip: 200.45.67.89" -A 'Arachni/v1'
and once that shows on the traced, blocking the IP and checking that the next request is blocked.

No release note since this only adds internal scaffolding without any usable feature, next two PRs will add one each.

Signed-off-by: Juanjo Alvarez <juanjo.alvarezmartinez@datadoghq.com>

## Checklist

Note: IP blocking docs will be published once the feature is announced publicly.

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/constants.py
```diff
@@ -58,3 +58,29 @@
...
+justify-content: center; -webkit-box-align: center; -ms-flex-align: center; align-items: center;
+-ms-flex-line-pack: center; align-content: center; flex-direction: column}p{font-size: 18px;
+line-height: normal; color: #646464; font-family: sans-serif; font-weight: 400}a{color:
+#4842b7}footer{width: 100%; text-align: center}footer p{font-size: 16px}</style></head><body>
+<main> <p>Sorry, you cannot access this page. Please contact the customer service team.</p></main>
+<footer> <p>Security provided by
+<a href="https://www.datadoghq.com/product/security-platform/application-security-monitoring/"
+target="_blank">Datadog</a></p></footer></body></html>"""
+
+# noqa: E501
```
brettlangdon, line 82: what line is this meant to be associated with?

ddtrace/constants.py
```diff
@@ -58,3 +58,29 @@
 AUTO_KEEP = 1
 # Use this to explicitly inform the backend that a trace should be kept and stored.
 USER_KEEP = 2
+
+APPSEC_BLOCKED_RESPONSE_HTML = """
```
brettlangdon, line 62: I wonder if these should be in `ddtrace/internal/constants.py` instead?

Not a strong opinion, feel free to resolve if you are ok with them being in the public API.

ddtrace/internal/remoteconfig/client.py
```diff
@@ -227,19 +227,23 @@ def base64_to_struct(val, cls):
...
         self._last_error = None  # type: Optional[str]
         self._backend_state = None  # type: Optional[str]
 
-    def register_product(self, product_name, func=None):
+    def register_product(self, product_name, func):
         # type: (str, Optional[ProductCallback]) -> None
         if func is not None:
             self._products[product_name] = func
         else:
             self._products.pop(product_name, None)
```
brettlangdon, line 241:
```suggestion
        # type: (str, ProductCallback) -> None
        self._products[product_name] = func
```

is this what you intended?

I see we are adding `unregister_product`, which should cover if someone wants to explicitly remove it.

I also see that `self._products` allows `None` as the value, is there a scenario where we register a product but not a callback?

tests/appsec/test_appsec_utils.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def test_get_blocked_template_no_env_var_html():
+    with override_env(dict(DD_APPSEC_HTTP_BLOCKED_TEMPLATE_HTML="", DD_APPSEC_HTTP_BLOCKED_TEMPLATE_JSON="")):
+        assert utils._get_blocked_template("text/html") == constants.APPSEC_BLOCKED_RESPONSE_HTML
+
+
+def test_get_blocked_template_no_env_var_json():
+    with override_env(dict(DD_APPSEC_HTTP_BLOCKED_TEMPLATE_HTML="", DD_APPSEC_HTTP_BLOCKED_TEMPLATE_JSON="")):
+        assert utils._get_blocked_template("other") == constants.APPSEC_BLOCKED_RESPONSE_JSON
```
brettlangdon, line 28: parametrizing could clean up/reduce duplicated code a bunch.


mild suggestion, I don't have a strong opinion, just calling out the option.

tests/appsec/test_appsec_utils.py
```diff
@@ -0,0 +1,61 @@
...
+
+def test_get_blocked_template_user_file_missing_html():
+    with override_env(dict(DD_APPSEC_HTTP_BLOCKED_TEMPLATE_HTML="missing.html")):
+        assert utils._get_blocked_template("text/html") == constants.APPSEC_BLOCKED_RESPONSE_HTML
+        assert utils._get_blocked_template("") == constants.APPSEC_BLOCKED_RESPONSE_JSON
+
+
+def test_get_blocked_template_user_file_missing_json():
+    with override_env(dict(DD_APPSEC_HTTP_BLOCKED_TEMPLATE_JSON="missing.json")):
+        assert utils._get_blocked_template("") == constants.APPSEC_BLOCKED_RESPONSE_JSON
```
brettlangdon, line 42: we should test with `application/json` as well

brettlangdon: `text/html` is what we use as the trigger in the code to determine html or json, but we should make sure the tests reflect real use cases rather than be written around implementation details of the code.

tests/appsec/test_appsec_utils.py
```diff
@@ -0,0 +1,61 @@
...
+        assert utils._get_blocked_template("text/html") == constants.APPSEC_BLOCKED_RESPONSE_HTML
+
+
+def test_get_blocked_template_user_file_exists_html():
+    template_path = os.path.join(os.path.dirname(__file__), "blocking_template_html.html")
```
brettlangdon, line 47: `pytest` has a `tmpdir` fixture to help with this:

https://github.com/DataDog/dd-trace-py/blob/97deadb4be6902c07536c4fc0a788b934641cadb/tests/integration/test_integration.py#L93-L106

tests/contrib/django/test_django_appsec.py
```diff
@@ -277,8 +277,6 @@ def test_django_path_params(client, test_spans, tracer):
 
 def test_django_useragent(client, test_spans, tracer):
     with override_global_config(dict(_appsec_enabled=True)):
-        tracer._appsec_enabled = True
-        tracer.configure(api_version="v0.4")
```
brettlangdon, line 281: unrelated?

tests/tracer/test_trace_utils.py
```diff
@@ -522,73 +522,53 @@ def test_set_http_meta_case_sensitive_headers_notfound(mock_store_headers, span,
...
+        ("via", {"via": "8.8.8.8"}, "8.8.8.8"),
         (
             "via",
             {"x-forwarded-for": "4.4.4.4", "via": "8.8.4.4"},
-            ["runtime-id", "network.client.ip", http.CLIENT_IP],
             "8.8.4.4",
         ),
     ],
 )
-def test_set_http_meta_headers_ip(
```
brettlangdon, line 576: do we have tests elsewhere that the appropriate tags get adds to spans when we have a request ip ?

tests/internal/remoteconfig/test_remoteconfig.py
```diff
@@ -172,6 +172,41 @@ def test_remote_configuration_check_deprecated_override():
...
+    with override_env(dict(DD_REMOTE_CONFIG_POLL_INTERVAL_SECONDS="0.1")):
+        mock_check_remote_config_enable_in_agent.return_value = True
+        mock_send_request.return_value = get_mock_encoded_msg(
+            b'{"rules_data": [{"data": [{"expiration": 1662804872, "value": "127.0.0.0"}, '
+            b'{"expiration": 1662804872, "value": "52.80.198.1"}], "id": "blocking_ips", '
+            b'"type": "ip_with_expiration"}]}'
+        )
+        rc = RemoteConfig()
+        rc.register(ASM_FEATURES_PRODUCT, callback._reload_features)
+        sleep(0.15)
```
brettlangdon, line 195: can we manually trigger a poll?

brettlangdon (commented):
I still have a few things to review, but here are some initial nits/comments.

tests/appsec/test_appsec_utils.py
juanjux: True. Will be in next commit.

tests/appsec/test_appsec_utils.py
juanjux: It's not a temporal dir but the dir where the test (and the fixture) is.

ddtrace/constants.py
juanjux: Or better yet, `ddtrace.appsec._constants` (on next commit)

ddtrace/constants.py
juanjux: The line above before it was reformatted, removed on next commit.

ddtrace/constants.py
juanjux: Mmmm I think some users could be interested on checking where the default blocking response comes, but this being Python not being in the public API would not really put a stop to it.

juanjux: Moved in next commit.

tests/appsec/test_appsec_utils.py
brettlangdon: oh... nvm. I read this too quickly. I saw the open calls, and was thinking we were writing content to a file. Sorry about that.

tests/tracer/test_trace_utils.py
juanjux: Some of the `test_set_http_meta_*`  in this file and in `test_django_appsec` and some other tests in django and flask. In this case the test wants to check the `_get_request_header_client_ip` which doesn't set any tags thus I removed them and the call of `set_http_meta` that didn't make sense for testing it (it was also needed for some unrelated thing related to the changes in these PRs that doesn't come to my mind now).

tests/appsec/test_appsec_utils.py
juanjux: For simple and not checking many cases tests like this I prefer the version without indirections, I just can read them better.

ddtrace/internal/remoteconfig/client.py
juanjux: Some mypy fight but I solved it and now it's like before.

tests/internal/remoteconfig/test_remoteconfig.py
juanjux: No idea. Will ask Alberto on monday.

ddtrace/internal/remoteconfig/client.py
```diff
@@ -370,17 +378,22 @@ def _process_response(self, data):
...
                 callback_action = False
 
-            callback = self._products[config.product_name]
-
-            try:
-                callback(config, callback_action)
-            except Exception:
-                log.debug("error while removing product %s config %r", config.product_name, config)
-                continue
+            callback = self._products.get(config.product_name)
```
avara1986, line 381: we check if the product exists [here](https://github.com/DataDog/dd-trace-py/pull/5032/files#diff-7f727a2a181fcfac54ea16cde6b40dc992e8b49d5c898b85400ee361c22889c8R342), we don't need `.get`

I think `if callback` it isn't needed

tests/internal/remoteconfig/test_remoteconfig.py
avara1986: @brettlangdon If I understand correctly, we test it in Flask PR:
https://github.com/DataDog/dd-trace-py/pull/5024/files#diff-bab792ffca2b00287b885d7e3d23f474135e6f5dd8aa81002f3be06cfecadad3

Or we have unit tests in 1.x:
https://github.com/DataDog/dd-trace-py/blob/1.x/tests/internal/remoteconfig/test_remoteconfig.py#L142
https://github.com/DataDog/dd-trace-py/blob/1.x/tests/internal/remoteconfig/test_remoteconfig_client.py#L85

Do you mean that?

brettlangdon: I just mean, instead of setting the polling sleep interval really low, and then sleeping, can we just manually trigger a poll?

e.g. do we have `rc.poll()`/`rc._run()` or something that will make the remote config client do what we need it to without worrying about sleeping/timing stuff?

ddtrace/internal/remoteconfig/client.py
juanjux: Code updated :+1:

juanjux: Code updated :+1:

tests/internal/remoteconfig/test_remoteconfig.py
avara1986: Ok, I understand :smiley: I think it's ok to test the Remote Config worker in background. I think it's more accurate this way rather than `RemoteConfig.request()`

emmettbutler (commented):
I don't have anything to add that other reviewers haven't already mentioned. Please request another review from me when you've resolved all conversations. Great work, and thanks!

ddtrace/appsec/_remoteconfiguration.py
```diff
@@ -30,13 +30,14 @@ def enable_appsec_rc():
     # type: () -> None
     # Import tracer here to avoid a circular import
     from ddtrace import tracer
+    from ddtrace.internal.remoteconfig import RemoteConfig
+    from ddtrace.internal.remoteconfig.constants import ASM_DATA_PRODUCT
+    from ddtrace.internal.remoteconfig.constants import ASM_FEATURES_PRODUCT
 
     if _appsec_rc_features_is_enabled():
-        from ddtrace.internal.remoteconfig import RemoteConfig
```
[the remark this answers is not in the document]
juanjux: Imports are still local, in `enable_appsec_rc` and `appsec_1click_activation` so it should not undo it.

ddtrace/appsec/_remoteconfiguration.py
```diff
@@ -95,7 +97,7 @@ def _reload_features(metadata, features):
 
         if features is not None:
             log.debug("Updating ASM Remote Configuration: %s", features)
-            _appsec_rules_data(tracer, features)
             _appsec_1click_activation(tracer, features)
+            _appsec_rules_data(tracer, features)
```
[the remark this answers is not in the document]
juanjux: Yep. Order matters. Added a comment.

ddtrace/internal/remoteconfig/client.py
```diff
@@ -239,19 +239,23 @@ def base64_to_struct(val, cls):
...
+        self._products = dict()  # type: MutableMapping[str, Optional[ProductCallback]]
         self._applied_configs = dict()  # type: Mapping[str, ConfigMetadata]
         self._last_targets_version = 0
         self._last_error = None  # type: Optional[str]
         self._backend_state = None  # type: Optional[str]
 
-    def register_product(self, product_name, func=None):
-        # type: (str, Optional[ProductCallback]) -> None
+    def register_product(self, product_name, func):
+        # type: (str, ProductCallback) -> None
```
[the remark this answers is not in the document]
juanjux: Updated this file to current 1.x since the new merged PRs have a more correct version.

ddtrace/internal/remoteconfig/client.py
```diff
@@ -239,19 +239,23 @@ def base64_to_struct(val, cls):
         self.converter.register_structure_hook(SignedRoot, base64_to_struct)
         self.converter.register_structure_hook(SignedTargets, base64_to_struct)
 
-        self._products = dict()  # type: MutableMapping[str, ProductCallback]
+        self._products = dict()  # type: MutableMapping[str, Optional[ProductCallback]]
```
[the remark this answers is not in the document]
juanjux: Ditto.

ddtrace/internal/remoteconfig/client.py
```diff
@@ -344,7 +347,12 @@ def _process_response(self, data):
         if last_targets_version is None or targets is None:
             log.debug("No targets in configuration payload")
             for callback in self._products.values():
```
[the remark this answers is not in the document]
juanjux: Ditto.

emmettbutler (approved):

avara1986 (approved):

gnufede (approved):

PR #5038, opened 2023-02-03
refactor(telemetry): improve maintainability and performance

mabdinur:
This PR was opened to stage all improvements to the telemetry writer. Feel free to push suggestions and add to the PR description. Changes: 

- Introduces an atomic counter to increment TelemetryWriter._sequence
      - With this change the writer does not need to acquire a lock to append to the events and integrations queues.
      - Note: this counter is only atomic in CPython due to an implementation detail in CPython and `itertools.count`. However, this change only impacts system tests and so should not impact users that use different distributions of Python.
- Moves `TelemetryWriter._create_headers(..)` and `TelemetryWriter._send_request(...)` methods to a new class: `TelemetryClient`. 
- Renames `TelemetryWriter.app_started_event` to `TelemetryWriter._app_started_event` 
- Renames `TelemetryWriter.app_heartbeat_event` to `TelemetryWriter._app_heartbeat_event`
- Removes agent_url arg from `TelemetryWriter.__init__()`
    - This argument is never used. This is not a breaking change since this object is only used in internal modules and tests.
- Removes `TelemetryWriter._create_telemetry_request(..)` and moves this logic to `TelemetryWriter.add_event()`
- Removes `TelemetryWriter._DD_TELEMETRY_WRITER_ADDITIONAL_HEADERS` from `TelemetryWriter.__init__()`
   - TelemetryWriter.__init__() is always called before this environment variable is set in tests, so it serves no purpose.
- Previously the value of the `debug` field in the body of Telemetry Events was serialized to `True`/`False`, these values should be lower case (`false`/`true`). This is a very minor nit that should not impact the backend. 
- Updates (and hopefully improves) inline documentation

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


avara1986:
:clap: :clap: nice refactor!!
Could we split client into new file? such as https://github.com/DataDog/dd-trace-py/tree/1.x/ddtrace/internal/remoteconfig WDYT?

ddtrace/internal/telemetry/writer.py
```diff
@@ -32,25 +32,63 @@ def _get_interval_or_default():
...
+
+    def url(self):
+        return "%s/%s" % (self._agent_url, self._endpoint)
+
+    def send_event(self, request):
+        # type: (Dict) -> Optional[httplib.HTTPResponse]
+        """Sends a telemetry request to the trace agent"""
+        resp = None
+        try:
+            rb_json = self._encoder.encode(request)
```
avara1986, line 52: What's the benefit of `JSONEncoderV2()`? should we use this class too in: https://github.com/DataDog/dd-trace-py/blob/1.x/ddtrace/internal/remoteconfig/client.py#L424

mabdinur: Honestly there is no real benefit here aside from aligning with existing writers. This encoder is used in the logs writer to format tracers. Here we only use the `encode` method which simply calls `json.JSONEncoder.encode()` This just call json.

mabdinur:
> 👏 👏 nice refactor!! Could we split client into new file? such as https://github.com/DataDog/dd-trace-py/tree/1.x/ddtrace/internal/remoteconfig WDYT?

I would prefer to the keep the client in this module since it will only be used by the TelemetryWriter. Also since the TelemetryClient is internal to the module we don't need to add/edit unit tests. The existing test coverage should be sufficient.

ddtrace/internal/telemetry/writer.py
```diff
@@ -126,31 +133,16 @@ def periodic(self):
...
-                log.debug(
-                    "failed to send telemetry to the Datadog Agent at %s/%s.",
-                    self._agent_url,
-                    self.ENDPOINT,
-                    exc_info=True,
-                )
+            self._app_heartbeat_event()
+
+        telemetry_events = self._flush_events_queue()
+
```
mabdinur, line 139:
```suggestion
        telemetry_events = self._flush_events_queue()
```

ddtrace/internal/telemetry/writer.py
```diff
@@ -187,21 +187,20 @@ def add_integration(self, integration_name, auto_enabled):
...
+        integration = {
+            "name": integration_name,
+            "version": "",
+            "enabled": True,
+            "auto_enabled": auto_enabled,
+            "compatible": True,
+            "error": "",
+        }
+        self._integrations_queue.append(integration)
```
mabdinur, line 201:
```suggestion
        if self._enabled:
            integration = {
                "name": integration_name,
                "version": "",
                "enabled": True,
                "auto_enabled": auto_enabled,
                "compatible": True,
                "error": "",
            }
            self._integrations_queue.append(integration)
```

ddtrace/internal/telemetry/writer.py
```diff
@@ -165,19 +157,27 @@ def _stop_service(self, *args, **kwargs):
...
-        :param Dict payload: stores a formatted telemetry request
+        :param Dict payload: stores a formatted telemetry event
         :param str payload_type: The payload_type denotes the type of telmetery request.
             Payload types accepted by telemetry/proxy v1: app-started, app-closing, app-integrations-change
         """
-        with self._lock:
-            if not self._enabled:
-                return
+        if not self._enabled:
+            return
```
mabdinur, line 167:
```suggestion
            event = {
                "tracer_time": int(time.time()),
                "runtime_id": get_runtime_id(),
                "api_version": "v1",
                "seq_id": next(self._sequence),
                "debug": str(self._debug).lower(),
                "application": get_application(config.service, config.version, config.env),
                "host": get_host_info(),
                "payload": payload,
                "request_type": payload_type,
            }
            self._events_queue.append(event)
```

ddtrace/internal/telemetry/writer.py
```diff
@@ -165,19 +157,27 @@ def _stop_service(self, *args, **kwargs):
...
 
-        :param Dict payload: stores a formatted telemetry request
+        :param Dict payload: stores a formatted telemetry event
         :param str payload_type: The payload_type denotes the type of telmetery request.
             Payload types accepted by telemetry/proxy v1: app-started, app-closing, app-integrations-change
         """
-        with self._lock:
-            if not self._enabled:
-                return
+        if not self._enabled:
```
mabdinur, line 166:
```suggestion
        if self._enabled:
```

ddtrace/internal/telemetry/writer.py
```diff
@@ -165,19 +157,27 @@ def _stop_service(self, *args, **kwargs):
...
+            "runtime_id": get_runtime_id(),
+            "api_version": "v1",
+            "seq_id": next(self._sequence),
+            "debug": str(self._debug).lower(),
+            "application": get_application(config.service, config.version, config.env),
+            "host": get_host_info(),
+            "payload": payload,
+            "request_type": payload_type,
+        }
+        self._events_queue.append(event)
```
mabdinur, line 180:
```suggestion
```

ddtrace/internal/telemetry/writer.py
```diff
@@ -237,34 +232,12 @@ def _app_closing_event(self):
...
         self.add_event(payload, "app-integrations-change")
 
-    def _create_headers(self, payload_type):
-        # type: (str) -> Dict
-        """Creates request headers"""
-        headers = self._headers.copy()
-        headers["DD-Telemetry-Request-Type"] = payload_type
-        return headers
-
-    def _create_telemetry_request(self, payload, payload_type, sequence_id):
```
mabdinur, line 253: moved to 165

ddtrace/internal/telemetry/writer.py
```diff
@@ -237,34 +232,12 @@ def _app_closing_event(self):
...
     def _app_integrations_changed_event(self, integrations):
         # type: (List[Dict]) -> None
-        """Adds a Telemetry request which sends a list of configured integrations to the agent"""
+        """Adds a Telemetry event which sends a list of configured integrations to the agent"""
         payload = {
             "integrations": integrations,
         }
         self.add_event(payload, "app-integrations-change")
 
-    def _create_headers(self, payload_type):
```
mabdinur, line 246: moved to _TelemetryClient

ddtrace/internal/telemetry/writer.py
```diff
@@ -126,31 +133,15 @@ def periodic(self):
 
         if not self._events_queue:
             # Optimization: only queue heartbeat if no other events are queued
-            self.app_heartbeat_event()
-
-        telemetry_requests = self._flush_events_queue()
-
-        for telemetry_request in telemetry_requests:
```
mabdinur, line 133: moved to _TelemetryClient

ddtrace/internal/telemetry/writer.py
```diff
@@ -32,25 +32,67 @@ def _get_interval_or_default():
...
+            log.debug("failed to send telemetry to the Datadog Agent at %s.", self.url, exc_info=True)
+        finally:
+            conn.close()
+        return resp
+
+    def get_headers(self, request):
+        # type: (Dict) -> Dict
+        """Get all telemetry api v1 request headers"""
+        headers = self._headers.copy()
+        headers["DD-Telemetry-Debug-Enabled"] = request["debug"]
```
Yun-Kim, line 75: Looks like `"DD-Telemetry-Debug-Enabled": str(self._debug).lower()` isn't transferred directly to `_TelemetryClient._headers`. Could you help me understand how this accomplishes that here?

ddtrace/internal/telemetry/writer.py
```diff
@@ -32,25 +32,67 @@ def _get_interval_or_default():
...
+        self._endpoint = endpoint
+        self._encoder = JSONEncoderV2()
+        self._headers = {
+            "Content-type": "application/json",
+            "DD-Telemetry-API-Version": "v1",
+        }
+
+    @property
+    def url(self):
+        return "%s/%s" % (self._agent_url, self._endpoint)
```
Yun-Kim, line 48: Why is `agent_url` fixed to always use `get_trace_url()` now?

ddtrace/internal/telemetry/writer.py
```diff
@@ -59,65 +101,30 @@ def __init__(self, agent_url=None):
         self._debug = asbool(os.environ.get("DD_TELEMETRY_DEBUG", "false"))
         forksafe.register(self._fork_writer)
 
-        self._headers = {
-            "Content-type": "application/json",
-            "DD-Telemetry-API-Version": "v1",
-            "DD-Telemetry-Debug-Enabled": str(self._debug).lower(),
-        }  # type: Dict[str, str]
-        additional_header_str = os.environ.get("_DD_TELEMETRY_WRITER_ADDITIONAL_HEADERS")
```
Yun-Kim, line 67: Even if `_DD_TELEMETRY_WRITER_ADDITIONAL_HEADERS` is set after the `TelemetryWriter` is initialized in tests, do we expect them to not be set by users pre-application startup? Should we be removing this entirely, or is this an internal environment variable?

ddtrace/internal/telemetry/writer.py
```diff
@@ -165,19 +156,25 @@ def _stop_service(self, *args, **kwargs):
...
-            request = self._create_telemetry_request(payload, payload_type, self._sequence)
-            self._sequence += 1
-            self._events_queue.append(request)
+        if self._enabled:
+            event = {
+                "tracer_time": int(time.time()),
+                "runtime_id": get_runtime_id(),
+                "api_version": "v1",
+                "seq_id": next(self._sequence),
+                "debug": str(self._debug).lower(),
```
Yun-Kim, line 171: Why are we adding these casts to lowercase string here? Previously we were just using the value of `self._debug` as is. Is there a reason for this change?

ddtrace/internal/telemetry/writer.py
```diff
@@ -165,19 +156,25 @@ def _stop_service(self, *args, **kwargs):
...
+                "application": get_application(config.service, config.version, config.env),
+                "host": get_host_info(),
+                "payload": payload,
+                "request_type": payload_type,
+            }
+            self._events_queue.append(event)
```
Yun-Kim, line 177: Does it matter the event's position on the `_events_queue`? It seems to me that even if the `seq_id` is obtained and incremented atomically this doesn't necessarily ensure that the event is placed onto the queue in the exact same position as it would have previous to this change.

Yun-Kim (commented):
Using `itertools.count` is only an atomic operation in CPython due to an implementation detail in CPython. I'm worried this will break on the off chance a customer using any other distribution.

ddtrace/internal/telemetry/writer.py
mabdinur: The `agent_url` argument in Telemetry.__init__(.....) was never used, so we always defaulted to using get_trace_url() anyway. The telemetry only submits payloads to the datadog agent. We can configure this value using the DD_TRACE_AGENT_URL configuration.  I'll update the PR description to include this

ddtrace/internal/telemetry/writer.py
mabdinur: No users will never set this configuration. This configuration is used by the test agent. This code block shouldn't have been added in the first place because it does nothing.

ddtrace/internal/telemetry/writer.py
mabdinur: self._debug is serialized to `True`/`False` by the json encoder but we set the `DD-Telemetry-Debug-Enabled` header to `true`/`false`. This just makes the values consistent but this isn't a "fix" since the backend can handle both values.

ddtrace/internal/telemetry/writer.py
mabdinur: Nope the position in the events queue does not matter. The only thing that matters is the timestamp and sequence ids in payloads. Also the sequence id is not used by the backend it is only used system tests.

ddtrace/internal/telemetry/writer.py
mabdinur: The "DD-Telemetry-Debug-Enabled" header is now set in `TelmeetryClient.get_headers()` instead of TelemetryWriter.__init__(). The value set should always be the same since 
[get_headers](https://github.com/DataDog/dd-trace-py/blob/1313f54e4f568f6eba1f2e048b204da437d48cb4/ddtrace/internal/telemetry/writer.py#L56) is called before every telemetry request is sent

mabdinur: Yup they should be the same value. Right now they are not (now when the request body has `debug:False` the value in the header is `DD-Telemetry-Debug-Enabled:false`). This change makes sure they're consistent. This isn't a fix since the backend can accept both values. The inconsistency just didn't sit right with me

Yun-Kim:
I still have a concern that this will only work as intended for CPython, as `itertools.count` is only an atomic operation in and due to an implementation detail in CPython. This will probably break for users that do not use CPython.

mabdinur:
> I still have a concern that this will only work as intended for CPython, as `itertools.count` is only an atomic operation in and due to an implementation detail in CPython. This will probably break for users that do not use CPython.

Honestly using a non atomic counter is not a deal. The sequence field is only used in system tests to validate that the correct payload was received. Best effort is good enough here

ddtrace/internal/telemetry/writer.py
```diff
@@ -59,65 +101,30 @@ def __init__(self, agent_url=None):
...
-                    sw.elapsed(),
-                    self._agent_url,
-                    self.ENDPOINT,
-                    resp.status,
-                )
-                return resp
-            finally:
-                conn.close()
+        # counter representing the number of events sent by the writer
+        self._sequence = itertools.count(1)
```
P403n1x87, line 105: Just a reminder that atomicity here is an implementation detail of `itertools.count` and that this could change at any time.

P403n1x87:
Can we have a way to quantify the performance improvements? Perhaps a CPU profile or some benchmarks?

emmettbutler (approved):
Thanks for making our codebase a little cleaner! This is a great change.

mabdinur:
> Can we have a way to quantify the performance improvements? Perhaps a CPU profile or some benchmarks?

There are no significant performance improvement right now but the ASM team is planning to queue telemetry events on span finish ([PR](https://github.com/DataDog/dd-trace-py/pull/5046/files#diff-6327979a325e44c1fa8d693f7d510b4d381d750ef95f12037e081c39081872ba)). That PR will increase lock contention.

ddtrace/internal/telemetry/writer.py
```diff
@@ -59,65 +101,30 @@ def __init__(self, agent_url=None):
...
-                    len(rb_json),
-                    sw.elapsed(),
-                    self._agent_url,
-                    self.ENDPOINT,
-                    resp.status,
-                )
-                return resp
-            finally:
-                conn.close()
+        # counter representing the number of events sent by the writer
```
mabdinur, line 104:
```suggestion
        # Counter representing the number of events sent by the writer. Here we are relying on the atomicity 
        # of `itertools.count()` which is a CPython implementation detail. The sequence field in telemetry payloads is 
        # only used in tests and is not a required to process Telemetry events.
```

ddtrace/internal/telemetry/writer.py
```diff
@@ -59,65 +101,32 @@ def __init__(self, agent_url=None):
...
-                    self._agent_url,
-                    self.ENDPOINT,
-                    resp.status,
-                )
-                return resp
-            finally:
-                conn.close()
+        # Counter representing the number of events sent by the writer. Here we are relying on the atomicity
+        # of `itertools.count()` which is a CPython implementation detail. The sequence field in telemetry payloads is 
+        # only used in tests and is not a required to process Telemetry events.
```
mabdinur, line 106:
```suggestion
        # of `itertools.count()` which is a CPython implementation detail. The sequence field in telemetry
        # payloads is only used in tests and is not a required to process Telemetry events.
```

Yun-Kim:
Everything looks great, but I'd like to discuss any possible risks from using a CPython-dependent implementation detail one last time.

> Honestly using a non atomic counter is not a [big] deal. The sequence field is only used in system tests to validate that the correct payload was received. Best effort is good enough here

I understand system-tests uses CPython so we should be fine for now, but is there anywhere else `TelemetryWriter._sequence` is used or impacts? If a user with PyPy (or any other non-CPython distribution) were to use this and in case of a race condition, would this cause any potential issues to a user's application? Taking it one step further, by your logic what's stopping us from even using a lock/atomic counter here at all then?

mabdinur:
> Everything looks great, but I'd like to discuss any possible risks from using a CPython-dependent implementation detail one last time.
> 
> > Honestly using a non atomic counter is not a [big] deal. The sequence field is only used in system tests to validate that the correct payload was received. Best effort is good enough here
> 
> I understand system-tests uses CPython so we should be fine for now, but if that were to change for whatever reason, are we okay with breaking system-tests? It seems to me that if that were to happen we would just have to revert back to using locks, which makes me feel this change is somewhat unnecessary for the headache it would potentially cause later on.
> 
> I also looked at the usage of `TelemetryWriter._sequence` in our codebase and couldn't find any other impacts in our codebase. I'm okay with making this change only if we're okay with the risk I mentioned above (breaking system-tests), and if we're 100% confident this attribute will not impact users / be used outside of system-tests. From what I've checked it looks like a test-only attribute, but I'd like your confirmation just in case before we go ahead with this.

I have no strong opinions here but this field is very insignificant. Worst case we can update the system tests to avoid using this field and just remove this field from the telemetry payload. Dealing with locks is a pain, I rather avoid them unless it's absolutely necessary.

ddtrace/internal/telemetry/writer.py
```diff
@@ -59,65 +101,32 @@ def __init__(self, agent_url=None):
...
-                    self._agent_url,
-                    self.ENDPOINT,
-                    resp.status,
-                )
-                return resp
-            finally:
-                conn.close()
+        # Counter representing the number of events sent by the writer. Here we are relying on the atomicity
+        # of `itertools.count()` which is a CPython implementation detail. The sequence field in telemetry
+        # payloads is only used in tests and is not a required to process Telemetry events.
```
Yun-Kim, line 106:
```suggestion
        # payloads is only used in tests and is not required to process Telemetry events.
```

ddtrace/internal/telemetry/writer.py
```diff
@@ -188,10 +187,8 @@ def add_integration(self, integration_name, auto_enabled):
         :param str integration_name: name of patched module
         :param bool auto_enabled: True if module is enabled in _monkey.PATCH_MODULES
         """
-        with self._lock:
-            if self._enabled is not None and not self._enabled:
-                return
-
+        if self._enabled is None or self._enabled:
```
Yun-Kim, line 190: Out of curiosity, can you explain why removing the `if self._enabled is None` resulted in application issues (failed gunicorn tests)? Thanks!

Yun-Kim (approved):

ddtrace/internal/telemetry/writer.py
Yun-Kim, line 190: Out of curiosity, can you explain why removing the if self._enabled is None resulted in application issues (failed gunicorn tests)? Thanks!

Yun-Kim (approved):
Still concerned about relying on CPython implementation details for atomic operations, but will approve based on the fact that this doesn't affect users and only system-tests.

ddtrace/internal/telemetry/writer.py
mabdinur: In this commit I refactored the `if self._enabled` check to make it more readable (ex: convert `if not enabled: return` to `if enabled: do_work .....`): [183dc254](https://github.com/DataDog/dd-trace-py/pull/5038/commits/69f574c50f77d457b555fe0ce8ce50ea183dc254). In this refactor I forgot that self._enabled boolean is an optional boolean and could be None. This bug prevented integrations from being queued on application start up which caused some ddtrace run test to fail. The failing gunicorn test asserted that a memory allocation error was raised.

emmettbutler (approved):

PR #5046, opened 2023-02-06
feat(telemetry): update telemetry to report metrics

avara1986:
# Description
Split telemetry worker: One Worker to report events (heartbeat, startup...) Other to report metrics.
Class methods to report distributions, gauge, count and rate Metrics type. 
ASM Metrics such as `waf.requests`, `waf.duration`, etc.

# Motivation
ASM needs to report information about 

# Testing strategy
See Slack

# PRs related:
System tests: https://github.com/DataDog/system-tests/pull/823

This PR comes from https://github.com/DataDog/dd-trace-py/pull/4943 cc @mabdinur 

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/internal/telemetry/metrics.py
```diff
@@ -0,0 +1,129 @@
...
+    A count type adds up all the submitted values in a time interval. This would be suitable for a
+    metric tracking the number of website hits, for instance.
+    """
+
+    def add_point(self, value=1.0):
+        # type: (float) -> None
+        """adds timestamped data point associated with a metric"""
+        timestamp = int(time.time())
+        # self._count += 1.0
+        # self._points = [(timestamp, self._count)]
```
christophe-papazian, line 95: maybe remove the comment lines that are not used anymore ?

ddtrace/internal/telemetry/metrics_namespaces.py
```diff
@@ -0,0 +1,58 @@
...
+        self._metrics_data = {TELEMETRY_TRACER: {}, TELEMETRY_APPSEC: {}}
+
+    def get(self):
+        # type: () -> Dict[str, Any]
+        namespace_metrics = self._metrics_data.copy()
+        return namespace_metrics
+
+    def validate_type_metric(self, metric_check, metric_to_validate):
+        # type: (Metric, Metric) -> bool
+        if metric_check.__class__ == metric_to_validate.__class__:
```
christophe-papazian, line 35: why checking only the class ?

ddtrace/internal/telemetry/metrics_namespaces.py
```diff
@@ -0,0 +1,58 @@
...
+        if not self.validate_type_metric(metric_check, metric):
+            raise TelemetryTypeError(
+                (
+                    'Error: metric with name "%s" and type "%s" exists. You can\'t create a new metric '
+                    'with this name an type "%s"'
+                )
+                % (name, metric.type, metric_check.type)
+            )
+        metric.add_point(value)
+        self._metrics_data[namespace][metric.id] = metric
```
christophe-papazian, line 58: Maybe use set instead of dict for namespace, as the key here is meaningless, and sets will use the id from hash anyway.

avara1986: Each metric is sent in a single payload group by namespace (appsec, tracer...) we use it to aggragate the metrics for each namespace :+1:

ddtrace/internal/telemetry/metrics_namespaces.py
christophe-papazian: ok, but the way it is done here is based on id, not name. So two metrics with the same name but different tags will still be possible. is this ok?

avara1986: yes, two metrics with the same name but different tags it will be reported as two different metrics (check it https://www.datadoghq.com/blog/the-power-of-tagged-metrics/#whats-a-metric-tag)

But we CAN NOT have two metrics with the same name and different type

christophe-papazian: ok, this looks perfect!

christophe-papazian (dismissed):

ddtrace/internal/telemetry/metrics_namespaces.py
```diff
@@ -0,0 +1,59 @@
...
+        """
+        Queues metric
+        """
+        name = "dd.app_telemetry." + namespace + "." + name
+        metric_check = self.metric_class[metric_type](  # type: ignore[abstract]
+            namespace, name, tags=tags, common=True, interval=interval
+        )
+        metric = self._metrics_data[namespace].get(metric_check.id, metric_check)
+        if not self.validate_type_metric(metric_check, metric):
+            raise TelemetryTypeError(
```
mabdinur, line 51: Is this error caught by the telemetry writer? Could this error crash an application?

avara1986: This error raises when you call `telemetry_writer.add_[type]_metric()`. In this case, a crash error it's ok to catch the problem while we're developing. If not, we've the risk to add a metric with errors/not working, merge and add it into a release

ddtrace/internal/constants.py
```diff
@@ -16,3 +16,8 @@
 W3C_TRACESTATE_ORIGIN_KEY = "o"
 W3C_TRACESTATE_SAMPLING_PRIORITY_KEY = "s"
 DEFAULT_SERVICE_NAME = "unnamed_python_service"
+
+TELEMETRY_TRACER = "tracers"
+TELEMETRY_APPSEC = "appsec"
```
emmettbutler, line 21: A slightly more descriptive naming convention on these constants would help readers understand their intended purpose. `TELEMETRY_TAG_TRACER`, maybe?

ddtrace/appsec/processor.py
```diff
@@ -297,13 +299,21 @@ def on_span_finish(self, span):
             if info.errors:
                 span.set_tag_str(APPSEC_EVENT_RULE_ERRORS, json.dumps(info.errors))
             span.set_tag_str(APPSEC_EVENT_RULE_VERSION, info.version)
+            # telemetry_writer.add_gauge_metric(TELEMETRY_APPSEC, "event_rules.version", info.version, {})
```
emmettbutler, line 302: What's the plan with these commented lines?

tests/telemetry/test_telemetry_metrics.py
```diff
@@ -0,0 +1,255 @@
...
+        "series": expected_series,
+    }
+    assert events[0]["request_type"] == TELEMETRY_TYPE_GENERATE_METRICS
+
+    assert events[0] == _get_request_body(payload, TELEMETRY_TYPE_GENERATE_METRICS, seq_id)
+
+
+@pytest.mark.skipif(sys.version_info < (3, 6), reason="mock.ANY doesn't works in py3.5 or lower")
+def test_send_metric_flush_and_series_is_restarted(test_agent_session_telemetry_metrics, mock_time):
+    """A datapoint is at least: a metric name, a metric value, and the time at which the value was collected.
```
emmettbutler, line 33: This docstring didn't help me understand the function. You might consider removing it here and in the few other places it's duplicated.

tests/telemetry/test_telemetry_metrics.py
```diff
@@ -0,0 +1,255 @@
...
+            "type": "count",
+        },
+    ]
+
+    _assert_metric(test_agent_session_telemetry_metrics, expected_series)
+    test_agent_session_telemetry_metrics.clear()
+
+    telemetry_writer.add_count_metric(TELEMETRY_TRACER, "test-metric2", 1, {"a": "b"})
+
+    _assert_metric(test_agent_session_telemetry_metrics, expected_series, seq_id=2)
```
emmettbutler, line 56: I don't understand how this assertion maps onto the idea of "flush and series is restarted". Is the key that the `points` array only has one element in it, indicating that another datapoint was not added to the series?

tests/telemetry/test_telemetry_metrics.py
```diff
@@ -0,0 +1,255 @@
...
+
+
+@pytest.mark.skipif(sys.version_info < (3, 6), reason="mock.ANY doesn't works in py3.5 or lower")
+def test_send_metric_datapoint_equal_type_and_tags_expected_1_serie(test_agent_session_telemetry_metrics, mock_time):
```
emmettbutler, line 60:
```suggestion
def test_send_metric_datapoint_equal_type_and_tags_yields_single_series(test_agent_session_telemetry_metrics, mock_time):
```

tests/telemetry/test_telemetry_metrics.py
```diff
@@ -0,0 +1,255 @@
...
+            "tags": {"a": "b"},
+            "type": "count",
+        },
+    ]
+
+    _assert_metric(test_agent_session_telemetry_metrics, expected_series)
+
+
+@pytest.mark.skipif(sys.version_info < (3, 6), reason="mock.ANY doesn't works in py3.5 or lower")
+def test_send_metric_datapoint_equal_type_different_tags_expected_3_serie(
```
emmettbutler, line 85:
```suggestion
def test_send_metric_datapoint_equal_type_different_tags_yields_multiple_series(
```

tests/telemetry/test_telemetry_metrics.py
```diff
@@ -0,0 +1,255 @@
...
+    """A datapoint is at least: a metric name, a metric value, and the time at which the value was collected.
+    But in Datadog, a datapoint also includes tags, which declare all the various scopes the datapoint belongs to
+    https://www.datadoghq.com/blog/the-power-of-tagged-metrics/#whats-a-metric-tag
+    """
+    telemetry_writer = test_agent_session_telemetry_metrics.telemetry_writer
+    telemetry_writer.add_count_metric(TELEMETRY_TRACER, "test-metric", 1, {"a": "b"})
+    with pytest.raises(TelemetryTypeError) as e:
+        telemetry_writer.add_gauge_metric(TELEMETRY_TRACER, "test-metric", 1, {"a": "b"})
+
+    assert e.value.args[0] == (
```
emmettbutler, line 142: I think this assertion needs to be within the `with` block to see the `e` variable. It's at least slightly clearer to the reader that way.

tests/telemetry/test_telemetry_metrics.py
```diff
@@ -0,0 +1,255 @@
...
+            "points": [[1642544540, 6.0]],
+            "tags": {},
+            "type": "count",
+        },
+    ]
+
+    _assert_metric(test_agent_session_telemetry_metrics, expected_series)
+
+
+def test_send_metric_datapoint_equal_tags_different_type_expected_1_serie(
```
emmettbutler, line 130:
```suggestion
def test_send_metric_datapoint_equal_tags_different_type_throws_error(
```

ddtrace/internal/telemetry/metrics_namespaces.py
```diff
@@ -0,0 +1,59 @@
...
+
+class TelemetryTypeError(Exception):
+    pass
+
+
+class MetricNamespace:
+    _metrics_data = {TELEMETRY_TRACER: {}, TELEMETRY_APPSEC: {}}  # type: NamespaceMetricType
+    metric_class = {"count": CountMetric, "gauge": GaugeMetric, "rate": RateMetric}
+
+    def _flush(self):
```
emmettbutler, line 25: `reset` might be a more appropriate name for this functionality, if I understand correctly what it's doing.

ddtrace/internal/telemetry/metrics.py
```diff
@@ -0,0 +1,131 @@
...
+from ..hostname import get_hostname
+
+
+MetricType = Literal["count", "gauge", "rate"]
+MetricTagType = Dict[str, str]
+
+
+class Metric(six.with_metaclass(abc.ABCMeta)):
+    """
+    stores metrics which will be sent to the Telemetry Intake metrics to the Datadog Instrumentation Telemetry Org
```
emmettbutler, line 21: What does this mean?

ddtrace/internal/telemetry/metrics.py
```diff
@@ -0,0 +1,131 @@
...
+    """
+
+    HOST_NAME = get_hostname()
+    metric_type = ""
+
+    def __init__(self, namespace, name, tags, common, interval=None):
+        # type: (str, str, MetricTagType, bool, Optional[float]) -> None
+        """
+        name: metric name
```
emmettbutler, line 30: This comment doesn't add any information one can't get from the variable name. I'd probably remove it.

ddtrace/internal/telemetry/metrics.py
```diff
@@ -0,0 +1,131 @@
...
+        common: set to True if a metric is common to all tracers, false if it is python specific
```
emmettbutler, line 31: Maybe you could name this argument `is_common_to_all_tracers`.

ddtrace/internal/telemetry/metrics.py
```diff
@@ -0,0 +1,131 @@
...
+        interval: field set for gauge and rate metrics, any field set is ignored for count metrics (in secs)
+        """
+        self.name = name
+        self.common = common
+        self.interval = interval
+        self._roll_up_interval = interval
```
emmettbutler, line 37:
```suggestion
        self.interval = self._roll_up_interval = interval
```

ddtrace/internal/telemetry/metrics.py
```diff
@@ -0,0 +1,131 @@
...
+    def __hash__(self):
+        return self.id
+
+    @abc.abstractmethod
+    def add_point(self, value=1.0):
+        # type: (float) -> None
+        """adds timestamped data point associated with a metric"""
+        pass
+
+    def set_tags(self, tags):
```
emmettbutler, line 59: It might not be particularly intuitive to a caller that this function replaces `self._tags` rather than `extend`ing it. You might either clearly document that fact or change the implementation to use `extend`.

ddtrace/internal/telemetry/metrics.py
```diff
@@ -0,0 +1,131 @@
...
+    Choosing the correct metric type ensures accurate data.
+    """
+
+    metric_type = "gauge"
+
+    def add_point(self, value=1.0):
+        # type: (float) -> None
+        """adds timestamped data point associated with a metric"""
+        timestamp = int(time.time())
+        self._points = [(timestamp, float(value))]
```
emmettbutler, line 112: Looks like this replaces the entire `_points` array with a single-element array. Why is that?

emmettbutler (changes requested):
Looking good, @avara1986! I took a close look at this today because the Python APM group was tagged for a review. I tried to be as helpful as possible in my comments. Please ask any questions you have and tell me if you'd like me to make any adjustments to my feedback approach. Thank you for the pull request!

ddtrace/internal/constants.py
avara1986: You're right, I think the correct name would be `TELEMETRY_NAMESPACE_TAG_TRACER`

ddtrace/appsec/processor.py
avara1986: I added a TODO below:
https://github.com/DataDog/dd-trace-py/pull/5046/files/49c7b6f056188c30450615d1349b0165e161309a#diff-0b32c6cdae8bdd01786317bcabebf73dbb57e535adab74c94858fda05da411efR610

I'm going to add the same TODO here

ddtrace/internal/telemetry/metrics_namespaces.py
avara1986: I set _flush name because TelemetryWriter use the same terms to retrieve and reset the data

ddtrace/internal/telemetry/writer.py
```diff
@@ -53,18 +60,22 @@ def __init__(self, agent_url=None):
...
         self._headers = {
             "Content-type": "application/json",
             "DD-Telemetry-API-Version": "v1",
+            "DD-Client-Library-Language": "python",
             "DD-Telemetry-Debug-Enabled": str(self._debug).lower(),
+            "DD-Agent-Env": "config.env",  # TODO
+            "DD-Agent-Hostname": "",  # TODO
         }  # type: Dict[str, str]
         additional_header_str = os.environ.get("_DD_TELEMETRY_WRITER_ADDITIONAL_HEADERS")
+        self._debug = asbool(os.environ.get("DD_TELEMETRY_DEGUG", "false"))
```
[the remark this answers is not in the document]
avara1986: Ups, I think it was a wrong automatic merge :S

ddtrace/internal/telemetry/metrics.py
```diff
@@ -0,0 +1,131 @@
...
+    def add_point(self, value=1.0):
+        # type: (float) -> None
+        """adds timestamped data point associated with a metric"""
+        timestamp = int(time.time())
+        self._points.append((timestamp, float(value)))
+
+
+class GaugeMetric(Metric):
+    """
+    A gauge type takes the last value reported during the interval. This type would make sense for tracking RAM or
```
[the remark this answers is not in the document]
avara1986: DD developers. If we want to add new metrics I think it will useful to have got the description of the type of metrics close :+1:

tests/telemetry/test_telemetry_metrics.py
avara1986: When we add a metric, this data is stored in the telemetry_writer instance. When we run telemetry_writer.periodic() the information is sent to the agent and flush the queue.

We want to test isolated elements in the queue and ensure the queue is empty

ddtrace/internal/telemetry/metrics.py
avara1986: cause GaugeMetric we want to store the last measure we take

ddtrace/internal/constants.py
```diff
@@ -16,7 +16,12 @@
 W3C_TRACESTATE_ORIGIN_KEY = "o"
 W3C_TRACESTATE_SAMPLING_PRIORITY_KEY = "s"
 DEFAULT_SERVICE_NAME = "unnamed_python_service"
-# Used to set the name of an integration on a span
```
gnufede, line 19: You're removing a comment here

ddtrace/internal/telemetry/metrics.py
avara1986: e.g: https://docs.datadoghq.com/metrics/#what-are-metrics

juanjux (approved):

gnufede (approved):

ddtrace/internal/constants.py
```diff
@@ -16,6 +16,7 @@
 W3C_TRACESTATE_ORIGIN_KEY = "o"
 W3C_TRACESTATE_SAMPLING_PRIORITY_KEY = "s"
 DEFAULT_SERVICE_NAME = "unnamed_python_service"
+
```
mabdinur, line 19:
```suggestion
```

ddtrace/internal/remoteconfig/client.py
```diff
@@ -25,8 +25,7 @@
 from ddtrace.internal.remoteconfig.constants import REMOTE_CONFIG_AGENT_ENDPOINT
 from ddtrace.internal.runtime import container
 from ddtrace.internal.utils.time import parse_isoformat
-
-from ..utils.version import _pep440_to_semver
+from ddtrace.internal.utils.version import _pep440_to_semver
```
mabdinur, line 28: Is this change required?

ddtrace/internal/telemetry/metrics.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+class CountMetric(Metric):
+    """
+    A count type adds up all the submitted values in a time interval. This would be suitable for a
+    metric tracking the number of website hits, for instance.
+    """
+
+    metric_type = TELEMETRY_METRIC_TYPE_COUNT
+    _points = []
```
mabdinur, line 98: _points is an attribute in the parent class (Metric). We shouldn't override it here.

```suggestion
```

ddtrace/internal/telemetry/metrics.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+class DistributionMetric(Metric):
+    """
+    The rate type takes the count and divides it by the length of the time interval. This is useful if you’re
+    interested in the number of hits per second.
+    """
+
+    metric_type = TELEMETRY_METRIC_TYPE_DISTRIBUTIONS
+    _points = []
```
mabdinur, line 157: _points is an attribute in the parent class (Metric). We shouldn't override it here.

```suggestion
```

tests/telemetry/conftest.py
```diff
@@ -106,3 +106,19 @@ def test_agent_session(telemetry_writer, request):
         telemetry_writer.periodic()
         del telemetry_writer._client._headers["X-Datadog-Test-Session-Token"]
         del os.environ["_DD_TELEMETRY_WRITER_ADDITIONAL_HEADERS"]
+
+
+@pytest.fixture
+def test_agent_session_telemetry_metrics(test_agent_session):
+    test_agent_session.telemetry_writer._flush_namespace_metrics()
+    test_agent_session.telemetry_writer._flush_events_queue()
+    test_agent_session.telemetry_writer._flush_integrations_queue()
```
mabdinur, line 115: We should update telemetry_writer._reset() to flush metrics queues

```suggestion
    test_agent_session.telemetry_writer._reset()
```

tests/telemetry/test_writer.py
```diff
@@ -36,7 +29,7 @@ def test_add_event(telemetry_writer, test_agent_session):
     assert len(requests) == 1
     assert requests[0]["headers"]["Content-Type"] == "application/json"
     assert requests[0]["headers"]["DD-Telemetry-Request-Type"] == payload_type
-    assert requests[0]["headers"]["DD-Telemetry-API-Version"] == "v1"
+    assert requests[0]["headers"]["DD-Telemetry-API-Version"] == "v2"
```
mabdinur, line 32: Can we add test coverage for the new headers telemetry headers

ddtrace/internal/telemetry/writer.py
```diff
@@ -74,6 +88,9 @@ def get_headers(self, request):
         headers = self._headers.copy()
         headers["DD-Telemetry-Debug-Enabled"] = request["debug"]
         headers["DD-Telemetry-Request-Type"] = request["request_type"]
+        headers["DD-Agent-Hostname"] = get_host_info()["hostname"]
```
mabdinur, line 91: api version is also defined in the request body, it might make sense to keep both values consistent. 

```suggestion
        headers["DD-Telemetry-API-Version""] = request["api_version"]
        headers["DD-Agent-Hostname"] = get_host_info()["hostname"]
```

ddtrace/internal/telemetry/writer.py
```diff
@@ -156,8 +186,61 @@ def _stop_service(self, *args, **kwargs):
...
+                    }
+                    log.debug("%s request payload: %s", payload_type, payload)
+                    if payload_type == TELEMETRY_TYPE_DISTRIBUTION:
+                        self.add_event(payload, TELEMETRY_TYPE_DISTRIBUTION)
+                    elif payload_type == TELEMETRY_TYPE_GENERATE_METRICS:
+                        self.add_event(payload, TELEMETRY_TYPE_GENERATE_METRICS)
+
     def add_event(self, payload, payload_type):
-        # type: (Dict, str) -> None
+        # type: (Dict[str, Any], str) -> None
```
mabdinur, line 243: api version on line 255 should be v2

mabdinur (commented):
Called out some nits, but overall looks good to me. Seeing how this is a high impact change can we run this branch on a sample app and see if we collect accurate metrics or release this feature behind a feature flag? 

The instrumentation telemetry datadog org will aggregate metrics across tracer versions. If the first release of this feature submits inaccurate metrics we can expect to receive these values for the foreseeable future.

tests/telemetry/test_writer.py
avara1986: :+1:

ddtrace/internal/telemetry/writer.py
avara1986: :ok_hand:

ddtrace/internal/telemetry/metrics.py
avara1986: :+1:

ddtrace/internal/remoteconfig/client.py
avara1986: No, it isn't, rollback

ddtrace/internal/telemetry/writer.py
avara1986: :ok_hand:

tests/telemetry/conftest.py
avara1986: Updated with the last refactor, thanks for the suggestion!

ddtrace/internal/remoteconfig/client.py
avara1986: :+1:

ddtrace/internal/telemetry/writer.py
```diff
@@ -70,50 +88,68 @@ def send_event(self, request):
...
+
+        :param Dict payload: stores a formatted telemetry event
+        :param str payload_type: The payload_type denotes the type of telmetery request.
+            Payload types accepted by telemetry/proxy v1: app-started, app-closing, app-integrations-change
+        """
+        if self._enabled:
+            event = {
+                "tracer_time": int(time.time()),
+                "runtime_id": get_runtime_id(),
+                "api_version": "v1",
```
mabdinur, line 144: Something to call out. I spoke to @mehulsonowal and it appears api v2 is not fully backwards compatible with v1. We should avoid bumping the telemetry writer to v2 in this PR (if possible) or scope api v2 to just metrics events.

avara1986: yes, I noticed if I update all telemetry requests to v2 system tests start to fail. I rollback to v1 and I tested manually and everything works fine. On the other hand, I split TelemetryWriter into TelemetryWriter (inverval of 60s and we use it to report heartbeat, startapp, close...) and TelemetryMetricsWriter (interval of 10s and to report generate-metrics and distributions). 

Thank you so much for your tip!

.github/CODEOWNERS
```diff
@@ -27,6 +27,7 @@ tests/debugging/                    @DataDog/debugger-python
 ddtrace/appsec/                     @DataDog/asm-python
 tests/appsec/                       @DataDog/asm-python
 tests/contrib/flask/test_flask_appsec.py    @DataDog/asm-python
+tests/contrib/flask/test_flask_appsec_telemetry.py    @DataDog/asm-python
```
gnufede, line 30: I think this should be covered by this line below:
```
tests/contrib/*/test*appsec*.py @DataDog/asm-python
```

avara1986: :+1:

juanjux (dismissed):

ddtrace/appsec/iast/cmake-build-debug/CMakeFiles/clion-log.txt
```diff
@@ -0,0 +1 @@
+CMakeLists.txt not found in /home/alberto.vara/projects/dd-python/dd-trace-py/ddtrace/appsec/iast Select CMakeLists.txt
```
mabdinur, line 1: Should we remove this file from this PR and add it to `.gitignore`?

ddtrace/appsec/processor.py
```diff
@@ -190,6 +192,57 @@ def _update_rules(self, new_rules):
         except TypeError:
             log.debug("Error updating ASM rules", exc_info=True)
 
+    def _set_metrics(self):
```
mabdinur, line 195: Ideally this PR should introduce the functionality to queue and send telemetry metrics. Then in a follow up PR we can use the telemetry writer to send metrics in different products. Can we decouple adding the telemetry metrics writer change from sending telemetry metrics in ASM products? 

This way if there is an issue with the telemetry writer (performance or a bug) we can just revert/fix the PRs which send telemetry and not all the logic in this PR. This is just a suggestion, if we need this functionality merged ASAP then I am fine with leaving this PR as is

mabdinur, line 195: We should add test coverage for test metrics. Right now there is no regression tests for appsec processor metrics, the values can change/break at any point and we won't be notified.

avara1986: Thank you so much @mabdinur , Nice idea to add performance tests :+1: 
About integration and regression tests is it enough with these tests?
- https://github.com/DataDog/dd-trace-py/pull/5046/files#diff-e56afc084b0d6a7e0a5d1c16ead0bc8d9916c99127c9c1d2224130b9e940e11eR1
- https://github.com/DataDog/dd-trace-py/pull/5046/files#diff-c210bd10d4e51b5ab136feb9a72477acc700d4183c3c59da7b942fc0de6d5d53R1
- https://github.com/DataDog/system-tests/pull/890

DM me if you want to share your thoughts and discuss the best approach :upside_down_face:

avara1986: Thanks! your suggestion is a good approach! If I start now to split the PR in smaller PRs I'll became mad :sweat_smile: What do you think about this idea?
- I'm continue in Telemetry Metrics in this branch
- When we're ready for the final review and merge I can split this PR into "Telemetry Metrics Core" and "ASM Telemetry Metrics"

is it OK for you? I'd love to know if this approach is quite so smooth for all of us :+1:

mabdinur: Yup that sounds great to me, it would be nice to get @Yun-Kim feedback on the performance benchmarks. Thanks for sticking with this. This PR looks good to me. If you can get another thumbs up from another review then we can split up and merging this change.

mabdinur:
LGTM, thanks for sticking with this 😄

ddtrace/appsec/processor.py
Yun-Kim: I'm curious why this change results in a regression in the `span-add-metrics` microbenchmark scenario, which specifically only call `span.set_metrics()` and shouldn't be impacted by telemetry, unless I'm missing something. @avara1986 do you have any clues as to why we're seeing this?

avara1986: ey Yun! no, I haven't. It's strange because the microbenchmark didn't report anything in the last commits, even after adding specific benchmark tests for Telemetry Metrics. `span-add-metrics` reports a regression after a merge of 1.x in this branch. But I merged again 1.x into this branch and the regression alert disappear. Flaky? I figure out that to be sure :+1:

emmettbutler (approved):

PR #5053, opened 2023-02-06
feat(tracing): add db.row_count to pymongo integration

wconti27:
Adds `db.row_count` to pymongo integration to match new Datadog Span spec.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


emmettbutler:
@wconti27 do you want a review on this while it's in its "draft" state, or should I wait?

ddtrace/contrib/pymongo/client.py
```diff
@@ -286,3 +309,19 @@ def _set_query_metadata(span, cmd):
...
+    search_key = "Batch"
+    try:
+        if not cursor:
+            cursor = result.docs[0].get("cursor")
+        # results returned in batches, get len of each batch
+        rowcount = sum([len(documents) for batch_key, documents in cursor.items() if search_key in batch_key])
+        span.set_metric(db.ROWCOUNT, rowcount)
+        return span
+
+    except Exception:
```
emmettbutler, line 324: Being more specific here about what type of exception is expected would make this code easier to maintain and learn.

emmettbutler (changes requested):

ddtrace/contrib/pymongo/client.py
wconti27: Changed this function and instead I just always pass the cursor object to the function. The try except was being used for my debugging purposes and I've also removed that.

emmettbutler (approved):

ddtrace/contrib/pymongo/client.py
```diff
@@ -133,6 +137,9 @@ def run_operation(self, sock_info, operation, *args, **kwargs):
                 result = self.__wrapped__.run_operation(sock_info, operation, *args, **kwargs)
                 if result and result.address:
                     set_address_tags(span, result.address)
+                if result and self._is_query(operation):
+                    cursor = result.docs[0].get("cursor")
```
mabdinur, line 141: Can the cursor be None? Can the docs array be empty? If so this can raise an exception: https://github.com/mongodb/mongo-python-driver/blob/3.5.1/pymongo/helpers.py#L142

ddtrace/contrib/pymongo/client.py
```diff
@@ -211,7 +233,7 @@ def write_command(self, *args, **kwargs):
         with self.__trace(cmd) as s:
             result = self.__wrapped__.write_command(*args, **kwargs)
             if result:
-                s.set_metric(mongox.ROWS, result.get("n", -1))
+                s.set_metric(db.ROWCOUNT, result.get("n", -1))
```
mabdinur, line 236: Is mongox.ROWS used anywhere else? If not, we should deprecate the constant

ddtrace/contrib/pymongo/client.py
```diff
@@ -285,3 +307,11 @@ def _set_query_metadata(span, cmd):
         span.resource = "{} {} {}".format(cmd.name, cmd.coll, q)
     else:
         span.resource = "{} {}".format(cmd.name, cmd.coll)
+
+
+def set_query_rowcount(span, cursor):
+    search_key = "Batch"
```
mabdinur, line 313: search_key should be a constant

ddtrace/contrib/pymongo/client.py
```diff
@@ -285,3 +307,11 @@ def _set_query_metadata(span, cmd):
...
+    # results returned in batches, get len of each batch
+    rowcount = sum([len(documents) for batch_key, documents in cursor.items() if search_key in batch_key])
```
mabdinur, line 315: Iterating over the cursor will pop data entries from the cursor object: https://github.com/mongodb/mongo-python-driver/blob/3.5.1/pymongo/cursor.py#L1138, this will consume the cursor iterable and the tracer will return an empty cursor object to users. We should avoid adding this side effect (I am not 100% sure about this, we can confirm this hypothesis by adding a test case). If this is actually the case we can avoid this issue by using `cursor.clone().items()`.

emmettbutler (approved):

ddtrace/contrib/pymongo/client.py
```diff
@@ -159,6 +169,18 @@ def send_message_with_response(self, operation, *args, **kwargs):
...
+                    if VERSION < (3, 2, 0):
+                        docs = _unpack_response(response=result.data)
+                        span.set_metric(db.ROWCOUNT, docs["number_returned"])
+                    elif (3, 2, 0) <= VERSION < (3, 6, 0):
+                        docs = _unpack_response(response=result.data)
+                        cursor = docs["data"][0]["cursor"]
+                        set_query_rowcount(cursor=cursor, span=span)
+                    else:
+                        docs = result.data.unpack_response()
+                        cursor = docs[0]["cursor"]
```
mabdinur, line 182: can we use docs[0]["number_returned"] here instead of the cursor? Also can docs be empty?

wconti27: I don't believe `number_returned` is defined for >3.6. Ill look into this

ddtrace/contrib/pymongo/client.py
```diff
@@ -146,6 +153,9 @@ def run_operation_with_response(self, sock_info, operation, *args, **kwargs):
                 result = self.__wrapped__.run_operation_with_response(sock_info, operation, *args, **kwargs)
                 if result and result.address:
                     set_address_tags(span, result.address)
+                if result and self._is_query(operation):
+                    cursor = result.docs[0].get("cursor")
+                    set_query_rowcount(cursor=cursor, span=span)
```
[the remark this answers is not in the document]
wconti27: shoud be fine

ddtrace/contrib/pymongo/client.py
wconti27: nope, added deprecation

ddtrace/contrib/pymongo/client.py
wconti27: to add to this, some of the tests already iterate through the whole cursor, i've added the same iteration to the added rowcount test so ensure the cursor is still alive and unaffected.

ddtrace/contrib/pymongo/client.py
wconti27: Just added checks for all of this, if anything is None or not as expected we abort and don't set tag

ddtrace/contrib/pymongo/client.py
wconti27: I thought this as well but haven't noticed of these side affects, atleast within Python (PHP cursors behaved as expected and iterating resulted in missing documents). For most of this Pymongo client code, the "cursor" object stored within the result isn't an actual cursor, just a plain dictionary.

wconti27: Also, we already have tests that iterate through our result cursor and check that the length is as expected.

ddtrace/contrib/pymongo/client.py
```diff
@@ -159,6 +169,18 @@ def send_message_with_response(self, operation, *args, **kwargs):
...
                 if result and result.address:
                     set_address_tags(span, result.address)
+                if result and self._is_query(operation):
+                    if VERSION < (3, 2, 0):
+                        docs = _unpack_response(response=result.data)
+                        span.set_metric(db.ROWCOUNT, docs["number_returned"])
+                    elif (3, 2, 0) <= VERSION < (3, 6, 0):
+                        docs = _unpack_response(response=result.data)
+                        cursor = docs["data"][0]["cursor"]
+                        set_query_rowcount(cursor=cursor, span=span)
```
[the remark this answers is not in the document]
wconti27: Upon further debugging, the number_returned attribute is for the number of returned batches for the cursor, and therefore can't be used for row_count.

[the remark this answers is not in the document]
wconti27: Leaving as is, but added catches for any exceptions so if anything is off, the tag isn't set and code execution continues as normal

emmettbutler (approved):

PR #5061, opened 2023-02-07
chore(tracing): add `db.system` tag to redis integrations

jbertran:
## Description

Part of the effort to unify tag names across integrations and languages involve setting the `db.system` tag when tracing DB and KV store integrations.

This PR sets the tag on all redis integrations:
* `aioredis`
* `aredis`
* `redis`
* `rediscluster`
* `yaaredis`

All integration tests that check tags now _also_ check for the `db.system` tag. Those that did not (because the integration tested inherits its behaviour from another) have no test changes.

All relevant snapshots are updated with the addition of the new expected tag.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/contrib/aioredis/patch.py
```diff
@@ -12,13 +12,17 @@
...
 from ...ext import SpanTypes
+from ...ext import db
 from ...ext import net
 from ...ext import redis as redisx
 from ...internal.utils.formats import stringify_cache_args
 from ..redis.util import _trace_redis_cmd
 from ..redis.util import _trace_redis_execute_pipeline
 
 
+DBMS_NAME = "redis"
```
mabdinur, line 23: We can move this constant to ddtrace/ext/redis.py or just use `ddtrace.ext.redis.APP`  (I don't think this constant is even used anymore but it has the value "redis")

releasenotes/notes/add-db-system-tag-redis-4a67baedfa7c2ade.yaml
```diff
@@ -0,0 +1,5 @@
+---
+features:
+  - |
+    tracing: The "db.system" tag is reported with value "redis" by redis integrations: aioredis, aredis,
+    redis, rediscluster, yaaredis.
```
mabdinur, line 5: Should we even add a release note? Would users want/need to know about this tag?

```suggestion
    tracing: Adds ``db.system`` span tag to the following integrations: aioredis, aredis,
    redis, rediscluster, yaaredis.
```

jbertran: I wondered as well. My reasoning was that since I’m splitting PRs to make diffs manageable, I want to be able to track which integrations report the new tag.

ddtrace/contrib/aioredis/patch.py
jbertran: I completely missed that this module was there, I’ll use that value.

releasenotes/notes/add-db-system-tag-redis-4a67baedfa7c2ade.yaml
jbertran: Then again, this means I'll create a lot of noise in release notes since I'm splitting in 5 PRs or more... The tag is useful, not crucial, so I'm happy with no release note.

gnufede (dismissed):
LGTM

emmettbutler (approved):

ddtrace/contrib/aioredis/patch.py
```diff
@@ -120,6 +121,7 @@ def traced_13_execute_command(func, instance, args, kwargs):
         child_of=parent,
     )
     span.set_tag_str(COMPONENT, config.aioredis.integration_name)
+    span.set_tag(db.SYSTEM, redisx.APP)
```
Yun-Kim, line 124: If `redisx.APP` is a text type, can we replace this with the more efficient `span.set_tag_str` method?
```suggestion
    span.set_tag_str(db.SYSTEM, redisx.APP)
```

Yun-Kim (changes requested):
One request but otherwise looks good!

releasenotes/notes/add-db-system-tag-redis-4a67baedfa7c2ade.yaml
Yun-Kim: This is a user-facing change, so users may be confused without a release note if all of a sudden they see new `db.system` tags in their traces. We can alternatively have one release note that encompasses all of these split PRs 😄

jbertran:
>     * Add a release note as this will add a new tag to redis integrations by default (unless I'm understanding wrong)

As we discussed offline, I'll make a release note PR after all the integrations are done to avoid being too noisy.

Yun-Kim (dismissed):

gnufede (approved):

mabdinur (approved):

PR #5077, opened 2023-02-08
chore(asm): enable ast patching at module import

gnufede:
## Motivation

The IAST logic works by patching the code at AST level during import time, in order to swap string manipulation functions to aspects that will also do taint tracking.

In order to do so, we need to apply a different logic at our import mechanism when executing a module, so the file imported can be patched at AST level. 

These changes, along with the imports required will only be performed if the ``DD_IAST_ENABLED`` env var is present, the added tests check that part.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/appsec/iast/_util.py
```diff
@@ -1,18 +1,18 @@
 #!/usr/bin/env python3
-from functools import cache
 import sys
 
 from ddtrace.internal.logger import get_logger
 from ddtrace.settings import _config as config
+from ddtrace.vendor.psutil._compat import lru_cache
```
P403n1x87, line 6: We have the `callonce` helper in our internal `cache` module that might be better suited for the job here

ddtrace/internal/module.py
```diff
@@ -189,7 +197,18 @@ def _create_module(self, spec):
         return self.loader.create_module(spec)
 
     def _exec_module(self, module):
-        self.loader.exec_module(module)
+        patched_source = None
+        if IS_IAST_ENABLED and _should_iast_patch(module.__name__):
```
P403n1x87, line 201: I'm reluctant to have product-specific logic into a generic module. Can we think of a different approach to this? I think this is (perhaps just partly) tackled in #4294

gnufede: Well, since this is called at import time, and we need to do the `exec_module` differently, I found this to be the most simple and explicit way to do it. I'm open to suggestions to do it differently, but think we'll end up patching all the customer's code and then some third-party modules as well.

avara1986 (dismissed):

P403n1x87 (changes requested):
We cannot import product-specific modules in `ddtrace.internal.module` as this can become a potential source of circular imports. A different approach is required here. Either an alternative to the AST patching, if possible, or a new extended `ModuleWatchdog` API.

ddtrace/internal/module.py
juanjux: @P403n1x87 so you want for the IAST part to do a different `_ImportHookChainedLoader` on a different module? I think this would be harder to follow and debug in case of problems with imports. Having the patching code here is IMO easier to follow and being under an "if IS_IAST_ENABLED" (like all IAST code will be) I don't see the problem.

juanjux:
> We cannot import product-specific modules in `ddtrace.internal.module` as this can become a potential source of circular imports. A different approach is required here. Either an alternative to the AST patching, if possible, or a new extended `ModuleWatchdog` API.

AST patching is not going away since this is the ASM-approved, tested and planned way to do for Python. 

I understand this can be unsettling but consider that everything under IAST related code will **always** be under a `if IAST` conditions, including this code, and for Python IAST is not even in private beta so users testing this will know there could be some rough edges but should never affect users not using it.

Circular imports are avoided because the function [should_iast_patch](https://github.com/DataDog/dd-trace-py/pull/5059/files#diff-7c9939e1d5c1efada2f31df3d25b9fc8048b6d4daab00524708f42f75020dbecR36) avoid patching any `ddtrace` module and uses an allow-list (for now very limit, even trough that could be configured in the future and automatically patch user code by default).

ddtrace/internal/module.py
P403n1x87: I can see a dedicated `Loader`/`Finder` in `sys.meta_path` as a much cleaner approach to this. To me this looks more like a hack inside `ModuleWatchdog`, whose purpose is to provide module import hooks. But like I said in my previous comment, I would be happy if this change proposed an extended API for the `ModuleWatchdog` hooks whereby we can accommodate for AST patching prior to execution. Then IAST could use that API instead in its internal area.

As for the IAST guard, I don't see how we could add a guard for the top-level import
~~~
from ddtrace.appsec.iast._util import _is_iast_enabled
~~~
which would then be the potential source of circular imports in the future.

I'm still not entirely convinced the approach proposed here is the best we can go for.

emmettbutler: > I can see a dedicated Loader/Finder in sys.meta_path as a much cleaner approach to this.

I second this. This could be a good opportunity to expand the interface of the import hook mechanism in such a way that it could be usable by other products in the future.

> Having the patching code here is IMO easier to follow...I don't see the problem

The clearest reason that I see to avoid putting product-specific code in `internal` is the increased cognitive load that it would place on future editors of `internal/module.py`. When changing this file, our job is made easier by the fact that we don't have to think about specific products as long as certain guarantees are maintained. With this change, people wanting to change this file would have to understand how IAST fits into that picture in greater detail than they would otherwise. For an internal module that's depended on by multiple products, I think it's worthwhile to be very careful about increasing cognitive load.

emmettbutler (changes requested):

ddtrace/appsec/iast/_util.py
```diff
@@ -1,18 +1,15 @@
 #!/usr/bin/env python3
-from functools import cache
```
gnufede, line 2: NB: I removed these because they log some deprecation warnings and that produces some test failures as the tests expect stdout and stderr to be empty.

ddtrace/internal/module.py
gnufede: I've changed the approach to remove the product import from the internal module, but unsure if the new approach is the right one. Resolving this conversation for now

ddtrace/appsec/iast/_util.py
gnufede: I tried our `callonce` from our internal `cache` but I removed it because it logs some deprecation warnings and that produces some test failures as the tests expect stdout and stderr to be empty.

ddtrace/internal/module.py
```diff
@@ -145,9 +146,10 @@ def find_loader(fullname):
 
 
 class _ImportHookChainedLoader(Loader):
-    def __init__(self, loader):
-        # type: (Loader) -> None
+    def __init__(self, loader, module_spec=None):
+        # type: (Loader, Optional[ModuleSpec]) -> None
         self.loader = loader
+        self.module_spec = module_spec
```
emmettbutler, line 152: Why is this added? Does it do something magical to `Loader` objects based on an implementation detail?

emmettbutler (approved):
I think this approach is more maintainable than the previous. Nice work!

ddtrace/internal/module.py
gnufede: You're right, I was misusing the `Loader` object there, the `Finder` is what I actually needed

ddtrace/appsec/iast/loader.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+IS_IAST_ENABLED = _is_iast_enabled()
+
+
+def _exec_iast_patched_module(module_watchdog, module):
+    patched_source = None
+    if IS_IAST_ENABLED:
+        log.debug("IAST enabled")
+        module_path, patched_source = astpatch_source(module.__name__, None)
```
P403n1x87, line 21: if we pass just the module, e.g.
```suggestion
        module_path, patched_source = astpatch_source(module)
```
then we might be able to use `ddtrace.internal.module.origin` to retrieve the module path and avoid the code for getting the spec and then the origin in `ast_patching.py`?

ddtrace/appsec/iast/_ast/ast_patching.py
```diff
@@ -58,30 +58,28 @@ def visit_ast(
...
         # Get the module path from the module name (foo.bar -> foo/bar.py)
-        loader = pkgutil.get_loader(module_name)
+        module_watchdog = ModuleWatchdog()
 
-        assert loader
-        if hasattr(loader, "module_spec"):
-            detected_module_path = loader.module_spec.origin
+        detected_module_spec = module_watchdog.find_spec(module_name)
+        if detected_module_spec and hasattr(detected_module_spec, "origin"):
+            detected_module_path = detected_module_spec.origin
```
P403n1x87, line 74: would the `origin` helper from `ddtrace.internal.module` work here

ddtrace/internal/module.py
```diff
@@ -189,7 +193,22 @@ def _create_module(self, spec):
...
+
+        for _ in sys.meta_path:
+            if isinstance(_, ModuleWatchdog):
+                for cond, hook in _._pre_exec_module_hooks:
+                    if isinstance(cond, str) and cond == module.__name__ or cond(module.__name__):
+                        pre_exec_hooks.append(hook)
+
+        if pre_exec_hooks:
+            for hook in pre_exec_hooks:
+                hook(self, module)
```
P403n1x87, line 208: maybe we can wrap this around `try...except` and raise an exception if we couldn't run the hook succesfully. Then we use this signal to call the default loader instead?

ddtrace/appsec/iast/loader.py
```diff
@@ -0,0 +1,32 @@
...
+
+    if patched_source:
+        # Patched source is executed instead of original module
+        compiled_code = compile(patched_source, module_path, "exec")
+        exec(compiled_code, module.__dict__)
+    else:
+        module_watchdog.loader.exec_module(module)
```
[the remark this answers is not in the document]
gnufede: Look at all the different branches that return ``"", ""`` here: https://github.com/DataDog/dd-trace-py/blob/1.x/ddtrace/appsec/iast/_ast/ast_patching.py#L59

ddtrace/appsec/iast/loader.py
gnufede: Thanks for the idea! I prefer keeping this signature because it's easier to test (it just needs the name and path of the modules, and not a Module object), and set up an intermediate function (``astpatch_module``) to retrieve the module origin as you suggest and pass it to this one

ddtrace/internal/module.py
gnufede: Hmmm... after giving it a thought... what if the exception is there when you execute a module? ImportError, RuntimeError, OSError... then we'd execute it twice. I believe the hook should capture its exceptions so the original ones are still propagated

ddtrace/appsec/iast/_ast/ast_patching.py
```diff
@@ -56,32 +58,35 @@ def visit_ast(
...
-        loader = pkgutil.get_loader(module_name)
+        module_watchdog = ModuleWatchdog._instance
 
-        assert loader
-        if hasattr(loader, "module_spec"):
-            detected_module_path = loader.module_spec.origin
+        assert module_watchdog
+        detected_module_spec = module_watchdog.find_spec(module_name)
+        if detected_module_spec and hasattr(detected_module_spec, "origin"):
+            detected_module_path = detected_module_spec.origin
```
P403n1x87, line 81: is this still required? Is `astpatch_source` called in other ways other than via the new `astpatch_module` helper, which is already passing in the module origin?

ddtrace/appsec/iast/loader.py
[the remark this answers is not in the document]
P403n1x87: Would it be possible to raise an exception here that would be handled in the `_exec_module` from the `ddtrace.internal.module` to avoid manually calling from the loader? Or could that be a potential issue if we want to pipe multiple transformers in the future? I'd really love if we could get rid of this line 🙂

ddtrace/appsec/iast/__init__.py
```diff
@@ -28,6 +28,8 @@ def wrapped_function(wrapped, instance, args, kwargs):
     )
     return wrapped(*args, **kwargs)
 """  # noqa: RST201, RST213, RST210
+# This import will trigger the change from regular loader to IAST loader if required
```
Kyle-Verhoog, line 31: Can we avoid doing import behaviour? Having logic run at import time is implicit and quite error prone. Could we instead have a function that is called at an explicit point in code?

ddtrace/appsec/iast/_ast/ast_patching.py
gnufede: Refactored and removed!

ddtrace/appsec/iast/loader.py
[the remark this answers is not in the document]
P403n1x87: In fact, thinking a bit more about this made me realise that the current design is flawed 🙁 . We cannot execute a module here, either with `exec` or with `loader.exec_module`. If every pre-exec hook we register does that, then we would be executing a module more than once, which is wrong by Python standards. So we either execute one hook, or we go back to the idea of having hook take the AST and return a modified AST that is then executed by `_exec_module` in `ddtrace.internal.module`.

P403n1x87 (changes requested):

ddtrace/appsec/iast/__init__.py
gnufede: Sure, I can move it to ``ddtrace.bootstrap.sitecustomize`` for example

ddtrace/appsec/iast/loader.py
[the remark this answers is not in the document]
gnufede: OK, just to unblock this PR, how about checking there is a single hook for the module, and tackle the chained transformations in a future PR? I propose this because we don't need to hurry with the chained transformations AFAIK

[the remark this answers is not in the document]
juanjux: Yup. To give more context, we don't expect to have more than one transformation for the IAST part, of course things could change in the future but for the first version I think limiting (and checking that only run) one hook per module should be fine.

ddtrace/internal/module.py
```diff
@@ -189,7 +193,25 @@ def _create_module(self, spec):
...
+                            # Several pre-exec hooks could match, we keep the first one
+                            pre_exec_hook = hook
+                            break
+                except Exception:
+                    log.debug("Exception happened while processing pre_exec_module_hooks", exc_info=True)
+
+        if pre_exec_hook:
+            pre_exec_hook(self, module)
+        else:
+            self.loader.exec_module(module)
```
P403n1x87, line 213:
```suggestion
        for _ in sys.meta_path:
            if isinstance(_, ModuleWatchdog) and not pre_exec_hook:
                try:
                    for cond, hook in _._pre_exec_module_hooks:
                        if isinstance(cond, str) and cond == module.__name__ or cond(module.__name__):
                            # Several pre-exec hooks could match, we keep the first one
                            hook(self, module)
                            return
                except Exception:
                    log.debug("Exception happened while processing pre_exec_module_hooks", exc_info=True)
                    break

        self.loader.exec_module(module)
```

ddtrace/internal/module.py
```diff
@@ -494,6 +517,22 @@ def unregister_module_hook(cls, module, hook):
         except ValueError:
             raise ValueError("Hook %r not registered for module %r" % (hook, module))
 
+    @classmethod
+    def register_pre_exec_module_hook(cls, cond, hook):
```
P403n1x87, line 521: Would we ever have the need to unregister these hooks? Perhaps if IAST is disabled at some point?

gnufede: That would only really make sense if IAST is enabled and disabled before the module is executed, because if it is enabled and executed, its code would be AST-patched. AFAIK, the planned way to "disable" IAST is to do early returns in the IAST swapped functions

ddtrace/internal/module.py
P403n1x87: ah right! I lost sight of the code below! 😢 then we can exit the outer loop early?
```suggestion
        pre_exec_hook = None

        for _ in sys.meta_path:
            if pre_exec_hook is not None:
                pre_exec_hoo(self, module)
                break
            if isinstance(_, ModuleWatchdog):
                try:
                    for cond, hook in _._pre_exec_module_hooks:
                        if isinstance(cond, str) and cond == module.__name__ or cond(module.__name__):
                            # Several pre-exec hooks could match, we keep the first one
                            pre_exec_hook = hook
                            break
                except Exception:
                    log.debug("Exception happened while processing pre_exec_module_hooks", exc_info=True)
        else:
            self.loader.exec_module(module)
```

gnufede: In this case we'll not execute ``self.loader.exec_module(module)`` (and we should) if there are hooks and none of them matches the ``cond`` condition.

I believe we can do:
```
        for _ in sys.meta_path:
            [...]
            if pre_exec_hook is not None:
                break
```
to exit the outer loop early, but then keep the
```
        if pre_exec_hook:
            pre_exec_hook(self, module)
        else:
            self.loader.exec_module(module)
```

P403n1x87 (approved):

avara1986 (approved):

juanjux (approved):

emmettbutler (approved):

christophe-papazian (approved):

PR #5088, opened 2023-02-09
fix(rcm): stuck child processes when application creates forks

avara1986:
# Description
Workaround (II) for remote config to work properly with gevent. We found that Remote Config create stuck/zombie child processes if an application with `gevent` uses `os.fork`

# What we do?
- Remove RemoteConfig.worker, because `worker` has `PeriodicService` as parent and `PeriodicService` use a worker too.
- Remove duplicates locks
- Add methods like `_stop_service`, `stop`... in the same way of `TelemetryWriter` and `SpanWriter`
- Remove "enable()" behavior. One of the previous fix (#4749) ensures the Remote config worker thread is launched at the very end, not interrupting the gevent patching mechanism. with forksafe.register(cls._restart) we were calling 2 times to remoteconfig.enable: one at fork time (and it raises the same error of the previous fix) and once again at register the products in the child process (this call is ok)

**Caveat:** this fix works with env variable `DD_GEVENT_PATCH_ALL=true`

## How to reproduce the error

- Create a flask application:
```
import os
import time

from flask import Flask

app = Flask(__name__)

def start():
    pid1 = os.fork()
    if pid1 == 0:
        os.setsid()
        x = 2
        while x > 0:
            time.sleep(0.2)
            print("child pid = ", pid1)
            x -= 1
    else:
        os.waitpid(pid1, 0)

@app.route('/')
def index():
    start()
    return "OK"
```
- start an agent
- add this environment variables:
```
DD_GEVENT_PATCH_ALL=true
DD_REMOTE_CONFIGURATION_ENABLED=true
```
- run the application with
```
 ddtrace-run gunicorn -w 3 -k gevent app:app
```
- Add traffic. e.g:
```
#!/bin/bash
while :
do
    curl -s -k 'GET'  http://127.0.0.1:8000/
done
```
- Check your processes, you should have 4-5 gunicorn workers, if you see more than this number, you have dead processes. The could remain even after stopping gunicorn.

```
ps -aux | grep gunicorn
alberto+  132031  1.4  0.1 115840 63280 pts/0    S+   16:33   0:00 /py-39/bin/gunicorn -c gunicorn_conf.py app:app
alberto+  132070  0.4  0.2 201700 65172 pts/0    Sl+  16:33   0:00 /py-39/bin/gunicorn -c gunicorn_conf.py app:app
alberto+  132071  0.3  0.2 201716 65192 pts/0    Sl+  16:33   0:00 /py-39/bin/gunicorn -c gunicorn_conf.py app:app
alberto+  132072  0.3  0.2 201712 65184 pts/0    Sl+  16:33   0:00 /py-39/bin/gunicorn -c gunicorn_conf.py app:app
```
Another way to check is run apachebenchmark (`ab`) like: 

```
ab  -n 1000000 -c 100 http://127.0.0.1:8001/
```

And running the script in another console. Eventually, the Flask app will stop responding (usually with 7 processes running on my machine).

You could check the gevent infinite lock with:
```
gdb -p 132072
```
or 
```
py-spy record --native  --rate 300 -o my-flamegraph-result.svg --pid 132072
```

## Checklist

## Reviewer Checklist

- [ ] Title is accurate.


ddtrace/internal/remoteconfig/__init__.py
```diff
@@ -1,22 +1,60 @@
...
+        )
+    return float(
+        os.getenv(
+            "DD_REMOTE_CONFIG_POLL_INTERVAL_SECONDS",
+            default=os.getenv("DD_REMOTECONFIG_POLL_SECONDS", default=DEFAULT_REMOTECONFIG_POLL_SECONDS),
+        )
+    )
+
+
+class RemoteConfigWriter(periodic.PeriodicService):
```
P403n1x87, line 37:
```suggestion
class RemoteConfigPoller(periodic.PeriodicService):
```

ddtrace/internal/remoteconfig/__init__.py
```diff
@@ -34,49 +72,56 @@ def _check_remote_config_enable_in_agent(cls):
...
-    def disable(cls):
-        # type: () -> None
-        with cls._worker_lock:
-            if cls._worker is not None:
-                cls._worker.stop()
-                cls._worker = None
 
-                forksafe.unregister(cls._restart)
-                atexit.unregister(cls.disable)
+remoteconfig_writer = RemoteConfigWriter()
```
P403n1x87, line 127:
```suggestion
remoteconfig_poller = RemoteConfigPoller()
```

ddtrace/internal/remoteconfig/client.py
```diff
@@ -245,6 +245,9 @@ def base64_to_struct(val, cls):
         self._last_error = None  # type: Optional[str]
         self._backend_state = None  # type: Optional[str]
 
+    def exists_products(self):
+        return len(self._products) > 0
```
P403n1x87, line 249:
```suggestion
    def has_products(self):
        return bool(self._products)
```

juanjux (approved):

ddtrace/internal/agent.py
```diff
@@ -155,6 +155,9 @@ def _healthcheck():
         _conn.request("GET", "info", {}, {"content-type": "application/json"})
         resp = _conn.getresponse()
         data = resp.read()
+    except Exception:
```
emmettbutler, line 158: A more specific exception handler here would help readers understand what to expect from `_healthcheck`

emmettbutler (changes requested):

emmettbutler:
Could the performance regressions from the bot comment above be related to changes in the default sleep time for the remoteconfig worker? We need to understand what's causing them, at least.

avara1986:
> Could the performance regressions from the bot comment above be related to changes in the default sleep time for the remoteconfig worker? We need to understand what's causing them, at least.

I think is flaky, most of the PRs have the same report

ddtrace/internal/agent.py
avara1986: :+1:

juanjux (dismissed):

ddtrace/internal/writer.py
```diff
@@ -532,7 +533,10 @@ def write(self, spans=None):
                     # are initialized
                     if asbool(os.getenv("DD_INSTRUMENTATION_TELEMETRY_ENABLED", True)):
                         telemetry_writer.enable()
-                    # appsec remote config should be enabled/started after the global tracer and configs
+                    # remote config should be enabled/started after the global tracer and configs are initialized
+                    if asbool(os.environ.get("DD_REMOTE_CONFIGURATION_ENABLED", "true")):
```
P403n1x87, line 537: I think this implies that DI without a tracer won't work, so this change has the potential of breaking DI.

avara1986: We need this approach to fix the gevent problem with RC, we update too Debugger part, isn't it enough to work with Debugger too? What do you propose?

majorgreys: @P403n1x87 Is this a blocker? Can we verify whether DI breaks with this change?

emmettbutler (approved):

emmettbutler:
I pulled out the regression test from this change and ran it against 1.x (you can see the diff in https://github.com/DataDog/dd-trace-py/pull/5130), and I found that the test passes. That means I can't really tell what issue this change is intended to solve. I'd appreciate it if anyone watching this pull request could tell me what I'm missing and how the test replicates an error.

juanjux (dismissed):

ddtrace/internal/writer.py
avara1986: @P403n1x87  Do you check this change? :point_right:  https://github.com/DataDog/dd-trace-py/pull/5088/files#diff-d14a7422eb9f2743a3db00be1dbc2245ca0706fc06c1120b5458516af389322fR251

All Debugger tests work fine with that change

P403n1x87: Ah no I missed that, but I think it unfortunately creates more problems than it solves:
- The way of enabling DI is with `DD_DYNAMIC_INSTRUMENTATION_ENABLED`; having to set another variable to enable DI is not the behaviour that we want to support.
- This might solve the issue for ASM, but if someone enables DI then we would have the same problem ASM is trying to fix.

gnufede (dismissed):

ddtrace/internal/writer.py
avara1986: You don't need to add another environment variable, `DD_REMOTE_CONFIGURATION_ENABLED` is True by default but we need this environment variable to run tests isolated.



> * This might solve the issue for ASM, but if someone enables DI then we would have the same problem ASM is trying to fix.

Bugs one by one :stuck_out_tongue: We fixed ASM/APM in this PR, we could merge this PR and #5105 fix this bug in DI, right? :+1:

emmettbutler (dismissed):

PR #5097, opened 2023-02-10
chore(ciapp): print stack trace for git errors only in debug

albertvaka:
Makes the error less scary by not printing the stack trace unless we are in debug mode.

## Testing strategy

Run `pytest` with the tracer (`DD_SERVICE=my-python-app`) in a project that is not in a git repository (ie: anywhere where `git status` would fail). The error should be printed but no stacktrace. Run again with `DD_CALL_BASIC_CONFIG=true DD_TRACE_DEBUG=true` to see the stacktrace.

This should not have any other side-effect.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/ext/ci.py
```diff
@@ -102,8 +102,9 @@ def tags(env=None, cwd=None):
         git_info[WORKSPACE_PATH] = git.extract_workspace_path(cwd=cwd)
     except git.GitNotFoundError:
         log.error("Git executable not found, cannot extract git metadata.")
-    except ValueError:
-        log.error("Error extracting git metadata, received non-zero return code.", exc_info=True)
+    except ValueError as e:
+        debug_mode = asbool(os.getenv("DD_TRACE_DEBUG", default=False))
```
brettlangdon, line 106:
```suggestion
        debug_mode = log.isEnabledFor(logging.DEBUG)
```

You'll need to add `import logging` above.

This will help catch it if someone configured debug logging not via `DD_TRACE_DEBUG`.

(It might also be faster than asbool + os.getenv, but I am not super worried about that in this code path.

albertvaka: Thanks, I've updated this. There are a couple places in the code were we still check for `DD_TRACE_DEBUG` directly, though, in case you want to take a look (you can grep for `os.environ.get("DD_TRACE_DEBUG"` and for `os.getenv("DD_TRACE_DEBUG"`)

emmettbutler (approved):

albertvaka:
@brettlangdon Added manual test strategy to the PR description

ddtrace/ext/git.py
```diff
@@ -152,7 +152,8 @@ def extract_git_metadata(cwd=None):
         log.error("Git executable not found, cannot extract git metadata.")
     except ValueError as e:
         debug_mode = log.isEnabledFor(logging.DEBUG)
-        log.error("Error extracting git metadata: %s", str(e), exc_info=debug_mode)
+        stderr = str(e)
```
mabdinur, line 155: ?
```suggestion
        stderr = compat.ensure_text(stderr).strip()
```

albertvaka: `ensure_text` only works for text, but e is of type `ValueError`. Casting a `ValueError` to string returns the error message passed to the exception when it was created. This works in python2 and 3. Note that the error message has already been passed through `ensure_text` earlier.

albertvaka:
```suggestion
        stderr = str(e) # Returns the message passed to the exception when it was created
```

albertvaka: We can add a comment to explain this, feel free to commit the above suggestion if you think it's worth it.

mabdinur (approved):

brettlangdon (approved):

brettlangdon (approved):

juanjux (approved):

PR #5099, opened 2023-02-10
fix(rcm): fix stuck processes on os.fork with gevent

juanjux:
## Description

Fix forked processes getting stuck when running under gevent.

## How

- Reduce the range of the locks in `RemoteConfig`. Since the `RemoteConfigWorker` `start` and `stop` methods already have locks, only protect the assignment to the class member.

- Remove call to `restart` on fork, the `AgentWriter` now takes care of it.

- Implement the `stop_service` and `shutdown` methods for `RemoteConfigWorker` which ensure proper cleanup without zombies. :zombie_man: :zombie_woman: 

- Add a regression test.

## Checklist

## Reviewer Checklist

- [ ] Title is accurate.


ddtrace/internal/writer.py
```diff
@@ -36,6 +37,7 @@
 from .encoding import JSONEncoderV2
 from .encoding import MSGPACK_ENCODERS
 from .logger import get_logger
+from .remoteconfig import RemoteConfig
```
gnufede, line 40: nit: duplicated import

tests/internal/remoteconfig/test_remoteconfig.py
```diff
@@ -124,7 +131,7 @@ def test_remote_config_forksafe():
 
         if os.fork() == 0:
             assert RemoteConfig._worker is not None
-            assert RemoteConfig._worker is not parent_worker
+            assert RemoteConfig._worker is parent_worker
```
emmettbutler, line 134: Does this mean that the new intended behavior is for both parent and child processes to share the same `RemoteConfig._worker`?

emmettbutler (approved):

emmettbutler (commented):
What do you think is the explanation for the performance regressions?

ddtrace/internal/writer.py
```diff
@@ -532,6 +533,10 @@ def write(self, spans=None):
                     # are initialized
                     if asbool(os.getenv("DD_INSTRUMENTATION_TELEMETRY_ENABLED", True)):
                         telemetry_writer.enable()
+
+                    if asbool(os.getenv("DD_REMOTE_CONFIGURATION_ENABLED", True)):
+                        RemoteConfig.enable()
```
Yun-Kim, line 538: Can you help me understand how the `AgentWriter` takes care of the call to restart on fork now?

Yun-Kim (commented):
Can we get a more detailed description of what the problem was on Gevent? It's not exactly clear why we need to do these things to fix forked processes getting stuck.

I also cannot tell after looking through the changes why we're seeing a significant bump in memory usage (rss in the benchmark scenarios), but it appears consistent enough across most of the benchmark scenarios. I'd like to get some discussion on this before we merge this.

juanjux:
Update: set it to draft for now. Going to investigate if I can find the problem with the forksafe module with gevent since that seems to be where the real problem lies (forksafe = fails, remove it = works).

P403n1x87:
The fix for this issue is essentially in #5105

Issue #3121, opened 2022-01-10
Intermittent `RuntimeError: the memalloc module was not started` error


cyface:
### Which version of dd-trace-py are you using?

ddtrace==0.57.0

### Which version of the libraries are you using?

You can copy/paste the output of `pip freeze` here.

arthurio:
We are facing the same issue with a Fastapi app. I'd be happy to share logs/config you need.

jd:
Could you share:
1. which Web server you're running?
2. Its configuration?
3. If you're using gevent?

cyface:
I am using nginx feeding into gunicon/gevent with regular proxy setup. Not sure which configurations would be helpful to you?

arthurio:
We have 2 different servers (not behind nginx).

**flask**
1. Gunicorn
2. `--worker-class=gthread`
3. No

**fastapi**
1. Gunicorn
2. `--worker-class=uvicorn.workers.UvicornWorker`
3. No

roniemartinez:
+1 We have the same issue with Gunicorn + Django

visaals:
+1 same in `0.59.1` we are using sanic with profiling enabled: `DD_PROFILING_ENABLED=true`

```
datadog~=0.44.0
ddtrace~=0.59.1
requests~=2.27.1
sanic~=19.12.2
```

<img width="1343" alt="image" src="https://user-images.githubusercontent.com/19598277/159076397-a20fc9dc-4f0b-4332-a4c0-0dba795b8b6b.png">

FlavioAlexander:
Currently, having the same issue

luislew:
Same issue, with Flask/gunicorn/gevent, using ddtrace 1.2.1:
```

RuntimeError: the memalloc module was not started
  File "ddtrace/internal/periodic.py", line 70, in run
    self._target()
  File "ddtrace/profiling/collector/__init__.py", line 42, in periodic
    for events in self.collect():
  File "ddtrace/profiling/collector/memalloc.py", line 145, in collect
    events, count, alloc_count = _memalloc.iter_events()
```

rgilkey:
Same issue with Flask/gunicorn/gevent as well using ddtrace 1.4.4:
<img width="766" alt="image" src="https://user-images.githubusercontent.com/3894011/201698121-ef82b8fb-f191-4a7b-9a8f-2841c37e57bf.png">

rgilkey:
[...]

PR #5105, opened 2023-02-10
fix(internal): remove nogevent compatibility layer

emmettbutler:
This pull request removes the `nogevent` compatibility layer from the library and tests. It also changes the behavior of some feature flags related to gevent and adjusts a whole lot of tests to work with these changes.

This change was manually tested on staging and shown to fix the issue we were investigating there. See noteboook 4782453 in the ddstaging datadoghq account for a detailed look at the metrics collected during that test.

## What Changed?

* `sitecustomize.py`: refactored module cloning logic; changed the default of `DD_UNLOAD_MODULES_FROM_SITECUSTOMIZE` to `auto` meaning it runs when gevent is installed; unloaded a few additional modules like `time` and `attrs` that were causing tests and manual checks to fail; deprecated `DD_GEVENT_PATCH_ALL` flag
* Removed various hacks and layers that had intended to fix gevent compatibility, including in `forksafe.py`, `periodic.py`, `ddtrace_gevent_check.py`,  and `nogevent.py`
* `profiling/`: adjusted all uses of the removed module `nogevent` to work with `threading`
* Adjusted tests to work with these removals

We tried to separate some of these changes into other pull requests, which you can see linked in the discussion below. Because of how difficult it is to replicate the issue we're chasing outside of the staging environment, we decided to minimize the number of variables under test and combine the various changes into a single pull request. This makes it a bit harder to review, which we've tried to mitigate with the checklist above.

## Risk

The main risk in this change is the change to the default behavior of module cloning. We've mitigated this risk with the automated test suite as well as manual testing described in the notebook above.

**Why doesn't this change put all new behavior behind a feature flag and leave the default behavior untouched?**

The main reason for this decision is pragmatic: it's really hard to test for the issue this solves, requiring a turnaround time of about an hour to get feedback from changes. The secondary reason is that the `nogevent` layer is highly coupled to the rest of the library's code, and putting it behind a feature flag is a significant and nontrivial effort. The third reason is that full support of all of the various configurations and combinations with other tools that gevent can be used in is a goal that we could probably spend infinite time on if we chose to. Given this, we need to intentionally set a goal that solves the current and likely near-future issues as completely as possible, make it the default behavior, and call this effort "done". @brettlangdon @P403n1x87 @Yun-Kim and I are in agreement that the evidence in noteboook 4782453 in the ddstaging datadoghq account is enough to justify this change to the default behavior.

## Performance Testing

Performance testing with a sample flask application (notebook 4442578) shows no immediately noticeable impact from tracing.

Dynamic instrumentation seems to cause slow-downs, and the reason has been tracked down to joining service threads on shutdown. Avoiding the joins cures the problem, but further testing is required to ensure that DI still behaves as intended. Note that the tracer is already implemented this way, and this is probably why we don't see it impacting process shutdown. 

Profiling also shows a slight slow-down in short-lived processes, when enabled, but this is an already existing issue. This seems to be due to retrieving the response from the agent after the payload has been uploaded. A potential solution to this might be offered by asynchronous processing with libdatadog.

The following are the details of the scenario used to measure the performance under different configurations.
The application is the simple Flask app of the issue reproducer:

```
# app.py

import os
import time

from ddtrace.internal.remoteconfig import RemoteConfig
from flask import Flask

app = Flask(__name__)


def start():
    pid1 = os.fork()
    if pid1 == 0:
        os.setsid()
        x = 2
        while x > 0:
            time.sleep(0.2)
            x -= 1
    else:
        os.waitpid(pid1, 0)


@app.route("/")
def index():
    start()
    return "OK" if RemoteConfig._worker is not None else "NOK"
```

We can control what products to start with the following run.sh script

```
#!/bin/bash

source .venv/bin/activate

export DD_DYNAMIC_INSTRUMENTATION_ENABLED=true
export DD_PROFILING_ENABLED=false
export DD_TRACE_ENABLED=false

export DD_ENV=gab-testing
export DD_SERVICE=flask-gevent

ddtrace-run gunicorn -w 3 -k gevent app:app

deactivate
```

To run the app we create a virtual environment with

```
python3.9 -m venv .venv
source .venv/bin/activate
pip install flask gevent gunicorn
pip install -e path/to/dd-trace-py
deactivate
```

and then invoke the script, adjusting the exported variables as required

```
chmod +x run.sh
./run.sh
```

In another terminal we can check the average response time by sending requests to the application while running. With the following simple k6 script

```
import http from 'k6/http';

export default function () {
  http.get('http://localhost:8000');
}
```

We invoke k6 with

```
k6 run -d 30s script.js
```

and look for this line in the output

```
http_req_duration..............: avg=335.68ms min=119.56ms med=418.76ms max=451.49ms p(90)
```

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


emmettbutler:
The big question with this change is whether or not it resolves [incident 18605](https://app.datadoghq.com/incidents/18605/remediation) when combined with module cloning.

tests/profiling/collector/test_memalloc.py
```diff
@@ -5,13 +5,14 @@
 
 import pytest
 
+from ddtrace.internal import compat
+
```
P403n1x87, line 9:
```suggestion
```

tests/profiling/collector/test_memalloc.py
```diff
@@ -132,12 +132,12 @@ def test_iter_events_multi_thread():
         assert size >= 1  # size depends on the object size
         if last_call[2] == "<listcomp>" and last_call[1] == _ALLOC_LINE_NUMBER:
             assert last_call[0] == __file__
-            if thread_id == nogevent.main_thread_id:
+            if thread_id == compat.main_thread.ident:
```
P403n1x87, line 135:
```suggestion
            if thread_id == threading.main_thread().ident:
```

tests/profiling/collector/test_memalloc.py
```diff
@@ -132,12 +132,12 @@ def test_iter_events_multi_thread():
...
                 count_object += 1
                 assert stack[1][0] == __file__
                 assert stack[1][1] == _ALLOC_LINE_NUMBER
                 assert stack[1][2] == "_allocate_1k"
-            elif thread_id == t.ident:
+            if thread_id == t.ident:
```
P403n1x87, line 140:
```suggestion
            elif thread_id == t.ident:
```

P403n1x87:
In the riotfile we need to make sure we are running the `gunicorn` test suite with any versions of Python that we support (with the possible exception of 3.11).

tests/internal/test_forksafe.py
```diff
@@ -288,61 +288,91 @@ def fn():
...
-        err=None,
-        env=dict(_DD_TRACE_GEVENT_HUB_PATCHED="true"),
-    )
-    def test_gevent_reinit_patch():
-        import os
-        import sys
+@pytest.mark.subprocess(
+    out=("CTCTCT" if sys.platform == "darwin" or (3,) < sys.version_info < (3, 7) else "CCCTTT"), err=None
+)
+def test_gevent_gunicorn_behaviour():
```
P403n1x87, line 294: We probably get the best value out of this test if we run it under ddtrace-run and let the library do the module clean-up by setting `DD_UNLOAD_MODULES_FROM_SITECUSTOMIZE`

Yun-Kim:
@emmettbutler can we relax the gunicorn test suite in `riotfile.py` to include all Python versions so we can see the results of this change on CI?

emmettbutler:
Absolutely @Yun-Kim. Would you like me to do that?

Yun-Kim:
@emmettbutler I can add it 👍

P403n1x87:
There is something going on with the current `sitecustomize.py`. If I use the one from https://github.com/P403n1x87/dd-trace-py/blob/2bb028b35010847893a764a8a641d79b00ea1454/ddtrace/bootstrap/sitecustomize.py I get the tests to behave as expected on legacy versions of CPython (e.g. 3.6). I'm currently trying to understand why that's the case. So far the issue seems to be with `gevent.events` importing `pkg_resources` after fork, which causes a deadlock for reasons. This  does not happen with the linked `sitecustomize.py`.

releasenotes/notes/nogevent-6f2892cb412f987f.yaml
```diff
@@ -0,0 +1,9 @@
+---
+deprecations:
+  - |
+    internal: `DD_GEVENT_PATCH_ALL` is deprecated and will be removed in the next major verison. Gevent compatibility is now automatic
```
P403n1x87, line 4: should we add a deprecation warning if we detect that the variable is still set to let the users know the variable is no longer needed? @brettlangdon WDYT?

ddtrace/bootstrap/sitecustomize.py
```diff
@@ -142,6 +51,13 @@ def cleanup_loaded_modules(aggressive=False):
 
 log = get_logger(__name__)
 
+if os.environ.get("DD_GEVENT_PATCH_ALL", "false").lower() in ("true", "1"):
```
P403n1x87, line 54: 👍  I think we can just check that `DD_GEVENT_PATCH_ALL` is in `os.environ` as we only care about whether it is set (to whatever) or not. Should we also include a test to check that we get the warning?

riotfile.py
```diff
@@ -2604,9 +2604,13 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
             pkgs={"requests": latest, "gevent": latest},
             venvs=[
                 Venv(
-                    pys=select_pys(min_version="3.8"),
+                    pys=select_pys(min_version="3.5"),
```
P403n1x87, line 2607: we should really aim at including 2.7 here if we can

emmettbutler:
Here's the error from the subprocess in one of the failing `profile` tests:
```
[2023-02-21 17:15:31 +0000] [21390] [ERROR] Exception in worker process                                                                                                                                                                                              
  File "/root/project/.tox/py27-profile-gevent/lib/python2.7/site-packages/gunicorn/arbiter.py", line 586, in spawn_worker                                                                                                                                                                                                  
    worker.init_process()                                                                                                                                                                                                                                                                                                   
  File "/root/project/.tox/py27-profile-gevent/lib/python2.7/site-packages/gunicorn/workers/ggevent.py", line 196, in init_process                                                                                                                                                                                          
    self.patch()                                                                                                                                                                                                                                                                                                            
  File "/root/project/.tox/py27-profile-gevent/lib/python2.7/site-packages/gunicorn/workers/ggevent.py", line 65, in patch                                                                                                                                                                                                  
    monkey.patch_all(subprocess=True)                                                                                                                                                                                                                                                                                       
  File "/root/project/.tox/py27-profile-gevent/lib/python2.7/site-packages/gevent/monkey.py", line 1263, in patch_all                                                                                                                                                                                                       
    patch_thread(Event=Event, _warnings=_warnings)                                                                                                            
  File "/root/project/.tox/py27-profile-gevent/lib/python2.7/site-packages/gevent/monkey.py", line 200, in ignores                                            
    return func(*args, **kwargs)                                                                                                                                                                                                                                                                                            
  File "/root/project/.tox/py27-profile-gevent/lib/python2.7/site-packages/gevent/monkey.py", line 779, in patch_thread                                                                                                                                                                                                     
    _notify_did_subscribers=False)                                                                                                                                                                                                                                                                                          
  File "/root/project/.tox/py27-profile-gevent/lib/python2.7/site-packages/gevent/monkey.py", line 468, in _patch_module                                                                                                                                                                                                    
    _call_hooks=_call_hooks)                                                                                                                                                                                                                                                                                                
  File "/root/project/.tox/py27-profile-gevent/lib/python2.7/site-packages/gevent/monkey.py", line 404, in patch_module                                       
    __call_module_hook(source_module, 'will', target_module, items, _warnings)                                                                                                                                                                                                                                              
  File "/root/project/.tox/py27-profile-gevent/lib/python2.7/site-packages/gevent/monkey.py", line 331, in __call_module_hook                                                                                                                                                                                               
    func(module, items, warn)                                                                                                                                                                                                                                                                                               
  File "/root/project/.tox/py27-profile-gevent/lib/python2.7/site-packages/gevent/threading.py", line 230, in _gevent_will_monkey_patch                                                                                                                                                                                     
    main_thread = main_native_thread()                                                                                                                                                                                                                                                                                      
  File "/root/project/.tox/py27-profile-gevent/lib/python2.7/site-packages/gevent/threading.py", line 167, in main_native_thread                                                                                                                                                                                            
    assert len(main_threads) == 1, "Too many main threads"                                                                                                                                                                                                                                                                  AssertionError: Too many main threads                                                                                                                                                                                                                                                                                       [2023-02-21 17:15:31 +0000] [21390] [INFO] Worker exiting (pid: 21390)                                                                                                                                                                                                                                                      /root/project/.tox/py27-profile-gevent/lib/python2.7/site-packages/google/protobuf/descriptor.py:167: RuntimeWarning: Parent module 'google.protobuf' not found while handling absolute import                                                                                                                                from google.protobuf import descriptor_pb2                                                                                                                                                                                                                                                                                [2023-02-21 17:15:32 +0000] [21371] [INFO] Shutting down: Master                                                                                                                                                                                                                                                            
[2023-02-21 17:15:32 +0000] [21371] [INFO] Reason: Worker failed to boot.                                                                                                                                                                                                                                                   
Exception KeyError: KeyError(281473538485024,) in <module 'threading' from '/root/.pyenv/versions/2.7.18/lib/python2.7/threading.pyc'> ignored                                                                                                                                                                              
/root/project/.tox/py27-profile-gevent/lib/python2.7/site-packages/google/protobuf/descriptor.py:167: RuntimeWarning: Parent module 'google.protobuf' not found while handling absolute import                                                                                                                              
  from google.protobuf import descriptor_pb2                                                       
```
I've locally tried depending on older versions of gevent and gunicorn but the issue persists. @P403n1x87 do you have any ideas about this?

P403n1x87:
> I've locally tried depending on older versions of gevent and gunicorn but the issue persists. @P403n1x87 do you have any ideas about this?

I can't tell whether the issue here is `google.protobuf` or the `KeyError`. One thing to try perhaps is adding `google` to the modules to keep for Python 2

tests/profiling/collector/test_stack.py
```diff
@@ -179,36 +172,65 @@ def _fib(n):
 
 
 @pytest.mark.skipif(not TESTING_GEVENT, reason="Not testing gevent")
-def test_collect_gevent_thread_task():
-    r = recorder.Recorder()
-    s = stack.StackCollector(r)
+def test_collect_gevent_thread_task(ddtrace_run_python_code_in_subprocess):
+    _, stderr, status, _ = ddtrace_run_python_code_in_subprocess(
```
P403n1x87, line 176: we can use the `pytest.mark.subprocess(ddtrace_run=True)` marker here I believe

brettlangdon:
Is this still a draft?

emmettbutler:
@brettlangdon yes. once the current test run finishes i'll promote it to mergeable.

tests/profiling/collector/test_stack.py
```diff
@@ -246,32 +262,6 @@ def collect(self):
         return []
 
 
-@pytest.mark.skipif(not TESTING_GEVENT, reason="Not testing gevent")
-@pytest.mark.parametrize("ignore", (True, False))
-def test_ignore_profiler_gevent_task(monkeypatch, ignore):
```
Yun-Kim, line 251: @P403n1x87 why was this test removed? I'm not sure where in this test gevent actually is introduced so I'm confused about the context of this test...

ddtrace/contrib/gevent/__init__.py
```diff
@@ -5,11 +5,8 @@
 The integration patches the gevent internals to add context management logic.
 
 .. note::
-    If :ref:`ddtrace-run<ddtracerun>` is being used set ``DD_GEVENT_PATCH_ALL=true`` and
-    ``gevent.monkey.patch_all()`` will be called as early as possible in the application
-    to avoid patching conflicts.
-    If ``ddtrace-run`` is not being used then be sure to call ``gevent.monkey.patch_all``
-    before importing ``ddtrace`` and calling ``ddtrace.patch`` or ``ddtrace.patch_all``.
+    If ``ddtrace-run`` is not being used then be sure to ``import ddtrace.bootstrap.sitecustomize``
```
brettlangdon, line 8: this is interesting, since the normal configuration we try to push people towards is `from ddtrace import patch_all; patch_all()`.

do these changes change that requirement and instead require that people import the sitecustomize module? (if yes, then I don't think I love the name very well, maybe we can create `ddtrace/auto.py` which is just `import ddtrace.bootstrap.sitecustomize`, this would match `import ddtrace.profiling.auto`)

ddtrace/contrib/gunicorn/__init__.py
```diff
@@ -1,36 +1,3 @@
-"""
-**Note:** ``ddtrace-run`` and Python 2 are both not supported with `Gunicorn <https://gunicorn.org>`__.
```
brettlangdon, line 2: Do we still need any instructions regarding if/how to use `ddtrace-run` or manually including `ddtrace.bootstrap.sitecustomize` ?

Is including `import ddtrace.bootstrap.sitecustomize in your `app.py` good enough for this to work? Would that be after `gunicorn` imports and patches `gevent` ?

ddtrace/contrib/gevent/__init__.py
brettlangdon: should we add a check to `ddtrace.bootstrap.sitecustomize` or `ddtrace.__init__` to check if `gevent` has been loaded before `ddtrace` and log a warning that you want to `import ddtrace` first? (or, should we have a check that the module cleanup has happened / we properly have the cloned modules we need?)

ddtrace/profiling/collector/_task.pyx
```diff
@@ -1,25 +1,35 @@
...
-    from greenlet import getcurrent
-    from greenlet import greenlet
-    from greenlet import settrace
-except ImportError:
-    _gevent_tracer = None
-else:
+_gevent_tracer = None
+
+
+@when_imported("gevent")
```
brettlangdon, line 14: Should we be using `ModuleWatchdog` for this?

I'd have to look to remind myself, but iirc `@when_imported` doesn't always work.

cc @P403n1x87

ddtrace/profiling/collector/memalloc.py
```diff
@@ -148,7 +148,13 @@ def snapshot(self):
         )
 
     def collect(self):
-        events, count, alloc_count = _memalloc.iter_events()
+        try:
+            events, count, alloc_count = _memalloc.iter_events()
```
brettlangdon, line 152: is this a related change?

ddtrace/profiling/collector/stack.pyx
```diff
@@ -296,14 +292,12 @@ cdef collect_threads(thread_id_ignore_list, thread_time, thread_span_links) with
...
 cdef stack_collect(ignore_profiler, thread_time, max_nframes, interval, wall_time, thread_span_links, collect_endpoint):
-
-    if ignore_profiler:
-        # Do not use `threading.enumerate` to not mess with locking (gevent!)
-        thread_id_ignore_list = {thread_id
-                                 for thread_id, thread in threading._active.items()
-                                 if getattr(thread, "_ddtrace_profiling_ignore", False)}
-    else:
-        thread_id_ignore_list = set()
+    # Do not use `threading.enumerate` to not mess with locking (gevent!)
```
brettlangdon, line 295: is this comment still valid? if we have our own copy of the threading module then it shouldn't matter?

ddtrace/profiling/recorder.py
```diff
@@ -1,11 +1,11 @@
 # -*- encoding: utf-8 -*-
 import collections
+import threading
```
brettlangdon, line 3: `as ddtrace_threading` ?

ddtrace_gevent_check.py
```diff
@@ -1,13 +0,0 @@
-import sys
-import warnings
-
-
-def gevent_patch_all(event):
-    if "ddtrace" in sys.modules:
-        warnings.warn(
-            "Loading ddtrace before using gevent monkey patching is not supported "
```
brettlangdon, line 8: do we need a warning for the opposite of this?

brettlangdon (commented):
I didn't check any test files. My main feedback is I feel like this is made to work well with `ddtrace-run` + `ddtrace.bootstrap.sitecustomize`. I'd like us to consider if/how we can make it work with `from ddtrace import patch_all; patch_all()` since there are a lot of users who are going to be doing that approach; if we can help them as well, let's do it.

ddtrace/contrib/gevent/__init__.py
P403n1x87: My proposal would be to take the ddtrace-specific logic into `ddtrace.bootstrap.preload` and leave the boilerplate code in `sitecustomize.py`. Then users can just do
~~~
import ddtrace.bootstrap.preload  # noqa
~~~
Or we could have `ddtrace.bootstrap._preload` define the `preload` function that can be called by the boilerplate code in `sitecustomize.py`, and have users do
~~~
from ddtrace.bootstrap import preload

preload()
~~~

ddtrace/profiling/collector/_task.pyx
P403n1x87: We could, but this seems to work just fine for now. Perhaps we can refactor if we push for dropping the vendored `wrapt` in favour of our `internal` stuff?

ddtrace/profiling/collector/memalloc.py
P403n1x87: We could take this out into its own PR. This prevents a race between the periodic memory collector service and the `_memalloc.stop` clean-up on shutdown from being noisy. This is not related to just gevent and might actually fix #3121.

ddtrace/profiling/collector/stack.pyx
P403n1x87: This sounds about right, but I haven't tested it. I'd be happy to make this a follow-up task. WDYT?

ddtrace_gevent_check.py
P403n1x87: It would be nice to have but we would need a way to find out if gevent-s `patch_all` was called without causing side-effects.

ddtrace/profiling/recorder.py
P403n1x87: I don't think it is necessary here. The only place where it would be required is where we have the need to look up the actual `threading` module used by the application from `sys.modules`, which we would assign to the variable `threading`.

ddtrace/contrib/gevent/__init__.py
P403n1x87: Example here: https://github.com/DataDog/wilma/tree/main/wilma/_bootstrap

brettlangdon: but why should users have to import/call preload vs just importing the library?

is there not behavior there we want to occur for all usages of ddtrace? (specifically module cloning, there are probably some ddtrace-run specifics we don't want to happen on import of `ddtrace`)

ddtrace/profiling/collector/_task.pyx
brettlangdon: what is the impact is this doesn't work properly? we'd be missing gevent task associations with profiles?

ddtrace/profiling/recorder.py
brettlangdon: I am just wondering for consistency, being clear about which/when we have a cloned version of `threading` vs the non-cloned one.

ddtrace_gevent_check.py
brettlangdon: 🤔 yeah... probably would need to check if gevent is already loaded in `sys.modules`, and if it is, we can check if patching has been called yet, all before calling module cloning ?

But that doesn't help the case when just doing `import ddtrace`

ddtrace/contrib/gevent/__init__.py
P403n1x87: I suppose we _could_ ensure we perform the necessary steps on import, but then we would be relying on import side-effects. My feeling is that in general we tend to avoid those as much as possible.

ddtrace/profiling/collector/_task.pyx
P403n1x87: Yes I think we would miss greenlets and won't collect their stacks

ddtrace/profiling/collector/memalloc.py
P403n1x87: https://github.com/DataDog/dd-trace-py/pull/5182

ddtrace/profiling/collector/_task.pyx
brettlangdon: ok, I'd say the risk then is fairly low. possibly bad product experience when using gevent, but no ill effects to the application it self 👍🏻

ddtrace/contrib/gunicorn/__init__.py
P403n1x87: Ideally we'd want to run the sitecustomize logic as soon as possible if gevent is being used, so the best thing would be to use `ddtrace-run`. I'm not sure what would happen if we are squeezed in between gunicorn and gevent. Would there be anything preventing a user from using `ddtrace-run` after this change? Note that they could still import from `ddtrace` and perform any configuration they like afterwards.

ddtrace/contrib/gevent/__init__.py
brettlangdon: This PR should probably include new documentation about exactly how to use `ddtrace` with `gevent`. Including every step needed to run with `ddtrace-run` or manual patching with any requirements.

I feel like the one thing in this PR is "when using ddtrace-run it just works™️", but I want to make sure we have proper documentation/required steps for other cases as well since not everyone can/will use `ddtrace-run`.

ddtrace/contrib/gunicorn/__init__.py
brettlangdon: Some users cannot modify the command that is executed to run their app, so they **must** use `from ddtrace import patch_all; patch_all()` for loading/configuring the tracer.

ddtrace/profiling/recorder.py
P403n1x87: Ah I see. I guess you're suggesting doing this for every module that we need to disambiguate? Otherwise we would have to do it for every module that we import.

ddtrace/contrib/gunicorn/__init__.py
P403n1x87: In those cases we would have to ensure that the sitecustomize logic is triggered before something like `gevent.monkey.patch_all()` is called. I think there are ways to tell things like `gunicorn` to pre-execute some code? I think we would have to leverage something like that.

ddtrace/profiling/recorder.py
brettlangdon: yeah, I want to make sure the code is clear when we review/update the code later. are we dealing with one of our clones or the shared version?

ddtrace/contrib/gunicorn/__init__.py
brettlangdon: good to write up the docs now before we release this change :)

Yun-Kim: @P403n1x87 are you talking about adding this sitecustomize logic to gunicorn's `pre_fork()` hook?
For ddtrace/gevent users that don't use gunicorn, does `import ddtrace.bootstrap.sitecustomize` before they call `gevent.monkey.patch_all()` work as a solution?

ddtrace/contrib/gevent/__init__.py
Yun-Kim: > My proposal would be to take the ddtrace-specific logic into `ddtrace.bootstrap.preload` and leave the boilerplate code in `sitecustomize.py`. 
@P403n1x87 @brettlangdon can we just do this but import this in `ddtrace.__init__.py` so that this will automatically run when users call `from ddtrace import patch_all`?

tests/profiling/collector/test_memalloc.py
```diff
@@ -163,11 +161,11 @@ def test_memory_collector():
...
-            assert event.thread_id == nogevent.main_thread_id
+            assert event.thread_id == threading.main_thread().ident
             assert event.thread_name == "MainThread"
             count_object += 1
             assert event.frames[2][0] == __file__
-            assert event.frames[2][1] == 154
+            assert event.frames[2][1] == 152
             assert event.frames[2][2] == "test_memory_collector"
 
     assert count_object > 0
```
Yun-Kim, line 171: Should we also remove the `@pytest.mark.skipif(TESTING_GEVENT...)` for the subsequent test `test_memory_collector_ignore_profiler`?

ddtrace/contrib/gevent/__init__.py
emmettbutler: @Yun-Kim I think that approach risks circular imports, but I could be wrong.

To @brettlangdon's question

> do these changes change that requirement and instead require that people import the sitecustomize module?

Yes, they do. I've committed a suggestion of how to handle this based on your first comment in this thread. https://github.com/DataDog/dd-trace-py/pull/5105/commits/81569f1f94b08a245c6fc9f84bdbc70b360ef91fEveryone paying attention, please let me know what you think.

ddtrace/contrib/gunicorn/__init__.py
emmettbutler: I updated this docstring with the guidance nonusers of ddtrace-run need to follow.

ddtrace/bootstrap/sitecustomize.py
```diff
@@ -162,6 +79,48 @@ def update_patched_modules():
...
+        if any(m == _ or m.startswith(_ + ".") for _ in KEEP_MODULES):
+            continue
+
+        if PY2:
+            KEEP_MODULES_PY2 = frozenset(["encodings", "codecs"])
+            if any(m == _ or m.startswith(_ + ".") for _ in KEEP_MODULES_PY2):
+                continue
+            # Store a reference to deleted modules to avoid them being garbage
+            # collected
+            _unloaded_modules.append(sys.modules[m])
```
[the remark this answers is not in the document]
emmettbutler: @Yun-Kim I've searched pretty hard for Python documentation answering this question and haven't been able to find any. Probably the way to find out is to turn it off under 3.x and see what breaks.

ddtrace/contrib/gunicorn/__init__.py
```diff
@@ -1,33 +1,10 @@
...
-  worker_class = "gevent"
-  bind = "8080"
-
-.. code-block:: bash
-
-  DD_GEVENT_PATCH_ALL=1 gunicorn --config gunicorn.conf.py path.to.my:app
+.. note::
+    If your Gunicorn server does not run inside of ``ddtrace-run``,
+    be sure to ``import ddtrace.auto`` as early as possible in your application's
+    lifecycle.
```
brettlangdon, line 7: Given that gunicorn + gevent worker will patch gevent, is there a place where we can be sure we get imported before gevent? 🤔 (am I understanding this correctly?)

Or do we need guidance of "... if using a gevent worker.... please use ddtrace-run" type thing?

releasenotes/notes/nogevent-6f2892cb412f987f.yaml
```diff
@@ -0,0 +1,9 @@
+---
+deprecations:
+  - |
+    internal: `DD_GEVENT_PATCH_ALL` is deprecated and will be removed in the next major version. Gevent compatibility is now automatic
```
brettlangdon, line 4:
```suggestion
    internal: ``DD_GEVENT_PATCH_ALL`` is deprecated and will be removed in the next major version. Gevent compatibility is now automatic
```

tests/internal/test_forksafe.py
```diff
@@ -288,61 +288,72 @@ def fn():
     assert exit_code == 42
 
 
-# FIXME: subprocess marks do not respect pytest.mark.skips
-if sys.version_info < (3, 11, 0):
+@pytest.mark.subprocess(
+    out=lambda _: Counter(_) == {"C": 3, "T": 4},
```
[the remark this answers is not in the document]
emmettbutler: @Yun-Kim I believe this change in expected value is intended to remove the need for `sys.platform` and `sys.version_info` checking and isn't the result of a change in functionality. The expectation now says "expect three Cs and four Ts, order doesn't matter", which is what it said before with more words.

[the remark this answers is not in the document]
Yun-Kim: @emmettbutler I believe previously we were expecting either `CTCTCT` or `CCCTTT` so there was a missing T from the `TestService` I couldn't figure out why 🤔

ddtrace/contrib/gunicorn/__init__.py
emmettbutler: This guidance reflects the setup that's proven to work by the [tests](https://github.com/DataDog/dd-trace-py/blob/3053a9417435abe8e30654237b682ae382598bba/tests/contrib/gunicorn/wsgi_mw_app.py#L16). That gives me some confidence that I don't need to be certain of my understanding of gunicorn internals, which is relevant because my answer to "am I understanding this correctly" is "I don't know".

brettlangdon: ah, ok. I think that test answer my question then. that everything still works for that test when using a gevent worker, then 👍🏻 

I thought I remembered us digging into the gunicorn code and it was explicitly calling gevent patch_all for the user at a certain point in the app startup/forking lifecycle. so I was worried that there was an ordering/timing of things that could still be messed up.

But if we have the tests, great!

ddtrace/profiling/recorder.py
brettlangdon: I think the issue is that we have other places that do import threading as ddtrace_threading because we also need to import the non-cloned version to iterate/check for all active threads.

So my comment here was mostly about, should we have some kind of consistent/not-just-"threading" name to indicate in the code when we are using (at least expected to be using) a cloned module.

Do we think we could end up getting confused in the future by what is cloned vs not when reviewing/debugging code?


(not a strong opinion here, just musings)

ddtrace/contrib/gunicorn/__init__.py
Yun-Kim: To my understanding, as long as we perform our cleanup (module cloning) logic before `gevent.monkey.patch_all()` is called, this should mitigate the problems we've been seeing with gevent patched modules interacting with non-gevent patched modules. What I'm not sure about is if we require `ddtrace` to be imported before `gevent`. 

If @emmettbutler or @P403n1x87 aren't sure as well, then I'd be in favour of guiding users to please run `ddtrace-run`. WDYT?

ddtrace/profiling/recorder.py
emmettbutler: I agree with @Yun-Kim that it's sufficient for contributors to understand that modules imported in `ddtrace/` are probably not accessible to user code. I've committed a suggestion that splits the difference with an explanatory comment for contributors. https://github.com/DataDog/dd-trace-py/pull/5105/commits/20929f246dd5973d60f02e59049abdc185857c70

brettlangdon: 👍🏻 would be good to cover in a central architecture doc if/when we get one too

ddtrace/contrib/gevent/__init__.py
brettlangdon: lgtm


I had a concern about `ddtrace.auto` (yes, I now I suggested the name) thinking that `ddtrace.tracing.auto` might be better, but our sitecustomize really is a central "any product enablement" activation, so I think it works!

a potential future improvement idea could be to try and isolate all the automated setup for each product in `ddtrace.{profiling,tracing,appsec,di}.auto` modules which get used in `sitecustomize.py`. could be nice way to segment them (neat idea, not a priority)

tests/internal/test_forksafe.py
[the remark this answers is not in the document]
emmettbutler: @Yun-Kim oh right, that's true. 🤷

ddtrace/auto.py
```diff
@@ -0,0 +1 @@
+import ddtrace.bootstrap.sitecustomize  # noqa
```
P403n1x87, line 1: I'm not a big fan of this `auto` approach, but if the rest of the team is on board with it then I'm fine with it too.

ddtrace_gevent_check.py
emmettbutler: @brettlangdon I've suggested a fix for this in https://github.com/DataDog/dd-trace-py/pull/5105/commits/1040f1c2385e7a4f0e2c8a4fa0baf18fa3341846, please let me know if that addresses your concern.

ddtrace/bootstrap/sitecustomize.py
```diff
@@ -142,6 +54,21 @@ def cleanup_loaded_modules(aggressive=False):
...
+        postfix="There is no special configuration necessary to make ddtrace work with gevent.",
+        removal_version="2.0.0",
+    )
+if gevent.monkey.is_module_patched("sys"):
+    warnings.warn(
+        "Loading ddtrace after gevent.monkey.patch_all() is not supported and is "
+        "likely to break the application. Use ddtrace-run to fix this, or "
+        "import ddtrace.auto before calling gevent.monkey.patch_all().",
+        RuntimeWarning,
+    )
```
brettlangdon, line 69: we probably don't want to import gevent in here right? could it have side effects for an env which has gevent installed, but not used?

we could do something like this

```python
# Check if gevent has been imported already
if "gevent" in sys.modules:
    import gevent.monkey
    if gevent.monkey.is_module_patched("sys"):
        warnings.warn(...)
```

Aside, why check "sys" module? "threading" seems to be the more important one, but is `sys` always patched or something where it is more consistent to check?

ddtrace/bootstrap/sitecustomize.py
```diff
@@ -142,6 +52,24 @@ def cleanup_loaded_modules(aggressive=False):
 
 log = get_logger(__name__)
 
+if os.environ.get("DD_GEVENT_PATCH_ALL") is not None:
+    deprecate(
+        "The environment variable DD_GEVENT_PATCH_ALL is deprecated and will be removed in a future version. ",
+        postfix="There is no special configuration necessary to make ddtrace work with gevent.",
```
Yun-Kim, line 58:
```suggestion
        postfix="There is no special configuration necessary to make ddtrace work with gevent if using ddtrace-run. If not using ddtrace-run, import ddtrace.auto before calling gevent.monkey.patch_all().",
```

ddtrace/contrib/gevent/__init__.py
```diff
@@ -5,11 +5,8 @@
 The integration patches the gevent internals to add context management logic.
 
 .. note::
-    If :ref:`ddtrace-run<ddtracerun>` is being used set ``DD_GEVENT_PATCH_ALL=true`` and
-    ``gevent.monkey.patch_all()`` will be called as early as possible in the application
-    to avoid patching conflicts.
-    If ``ddtrace-run`` is not being used then be sure to call ``gevent.monkey.patch_all``
-    before importing ``ddtrace`` and calling ``ddtrace.patch`` or ``ddtrace.patch_all``.
+    If ``ddtrace-run`` is not being used then be sure to ``import ddtrace.auto``
+    before calling ``gevent.monkey.patch_all``.
```
Yun-Kim, line 9:
```suggestion
    before calling ``gevent.monkey.patch_all``.
    If ``ddtrace-run`` is being used then no additional configuration is required.
```

ddtrace/contrib/gunicorn/__init__.py
```diff
@@ -1,33 +1,10 @@
...
-
-  workers = 4
-  worker_class = "gevent"
-  bind = "8080"
-
-.. code-block:: bash
-
-  DD_GEVENT_PATCH_ALL=1 gunicorn --config gunicorn.conf.py path.to.my:app
+.. note::
+    If your Gunicorn server does not run inside of ``ddtrace-run``,
```
Yun-Kim, line 5:
```suggestion
    If your Gunicorn server does not run inside of ``ddtrace-run`` and uses ``gevent`` workers,
```

Yun-Kim (commented):
Mostly docs suggestions to make it clear for both use cases (ddtrace-run vs import ddtrace.patch_all)

ddtrace/contrib/gunicorn/__init__.py
P403n1x87, line 5:
```suggestion
    If you cannot wrap your Gunicorn server with the ``ddtrace-run`` command,
```

emmettbutler:
@P403n1x87 @Yun-Kim I applied all of the suggestions from your reviews and re-requested reviews from you. Thank you for all of your hard work on this!

P403n1x87:
@emmettbutler looks like black is not happy 🙁

P403n1x87 (approved):

ddtrace/bootstrap/sitecustomize.py
```diff
@@ -162,6 +91,52 @@ def update_patched_modules():
...
+
+    if not asbool(do_cleanup):
+        return
+
+    # Unload all the modules that we have imported, except for the ddtrace one.
+    # NB: this means that every `import threading` anywhere in `ddtrace/` code
+    # uses a copy of that module that is distinct from the copy that user code
+    # gets when it does `import threading`. The same applies to every module
+    # not in `KEEP_MODULES`.
+    KEEP_MODULES = frozenset(["atexit", "ddtrace", "asyncio", "concurrent", "typing", "logging", "attr"])
```
mabdinur, line 116: nit: I don't think we get much value from defining a frozen set here. Should we define this constant on the module level?

ddtrace/bootstrap/sitecustomize.py
```diff
@@ -162,6 +91,52 @@ def update_patched_modules():
...
+    for m in list(_ for _ in sys.modules if _ not in LOADED_MODULES):
+        if any(m == _ or m.startswith(_ + ".") for _ in KEEP_MODULES):
+            continue
+
+        if PY2:
+            KEEP_MODULES_PY2 = frozenset(["encodings", "codecs"])
```
mabdinur, line 122: same here

ddtrace/bootstrap/sitecustomize.py
```diff
@@ -142,6 +52,25 @@ def cleanup_loaded_modules(aggressive=False):
...
+    if gevent.monkey.is_module_patched("threading"):
+        warnings.warn(
+            "Loading ddtrace after gevent.monkey.patch_all() is not supported and is "
+            "likely to break the application. Use ddtrace-run to fix this, or "
+            "import ddtrace.auto before calling gevent.monkey.patch_all().",
+            RuntimeWarning,
+        )
+
+
 EXTRA_PATCHED_MODULES = {
```
mabdinur, line 74: This might be out of the scope of this PR but do we need `EXTRA_PATCHED_MODULES`? We define a list of integrations to patch here and in `ddtrace._monkey.py`. I am not sure if we need both

ddtrace/contrib/gunicorn/__init__.py
```diff
@@ -1,33 +1,10 @@
...
-
-  workers = 4
-  worker_class = "gevent"
-  bind = "8080"
-
-.. code-block:: bash
-
-  DD_GEVENT_PATCH_ALL=1 gunicorn --config gunicorn.conf.py path.to.my:app
+.. note::
+    If you cannot wrap your Gunicorn server with the ``ddtrace-run``command and
```
mabdinur, line 5: We removed the note stating that only `Python version >=3.6 and <=3.10` is supported. Is this documented somewhere else? Is this still accurate?

ddtrace/profiling/_threading.pyx
```diff
@@ -1,52 +1,64 @@
...
-    # so we will find them normally.
-    if thread_id == nogevent.main_thread_id:
-        return "MainThread"
-
-    # We don't want to bother to lock anything here, especially with eventlet involved 😓. We make a best effort to
-    # get the thread name; if we fail, it'll just be an anonymous thread because it's either starting or dying.
-    try:
-        return threading._active[thread_id].name
-    except KeyError:
+    # Do not force-load the threading module if it's not already loaded
```
mabdinur, line 13: In what scenarios will `threading` not be found in sys.modules? Do we expect this to happen frequently?

ddtrace/profiling/_threading.pyx
```diff
@@ -1,52 +1,64 @@
...
-            return threading._limbo[thread_id].name
+            return threading_mod._active[thread_id].name
         except KeyError:
-            return None
+            try:
+                return threading_mod._limbo[thread_id].name
+            except KeyError:
+                pass
+
+    return None
```
mabdinur, line 33: Looking at [ddtrace/profiling/_threading.pyi](https://github.com/DataDog/dd-trace-py/blob/4dc15e2722b8aec02aa81d50bbdd40e91232ba4e/ddtrace/profiling/_threading.pyi#L4) get_thread_name() should return a string. Here we return None.  Is this something we should address (this existed before, I just noticed now)?

ddtrace/profiling/collector/_lock.py
```diff
@@ -122,12 +122,8 @@ def acquire(self, *args, **kwargs):
...
                 pass
 
-    def release(
-        self,
-        *args,  # type: typing.Any
-        **kwargs  # type: typing.Any
-    ):
-        # type: (...) -> None
+    def release(self, *args, **kwargs):
+        # type (typing.Any, typing.Any) -> None
```
mabdinur, line 126: nit:
```suggestion
        # type: (typing.Any, typing.Any) -> None
```

ddtrace/profiling/collector/stack.pyx
```diff
@@ -2,7 +2,7 @@
 from __future__ import absolute_import
 
 import sys
-import threading
+import threading as ddtrace_threading
```
mabdinur, line 5: Do we need this alias? If so we should leave a comment.

releasenotes/notes/nogevent-6f2892cb412f987f.yaml
```diff
@@ -0,0 +1,9 @@
+---
+deprecations:
+  - |
+    internal: ``DD_GEVENT_PATCH_ALL`` is deprecated and will be removed in the next major version. Gevent compatibility is now automatic
```
mabdinur, line 4: ?
```suggestion
    gevent: ``DD_GEVENT_PATCH_ALL`` is deprecated and will be removed in the next major version. Gevent compatibility is now automatic
```

releasenotes/notes/nogevent-6f2892cb412f987f.yaml
```diff
@@ -0,0 +1,9 @@
...
+    and does not require extra configuration when running with ``ddtrace-run``. If not using ``ddtrace-run``, please import ``ddtrace.auto`` before calling ``gevent.monkey.patch_all()``.
+fixes:
+  - |
+    internal: This fix resolves an incompatibility between ddtrace and gevent that caused threads to hang in certain configurations, for example
```
mabdinur, line 8: ?
```suggestion
    gevent: This fix resolves an incompatibility between ddtrace and gevent that caused threads to hang in certain configurations, for example
```

mabdinur (commented):
Left some questions, feel free to resolve if the questions do not require follow up.

ddtrace/bootstrap/sitecustomize.py
emmettbutler: @mabdinur Can you help me understand the downside of using a frozenset in this way?

ddtrace/bootstrap/sitecustomize.py
emmettbutler: Probably not. We can remove it in a separate pull request.

ddtrace/contrib/gunicorn/__init__.py
emmettbutler: This pull request extends Python support to all versions that ddtrace supports.

ddtrace/profiling/_threading.pyx
P403n1x87: Good catch. We should fix this, but perhaps in a separate PR.

ddtrace/bootstrap/sitecustomize.py
P403n1x87: This constant is only used by this helper function, so I don't see much value in moving it to a global level.

ddtrace/contrib/gunicorn/__init__.py
P403n1x87: We should now support all versions of Python, with perhaps the only exception of Python 3.11, but only because gunicorn itself might not support that version yet

ddtrace/profiling/_threading.pyx
P403n1x87: Since we are unloading modules when gevent is detected, we don't know if and when the threading module will be loaded on every call to this function. We really want to ensure we use the `threading` module that the application uses but without force-loading, so our best bet is to look into `sys.modules`.

ddtrace/profiling/collector/stack.pyx
P403n1x87: with the current coding in this module I believe we don't. However it probably makes it clearer here that we are using the copy of the `threading` module that our library ends up using. This is only relevant when module cloning is triggered.

mabdinur (approved):

Yun-Kim (commented):
LGTM, if we can fix the typing suggestion then I'm good to 👍

ddtrace/profiling/collector/_lock.py
emmettbutler: I tried applying this, but it causes the mypy linting to fail.

emmettbutler:
@Yun-Kim that suggestion causes the linting to fail https://github.com/DataDog/dd-trace-py/pull/5105#discussion_r1123685418
maybe a fix for however typing is broken in that file is out of scope?

ddtrace/profiling/collector/_lock.py
Yun-Kim: @emmettbutler That's odd. Can we revert this to how it was prior to this PR then? i.e.: https://github.com/DataDog/dd-trace-py/blob/af0d087b2ba104845ddd3c7c794b5b1840fe9a44/ddtrace/profiling/collector/_lock.py#L125-L130

Yun-Kim (approved):

P403n1x87 (approved):

PR #5134, opened 2023-02-15
chore(asm): store block callable into ASM context

juanjux:
# Motivation

Store the blocking function into the ASM request context so it can be further called by future blocking features, specially UserId blocking since it'll be implemented as functions (`isUserBlocked`, `blockUser`, `blockUserIfBlocked`) that the user will call at any moment. 

# Design

Each framework will store the block callable on the `_asm_request_context`. Tests have been added to make the call to it from Django or Flask views, but the user should not call them directly (the User ID blocking will provide functions for these).

## Checklist

- [X] Author is aware of the performance implications of this PR as reported in the benchmarks PR comment.

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/appsec/_asm_request_context.py
```diff
@@ -59,6 +65,29 @@ def get_headers_case_sensitive():  # type: () -> bool
...
+    Calls or returns the stored block request callable, if set.
+    """
+    _callable = _DD_BLOCK_REQUEST_CALLABLE.get()
+    if _callable:
+        return _callable()
+
+    log.debug("Block request called but block callable not set by framework")
+
+
+def asm_request_context_set(remote_ip=None, headers=None, headers_case_sensitive=False, block_request_callable=None):
```
emmettbutler, line 89: What's going on here? This empty function looks like an unintentional change to me, but I may be missing something.

emmettbutler (approved):

ddtrace/appsec/_asm_request_context.py
emmettbutler, line 89: What's going on here? This empty function looks like an unintentional change to me, but I may be missing something.

emmettbutler (changes requested):

ddtrace/appsec/_asm_request_context.py
juanjux: Rebase fuckup and then going to do another urgent thing and forgot to set it as draft... thanks for the reminder.

juanjux:
Did an additional change: the django block callable will also block the request (by raising a django `PermissionDenied` exception that will be checked on the `patch.py` to add out custom message. This way it works like the Flask callable and the user doesn't have to return the result in the view, which makes it easier to use when called on nested contexts of the user code.

christophe-papazian (approved):

christophe-papazian (dismissed):

christophe-papazian (approved):

avara1986 (approved):

gnufede (approved):

ddtrace/appsec/_asm_request_context.py
juanjux: @emmettbutler this is fixed, could you remove the request for change?

emmettbutler (approved):

PR #5147, opened 2023-02-17
feat(ci): test merge queues: good build in merge group

jbertran:
# No review required!

This targets a test branch with no impact on `1.x`.

We want queue A-normal -> B-bad -> C-good

This is C-good, which is supposed to fix B-bad and succeed.

Checks:

- [x] check


mabdinur:
Should we close this PR?

PR #5171, opened 2023-02-22
chore(asm): keep waf context during request

christophe-papazian:
To avoid performance issues, this PR defines a way to keep the same WAF context for all calls to the WAF during a single request.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.

### DIFF
--- a/ddtrace/appsec/ddwaf/__init__.py
+++ b/ddtrace/appsec/ddwaf/__init__.py
@@ -140,9 +140,11 @@ if _DDWAF_LOADED:
             start = time.time()
 
             if self._ctx == 0:
+                LOGGER.warning("DDWaf failsafe to create the context")
                 self._ctx = ddwaf_context_init(self._handle)
 
             if self._ctx == 0:
+                LOGGER.error("DDWaf failure: no context created")
                 return DDWaf_result(None, [], 0, (time.time() - start) * 1e6)
 
             result = ddwaf_result()
--- a/ddtrace/appsec/ddwaf/__init__.py
+++ b/ddtrace/appsec/ddwaf/__init__.py
@@ -167,6 +167,7 @@ if _DDWAF_LOADED:
         # type: () -> text_type
         return ddwaf_get_version().decode("UTF-8")
 
+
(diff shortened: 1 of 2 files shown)


ddtrace/appsec/ddwaf/__init__.py
```diff
@@ -126,33 +139,34 @@ def run(
...
-                LOGGER.error("DDWaf failure to create the context")
+            if self._ctx == 0:
+                LOGGER.warning("DDWaf failsafe to create the context")
+                self._ctx = ddwaf_context_init(self._handle)
+
+            if self._ctx == 0:
+                LOGGER.error("DDWaf failure: no context created")
                 return DDWaf_result(None, [], 0, (time.time() - start) * 1e6)
 
+            result = ddwaf_result()
```
gnufede, line 150: This block was previously inside a try block, are we sure about removing that?

christophe-papazian: Yes, the try/finally was only present to free the context. We don't want that anymore.

gnufede (approved):

ddtrace/appsec/ddwaf/__init__.py
```diff
@@ -126,33 +139,34 @@ def run(
             # type: (...) -> DDWaf_result
             start = time.time()
 
-            ctx = ddwaf_context_init(self._handle)
-            if ctx == 0:
-                LOGGER.error("DDWaf failure to create the context")
+            if self._ctx == 0:
+                LOGGER.warning("DDWaf failsafe to create the context")
```
avara1986, line 143: is not this validation redundant with `_at_request_start`?

juanjux (approved):

ddtrace/appsec/ddwaf/__init__.py
christophe-papazian: It's a failsafe. Yes it's redondant but in the case that span_start was not called for any reason before the call to the WAF, it allows to process as usual without errors.

PR #5188, opened 2023-02-23
fix(dbm): support byte string queries

mabdinur:
Database monitoring and APM Linking only supports unicode sql queries. This PR extends support to sql queries with the type: `bytes`.

When `DD_DBM_PROPAGATION_MODE=full` the following code snippet raises an excpetion: 

```
import psycopg2
import ddtrace

ddtrace.patch(psycopg=True)
conn = psycopg2.connect(**POSTGRES_CONFIG)

# raises -> `TypeError: can only concatenate str (not "bytes") to str`
conn.cursor().execute(b"select 'blah'")
```

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ZStriker19:
Should it be `fix` rather than `chore` for the title?

ddtrace/propagation/_database_monitoring.py
```diff
@@ -25,23 +25,24 @@
 log = get_logger(__name__)
 
 
-def _default_sql_injector(dbm_comment, sql_statement):
+def default_sql_injector(dbm_comment, sql_statement):
     # type: (str, str) -> str
     try:
+        if isinstance(sql_statement, bytes):
+            return dbm_comment.encode("utf-8") + sql_statement
```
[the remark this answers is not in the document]
emmettbutler: Here's a sketch of my suggestion for a pragmatic way to sidestep the issue of other encodings:
```suggestion
            try:
                dbm_comment = dbm_comment.encode("utf-8", errors="strict")
            except UnicodeDecodeError:
                dbm_comment = ""
            return dbm_comment + sql_statement
```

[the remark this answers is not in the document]
ZStriker19: this seems good to me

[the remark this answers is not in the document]
emmettbutler: @ZStriker19 if you think this is ok, please commit it and let's see what the tests do.

ddtrace/propagation/_database_monitoring.py
```diff
@@ -25,23 +25,24 @@
 log = get_logger(__name__)
 
 
-def _default_sql_injector(dbm_comment, sql_statement):
+def default_sql_injector(dbm_comment, sql_statement):
     # type: (str, str) -> str
```
brettlangdon, line 29: can we update the type hints here:

we need to have a `Union[str, bytes]` right? and same for the return `Union[str, bytes]`

(I don't know how to force the return type to be the same as the input type for `sql_statement`, since that is the actual ideal implementation)

ddtrace/propagation/_database_monitoring.py
```diff
@@ -25,23 +25,27 @@
...
     try:
+        if isinstance(sql_statement, bytes):
+            try:
+                dbm_comment = dbm_comment.encode("utf-8", errors="strict")
+            except UnicodeDecodeError:
+                dbm_comment = ""
```
brettlangdon, line 35:
```suggestion
                dbm_comment = b""
```

?

brettlangdon: if we have a decoding error, then we'd be trying to concatenate `str + bytes` which is no good

ddtrace/propagation/_database_monitoring.py
```diff
@@ -25,23 +26,27 @@
 log = get_logger(__name__)
 
 
-def _default_sql_injector(dbm_comment, sql_statement):
-    # type: (str, str) -> str
+def default_sql_injector(dbm_comment, sql_statement):
+    # type: (Union[str, bytes], Union[str, bytes]) -> Union[str, bytes]
```
brettlangdon, line 30:
```suggestion
    # type: (str, Union[str, bytes]) -> Union[str, bytes]
```

emmettbutler (dismissed):

tests/internal/test_database_monitoring.py
```diff
@@ -114,3 +117,24 @@ def test_dbm_propagation_full_mode():
 
     # ensure that dbm tag is set (required for full mode)
     assert dbspan.get_tag(_database_monitoring.DBM_TRACE_INJECTED_TAG) == "true"
+
+
+def test_default_sql_injector(caplog):
```
brettlangdon, line 122: can we create a test case for when the dbm comment would throw a bytes encoding error?

this case would have caught the `dbm_comment = ""` "bug"

is this even possible?

ZStriker19 (dismissed):

ddtrace/propagation/_database_monitoring.py
```diff
@@ -25,23 +26,27 @@
...
     try:
+        if isinstance(sql_statement, bytes):
+            try:
+                dbm_comment = dbm_comment.encode("utf-8", errors="strict")
+            except UnicodeDecodeError:
+                dbm_comment = b""
         return dbm_comment + sql_statement
-    except TypeError:
+    except Exception:
```
brettlangdon, line 38: what else can get thrown here? wouldn't it always be a `TypeError`?

if we want to keep this as `Exception`, then we should probably include the Exception message in the error log, otherwise we cannot know why this try/except failed.

ZStriker19: Let's switch it back to `TypeError` now that Emmet's change catches the `UnicodeDecodeError`.

ddtrace/propagation/_database_monitoring.py
```diff
@@ -25,23 +26,27 @@
 log = get_logger(__name__)
 
 
-def _default_sql_injector(dbm_comment, sql_statement):
-    # type: (str, str) -> str
+def default_sql_injector(dbm_comment, sql_statement):
+    # type: (str, Union[str, bytes]) -> Union[str, bytes]
```
brettlangdon, line 30: I am not asking for this change, I was curious so looked it up.

```python
from typing import AnyStr

def default_sql_injector(dbm_comment, sql_statement):
    # type: (str, AnyStr) -> AnyStr
```

POC

```python
from typing import AnyStr


def concat(a: str, b: AnyStr) -> AnyStr:
    if isinstance(b, bytes):
        return a.encode("utf-8") + b
    return a + b


a = concat("", b"")
assert type(a) == bytes
b = concat("", "")
assert type(b) == str

```

tests/internal/test_database_monitoring.py
ZStriker19: decided it is not worth testing

brettlangdon: @ZStriker19 and I discussed this, it is going to be hard to get `dbm_comment` where it can raise a `UnicodeDecodeError`. The best option we have is probably having a separate file with a different file encoding so we can put non-utf-8 characters into `dbm_comment`, but then try to encode as `utf-8`.

but this is a lot of hoops to jump through to test this.

an alternative could be to mock over `dbm_comment.encode` and force the error as a side effect but... then is that even a good test scenario?



I think as long as the file encoding is `utf-8` or `ascii` then we shouldn't need the try/except around `dbm_comment.encode("utf-8")` since it _shouldnt_ fail, but keeping the try/except as a fail-safe is 👍🏻

ddtrace/propagation/_database_monitoring.py
brettlangdon: > or could just do except (TypeError, ValueError):

I meant, we could do this instead of the inner try/except block.

releasenotes/notes/dbm-support-sql-queries-with-the-type-bytes-30fb4b75ea2f8cf8.yaml
```diff
@@ -0,0 +1,4 @@
+---
+fixes:
+  - |
+    dbm: Support sql comments for byte strings.
```
mabdinur, line 4:
```suggestion
    dbm: Support sql queries with the type  ``byte``.
```

ddtrace/propagation/_database_monitoring.py
```diff
@@ -25,23 +26,27 @@
...
     try:
+        if isinstance(sql_statement, bytes):
+            try:
+                dbm_comment = dbm_comment.encode("utf-8", errors="strict")
+            except UnicodeDecodeError:
+                dbm_comment = b""
         return dbm_comment + sql_statement
     except TypeError:
```
mabdinur, line 38:
```suggestion
    except (TypeError, UnicodeDecodeError):
```

emmettbutler (approved):

ddtrace/propagation/_database_monitoring.py
```diff
@@ -25,23 +26,27 @@
...
+    # type: (str, Union[str, bytes]) -> Union[str, bytes]
     try:
+        if isinstance(sql_statement, bytes):
+            # raises a unicode error when sql statements with non utf-8 encodings
+            # are concatenated. This operation will log a warning.
+            sql_statement = sql_statement.decode("utf-8", errors="strict")
+            return  bytes(dbm_comment + sql_statement, "utf-8")
         return dbm_comment + sql_statement
-    except TypeError:
+    except (TypeError, UnicodeDecodeError, UnicodeEncodeError):
```
brettlangdon, line 38:
```suggestion
    except (TypeError, ValueError):
```

Both unicode errors derive from `ValueError`, if we wanted to be more permissive, but this is not a requirement, being explicit and narrow for expected types is 👍🏻

emmettbutler (dismissed):

ddtrace/propagation/_database_monitoring.py
```diff
@@ -25,23 +26,27 @@
...
-def _default_sql_injector(dbm_comment, sql_statement):
-    # type: (str, str) -> str
+def default_sql_injector(dbm_comment, sql_statement):
+    # type: (str, Union[str, bytes]) -> Union[str, bytes]
     try:
+        if isinstance(sql_statement, bytes):
+            # raises a unicode error when sql statements with non utf-8 encodings
+            # are concatenated. This operation will log a warning.
+            sql_statement = sql_statement.decode("utf-8", errors="strict")
+            return  bytes(dbm_comment + sql_statement, "utf-8")
```
brettlangdon: I don't think this is correct.

the "source" is `sql_statement` which can be provided as a `str` or a `bytes`, and we want to prepend the `str` `dbm_comment` to `sql_statement`. We shouldn't muck about with encoding/decoding `sql_statement`, we should make `dbm_comment` conform to what `sql_statement` is and then prepend.

wdyt?

emmettbutler: makes sense to me! https://github.com/DataDog/dd-trace-py/pull/5188/commits/8f156b4cde01c5034f38419ad510313c35373488

brettlangdon (dismissed):

emmettbutler (approved):

juanjux (approved):

emmettbutler (dismissed):

brettlangdon (approved):

emmettbutler:
I'm currently trying to regenerate snapshots to fix the failing tests.

ddtrace/propagation/_database_monitoring.py
```diff
@@ -25,23 +26,24 @@
...
-def _default_sql_injector(dbm_comment, sql_statement):
-    # type: (str, str) -> str
+def default_sql_injector(dbm_comment, sql_statement):
+    # type: (str, Union[str, bytes]) -> Union[str, bytes]
     try:
+        if isinstance(sql_statement, bytes):
+            return dbm_comment.encode("utf-8", errors="strict") + sql_statement
         return dbm_comment + sql_statement
-    except TypeError:
+    except (TypeError, UnicodeDecodeError, UnicodeEncodeError):
```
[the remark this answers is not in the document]
emmettbutler: Personally I prefer the explicitness of listing the exceptions. Doesn't matter too much either way.

emmettbutler (approved):

ZStriker19 (approved):

emmettbutler:
@Mergifyio backport 1.9

PR #5190, opened 2023-02-23
fix(aws_lambda): update signal handler to avoid wrapping multiple times

duncanista:
## Description
Fix a bug in which our signal might be wrapped multiple times, which cause it to call the the signal multiple times.

## Testing strategy
Updated unit tests which were causing this, I first added them in #5176, moved them here.

## Risk
This was only affecting unit testing, since this wouldn't have happened in production. This is because in testing we are setting the same handler for multiple tests, which the system would recognize as previously defined. In an AWS Lambda function, once the environment is dead, all signals are lost. So there will not be multiple wrapping.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


emmettbutler (dismissed):

ddtrace/contrib/aws_lambda/patch.py
```diff
@@ -32,21 +32,25 @@ def _crash_flush(_, __):
...
 
-    Returns the handler of the wrapped signal.
+    Returns a signal of type `sig` with function `_f`
     """
-    old_signal = None
-    if callable(signal.getsignal(sig)):
-        old_signal = signal.getsignal(sig)
+    old_signal = signal.getsignal(sig)
+    if not callable(old_signal) or old_signal == f:
+        old_signal = None
```
Yun-Kim, line 44: Can we simplify the logic into something like: 

```suggestion
    if not callable(old_signal) or old_signal == f:
        return signal.signal(sig, f)
```
This is the intended result if there is already a wrapped f on the signal, correct?

ddtrace/contrib/aws_lambda/patch.py
```diff
@@ -32,21 +32,25 @@ def _crash_flush(_, __):
...
 
     def wrap_signals(*args, **kwargs):
         if old_signal is not None:
             old_signal(*args, **kwargs)
         f(*args, **kwargs)
 
-    return signal.signal(sig, wrap_signals)
+    _f = wrap_signals if old_signal else f
```
Yun-Kim, line 51:
```suggestion
```
Let me know if this is functionally the same. To me, this is easier to follow.

ddtrace/contrib/aws_lambda/patch.py
duncanista: Yes, I changed it to your suggestions since it's more readable, thanks for the feedback!

ddtrace/contrib/aws_lambda/patch.py
duncanista: It worked as expected in the tests, and yeah, agree, easier to read!

PR #5215, opened 2023-02-28
feat(asgi): websocket support in asgi middleware

haozhun:
With the latest version of fastapi (0.92.0), websocket requests do not show up in DataDog APM traces.

The source code clearly shows that websocket support is intended because there are pre-existing mentions of websocket and ws in this file.

I patched the datadog package using the local changes in this PR, and found that websocket requests now show up in APM traces. However, any simple websocket app should suffice. Here's a simple minimal example: https://fastapi.tiangolo.com/advanced/websockets/

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


tests/contrib/fastapi/test_fastapi.py
```diff
@@ -633,6 +633,14 @@ def test_table_query_snapshot(snapshot_client):
     }
 
 
+@snapshot()
+async def test_traced_websocket(snapshot_client):
```
mabdinur, line 637:
```suggestion
def test_traced_websocket(snapshot_client):
```

ZStriker19:
Hi @haozhun , testing this manually I seem to be hitting an issue where if the websocket endpoint has an infinite loop, like in the fastapi example app you link to, the span continues on forever, until the webpage is closed or refreshed. With this I worry about having super long spans from a memory and usability standpoint (would users gain a lot from seeing these super long spans?). Perhaps instead we should try instrumenting `websocket.receive_text` & `websocket.send_text` etc. Another possibility would be to put this capability behind a flag. What are your thoughts on this?

haozhun:
> testing this manually I seem to be hitting an issue where if the websocket endpoint has an infinite loop, like in the fastapi example app you link to, the span continues on forever, until the webpage is closed or refreshed. 

To me, this isn't an issue. It's how it's supposed to work. You could say the same thing about HTTP long poll, HTTP streaming, or HTTP server-sent events. All three of which show up as plain GET or POST in DataDog, and people don't argue that DataDog should track only active compute/io. Ultimately, WebSocket or HTTP is just a tool. People use it in different ways, and such usage pattern may or may not be well suited for any particular visualization or tool.

> from a memory ... standpoint

I'm not sure what you mean. Maybe you're talking about performance aspects of DataDog implementation details. I'm not familiar with it and don't have anything to share.

> usability standpoint (would users gain a lot from seeing these super long spans?)

Yes, in my specific case, I'm interested. 

> Perhaps instead we should try instrumenting websocket.receive_text & websocket.send_text etc.

That doesn't sound right. People are likely interested in the compute or IO that leads up to the send_text. Tracking only send_text or receive_text doesn't seem interesting.

> Another possibility would be to put this capability behind a flag.

Sure. This is your call.

ZStriker19:
Hi @haozhun. I talked with the team about this. Let's put this feature behind a flag to collect some feedback from users first before we consider enabling it by default, add that flag to configuration.rst, and explain there that this could lead to spans that last for a very long time aka until the connection is closed, so users are aware of what to expect.

Regarding the memory question, I was thinking if you say, had thousands of connections from different clients, you'd have thousands of spans in memory, it does seem unlikely to cause a real issue though.

haozhun:
@ZStriker19, thanks for the update. I appreciate it. To clarify, when you say "Let's", are you asking that I do it, or are you suggesting that you will do it?

ZStriker19:
> @ZStriker19, thanks for the update. I appreciate it. To clarify, when you say "Let's", are you asking that I do it, or are you suggesting that you will do it?

Hi @haozhun  I can do it. Just not sure why I'm getting a policy violation error when making the call to the app endpoint for testing. Any insight? If not, I'll continue looking into it further. `starlette.websockets.WebSocketDisconnect: (1008, '')`

noudin-ledger:
Is this still a WIP?
It would be great to have this resolved :)

ZStriker19:
Trying to solve this test failure, I can't recreate locally despite having what appears to be the same environment (looked at riot dependencies and copied my local hash that matches them). It's also not intuitive to me how for this particular test, we could end up with a lack of propagation possibly, resulting in two traces rather than one from the changes that were made. I also wondered if it was perhaps just the operation name that had changed, since this PR does some work around that, but I couldn't find how that might happen. @mabdinur do you have any ideas here? https://app.circleci.com/pipelines/github/DataDog/dd-trace-py/53348/workflows/1de03822-d854-4e29-9fa9-f6d798e9866c/jobs/3397883

ZStriker19:
I think we're good to go!

tests/contrib/fastapi/test_fastapi.py
```diff
@@ -621,6 +622,29 @@ def test_table_query_snapshot(snapshot_client):
...
+def test_traced_websocket(test_spans, snapshot_app):
+    client = TestClient(snapshot_app)
+    with override_global_config(dict(_trace_asgi_websocket=True)):
+        with client.websocket_connect("/ws") as websocket:
+            data = websocket.receive_json()
+            assert data == {"test": "Hello WebSocket"}
+            spans = test_spans.pop_traces()
+            assert len(spans) == 1
+            span = spans[0][0]
+            assert span.service == "fastapi"
```
emmettbutler, line 635: I think these assertions are redundant with the snapshot comparison.

ddtrace/contrib/asgi/middleware.py
```diff
@@ -120,9 +124,14 @@ async def __call__(self, scope, receive, send):
...
-        resource = " ".join((scope["method"], scope["path"]))
-        operation_name = self.integration_config.get("request_span_name", "asgi.request")
-        operation_name = schematize_url_operation(operation_name, direction=SpanDirection.INBOUND, protocol="http")
+        resource = " ".join([method, scope["path"]])
+        if scope["type"] == "http":
+            operation_name = self.integration_config.get("request_span_name", "asgi.request")
+            operation_name = schematize_url_operation(operation_name, direction=SpanDirection.INBOUND, protocol="http")
+        else:
+            # in the case of websockets we don't currently schematize the operation names
+            operation_name = self.integration_config.get("request_span_name", "asgi.request")
```
emmettbutler, line 133: This is a duplicate of line 129. Deduplicating them would reduce the potential for future bugs in this code.

ddtrace/settings/config.py
```diff
@@ -446,6 +446,8 @@ def __init__(self):
         self.version = os.getenv("DD_VERSION", default=self.tags.get("version"))
         self.http_server = self._HTTPServerConfig()
 
+        self._trace_asgi_websocket = os.getenv("DD_TRACE_ASGI_WEBSOCKET", default=False)
```
emmettbutler, line 449: Putting this option on the `asgi` integration config [here](https://github.com/DataDog/dd-trace-py/blob/3b0311d865a03ebb92ac0ba9573166dbeb660f69/ddtrace/contrib/asgi/middleware.py#L31) would avoid leaking integration-specific code into the top-level config object.

emmettbutler (changes requested):
Looks great, nice work. The one thing I think it's important to change is nesting the new configuration option in the asgi integration config object.

ddtrace/settings/config.py
ZStriker19: Ah I hadn't thought about how that's organized, makes sense!

emmettbutler: I'm not really sure without digging in. You could give it a try and see if you can get the tests to pass. There are a few contrib-specific options in configuration.rst, so you wouldn't be breaking an existing rule by putting this new option at the global level. Even so, I think adding the config to multiple contribs could still be preferable for the sake of encapsulation. Your call.

ZStriker19: I went with adding it to each config that uses the middleware. Lmk what you think.

ddtrace/contrib/django/patch.py
```diff
@@ -69,6 +69,7 @@
         ),
         use_handler_resource_format=asbool(os.getenv("DD_DJANGO_USE_HANDLER_RESOURCE_FORMAT", default=False)),
         use_legacy_resource_format=asbool(os.getenv("DD_DJANGO_USE_LEGACY_RESOURCE_FORMAT", default=False)),
+        trace_asgi_websocket=os.getenv("DD_TRACE_ASGI_WEBSOCKET", default=False),
```
emmettbutler, line 72:
```suggestion
        trace_asgi_websocket=os.getenv("DD_ASGI_TRACE_WEBSOCKET", default=False),
```
This might follow existing naming conventions a bit more closely.

releasenotes/notes/asgi-support-websockets-f970956661062882.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    asgi: Trace websockets in asgi middleware behind feature flag. Set ``DD_TRACE_ASGI_WEBSOCK=true`` to enable.
```
emmettbutler, line 4:
```suggestion
    asgi: Trace websockets in asgi middleware behind feature flag. Set ``DD_ASGI_TRACE_WEBSOCKET=true`` to enable.
```
I missed this typo on the previous review, sorry about that.

emmettbutler (changes requested):

emmettbutler:
Looks like one of the newly-added tests fails sometimes. You might try a technique like [this](https://github.com/DataDog/dd-trace-py/commit/2c8384e32b25029700db48443a01c83f923dfdc2) to get around it.

ddtrace/contrib/asgi/middleware.py
```diff
@@ -28,7 +29,12 @@
 
 config._add(
     "asgi",
-    dict(service_name=config._get_service(default="asgi"), request_span_name="asgi.request", distributed_tracing=True),
+    dict(
+        service_name=config._get_service(default="asgi"),
+        request_span_name="asgi.request",
+        distributed_tracing=True,
+        trace_asgi_websocket=os.getenv("DD_ASGI_TRACE_WEBSOCKET", default=False),
```
mabdinur, line 36: We should rename this configuration to better align with existing conventions. 

We should also make `trace_asgi_websocket` "internal". This will give use flexibility and allow us to refactor this code. Setting this config via envar should be sufficient.

```suggestion
        _trace_asgi_websocket=os.getenv("DD_TRACE_ASGI_WEBSOCKET", default=False),
```

emmettbutler (approved):

Issue #5220, opened 2023-02-28
Lambda Layer is crashing because current_root_span is returning None


amersaw:
### Summary of problem
Today we start facing crashes on Lambda app that was using the [datadog-lambda-python](https://github.com/DataDog/datadog-lambda-python) extension. I checked the logs and found that crash caused by `ddtrace`. 

### Which version of dd-trace-py are you using?
1.9.0
The `datadog-lambda-python` package is using  the latest version of ddtrace (<2) 

### Which version of pip are you using?
pip 22.3.1

### Which libraries and their versions are you using?
```
datadog==0.44.0
datadog-lambda==4.68.0
ddsketch==2.0.4
ddtrace==1.9.0
```

### How can we reproduce your problem?
- Deploy a lambda app as described in [the following official documentation](https://docs.datadoghq.com/serverless/installation/python/?tab=containerimage)
- Couldn't determine when exactly the app will crash, but about 30% of requests are crashing where other requests working as expected.
- When the request crash I get the output log mentioned below.

### What is the result that you get?
The whole lambda request fails with the following Cloudwatch logs:

```
START RequestId: bac50653-2b72-49c7-86e7-ad6695169298 Version: $LATEST
Traceback (most recent call last):
File "/var/runtime/bootstrap", line 9, in <module>
main()
File "/var/runtime/bootstrap.py", line 346, in main
event_request = lambda_runtime_client.wait_next_invocation()
File "/var/runtime/lambda_runtime_client.py", line 57, in wait_next_invocation
response = self.runtime_connection.getresponse()
File "/var/lang/lib/python3.7/http/client.py", line 1373, in getresponse
response.begin()
File "/var/lang/lib/python3.7/http/client.py", line 319, in begin
version, status, reason = self._read_status()
File "/var/lang/lib/python3.7/http/client.py", line 280, in _read_status
line = str(self.fp.readline(_MAXLINE + 1), "iso-8859-1")
File "/var/lang/lib/python3.7/socket.py", line 589, in readinto
return self._sock.recv_into(b)
File "/var/task/ddtrace/contrib/aws_lambda/patch.py", line 47, in wrap_signals
f(*args, **kwargs)
File "/var/task/ddtrace/contrib/aws_lambda/patch.py", line 25, in _crash_flush
root_span.error = 1
AttributeError: 'NoneType' object has no attribute 'error'
[...]

PR #5224, opened 2023-02-28
fix(aws_lambda): warn when impending timeout but all traces have been finished

duncanista:
## Description

The root cause might be that customer is finishing all its spans, even the one provided by `datadog-lambda`. Then hitting an impending timeout, which makes the contrib `aws_lambda` to try to tag an error in the root span. After this, the contrib might try to close all unfinished spans too. Which in turn, if there's no root span, there will not be current span, therefore, adding a conditional for both root span and current span to act only if they are not `None`.

Discovered in issue #5220 

## Testing strategy
Added a unit test which finishes `aws_lambda` root span and then times out. Expecting not to crash, and to not tag the error.

## Risk
User will not be getting an Impending Timeout error tagged in their root span, but they will be receiving a warning message in their logs.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


releasenotes/notes/fix-aws-lambda-crashing-when-timeout-has-no-root-span-4a0fdacf707709a9.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    aws_lambda: This fix resolves an issue where the runtime would crash when no root span is found on an impending timeout. The root span provided by `datadog-lambda` is not recommended to be finished by the customer. But it will log a warning stating that no root span was found.
+
```
mabdinur, line 5:
```suggestion

```

ddtrace/contrib/aws_lambda/patch.py
```diff
@@ -20,14 +20,17 @@ def _crash_flush(_, __):
...
     root_span = tracer.current_root_span()
-    root_span.error = 1
-    root_span.set_tag_str(ERROR_MSG, "Datadog detected an Impending Timeout")
-    root_span.set_tag_str(ERROR_TYPE, "Impending Timeout")
+    if root_span is not None:
+        root_span.error = 1
+        root_span.set_tag_str(ERROR_MSG, "Datadog detected an Impending Timeout")
+        root_span.set_tag_str(ERROR_TYPE, "Impending Timeout")
+    else:
+        log.warning("An impending timeout was reached, but no root span was found. No error will be tagged.")
```
mabdinur, line 29: Is this unexpected?  When aws lambda is used should there always be an active root span? 
```suggestion
```

releasenotes/notes/fix-aws-lambda-crashing-when-timeout-has-no-root-span-4a0fdacf707709a9.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    aws_lambda: This fix resolves an issue where the runtime would crash when no root span is found on an impending timeout. The root span provided by `datadog-lambda` is not recommended to be finished by the customer. But it will log a warning stating that no root span was found.
```
mabdinur, line 4: Can we simplify this? Would this work?
```suggestion
    aws_lambda: Resolves an unhandled exception which occurs when no root span is found before a lambda timeouts.
```

ddtrace/contrib/aws_lambda/patch.py
duncanista: Yes, this is unexpected, since the `datadog-lambda` package always creates a wrapper span titled `aws.lambda`.

releasenotes/notes/fix-aws-lambda-crashing-when-timeout-has-no-root-span-4a0fdacf707709a9.yaml
```diff
@@ -0,0 +1,4 @@
+---
+fixes:
+  - |
+    aws_lambda: Resolves an exception not being handled, which occurs when no root span is found before a lambda times out.
```
mabdinur, line 4:
```suggestion
    aws_lambda: Resolves an exception not being handled, which occurs when no root span is found before a lambda times out.

```

ddtrace/contrib/aws_lambda/patch.py
duncanista: Update: customer explained how their lambda behaved and it doesn't appear to be falling in the description above, we might have to follow up asking them to file a support ticket to understand why this might be happening to them. Nonetheless, this is log is useful on its own.

mabdinur (approved):

PR #5231, opened 2023-02-28
chore(ci): cleanup riotfile by reducing tested contrib versions

Yun-Kim:
## TL;DR

This PR reduces the number of tested versions of third-party libraries in our riotfile, which should significantly reduce our CI workload and duration. Additionally, the number of parallel workers for contrib test suites have been reduced as we no longer need to run as many test suites and are bottlenecked by non-contrib test suites (tracer, profiler, benchmarks, integration).

In summary, most contrib test suite matrices have been cut in half, and by minimizing CI worker parallelism for contrib jobs (due to the `tracer/profiler/benchmarks` bottleneck jobs), we have cut CI costs by almost 50%. All this, without significant loss of coverage beyond 2 years. 

Note: This PR is not dropping official support for older versions, merely dropping the amount of testing we do for very old versions.

# Problem
Currently, the library's policy on support for third-party libraries is as follows, from `docs/contributing.rst`:
```
``ddtrace`` tries to support as many active versions of a library as possible.
The general rule is:
  - If the integration depends on internals of the library then test every
    minor version going back 2 years.
  - Else test each major version going back 2 years.
```
However, our docs also state the oldest versions of each integration that we provide support for, which conflicts with the above.

In reality, we have not maintained the lists of tested versions on our riotfile, if only to add newer versions to test matrices. This has resulted in our test matrices ballooning in size due to the sheer number of python versions to test as well as new releases from every single library we provide integrations for. This has three major problems:

1. We are over testing (almost every single minor version from beyond 2 years for certain libraries) without any significant gain in confidence or coverage, as most libraries' minor releases should not (in theory at least) have any breaking changes.
2. CI already takes a long time to run, and this will likely continue to increase as we have to support/test newer versions of third-party libraries that we provide integrations for. This means more developer time and CI costs growing with very little value that we get from testing 8-year old versions.
3. We do not have a strongly codified versioning policy for users. It's confusing which versions we support or should stop/keep supporting, which has contributed to the growing test matrix for the past few years.

# Proposal

This PR cleans up the riotfile in accordance to this proposed third-party library versioning testing support policy:

```
For each third-party library ddtrace provides integrations for:
- Test the latest major version. Specifically, test the min-max of the latest major version (the oldest and newest minor releases from the major version) going back 2 years.
- Test the latest minor release of non-latest major versions going back 2 years.
- If no new releases within the past 2 years, test the latest version.
- If support for legacy Python versions (2.7, 3.5, 3.6) was dropped in a major release, then test the latest minor release to support the legacy Python version only.
```
Let’s take for example the aiobotocore library, which has 3 major versions (0.x, 1.x, and 2.x) as of Feb 27, 2023: 
- 0.x is older than 2 years, so only test the latest minor version of 1.x (1.4) and the oldest and newest minor versions of 2.x (2.0, 2.4).
- aiobotocore dropped support for Python 3.5 in 0.12. So test 0.11 with Python 3.5 only.

# Result
With the proposed changes from this PR, none of the contrib test suites take more than 5 minutes to run with our existing parallelism. The result of this PR is that most contrib test suite matrices have been cut in half without significant loss of coverage beyond 2 years. 

Since there are non-contrib test suites that are current bottlenecks (e.g. `tracer, integration-agent, benchmarks, profiler` which run around 15+ minutes per run), we can reduce parallelism without losing any significant benefit. By minimizing the parallelism for the contrib jobs in CI, we have cut CI costs by almost 50%, while not gaining any time from running CI.

# Risk Management
This PR has loss of test coverage for much older (>4 years) major versions, as well as minor versions in between our new min/max testing. This might appear like a significant risk, but consider the following:

- In most cases, older major versions are still tested due to legacy Python support cases (Python 2.7, 3.5, 3.6). 
- For libraries that have not released a new version within the past 2 years, we are still testing the most recent version. The majority of the changes in PR 5231 was in dropping testing for major versions older than 4 years.
- Minor versions between the min/max testing have now been dropped, but by definition minor versions should not add any breaking changes that impact our loss of coverage. 

Most importantly, our library also has continued supporting older minor versions until now, and by continuing to test latest released minor versions for all integrations as they get released, we will continue to catch any incompatibility issues between minor versions as they get released, meaning there should be a minimal risk from this loss of coverage.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


emmettbutler (dismissed):
This is amazing!

brettlangdon:
@Yun-Kim `build_docs` is failing with this:

```
Warning, treated as error:
/root/project/docs/versioning.rst:49:malformed hyperlink target.
```

docs/contributing.rst
```diff
@@ -274,13 +274,14 @@ enabling and disabling the instrumentation and overriding the service name.
...
-The general rule is:
+The ``ddtrace`` library's support policy for third-party libraries we provide integrations for is as follows:
 
-  - If the integration depends on internals of the library then test every
-    minor version going back 2 years.
 
-  - Else test each major version going back 2 years.
+  - Test the oldest and latest minor versions of the most latest major version going back 2 years.
+
+  - Test the latest minor version of any previous major version going back 2 years.
```
brettlangdon, line 282: I wonder if we should re-word as "what we support" vs "what we test".

Since supporting every version going back 2 years vs how we test that can be different.

e.g. we aren't support only the oldest and latest minor versions, we are supporting everything in between, but we are only testing those versions.


I don't have better wording, so take this as just an optional nit for now.

docs/versioning.rst
```diff
@@ -46,6 +46,21 @@ Support levels
...
 
 
+.. _versioning_integration_library_support:
+
+Library support
+===============
+
+
+``ddtrace`` tries to support as many active versions of a library as possible, but in practice this is difficult
+due to the sheer maintenance burden of testing. The ddtrace library's support policy for third-party libraries that
```
brettlangdon, line 56:
```suggestion
``ddtrace`` tries to support as many active versions of a library as possible.
The ddtrace library's support policy for third-party libraries that
```


internal detail customers shouldn't have to worry about 🤔 

I am not convinced that my change is required, just sharing a thought.

docs/index.rst
```diff
@@ -40,11 +40,11 @@ contacting support.
 +--------------------------------------------------+---------------+----------------+
 |                   Integration                    |  |SUPPVER|    | |AUTO|    [1]_ |
 +==================================================+===============+================+
-| :ref:`aiobotocore`                               | >= 0.2.3      | No             |
```
brettlangdon, line 43: It seems like we are "dropping" support for some integrations here, these should likely have release notes for each... ?

riotfile.py
```diff
@@ -408,14 +408,14 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
...
-                                "gevent": ["~=20.9.0", "~=20.12.0", "~=21.1.0"],
+                                "gevent": ["~=21.1.0", latest],
                                 "greenlet": "~=1.0",
                             },
                         ),
                         Venv(
                             pys="3.10",
                             pkgs={
-                                "gevent": ["~=21.8.0"],
+                                "gevent": ["~=21.8.0", latest],
```
brettlangdon, line 418: we cannot combine this venv with the previous because 21.8.0 is when 3.10 support was added?

riotfile.py
```diff
@@ -476,6 +476,29 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
             pkgs={
                 "vertica-python": [">=0.6.0,<0.7.0", ">=0.7.0,<0.8.0"],
             },
+            # venvs=[
+            # FIXME: tests fail on vertica 1.x
```
brettlangdon, line 480: do we need these before we can merge?

brettlangdon: Same for the other FIXMEs ?

ddtrace/contrib/tornado/stack_context.py
```diff
@@ -10,7 +8,7 @@
 
 # tornado.stack_context deprecated in Tornado 5 removed in Tornado 6
 # instead use DefaultContextProvider with ContextVarContextManager for asyncio
-_USE_STACK_CONTEXT = not (sys.version_info >= (3, 7) and tornado.version_info >= (5, 0))
+_USE_STACK_CONTEXT = not tornado.version_info >= (5, 0)
```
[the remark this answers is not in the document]
Yun-Kim: I'm not sure why this version gated Python >= 3.7, tornado 5.0 introduced `stack_context` but 5.0 was still compatible with Python 2.7/3.5/3.6. AFAIK, this was an unnecessary version gate.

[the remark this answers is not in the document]
brettlangdon: 👍🏻

riotfile.py
Yun-Kim: The FIXMEs are all what we **should** be testing (was missing test coverage before), but we currently fail them as we don't have support for vertica/sqlalchemy/algoliasearch/... latest versions. 
I'm okay with merging this, I left these comments in to easily add testing once we add support in the future. I thought adding actual support was outside the scope of this PR.

emmettbutler (dismissed):

riotfile.py
brettlangdon: > (was missing test coverage before)

ok, got it, as long as we aren't removing test coverage for something and it is expected but unsupported 👍🏻

emmettbutler (approved):

releasenotes/notes/integration-support-drop-e45718d6befa78d5.yaml
```diff
@@ -0,0 +1,135 @@
...
+#    * Be concise
+#    * Include actionable steps with the necessary code changes
+#    * Include relevant links (bug issues, upstream issues or release notes, documentation pages)
+#    * Use full sentences with sentence-casing and punctuation.
+#    * Before using Datadog specific acronyms/terminology, a release note must first introduce them with a definition.
+#
+#    Release notes should not:
+#    * Be vague. Example: ``fixes an issue in tracing``.
+#    * Use overly technical language
+#    * Use dynamic links (``stable/latest/1.x`` URLs). Instead, use static links (specific version, commit hash) whenever possible so that they don't break in the future.
```
brettlangdon, line 26:
```suggestion
```

releasenotes/notes/integration-support-drop-e45718d6befa78d5.yaml
```diff
@@ -0,0 +1,135 @@
...
+prelude: >
+  While ``ddtrace`` aims to support as many versions of a library as possible, this is difficult in practice due to
+  the maintenance burden of testing all possible versions of a library. The ``ddtrace`` library's support policy
+  moving forward is to have testing support for:
```
brettlangdon, line 30:
```suggestion
  The ``ddtrace`` library's support policy moving forward is to have testing support for:
```

releasenotes/notes/integration-support-drop-e45718d6befa78d5.yaml
```diff
@@ -0,0 +1,135 @@
...
+    - The latest major version of a library, which includes testing the oldest and latest minor versions going back 2 years. 
+    - The latest minor version for previous major versions going back 2 years.
+    - Legacy Python versions will have testing support for the latest version to support that legacy Python version.
+  
+  .. note:
+    While compatibility for older versions has not changed, support for these older versions has ended moving forward 
+    and we recommend upgrading to the minimum supported version as indicated on the integration-specific release note below.
```
brettlangdon, line 37: link to index/integration supported page?

docs/index.rst
brettlangdon, line 43: we also need to update the corp docs: https://github.com/DataDog/documentation

we should also consider adding the note about library version support to the corp docs as well.

Yun-Kim: As per offline discussion: we are only dropping testing for older versions, not official support. We won't need to update public docs yet, at least until 2.0.

emmettbutler (approved):

mabdinur (approved):

PR #5297, opened 2023-03-10
chore: add release notes script

ZStriker19:
This script can generate release notes for patch, release candidate, and minor releases. How to setup and use it is included in the script. Note that release notes should still always be double checked by a human. Note that the script is dependent on our current formatting of release notes. If that changes then the script will need to be modified as well.

Examples of script output: 
`BASE=1.10 RC=1 python release.py`   
Releases · DataDogdd-trace-py 2023-03-14 at 4 59 01 PM

`BASE=1.9 PATCH=1 python release.py`
Releases · DataDogdd-trace-py 2023-03-14 at 4 58 06 PM

`BASE=1.10 python release.py`
Releases · DataDogdd-trace-py 2023-03-14 at 4 57 27 PM

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


scripts/release.py
```diff
@@ -0,0 +1,199 @@
...
+
+def create_release_draft():
+    #  Figure out the version of the library that you’re working on releasing grabbed with VERSION envar
+    base = os.getenv("BASE")
+    gh_token = os.getenv("GH_TOKEN")
+    rc = bool(os.getenv("RC"))
+    patch = bool(os.getenv("PATCH"))
+
+    if base is None:
+        raise ValueError("Need to specify the base version with envar e.g. BASE=1.10.0")
```
Yun-Kim, line 49: From this error message, it sounds like `BASE` needs to include the patch version. Is that right? Or does this error message need to be corrected?

scripts/release.py
```diff
@@ -0,0 +1,199 @@
...
+Optional:
+    RC - Whether or not this is a release candidate. e.g. RC=1 or RC=0
+    PATCH - Whether or not this a patch release. e.g. PATCH=1 or PATCH=0
+    
+Examples:
+Generate release notes for next release candidate version of 1.11: `BASE=1.11 RC=1 python release.py`
+
+Generate release notes for next patch version of 1.13: `BASE=1.13 PATCH=1 python release.py`
+
+Generate release notes for the 1.15 release: `BASE=1.15 python release.py`
```
Yun-Kim, line 32: what happens if you don't have a RC or patch specified for a minor version that doesn't exist? Does it generate an official draft release? Or does this error out?

scripts/release.py
```diff
@@ -0,0 +1,199 @@
...
+2. Create Github token: https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token#creating-a-personal-access-token-classic
+3. Give the Github token repo, user, audit_log, and project permissions.
+2. Add `export GH_TOKEN=<github token>` to the `.env` file.
+
+Usage:
+Required:
+    BASE - The base branch you are building your release candidate, patch, or minor release off of.
+        If this is a rc1, then just specify the branch you'll create after the release is published. e.g. BASE=1.9
+
+Optional:
```
Yun-Kim, line 23: Could we add an optional param to generate release notes without creating a draft release? I could see it being useful instead of manually using reno. If this is out of scope, feel free to ignore.

scripts/release.py
```diff
@@ -0,0 +1,199 @@
...
+            try:
+                rn_clean += "### %s\n\n%s" % (key, rns_dict_clean[key])
+            except KeyError:
+                continue
+
+        create_draft_release(branch=branch, name=name, tag=tag, dd_repo=dd_repo, rn=rn_clean, prerelease=False)
+
+
+def clean_rn(rn_raw):
+    return rn_raw.decode().split("## v")[0].replace("\n## Unreleased\n", "", 1).replace("# Release Notes\n", "", 1)
```
Yun-Kim, line 167: Does this remove all release notes generated other than the one of interest?

scripts/release.py
```diff
@@ -0,0 +1,199 @@
...
+        tags = dd_repo.get_tags()
+        rc_version = 0
+        for tag in tags:
+            try:
+                rc_num = re.findall(search, tag.name)[0][0]
+                rc_num = int(rc_num)
+            except (IndexError, ValueError, TypeError):
+                continue
+            if rc_num > rc_version:
+                rc_version = rc_num
```
Yun-Kim, line 70: Confused on what this loop is doing. Is it finding the most recent RC tag for the given minor version?

scripts/release.py
```diff
@@ -0,0 +1,199 @@
...
+
+        #  if this is the first rc for this base, we want to target 1.x
+        if rc_version == 0:
```
Yun-Kim, line 73: Just for clarity, is there any way to make the check be for `rc_version == 1` since RCs are indexed starting at 1, not 0?

scripts/release.py
```diff
@@ -0,0 +1,199 @@
...
+from github import Github
+
+
+"""This release notes script is built to create a release notes draft for release candidates, patches, and minor releases.
+
+Setup:
+1. Create a `.env` file in the scripts directory.
+2. Create Github token: https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token#creating-a-personal-access-token-classic
+3. Give the Github token repo, user, audit_log, and project permissions.
+2. Add `export GH_TOKEN=<github token>` to the `.env` file.
```
Yun-Kim, line 16: We can add exporting our Github tokens into personal `~/.zshrc` config files, no need for creating a `.env` file in scripts. I would prefer setup requiring github tokens being exported one way or another, not necessarily requiring an `.env` file. What do you think?

scripts/release.py
```diff
@@ -0,0 +1,199 @@
...
+
+Generate release notes for next patch version of 1.13: `BASE=1.13 PATCH=1 python release.py`
+
+Generate release notes for the 1.15 release: `BASE=1.15 python release.py`
+"""
+
+home_dir = os.path.expanduser("~")
+env_loaded = load_dotenv(dotenv_path="%s/.env" % home_dir)
+if not env_loaded:
+    raise ValueError("No envars were loaded from .env file. Please follow the instructions in the script.")
```
Yun-Kim, line 38: We can just do a `if not os.getenv("GH_TOKEN"):` check here instead of the `.env` file as mentioned in my above comment?

scripts/release.py
Yun-Kim: I was primarily confused by what `rc_version` means, I am suspecting that this is the most recent RC version instead of the RC version we want to generate release notes for. Can we clarify this? For example renaming `rc_version` to `most_recent_rc` and making `rc_version = most_recent_rc + 1`

scripts/release.py
```diff
@@ -0,0 +1,199 @@
...
+        rn_raw = generate_rn(branch)
+        rn = clean_rn(rn_raw)
+        create_draft_release(branch=branch, name=name, tag=tag, dd_repo=dd_repo, rn=rn, prerelease=True)
+
+    # patch release
+    elif patch:
+        # figure out the patch version we want
+        search = "v%s.((\d+))" % base
+        tags = dd_repo.get_tags()
+        patch_version = 1
```
Yun-Kim, line 91: Same request here. Can we rename `patch_version` to `most_recent_patch_version` and make `patch_version = most_recent_patch_version + 1`?

scripts/release.py
```diff
@@ -0,0 +1,199 @@
...
+    else:
+        name = "%s.0" % base
+        tag = "v%s" % name
+        branch = base
+        rn_raw = generate_rn(branch)
+        # get anything in unreleased section in case there were updates since the last RC
+        unreleased = (
+            rn_raw.decode().split("## v")[0].replace("\n## Unreleased\n", "", 1).replace("# Release Notes\n", "", 1)
+        )
+        unreleased = unreleased.split("###")
```
Yun-Kim, line 117: Can we extract this into a helper? This would help simplify what's happening.

scripts/release.py
```diff
@@ -0,0 +1,199 @@
...
+        try:
+            unreleased_sections = dict(section.split("\n\n-") for section in unreleased)
+        except ValueError:
+            unreleased_sections = {}
+        relevant_rns = []
+        if unreleased_sections:
+            relevant_rns.append(unreleased_sections)
```
Yun-Kim, line 124: Same here

scripts/release.py
```diff
@@ -0,0 +1,199 @@
...
+
+        rns = rn_raw.decode().split("## v")
+        split_str = "## v%s" % base
```
Yun-Kim, line 127: Do we need this variable?

scripts/release.py
```diff
@@ -0,0 +1,199 @@
...
+                    prelude_section[" Prelude"] = sections[0].split("\n\n", 1)[1]
+                    sections = sections[1:]
+                sections_dict = {**dict(section.split("\n\n", 1) for section in sections), **prelude_section}
+                relevant_rns.append(sections_dict)
+        # join all the sections from different relevant RCs together
+        keys = set().union(*relevant_rns)
+        rns_dict = {k: "".join(dic.get(k, "") for dic in relevant_rns) for k in keys}
+        rns_dict_clean = {}
+        for key in rns_dict.keys():
+            rns_dict_clean[key.lstrip()] = rns_dict[key]
```
Yun-Kim, line 144: Same here (extract into helper to simplify logic here). My main concern is that someone without context (for example, me) of how the PyGithub library works might be overwhelmed by what's going on, and extracting this into a helper might make more sense to readers in the future.

scripts/release.py
ZStriker19: It generates an official draft release. I'll add that example, thanks!

ZStriker19: Re-reading this, what do you think the expected behavior should be, should we error out? I think we shouldn't need to, since it should be quite obvious one the draft is on the release notes page for review if you specified the wrong version.

scripts/release.py
ZStriker19: Sure, I'll add that.

scripts/release.py
ZStriker19: Yes exactly, I'll comment it.

scripts/release.py
ZStriker19: Yes correct, it's finding the rc version we want to use by figuring out what the most recent rc of that minor version was and adding 1 to it. Yeah I'll clarify the variable names more. Thanks!

scripts/release.py
ZStriker19: Sure, I'll add the 1 earlier and do the comparison against 1 instead. That makes the logic clearer.

scripts/release.py
ZStriker19: That's great, I just had issues doing that initially so switched to the .env file. Let me give the .zshrc way another shot.

scripts/release.py
ZStriker19: I tested it, and we error out, which I think is fine behavior.

scripts/release.py
ZStriker19: seemingly not?

scripts/release.py
```diff
@@ -0,0 +1,221 @@
...
+        rn_raw = generate_rn(branch)
+        rn = clean_rn(rn_raw)
+        create_draft_release(branch=branch, name=name, tag=tag, dd_repo=dd_repo, rn=rn, prerelease=True)
+
+    # patch release
+    elif patch:
+        # figure out the patch version we want
+        search = r"v%s.((\d+))" % base
+        tags = dd_repo.get_tags()
+        latest_patch_version = 1
```
Yun-Kim, line 90: don't patch releases start with 0?

Yun-Kim (commented):
Looks good! Just left a couple suggestions

scripts/release.py
```diff
@@ -0,0 +1,221 @@
...
+
+Usage:
+Required:
+    BASE - The base branch you are building your release candidate, patch, or minor release off of.
+        If this is a rc1, then just specify the branch you'll create after the release is published. e.g. BASE=1.9
+
+Optional:
+    RC - Whether or not this is a release candidate. e.g. RC=1 or RC=0
+    PATCH - Whether or not this a patch release. e.g. PATCH=1 or PATCH=0
+    PRINT - Whether or not the release notes should be printed to CLI or be used to create a Github release. Default is 0 e.g. PRINT=1 or PRINT=0
```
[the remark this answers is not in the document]
ZStriker19: Yeah PRINT=0 will create a draft release, but it defaults to 0, so basically in my mind the norm will be to just not add the print envar to your command, because you want to generate release notes, not print them. If you do want to print them, then the option is there.

scripts/release.py
ZStriker19: I don't think so?

mabdinur (approved):
I have a few usability nits. This script works so we can make these enhancements in a follow up PR:
  - We should add instructions to create a venv and install PyGithub. I tried to using this scripts and it immediately raised an import error because I did not have this package. Also `pip install github` installs a completely different package. We should call this out.
  - nit: The release script only works if it is run in the `scripts/` directory. This could be a pain point if someone does not expect this. We should either call this out in the instructions or just detect a users current directory and make the script just work.
  - nit: Is there a way to update the release notes logic to remove the PATCH configuration? Say if I am on the 1.9 branch and 1.9.4 already exists, I shouldn't have to set PATCH=1 to get notes for 1.9.5. This might increase the complexity of the script but it would be nice to have default values that work for common scenarios. 
  - The scripts switches my branch and doesn't switch me back. So after I generate release notes I am on 1.9 or 1.10. The script isn't on these branches so I can re-run it. It's just not the best experience.

mabdinur:
By default pre-release is unchecked and latest-release is checked. Is there an api check/uncheck these boxes? It would be nice if the release script had this functionality
Screen Shot 2023-03-23 at 6 00 26 PM

![latest release checkbox]

ZStriker19:
> By default pre-release is unchecked and latest-release is checked. Is there an api check/uncheck these boxes? It would be nice if the release script had this functionality

There is, let me check it real quick, I'd like to make sure that is right, and will make another PR for your usability notes.

PR #5303, opened 2023-03-10
feat(kafka): add confluent_kafka integration

emmettbutler:
This PR adds automatic instrumentation for [confluent_kafka](https://github.com/confluentinc/confluent-kafka-python), which is the Python client for Apache Kafka v0.8 and above. 

Currently, this integration provides a wrapper for the `confluent_kafka.Producer` and `confluent_kafka.Consumer` classes, specifically providing tracing for `confluent_kafka.Producer.produce()` and `confluent_kafka.Consumer.poll()`.

This integration supports `confluent_kafka>=1.7.0`.

Notes: 
- consumer group_id as well as consumer/producer client_id are not publicly available attributes in the `confluent_kafka` API. While the `message_key, partition, and tombstone` attributes are set in our instrumentation, the former two attributes are unavailable until the `confluent_kafka` API changes to publicly include them in the future. 
- When a traced producer creates a message with `Producer.produce(topic)`, but no partition is explicitly set, the resulting span will have a default partition of `-1` to signify that the partition will be set by the built-in partitioner. `Consumer poll()` spans with messages will have the correct partition set as the tag as that information is available directly from the message.
- The traced `Consumer.poll()` span will have a boolean tag `kafka.received_message` to indicate if the `poll()` method resulted in a received message or not. If a message was received, then the span will include message-specific tags including `topic,partition,tombstone`.
- The traced `Consumer.poll()` (if a message was received) and `Producer.produce()` spans will include a boolean `tombstone` tag to indicate whether or not the message was a tombstone (empty/null message value).

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


P403n1x87:
Just a reminder to run `scripts/contrib-patch-tests.py` to generate patch tests for the new integration.

emmettbutler (commented):
Looking good!

ddtrace/contrib/kafka/patch.py
```diff
@@ -0,0 +1,124 @@
...
+            return message
+
+
+def patch():
+    if getattr(confluent_kafka, "_datadog_patch", False):
+        return
+    setattr(confluent_kafka, "_datadog_patch", True)
+
+    setattr(confluent_kafka, "Producer", TracedProducer)
+    setattr(confluent_kafka, "Consumer", TracedConsumer)
```
mabdinur, line 116: Instead of patching the Producer and Consumer classes can we patch the methods directly. Patching classes leads to some edge cases. It has bit us before with the TracedMongoClient.

If a user imports `confluent_kafka.Producer`  and/or `confluent_kafka.Consumer` then calls ddtrace.patch(kafka=True). We won't patch any Producer or Consumer classes that have already been imported.

Producer is not patched:
```
>>> from confluent_kafka import Producer, Consumer
>>> from ddtrace import patch
>>> patch(kafka=True)
>>> pd = Producer(dict())
%5|1679066975.054|CONFWARN|rdkafka#producer-1| [thrd:app]: No `bootstrap.servers` configured: client will not be able to connect to Kafka cluster
>>> type(pd)
<class 'cimpl.Producer'>
```

Producer is patched:
```
 >>> from ddtrace import patch
>>> patch(kafka=True)
>>> from confluent_kafka import Producer, Consumer
>>> pd = Producer(dict())
%5|1679066848.217|CONFWARN|rdkafka#producer-1| [thrd:app]: No `bootstrap.servers` configured: client will not be able to connect to Kafka cluster
>>> type(pd)
<class 'ddtrace.contrib.kafka.patch.TracedProducer'>
```

ddtrace/contrib/kafka/patch.py
```diff
@@ -0,0 +1,124 @@
...
+        if not pin or not pin.enabled():
+            return func(*args, **kwargs)
+
+        with pin.tracer.trace(
+            kafkax.CONSUME,
+            service=trace_utils.ext_service(pin, config.kafka),
+            span_type=SpanTypes.WORKER,
+        ) as span:
+            message = func(*args, **kwargs)
+            span.set_tag_str("messaging.system", kafkax.SERVICE)
```
mabdinur, line 93: "messaging.system" can be a constant. I can see us using this tag in other integrations.

ddtrace/contrib/kafka/patch.py
```diff
@@ -0,0 +1,124 @@
...
+            if rate is not None:
+                span.set_tag(ANALYTICS_SAMPLE_RATE_KEY, rate)
+            return func(*args, **kwargs)
+
+    # in older versions of confluent_kafka, bool(Producer()) evaluates to False,
+    # which makes the Pin functionality ignore it.
+    def __bool__(self):
+        return True
+
+    __nonzero__ = __bool__
```
mabdinur, line 72: Why does `bool(Producer())` evaluates to False in older versions? 
Was this a bug? Was this value used? 

I am scared about overriding default values just to use the `Pin` object. We might be introducing unwanted side effects.

tests/contrib/kafka/test_kafka.py
```diff
@@ -0,0 +1,228 @@
...
+from ddtrace import Pin
+from ddtrace.constants import ANALYTICS_SAMPLE_RATE_KEY
+from ddtrace.contrib.kafka.patch import patch
+from ddtrace.contrib.kafka.patch import unpatch
+from tests.contrib.config import KAFKA_CONFIG
+from tests.utils import TracerTestCase
+from tests.utils import override_config
+
+
+class TestKafkaPatch(TracerTestCase):
```
mabdinur, line 13: nit: I believe we're moving away from using the `TracerTestCase` class in our test suite. Can we add snapshot tests and use plain pytest functions.

ddtrace/contrib/kafka/patch.py
emmettbutler: @mabdinur unfortunately these methods can't be patched directly because they are [implemented in C](https://github.com/confluentinc/confluent-kafka-python/blob/bbe1225dc7d36df2c7e3e0f7476bb283fabcd1ff/src/confluent_kafka/src/confluent_kafka.c#L2698) and [directly imported](https://github.com/confluentinc/confluent-kafka-python/blob/bbe1225dc7d36df2c7e3e0f7476bb283fabcd1ff/src/confluent_kafka/__init__.py#L24).

This pull request implements a similar solution to [OpenTelemetry's](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/2629c121610693ad3add1e84e94d237733271403/instrumentation/opentelemetry-instrumentation-confluent-kafka/src/opentelemetry/instrumentation/confluent_kafka/__init__.py#L258).

ddtrace/contrib/kafka/patch.py
```diff
@@ -0,0 +1,124 @@
...
+            span.set_tag_str("messaging.system", kafkax.SERVICE)
+            span.set_tag_str(COMPONENT, config.kafka.integration_name)
+            span.set_tag_str(SPAN_KIND, SpanKind.CONSUMER)
+            span.set_tag_str(kafkax.RECEIVED_MESSAGE, str(message is not None))
+            if message is not None:
+                message_key = message.key() or ""
+                span.set_tag_str(kafkax.TOPIC, message.topic())
+                span.set_tag_str(kafkax.MESSAGE_KEY, ensure_text(message_key))
+                span.set_tag(kafkax.PARTITION, message.partition())
+                span.set_tag_str(kafkax.TOMBSTONE, str(len(message) == 0))
```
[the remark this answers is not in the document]
emmettbutler: Topic name and message key are about as sensitive as database table name in other integrations. Partition and tombstone are not sensitive at all.

ddtrace/contrib/kafka/patch.py
Yun-Kim: Isn't our patch guidance to patch before importing? AFAIK our `patch()` or `patch_all()` works by registering the patch hook to the module import, meaning users have to patch before importing `confluent_kafka`.

tests/contrib/kafka/test_kafka.py
Yun-Kim: I'd love to convert this to snapshots, but unsure how to do so since there's a non-deterministic number of spans generated when the consumer waits for a message. Do you know how to account for non-deterministicness in a snapshot test?

ddtrace/contrib/kafka/patch.py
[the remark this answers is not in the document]
Yun-Kim: Note that the message value itself is not stored as a tag, which would be the most risk of containing PII.

tests/contrib/kafka/test_kafka.py
```diff
@@ -0,0 +1,152 @@
...
+GROUP_ID = "test_group"
+BOOTSTRAP_SERVERS = "localhost:{}".format(KAFKA_CONFIG["port"])
+KEY = "test_key"
+if six.PY3:
+    PAYLOAD = bytes("hueh hueh hueh", encoding="utf-8")
+else:
+    PAYLOAD = bytes("hueh hueh hueh")
+
+
+class KafkaConsumerPollFilter(TraceFilter):
```
ZStriker19, line 24: With this filter, are we ever testing that the poll spans look as expected?

tests/contrib/kafka/test_kafka.py
```diff
@@ -0,0 +1,152 @@
...
+    message = None
+    while message is None:
+        message = consumer.poll(1.0)
+    unpatch()
+    consumer.close()
+
+
+@pytest.mark.snapshot
+def test_analytics_with_rate(producer, consumer):
+    with override_config("kafka", dict(analytics_enabled=True, analytics_sample_rate=0.5)):
```
ZStriker19, line 137: This tests a sample rate with a single span being produced? Oh, or we're just checking that the tag is there essentially?

tests/contrib/kafka/test_kafka.py
Yun-Kim: `confluent_kafka.Consumer.poll()` is often used in this pattern (see their own [example here](https://github.com/confluentinc/confluent-kafka-python/blob/97f08fe107d259eff6f9c281a61d92e204d2935d/examples/consumer.py#L93-L104)):
```python
while True:
    message = consumer.poll(1.0)
    if message:
        # do things
        break
```
Since consumers often poll repeatedly until they get a message, most `consumer.poll()` spans will not have a message associated with them, and the number of empty (message-wise) `poll()` spans is non-deterministic. To get around this in snapshot testing, I added a filter here to get rid of all `poll()` spans that don't have messages associated with them, so the snapshot should always test that there is a `producer.produce()` span and a `consumer.poll()` span that has a message.

tests/contrib/kafka/test_kafka.py
Yun-Kim: This is a similar snapshot compared to the rest of the tests, with 1 `producer.produce()` and 1 `consumer.poll()` spans. But this one specifically is just making sure that the analytics sample rate is included and reflects the overridden config sample rate.

ddtrace/contrib/kafka/patch.py
```diff
@@ -0,0 +1,125 @@
...
+        super(TracedConsumer, self).__init__(consumer)
+        Pin().onto(self)
+
+    def poll(self, *args, **kwargs):
+        func = self.__wrapped__.poll
+        pin = Pin.get_from(self)
+        if not pin or not pin.enabled():
+            return func(*args, **kwargs)
+
+        with pin.tracer.trace(
```
piochelepiotr, line 88: so usually, the code is:
```
message = consumer.consume()
doSomething(message)
doSomethingElse(message)
```
what would be nice (and what .NET allows for for eg) is to keep the span alive until the next call of consumer.consume(). That way, the span from consume is a parent span of `doSomething()` and `doSomethingElse` in my example. Is that something that would be possible to do?

ddtrace/contrib/kafka/patch.py
```diff
@@ -0,0 +1,125 @@
...
+            kafkax.CONSUME,
+            service=trace_utils.ext_service(pin, config.kafka),
+            span_type=SpanTypes.WORKER,
+        ) as span:
+            message = func(*args, **kwargs)
```
piochelepiotr, line 93: can we also tag the span with the consumer group ? And maybe message offset?

.circleci/config.yml
```diff
@@ -15,6 +15,8 @@ mongo_image: &mongo_image mongo:3.6
 httpbin_image: &httpbin_image kennethreitz/httpbin@sha256:2c7abc4803080c22928265744410173b6fea3b898872c01c5fd0f0f9df4a59fb
 vertica_image: &vertica_image sumitchawla/vertica:latest
 rabbitmq_image: &rabbitmq_image rabbitmq:3.7-alpine
+zookeeper_image: &zookeeper_image confluentinc/cp-zookeeper:7.3.1
+kafka_image: &kafka_image confluentinc/cp-kafka:7.3.2
```
mabdinur, line 19: Are these images used in circleci? If not, we should remove them

Yun-Kim: Will remove them, thanks for catching that!

ddtrace/contrib/kafka/patch.py
```diff
@@ -0,0 +1,125 @@
...
+            kafkax.PRODUCE,
+            service=trace_utils.ext_service(pin, config.kafka),
+            span_type=SpanTypes.WORKER,
+        ) as span:
+            span.set_tag_str(MESSAGING_SYSTEM, kafkax.SERVICE)
+            span.set_tag_str(COMPONENT, config.kafka.integration_name)
+            span.set_tag_str(SPAN_KIND, SpanKind.PRODUCER)
+            span.set_tag_str(kafkax.TOPIC, topic)
+            span.set_tag_str(kafkax.MESSAGE_KEY, ensure_text(message_key))
+            span.set_tag(kafkax.PARTITION, partition)
```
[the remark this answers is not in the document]
Yun-Kim: @piochelepiotr I wrote it such that we set the partition to -1 if no partition was explicitly set (i.e. using the built-in partitioner). 
From confluent-kafka's [documentation](https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/index.html#confluent_kafka.Producer.produce), it looks like `producer.produce()` returns None, so I'm not sure what you mean by getting the partition/offset in the return.

ddtrace/contrib/kafka/patch.py
Yun-Kim: Adding the message offset is no issue, but from confluent-kafka's [documentation](https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/index.html#confluent_kafka.Consumer), the consumer group_id not a publicly accessible from the `Consumer` instance. I tried looking for ways to extract that, but doesn't look trivial (i.e. have to rely on confluent_kafka internals which isn't confirmed to always behave in the same way and could change at any time).

.circleci/config.yml
mabdinur, line 19:
```suggestion
```

.circleci/config.yml
```diff
@@ -273,6 +275,7 @@ postgres_server: &postgres_server
     - POSTGRES_USER=postgres
     - POSTGRES_DB=postgres
 
+
```
mabdinur, line 278: nit:
```suggestion
```

tests/contrib/kafka/test_kafka.py
ZStriker19: Thanks for the nice explanation! I got that part, I just wanted to make sure that we're testing that the empty poll spans are looking as expected somewhere?

Yun-Kim: Good point, I'll see if I can add that as a test. Thanks for catching!

ddtrace/contrib/kafka/patch.py
Yun-Kim: @piochelepiotr this PR instruments `Consumer.poll()`, which is similar but not exactly the same as `consume()` AFAIK. Should we wrap `Consumer.consume()` instead? 

I'm not opposed to keeping the span alive, but my initial worry is for edge cases where `poll()` is called but no subsequent `poll()` is called. How do we know to finish the span without the next call of `consumer.poll()`?

ddtrace/contrib/kafka/patch.py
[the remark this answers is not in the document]
emmettbutler: @Yun-Kim I believe @piochelepiotr is referring to the fact that `Message` instances [contain the partition number](https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/index.html#confluent_kafka.Message.partition) from which they were consumed.

ddtrace/contrib/kafka/patch.py
```diff
@@ -0,0 +1,127 @@
...
+            span.set_tag_str(kafkax.TOMBSTONE, str(value is None))
+            span.set_tag(SPAN_MEASURED_KEY)
+            rate = config.kafka.get_analytics_sample_rate()
+            if rate is not None:
+                span.set_tag(ANALYTICS_SAMPLE_RATE_KEY, rate)
+            return func(*args, **kwargs)
```
emmettbutler, line 66: This approach might address @piochelepiotr's note about the "return".

```suggestion
            message = func(*args, **kwargs)
            if message:
                span.set_tag(kafkax.PARTITION, message.partition() or -1)
            return message
```

ddtrace/contrib/kafka/patch.py
emmettbutler: @Yun-Kim I believe `[consume](https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/index.html#confluent_kafka.Consumer.consume)()` is a convenience wrapper around `poll()` that returns a given number of messages rather than `poll()`'s 1-or-0 return guarantee.

It seems to me that we should not keep `poll()` spans alive because of the issue you mentioned. Separately, we could add instrumentation for `consume()` that closed spans either on the next `consume()` call or on `consume()` timeout. I'd personally be fine with putting that in a subsequent pull request.

mabdinur (dismissed):
LGTM 😄

ddtrace/contrib/kafka/patch.py
Yun-Kim: @emmettbutler agreed, I think this behavior doesn't need to block this PR and can be implemented in a future PR after spec-ing out the behavior more thoroughly.

ZStriker19 (approved):

PR #5326, opened 2023-03-14
feat(tracing): support 128-bit trace ids [does not include support for b3 and w3c]

mabdinur:
## Description 

Currently Datadog spans are correlated using 64 bit trace ids. However, as industry standards have sprung up around distributed tracing ([OpenTracing](https://github.com/opentracing/specification/blob/master/rfc/trace_identifiers.md#trace-context-http-headers), [OpenCensus](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/Span.md#traceid), and now [OpenTelemetry](https://opentelemetry.io/docs/reference/specification/trace/api/#spancontext)) the accepted standard length for trace IDs has settled at 128 bits. This PR introduced a configuration which can enable the generation and propagation of 128bit trace ids (`DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED`).

With this change 128bit trace ids will be `opt-in`.  128bit trace ids will become the default standard in a future PR.

### Format

Current format: trace ids are 64bit integers with the following binary representation: <64 random bits>
Proposed format: trace ids are 128bit integers with the following binary representation: `<32-bit unix seconds><32 bits of zero><64 random bits>`

### Encoding

128bit integers are not compatible with ddtrace trace encoders and agent endpoints. As a workaround 128bit trace ids will be encoded in two fields. 
  - The 64 lowest order bits will be stored in the `trace_id` field in Json and MsgPack Encoders.
  - The 64 highest order bits will be encoded into hex and stored as a span tag with the key `_dd.p.tid`. 

#### Example

Span: `<Span(id=3822442170818112150,trace_id=133030438088573679178230931306392465947,parent_id=None,name=bit128)>`
trace_id as binary: `01100100000101001011101011111101000000000000000000000000000000001101100111101101100000011001111100101111111100100011001000011011`

trace_id lower order bits: `1101100111101101100000011001111100101111111100100011001000011011`
trace_id lower order bits as integer: `15703349996414972443`
trace_id higher order bits: `0110010000010100101110101111110100000000000000000000000000000000`
trace_id higher order bits as hex: `6414bafd00000000`

Encoded Span: `span_id=3822442170818112150` , `trace_id=15703349996414972443` , `name=bit128`, ` _meta={"_dd.tid": "6414bafd00000000")>`

### Distributed Tracing

This PR only adds support for propagating 128bit trace ids using the datadog propagation mode. Although b3 and w3c trace header formats support 128bit the Datadog tracer truncates all trace ids to 64 bits. Moving forward this is unacceptable. Supporting 128bit trace ids in b3 and w3c will be added in a future PR. 

#### Datadog Distributed Tracing Headers

Similar to the encoding example above the `x-datadog-trace-id`  header will propagate the 64 lower order bits as an integer and `x-datadog-tags`  header will propagate the higher order bits as hex (using the `_dd.tid` tag).

##### Example

Span: `<Span(id=3822442170818112150,trace_id=133030438088573679178230931306392465947,parent_id=None,name=bit128)>`

distributed tracing headers: `{"x-datadog-tags": "t.tid:6414bafd00000000",  "x-datadog-trace-id": 15703349996414972443, "x-datadog-parent-id": 3822442170818112150}`

### Sampling

The 64 lowest order bits are random but the 64 highest order bits correspond to the unix time and are not random. When 128bit trace ids are generated we should only use the lowest order 64 bits (random component) to determine whether a span should be sampled. This will ensure when trace ids mapped to values from 0 to 1 we get a uniform random distribution.  

## Testing Strategy

- Run the tracer test suite with `DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED=true`. This will ensure all tracing operations work as expected when 128bit trace ids are generated (ex: sampling, distributed tracing, encoding).
- Ensure all 128bit trace id system tests pass.
- Add integration tests for the following scenarios
  - 128bit trace ids are encoded without data loss and raising an OverflowError. 
        - trace_id field should only contain a 64bit integer and the remaining bits are stored in the `_dd.p.tid` tag.
  - 128bit trace ids are propagated by Datadog distributing tracing headers.
        - Ensure the full 128bit trace id propagated and reconstructed by downstream services.
  - Ensure Spans with 128bit trace ids are sampled at the expected rate

### Performance Testing

1. There is no performance regression when 128bit trace id generation is disabled. This is the default mode.
2. When 128bit trace id generation is enabled there is a ~60ns increase (578ns -> 640ns, 10%) to span creation (`Span.__init__(name, ......)`). This performance regression does not appear avoidable.
   - This overhead was measured on M1 using python 3.8. Results vary across platforms and python versions.

## Next Steps

1. Support 128bit trace id propagation in b3 and w3c headers
   - The code change in straight forward (ie avoid truncating incoming trace ids) but this change will require a significant refactor of existing tests.
2. Add support for the `DD_TRACE_128_BIT_TRACEID_LOGGING_ENABLED` environment variable 
   - This supports logging 64bit trace ids even when `DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED=true`
4. Enable 128bit trace id on a sample application (ex: internal services using the ddtrace library)
   - Ensure sampling rates are respected
   - Ensure logs correlation works for traces with 128bit trace ids (this is a concern raised in the RFC)
   - Ensure distributed tracing works across tracers that only support 64bit trace ids
   - Ensure the full 128bit trace id is reconstructed and is viewable in the Datadog product
5. Enable support for 128bit trace id by default and add public documentation.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/contrib/logging/patch.py
```diff
@@ -73,7 +74,13 @@ def _w_makeRecord(func, instance, args, kwargs):
...
 
     if span:
-        setattr(record, RECORD_ATTR_TRACE_ID, str(span.trace_id))
+        if span.trace_id > _MAX_UINT_64BITS and not ddtrace.config._128_bit_trace_logging_enabled:
+            # Ensure 64 bit trace id is logged (instead of the 64bit version)
+            trace_id = str(span._trace_id_64bits)
+        else:
+            trace_id = str(span.trace_id)
+
+        setattr(record, RECORD_ATTR_TRACE_ID, str(trace_id))
```
mabdinur, line 83: I plan on reverting this change. 

Logs ingestion will break if a 128bit trace id is set. As a workaround `DD_TRACE_128_BIT_TRACEID_LOGGING_ENABLED` was defined and will be set for False for the foreeable future. When logs ingestion is compatible with 128bit trace ids then we can remove this flag and log the full 128bit trace id. I am not sure if we want this hack but I left it in for now.

mabdinur, line 83:
```suggestion
```

I plan on reverting this change.

Logs ingestion will break if a 128bit trace id is set. As a workaround DD_TRACE_128_BIT_TRACEID_LOGGING_ENABLED was defined and will be set for False for the foreeable future. When logs ingestion is compatible with 128bit trace ids then we can remove this flag and log the full 128bit trace id. I am not sure if we want this hack but I left it in for now.

mabdinur:
> ## Benchmarks
> Comparing candidate commit [9c5a9e6](https://github.com/DataDog/dd-trace-py/commit/9c5a9e6362eea6c977a35e0b41fe7910628cf348) in PR branch `munir/implement-128-bit-trace-ids` with baseline commit [caca9d3](https://github.com/DataDog/dd-trace-py/commit/caca9d3b875d1a6d4d3708d948477d0df31572a9) in branch `1.x`.
> 
> Found 0 performance improvements and 2 performance regressions! Performance is the same for 54 cases.
> 
> #### scenario:span-start
> * 🟥 `execution_time` [+0.228ms; +0.260ms] or [+9.682%; +11.041%]
> 
> #### scenario:span-start-finish
> * 🟥 `execution_time` [+0.224ms; +0.252ms] or [+7.558%; +8.513%]

The only change I made to `Span.__init__` was replacing `self.trace_id = trace_id or _rand.rand64bits()`  with `self.trace_id = trace_id or _generate_trace_id()`. I benchmarked `_generate_trace_id()` locally and got the following results:

```
python3 -m timeit -s "from ddtrace.span import _generate_trace_id"  "_generate_trace_id()"
5000000 loops, best of 5: 93.6 nsec per loop
```

```
DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED=true python3 -m timeit -s "from ddtrace.span import _generate_trace_id"  "_generate_trace_id()"
1000000 loops, best of 5: 218 nsec per loop
```

Generating a 128bit trace id has an overhead of ~100nsec. Is this acceptable?

ddtrace/span.py
```diff
@@ -52,12 +52,14 @@
 log = get_logger(__name__)
 
 
-def _generate_trace_id():
+def _generate_trace_id():  # total time -> 198 nsec
     # type: () -> int
-    if config._128_bit_trace_id_enabled:
+    rand_int64 = _rand.rand64bits()  # generating random int -> 36.7ns
+    if config._128_bit_trace_id_enabled:  # 24 ns
+        unix_time_sec = int(time.time())  # getting unix time in seconds -> 96.7ns
```
mabdinur, line 59: time.time() -> ~32ns
int(........) -> ~50ns

ddtrace/sampler.py
```diff
@@ -44,8 +45,8 @@
 
 log = get_logger(__name__)
 
-MAX_TRACE_ID = 2 ** 64
-
+# deprecare and remove MAX_TRACE_ID
```
mabdinur, line 48:
```suggestion
# deprecate and remove MAX_TRACE_ID
```

ddtrace/settings/config.py
```diff
@@ -224,6 +224,10 @@ def __init__(self):
 
         self._telemetry_metrics_enabled = asbool(os.getenv("_DD_TELEMETRY_METRICS_ENABLED", default=False))
 
+        self._128_bit_trace_id_enabled = asbool(os.getenv("DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED", False))
+
+        self._128_bit_trace_logging_enabled = asbool(os.getenv("DD_TRACE_128_BIT_TRACEID_LOGGING_ENABLED", False))
+
```
mabdinur, line 230:
```suggestion
```

ddtrace/contrib/logging/patch.py
```diff
@@ -4,6 +4,7 @@
 
 import ddtrace
 
+from ...internal.constants import _MAX_UINT_64BITS
```
mabdinur, line 7:
```suggestion
```

ZStriker19:
What if, going forward we:
1. basically deprecate 64 bit trace ids and pad them with 0s to create 128 bit trace ids
2. change the encoders and agent to support 128 bit trace ids, and use those trace ids if available
3. encode and send to the agent with both 128 and 64 bit trace ids for backwards compatibility. Like for propagation `"x-datadog-trace-id_long": "<128bit trace id>", "x-datadog-trace-id":"64_bit_trace_id"` 

There might be a few other steps here, but I guess my reasoning for going about it this way would be that, if we're going to make 128 bit the standard way going forward, wouldn't this be a simpler step in that direction?

mabdinur:
> What if, going forward we:
> 
> 1. basically deprecate 64 bit trace ids and pad them with 0s to create 128 bit trace ids
> 2. change the encoders and agent and backend to support 128 bit trace ids, and use those 128 bit trace ids if available
> 3. encode and send to the agent with both 128 and 64 bit trace ids for backwards compatibility. Like for propagation `"x-datadog-trace-id_long": "<128bit trace id>", "x-datadog-trace-id":"64_bit_trace_id"`
> 
> There might be a few other steps here, but I guess my reasoning for going about it this way would be that, if we're going to make 128 bit the standard way going forward, wouldn't this be a simpler step in that direction?

Updating the agent and encoders to support 128bit trace ids is not trivial. 
 1. Msgpack does not support 128bit integers [issue-206](https://github.com/msgpack/msgpack/issues/206). We will need to encode the trace id as a string.
    - This will require creating a new agent endpoint that supports decoding 128bit trace ids. The agent is released every 6 weeks, so this change will not be released for awhile. 
 3. There is limit to the size and number of http headers that sent to some servers (I believe aws has limit of 10 headers for some services). Adding more datadog http headers is a deal breaker for some users.
 4. Making 128bit trace ids the default format in this PR is risky. We should ideally ship this feature behind a feature flag and work out the kinks over a few minor releases. Then we should deprecate/remove support for generating 64bit trace ids. 
 5. Trace ids are encoded as integers. We can't pad an integer with zeros (but we can do this for distributed tracing headers but this is not backwards compatible).

In my opinion the current approach provides a straight forward implementation while maintaining backwards compatibility.

ddtrace/propagation/http.py
```diff
@@ -270,7 +301,7 @@ def _extract(headers):
             log.debug(
                 (
                     "received invalid x-datadog-* headers, "
-                    "trace-id: %r, parent-id: %r, priority: %r, origin: %r, tags:%r"
+                    "trace-id: %s, parent-id: %r, priority: %r, origin: %r, tags:%r"
```
mabdinur, line 304:
```suggestion
                    "trace-id: %d, parent-id: %r, priority: %r, origin: %r, tags:%r"
```

mabdinur, line 304:
```suggestion
                    "trace-id: %r, parent-id: %r, priority: %r, origin: %r, tags:%r"
```

ddtrace/internal/processor/trace.py
```diff
@@ -130,6 +133,21 @@ def process_trace(self, trace):
...
+    """Processor that applies trace-level tags to the trace."""
+
+    def process_trace(self, trace):
+        # type: (List[Span]) -> List[Span]
+        if not trace or trace[0].trace_id <= _MAX_UINT_64BITS:
+            return trace
+
+        trace_id_hob = get_hob_from_128bit_traceid_as_hex(trace[0].trace_id)
+        for span in trace:
+            span._meta[_HIGHER_ORDER_TRACE_ID_BITS] = trace_id_hob
```
brettlangdon, line 147: oof.... can we have the agent or intake expand this tag to all spans instead of us doing it?

Seems weird we need to copy everywhere and not just sending once per-payload per-trace.

ddtrace/span.py
```diff
@@ -176,6 +183,10 @@ def _get_ctx_item(self, key):
             return None
         return self._store.get(key)
 
+    @property
+    def _trace_id_64bits(self):
+        return _MAX_UINT_64BITS & self.trace_id
```
brettlangdon, line 188:
```suggestion
        return convert_128bit_traceid_to_64bit(self.trace_id)
```

?

ddtrace/internal/utils/http.py
```diff
@@ -209,3 +210,13 @@ def w3c_encode_tag(args):
     tag_val = pattern.sub(replacement, tag_val)
     # replace = with ~ if it wasn't already replaced by the regex
     return tag_val.replace("=", "~")
+
+
+def convert_128bit_traceid_to_64bit(trace_id):
```
brettlangdon, line 215: since the other helper is getting the higher order 64-bits, should this be named to get the lower order 64-bits ?

ddtrace/propagation/http.py
```diff
@@ -197,11 +210,17 @@ def _inject(span_context, headers):
     @staticmethod
     def _extract(headers):
         # type: (Dict[str, str]) -> Optional[Context]
-        trace_id = _extract_header_value(
-            POSSIBLE_HTTP_HEADER_TRACE_IDS,
-            headers,
+        trace_id = int(
```
brettlangdon, line 213: the try/except from below is now removed for this.

we should be sure we have test cases that would cause this to fail (I am surprised no tests are failing).

ddtrace/propagation/http.py
```diff
@@ -197,11 +210,17 @@ def _inject(span_context, headers):
...
+            _extract_header_value(
+                POSSIBLE_HTTP_HEADER_TRACE_IDS,
+                headers,
+            )
+            or 0
         )
-        if trace_id is None:
+        if trace_id == 0 or trace_id > _MAX_UINT_64BITS:
+            log.debug(
+                "Invalid trace id: %d. `x-datadog-trace-id` must be greater than zero and less than 2**64 - 1", trace_id
```
brettlangdon, line 222: is `2**64 - 1` the max allowed?

does the message need to say "and less than 2**64" ? (since 2**64 - 1 should be inclusive)

ddtrace/span.py
```diff
@@ -137,7 +144,7 @@ def __init__(
         self.duration_ns = None  # type: Optional[int]
 
         # tracing
-        self.trace_id = trace_id or _rand.rand64bits()  # type: int
+        self.trace_id = trace_id or _generate_trace_id()  # type: int
```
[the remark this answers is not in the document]
mabdinur: I considered this however it adds a lot of complexity and some (but minimal) performance overhead (~4-8ns) when trace_id is accessed. I tried adding the following properties `Span._trace_id64lower` and `Span._trace_id64upper`. I converted Span.trace_id in to a derived property. When `trace_id` was accessed I used bitwise operations to combine `_trace_id64lower` and `_trace_id64upper` into one integer. However in this design `_trace_id64upper` was never used directly. The more performant approach was to generate one 128bit trace_id and convert it into hex directly (higher order bits are never sent as an integer). 

Another approach is to cache all three ids. But in this scenario if `trace_id`, `_trace_id64upper`, or `_trace_id64lower` is updated we will need to ensure all the values in sync. The additional complexity does not appear to be worth it. 

All in all, `Span.trace_id` is accessed in many places but `Span._trace_id64lower` is only accessed when a trace is encoded and `Span._trace_id64upper` will never be used directly. I think the current approach has the best performance and is the easiest to maintain. Thoughts?

ddtrace/propagation/http.py
mabdinur: Yup, I'll fix this 2**64 - 1 is a valid traced id (since trace ids are unsigned)

ddtrace/internal/processor/trace.py
mabdinur: We could make this a "trace level" tag (ie add this tag to span.context.meta), but I wasn't sure if there were edge cases where `Span.context is None` or if the trace id will be reconstructed by the backend/agent before the trace level tags are set. 

I'll ask for clarification on this.

ddtrace/propagation/http.py
mabdinur: Do you mean this line: https://github.com/DataDog/dd-trace-py/blob/v1.10.1/ddtrace/propagation/http.py#L260?
`0 or None` evaluates to `None` so this behavior is unchanged.

brettlangdon: what if someone sends `X-Datadog-Trace-Id: peanuts` ?

mabdinur: That would not be fun 😆

mabdinur: I'll revert this change

ddtrace/internal/processor/trace.py
brettlangdon: having it accessible from the Span is probably fine, I am just wondering if we can reduce the duplicating of the data on every single span, especially since it is a tag.

there should be a way to have this expansion applied downstream to save us the computational overhead/bytes on the wire.

ddtrace/span.py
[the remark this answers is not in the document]
brettlangdon: I think that reasoning makes a lot of sense based on the access pattern. if the full trace id is what is accessed majority of the time then 👍🏻

ddtrace/propagation/http.py
brettlangdon: (well, or let's be sure we have a failing test for that to validate that my assumption is correct, or that we handle that case)

ddtrace/propagation/http.py
```diff
@@ -197,11 +210,16 @@ def _inject(span_context, headers):
     @staticmethod
     def _extract(headers):
         # type: (Dict[str, str]) -> Optional[Context]
-        trace_id = _extract_header_value(
-            POSSIBLE_HTTP_HEADER_TRACE_IDS,
-            headers,
-        )
-        if trace_id is None:
+        trace_id_str = _extract_header_value(POSSIBLE_HTTP_HEADER_TRACE_IDS, headers) or ""
```
mabdinur, line 213: clean up

ddtrace/propagation/http.py
```diff
@@ -246,6 +264,20 @@ def _extract(headers):
                 }
                 log.debug("failed to decode x-datadog-tags: %r", tags_value, exc_info=True)
 
+        if config._128_bit_trace_id_enabled:
```
mabdinur, line 267: nit: check if meta is None

ddtrace/propagation/http.py
```diff
@@ -218,7 +236,7 @@ def _extract(headers):
             headers,
         )
 
-        meta = None
+        meta = {}
```
mabdinur, line 239: revert

ddtrace/sampler.py
```diff
@@ -44,8 +45,8 @@
 
 log = get_logger(__name__)
 
-MAX_TRACE_ID = 2 ** 64
-
+# MAX_TRACE_ID is no longer used. It should deprecated and removed.
+MAX_TRACE_ID = _MAX_UINT_64BITS
```
mabdinur, line 49: Keep max trace id

ddtrace/propagation/http.py
```diff
@@ -218,7 +238,7 @@ def _extract(headers):
...
```
mabdinur, line 241:
```suggestion
        meta = None
```

ddtrace/propagation/http.py
```diff
@@ -154,7 +158,16 @@ def _inject(span_context, headers):
             log.debug("tried to inject invalid context %r", span_context)
             return
 
-        headers[HTTP_HEADER_TRACE_ID] = str(span_context.trace_id)
+        if span_context.trace_id > _MAX_UINT_64BITS:
+            # set lower order 64 bits in `x-datadog-trace-id` header. For backwards compatibility these
+            # bits should be converted to a base 10 integer.
+            headers[HTTP_HEADER_TRACE_ID] = str(get_64_lowest_order_bits_as_int(span_context.trace_id))
+            # set higher order 64 bits in `_dd.p.tid` to propagate the full 128 bit trace id.
```
ZStriker19, line 165: We agreed upon this across tracers? Just want to make sure for propagation

ddtrace/sampler.py
```diff
@@ -44,8 +45,8 @@
 
 log = get_logger(__name__)
 
-MAX_TRACE_ID = 2 ** 64
-
+# MAX_TRACE_ID is no longer used. It should deprecated and removed.
```
ZStriker19, line 48: should we deprecate now, or leave for later?

ddtrace/settings/config.py
```diff
@@ -225,6 +225,8 @@ def __init__(self):
 
         self._telemetry_metrics_enabled = asbool(os.getenv("_DD_TELEMETRY_METRICS_ENABLED", default=False))
 
+        self._128_bit_trace_id_enabled = asbool(os.getenv("DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED", False))
```
ZStriker19, line 228: For b3 and w3c propagation, we just continue to use the lower 64 bit trace ids correct? Just want to make sure.

ddtrace/propagation/http.py
mabdinur: Yup, this is defined in the RFC and already implemented in Ruby

ddtrace/sampler.py
mabdinur: I rather do it in a follow up. This PR is already getting a bit long. I'll add a `TODO` to the comment.

mabdinur, line 48:
```suggestion
# TODO: deprecate and remove MAX_TRACE_ID, it is no longer used.
```

ddtrace/settings/config.py
mabdinur: I called this out in the next steps. I plan to address this in my next PR. We will only propagate the 128bit trace ids for w3c and b3 if `DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED=True`. Otherwise we'll continue to truncate it and use the lower 64 bits

ZStriker19 (approved):

Kyle-Verhoog (changes requested):
Did a quick pass and noticed a few things.

There should be documentation for `DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED`.

ddtrace/span.py
```diff
@@ -50,6 +51,12 @@
 log = get_logger(__name__)
 
 
+if config._128_bit_trace_id_enabled:
```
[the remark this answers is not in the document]
mabdinur: Ya I am not a big fan either but this is a performance optimization. We can discuss this offline

tests/tracer/test_writer.py
```diff
@@ -202,7 +202,7 @@ def test_drop_reason_buffer_full(self):
         statsd = mock.Mock()
         writer_metrics_reset = mock.Mock()
         with override_global_config(dict(health_metrics_enabled=False)):
-            writer = AgentWriter(agent_url="http://asdf:1234", buffer_size=5235, dogstatsd=statsd)
```
mabdinur, line 205: https://github.com/DataDog/dd-trace-py/pull/5326/commits/a4136ae4447aaa797efcd34510b023a61cf90c03 changes the size of the span object which causes the following [test failure](https://circleci.com/gh/DataDog/dd-trace-py/2139627). This fix is required to resolve the failing test.

ddtrace/internal/utils/http.py
```diff
@@ -209,3 +210,15 @@ def w3c_encode_tag(args):
     tag_val = pattern.sub(replacement, tag_val)
     # replace = with ~ if it wasn't already replaced by the regex
     return tag_val.replace("=", "~")
+
+
+def get_64_lowest_order_bits_as_int(large_int):
```
Kyle-Verhoog, line 215: these don't really make sense as `http` utilities, maybe make more sense in `span` or in `tracer` since they pertain to the implementation of trace ids specifically

docs/configuration.rst
```diff
@@ -66,6 +66,13 @@ The following environment variables for the tracer are supported:
      description: |
          Enables sending :ref:`telemetry <Instrumentation Telemetry>` events to the agent.
 
+   DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED:
+     type: Boolean
+     default: False
+     description: |
+         Historically datadog trace ids were 64 bit integers. This configuration enables the generation of 128 bit trace ids to better align
```
Kyle-Verhoog, line 73:
```suggestion
         This configuration enables the generation of 128 bit trace ids to better align
```

we don't need the context of the configuration option, just what it does and any limitations (like lack of b3/w3c propagation support)

ddtrace/sampler.py
```diff
@@ -44,8 +45,8 @@
 
 log = get_logger(__name__)
 
-MAX_TRACE_ID = 2 ** 64
-
+# TODO: deprecate and remove MAX_TRACE_ID, it is no longer used.
```
Kyle-Verhoog, line 48:
```suggestion
```

Kyle-Verhoog: also we're subtly changing this value now (from 2^64 to 2^64-1)

ddtrace/sampler.py
```diff
@@ -101,11 +102,11 @@ def __init__(self, sample_rate=1.0):
     def set_sample_rate(self, sample_rate):
         # type: (float) -> None
         self.sample_rate = float(sample_rate)
-        self.sampling_id_threshold = self.sample_rate * MAX_TRACE_ID
+        self.sampling_id_threshold = self.sample_rate * _MAX_UINT_64BITS
```
Kyle-Verhoog, line 105: are these changes all unnecessary since we're keeping MAX_TRACE_ID? can they be reverted?

ddtrace/sampler.py
```diff
@@ -44,8 +45,8 @@
 
 log = get_logger(__name__)
 
-MAX_TRACE_ID = 2 ** 64
-
+# TODO: deprecate and remove MAX_TRACE_ID, it is no longer used.
+MAX_TRACE_ID = _MAX_UINT_64BITS
```
Kyle-Verhoog, line 49: this constant is a little bit odd now. Trace IDs can technically be 128 bits now but this values will only be 64 bits. Maybe worth a comment about why it remains 64 bits even though 128 bit support is present?

ddtrace/internal/constants.py
```diff
@@ -18,6 +18,8 @@
 DEFAULT_SERVICE_NAME = "unnamed_python_service"
 # Used to set the name of an integration on a span
 COMPONENT = "component"
+_HIGHER_ORDER_TRACE_ID_BITS = "_dd.p.tid"
```
Kyle-Verhoog, line 21:
```suggestion
HIGHER_ORDER_TRACE_ID_BITS_KEY = "_dd.p.tid"
```

doesn't have to be internal since the module is internal

ddtrace/internal/constants.py
```diff
@@ -18,6 +18,8 @@
...
+_MAX_UINT_64BITS = (1 << 64) - 1
```
Kyle-Verhoog, line 22:
```suggestion
MAX_UINT_64BITS = (1 << 64) - 1
```
same here

ddtrace/internal/processor/trace.py
```diff
@@ -127,6 +130,10 @@ def process_trace(self, trace):
 
         ctx._update_tags(chunk_root)
         chunk_root.set_tag_str("language", "python")
+        # for 128 bit trace ids
+        if chunk_root.trace_id > _MAX_UINT_64BITS:
```
Kyle-Verhoog, line 134: this check bothers me a bit because it relies on the implementation of the "random" 128 bit trace id is generated. Since it uses the unix time stamp (a non-zero value) we're guaranteed that the generated 128bit id is > _MAX_UINT_64BITS. If implementation were to change and we truly generated random 128 bit trace ids then this check would not hold and this code would be broken.

I'd like to see a stronger variant that we use to establish that a span is using 128-bit ids. This could be done if we upfront stored the 128 trace id as a tag in the span on span creation and keep `span.trace_id` as just the 64 bit portion. Then here we would check the presence of that tag on the span.

WDYT @mabdinur?

Kyle-Verhoog: note that this also might be more performant as it would be less costly to access the 64 bit trace id. We also wouldn't have to change all the places that we reference `trace_id` in our code. Also also, this would not be a breaking change for users who are assuming that trace ids are 64 bits.

We could then introduce a backwards compatible `span.trace_id_128` to get the 128 bit id

ddtrace/propagation/http.py
```diff
@@ -154,7 +158,16 @@ def _inject(span_context, headers):
             log.debug("tried to inject invalid context %r", span_context)
             return
 
-        headers[HTTP_HEADER_TRACE_ID] = str(span_context.trace_id)
+        if span_context.trace_id > _MAX_UINT_64BITS:
```
Kyle-Verhoog, line 161: same as my above comment, if we take any action there

ddtrace/propagation/http.py
```diff
@@ -197,11 +210,18 @@ def _inject(span_context, headers):
...
+        trace_id_str = _extract_header_value(POSSIBLE_HTTP_HEADER_TRACE_IDS, headers)
+        if trace_id_str is None:
+            return None
+        try:
+            trace_id = int(trace_id_str)
+        except ValueError:
+            trace_id = 0
+
+        if trace_id == 0 or trace_id > _MAX_UINT_64BITS:
+            log.debug(
```
Kyle-Verhoog, line 222:
```suggestion
            log.warning(
```

probably want to warn here since this would be broken trace propagation

ddtrace/propagation/http.py
```diff
@@ -246,6 +266,20 @@ def _extract(headers):
...
+            # and set in the `x-datadog-trace-id` header (this was done for backwards compatibility).
+            # The 64 highest order bits are encoded in base 16 and store in the `_dd.p.tid` tag.
+            # Here we reconstruct the full 128 bit trace_id.
+            trace_id_hob_hex = meta.get(_HIGHER_ORDER_TRACE_ID_BITS)  # type: Optional[str]
+            if trace_id_hob_hex is not None:
+                # convert lowest order bits in trace_id to base 16
+                trace_id_lod_hex = "{:016x}".format(trace_id)
+                # combine highest and lowest order hex values to create a 128 bit trace_id
+                trace_id = int(trace_id_hob_hex + trace_id_lod_hex, 16)
+                # After the full trace id is reconstructed this tag is no longer required
```
Kyle-Verhoog, line 280: but we eventually set it again when we encode the trace?

ddtrace/internal/utils/http.py
mabdinur: I added these here because these utils were mainly used in propagation. But that changed so now it makes no sense. I’ll make this change

ddtrace/sampler.py
mabdinur: Yup I think 2**64 is incorrect. MAX TRACE ID should be equal to max uint64

ddtrace/internal/processor/trace.py
mabdinur: Even if we generate a traceid of say 10 when 128bit trace if generation is enabled we don’t need to set or extract from the tag containing higher order bits. We only need to set and extract _dd.p.tid if a trace id is greater than 2**64 - 1. 

Trace_id is referenced in many places in our library. If we introduce two interfaces for setting and accessing a trace ids (ex: traceid and traceid128) it would significantly increase the complexity of using Spans. The current approach was designed with this in mind. Also the higher order bits tags should not be set on every span. It should only be set on the chunk root span. This was defined in the RFC

mabdinur: TLDR this code won’t break even if we generate a 64bit trace id when DD_TRACE_128_Traceid_generation=true.

ddtrace/propagation/http.py
mabdinur: I made this change because I can see a scenario where the trace id in the context is manually overriden. Ie a user gets the current active context and does context.trace_id = 2**69. In this scenario we’ll keep a reference to a stale meta tag. Which isn’t a big deal but could lead to cryptic bugs if the code was refactored in the future.

ddtrace/sampler.py
mabdinur: The plan is to remove max trace id since it's misleading. With this change the max trace id is 2**128 -1

docs/configuration.rst
```diff
@@ -66,6 +66,12 @@ The following environment variables for the tracer are supported:
      description: |
          Enables sending :ref:`telemetry <Instrumentation Telemetry>` events to the agent.
 
+   DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED:
+     type: Boolean
+     default: False
+     description: |
+         This configuration enables the generation of 128 bit trace ids to better align with industry standards (ex: `OpenTelemetry <https://opentelemetry.io/docs/reference/specification/trace/api/>`_).
```
Kyle-Verhoog, line 73:
```suggestion
         This configuration enables the generation of 128 bit trace ids.
```

again, not worth saying _why_ more about what it is. Is it worth mentioning b3/w3c support isn't included yet, or is that out of scope of this setting?

riotfile.py
```diff
@@ -273,6 +273,15 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
             command="pytest {cmdargs} tests/tracer/ --ignore=tests/tracer/test_http.py",
             venvs=[
                 Venv(pys=select_pys()),
+                # This test variant ensures all tracer tests are compatible with 128bit trace ids.
+                # Once DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED is set to True by default this test variant
+                # should be removed (or DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED should be set to false).
+                Venv(
+                    pys=select_pys(),
+                    env={
+                        "DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED": "True",
```
Kyle-Verhoog, line 282: this means we're only running the http tests with `DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED=True` is that intentional?

should it be 
```suggestion
                        "DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED": ["", "True"],
```
?

ddtrace/internal/processor/trace.py
Kyle-Verhoog: To clarify, by "broken" I mean that we won't set the HOB tag in traces and then any assumptions around that would break eg. if the backend relies on that tag to determine if 128 bit trace ids are being generated. This is a weak example though. I just want to make sure that this is a safe condition to use.

👍 makes sense that the logic holds when _using_ the trace id.

Kyle-Verhoog (commented):
I think we should have a release note for this since we're adding a public setting

ddtrace/internal/utils/http.py
```diff
@@ -209,3 +210,15 @@ def w3c_encode_tag(args):
     tag_val = pattern.sub(replacement, tag_val)
     # replace = with ~ if it wasn't already replaced by the regex
     return tag_val.replace("=", "~")
+
+
+def _get_64_lowest_order_bits_as_int(large_int):
```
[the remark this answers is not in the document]
mabdinur: This is a weird utility, it is used to get the lowest order bits from Span.tracer_id (using in sampling, encoding) and Context.trace_id (used to send/parse distributed trace headers).

riotfile.py
mabdinur: Yes by default this configuration is false. I avoided this change due to the ci costs.

 It is not explicit but all the other tracer tests run with `"DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED": "false"`

mabdinur:
> To clarify, by "broken" I mean that we won't set the HOB tag in traces and then any assumptions around that would break eg. if the backend relies on that tag to determine if 128 bit trace ids are being generated. This is a weak example though. I just want to make sure that this is a safe condition to use.
> 
> 👍 makes sense that the logic holds when using the trace id.
> 

The RFC explictly states that the `_dd.p.tid` tag should not be set if there are no higher order bits. So this is inline with the spec. I think this condition is safe to use.

ddtrace/internal/processor/trace.py
mabdinur: The RFC states that the `_dd.p.tid` tag should only be set if there are higher order bits. So this is inline with the spec. I think it's safe condition to use.

ddtrace/sampler.py
Kyle-Verhoog: should we deprecate it right away since it's misleading now? Would be useful to the user to know that it is too

ddtrace/internal/processor/trace.py
Kyle-Verhoog: :+1: 👍

ddtrace/internal/utils/http.py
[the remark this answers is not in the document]
Kyle-Verhoog: right now there are two definitions of the function why not import the one from span here? Where it lives is not a blocker for me (so long as it's internal). It is a blocker that we have the same function in two places

riotfile.py
Kyle-Verhoog: Different code paths are taken so I think it's worth running both. CI cost can be addressed if it becomes an issue.

mabdinur, line 282:
```suggestion
                        "DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED": ["false", "true"],
```

releasenotes/notes/add-dd-128bit-traceid-releasenote-dae1a746134859bb.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    tracing: Adds support for 128 bit trace ids. To generate and propagate 128 bit trace ids using datadog distributed tracing headers set the following configuration: ``DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED=True``.   Support for B3 and W3C distributed tracing headers will be added in a future change.
```
mabdinur, line 4:
```suggestion
    tracing: Adds support for 128 bit trace ids. To generate and propagate 128 bit trace ids using datadog distributed tracing headers set the following configuration: ``DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED=True``. Support for B3 and W3C distributed tracing headers will be added in a future change.
```

ddtrace/internal/utils/http.py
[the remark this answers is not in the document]
mabdinur: Oh I completely missed that. Sorry, I thought this was the span module. I will fix this

ddtrace/propagation/http.py
```diff
@@ -154,7 +158,16 @@ def _inject(span_context, headers):
...
             return
 
-        headers[HTTP_HEADER_TRACE_ID] = str(span_context.trace_id)
+        if span_context.trace_id > _MAX_UINT_64BITS:
+            # set lower order 64 bits in `x-datadog-trace-id` header. For backwards compatibility these
+            # bits should be converted to a base 10 integer.
+            headers[HTTP_HEADER_TRACE_ID] = str(_get_64_lowest_order_bits_as_int(span_context.trace_id))
+            # set higher order 64 bits in `_dd.p.tid` to propagate the full 128 bit trace id.
+            # Note - The higher order bits must be encoded in hex
+            span_context._meta[_HIGHER_ORDER_TRACE_ID_BITS] = _get_64_highest_order_bits_as_hex(span_context.trace_id)
```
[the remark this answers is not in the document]
mabdinur: ddtrace.Context does not have a public api for settting and removing tags.  Using `ddtrace.Context._meta` is consistent with how we currently set/extract distributed tracing tags.

Kyle-Verhoog (approved):
probably also worth having some snapshot tests with 128-bit trace ids enabled but this would be a nice to have.

the rest of my comments are non-blocking. The buffer test case needing to be updated is a bit concerning and might be worth getting an answer to @mabdinur.

👏 on getting this over the finish line 🏁

tests/tracer/test_propagation.py
```diff
@@ -185,6 +217,48 @@ def test_extract(tracer):
...
 
 
+@pytest.mark.subprocess(
+    env=dict(DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED="true"),
+    parametrize={
+        "TEST_TRACE_ID": [
+            str(2 ** 128 - 1),
+            str(2 ** 127 + 1),
+            str(2 ** 65 - 1),
+            str(2 ** 64 + 1),
```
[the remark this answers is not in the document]
mabdinur: `2**64` is an interesting case. Here the higher order bits will be `0000000000000001` and the trace_id_64bit will be zero. trace_id_64bit can not be zero, this will break `x-datadog-trace-id` headers in older versions of the tracer.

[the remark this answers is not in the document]
mabdinur: That's why I used `2 ** 64 + 1` instead. Here the higher order bits will be set to `0000000000000001` and the trace_id_64bit will be `1`

tests/tracer/test_writer.py
```diff
@@ -202,7 +202,7 @@ def test_drop_reason_buffer_full(self):
...
+            writer = AgentWriter(agent_url="http://asdf:1234", buffer_size=5125, dogstatsd=statsd)
```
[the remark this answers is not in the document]
mabdinur: I did some digging the actual size difference is a decrease in span size by 20 bytes. This is consistent whether this test generates 20 spans or 40. So it makes sense that we see this behavior even when this feature is disabled.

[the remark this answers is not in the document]
mabdinur: The `_rand`  module is no longer referenced in `Span.__init__()`. Maybe the missing 20 bytes involved referencing that module in the Span.__code__ attribute? That's just a guess. I will look into this a future PR.

Kyle-Verhoog (approved):

ddtrace/sampler.py
Yun-Kim: Can we have a comment here explaining this? Thanks!

riotfile.py
```diff
@@ -273,6 +273,15 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
             command="pytest {cmdargs} tests/tracer/ --ignore=tests/tracer/test_http.py",
             venvs=[
                 Venv(pys=select_pys()),
```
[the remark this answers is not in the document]
mabdinur: Yup, Kyle suggested that we be explicit. I can run these tests for only python3.11, that should reduce the impact.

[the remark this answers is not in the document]
Yun-Kim: I meant that there's two `Venvs`, one that runs all tests, and the one you introduced which runs all tests twice (with the env variable set to True/False), we can just remove the original Venv as I suggested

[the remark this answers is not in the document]
Yun-Kim: Does this only apply for Python 3.11?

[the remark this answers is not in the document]
Yun-Kim: Sounds good then, I'm happy to just run with Python 3.11.

Yun-Kim (approved):

PR #5335, opened 2023-03-15
fix(aws_lambda): patch callable object handlers properly

### LINKED ISSUE #5284: The aws_lambda patch module can't handle callable object event handlers
### Summary of problem
NOTE: I'll be cross-posting this in [datadog-lambda-python](https://github.com/DataDog/datadog-lambda-python) project because I'm not sure whether this is in the domain of the Lambda layer or this library itself. This could either be solved in this library by checking if the function being wrapped is a callable class instance, or by the Lambda layer passing the `__call__` function reference instead of the callable class.

The issue stems from ddtrace (specifically the `aws_lambda` module) trying to patch the outer most Lambda handler functions, which can be any callable that accepts an event and context arguments, so either a standard function or a callable object, like a class instance that defines `__call__`.

The internal wrapper is passed the class instance itself instead of the `__call__` function however, so as a result, it fails as a class instance isn't going to have a `__code__` property. The `patch_all` the Lambda layer is executing is in a try/catch block and only runs once, so there's no errors being raised, but on every cold start there's a stack trace popping up in the logs. 

Platform: AWS Lambda
Runtime: Python 3.9
Serverless Framework Version: 3.28.0
Serverless Plugin Version: 5.20.0
DataDog Lambda Layer Version: 68
ddtrace Version: 1.9.1

ddtrace@1.9.1

pip@21.2.4

Output of `poetry show --no-dev`, excluded some internal libraries irrelevant to this and added `chalice` since that's packaged manually using their internal packaging mechanism.

attrs@21.4.0
aws-lambda-context@1.1.0
boto3@1.24.57
botocore@1.27.57
bytecode@0.14.0
cattrs@22.2.0
certifi@2022.6.15
charset-normalizer@2.1.1
ddsketch@2.0.4
ddtrace@1.9.1
elasticsearch@7.9.1
elasticsearch-dsl@7.3.0
envier@0.4.0
exceptiongroup@1.1.0
idna@3.3
jmespath@0.10.0
jsonschema@3.2.0
marshmallow@3.15.0
packaging@23.0
protobuf@4.22.0
pyrsistent@0.18.1
python-dateutil@2.8.2
python-json-logger@2.0.4
requests@2.28.1
requests-aws4auth@1.0
s3transfer@0.6.0
simplejson@3.17.2
six@1.16.0
tenacity@8.2.2
typing-extensions@4.3.0
urllib3@1.26.12
xmltodict@0.13.0
chalice@1.26.6

Deploy a lambda function where the event handler is a callable class instance, whether through the [chalice](https://github.com/aws/chalice) framework or something like this:
```
# app.py
from typing import Any

class LambdaHandler:
    def __call__(self, event: dict[str, Any], context: Any) -> dict[str, Any]:
        return {"hello": "world"}

app = LambdaHandler()
```
With `app.app` being the handler of the lambda function.

The following stack trace is thrown on every cold start as `patch_all` fails:
```
	File "/opt/python/lib/python3.9/site-packages/datadog_lambda/wrapper.py", line 177, in __init__
		patch_all()
	File "/opt/python/lib/python3.9/site-packages/datadog_lambda/patch.py", line 36, in patch_all
		patch_all_dd()
	File "/var/task/ddtrace/_monkey.py", line 190, in patch_all
		patch(raise_errors=False, **modules)
	File "/var/task/ddtrace/_monkey.py", line 230, in patch
		when_imported(module)(_on_import_factory(contrib, raise_errors=False))
	File "/var/task/ddtrace/vendor/wrapt/importer.py", line 284, in register
		register_post_import_hook(hook, name)
	File "/var/task/ddtrace/vendor/wrapt/decorators.py", line 470, in _synchronized
	File "/var/task/ddtrace/vendor/wrapt/importer.py", line 82, in register_post_import_hook
		hook(module)
	File "/var/task/ddtrace/_monkey.py", line 155, in on_import
		imported_module.patch()
	File "/var/task/ddtrace/contrib/aws_lambda/patch.py", line 156, in patch
		wrap(handler, wrapper)
	File "/var/task/ddtrace/internal/wrapping.py", line 428, in wrap
		f.__code__,
	AttributeError: 'EventSourceHandler' object has no attribute '__code__'
```

This operation should not fail, and either [ddtrace.internal.wrapping.wrap](https://github.com/DataDog/dd-trace-py/blob/1.x/ddtrace/internal/wrapping.py#L417) should account for callable classes, or `datadog_lambda` should be figuring out that the handler is a callable class instance and passing the `__call__` function reference instead of the class instance.

duncanista:
Fixes #5284.

Previously, the tracer would throw an error when trying to patch AWS Lambda functions (`aws_lambda` contrib) which handler was an instance of a callable class.
```python
class CallableHandler:
  def __call__(self, event, context):
    return _handler(event, context)

handler = CallableHandler()
# handler is going to be used by AWS as 
# handler(event, context), or handler(*args, **kwargs)
```

This error was caused since our patching wasn't expecting to patch instances of callable classes. This is use case was completely missed.

The patching now check it this is the case ONLY for file patching. This means that if the user is manually wrapping their AWS Lambda function as the following:
```
patched_handler = datadog_lambda_wrapper(handler)
```
No action is needed, since when a manual patching happens, we specifically patch `datadog-lambda` instead of the users code.

## Testing strategy
- Added unit tests for it.
- Manually tested deploying AWS Lambda functions with the following:
1. Custom class (described above).
2. Instrumenting custom class manually.
3. Using AWS Chalice instance as the handler.
4. Instrumenting AWS Chalice class manually.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


mabdinur:
Blocked by: https://github.com/DataDog/dd-trace-py/pull/5349

ddtrace/contrib/aws_lambda/patch.py
```diff
@@ -121,19 +121,27 @@ def _get_handler_and_module():
...
         (mod_name, handler_name) = parts
         modified_mod_name = _modify_module_name(mod_name)
         handler_module = import_module(modified_mod_name)
         handler = getattr(handler_module, handler_name)
-        return handler, handler_module, _datadog_instrumentation
+
+        # Check if handler is a class instance that can be called.
+        if not hasattr(handler, "__code__") and hasattr(handler, "__call__"):
+            class_name = type(handler).__name__
+            handler_module = getattr(handler_module, class_name)
```
mabdinur, line 141: `getattr(module, class_name)` this will return a reference to a class. Is this expected? If so we should rename `handler_module` or `class_name`

tests/contrib/aws_lambda/test_aws_lambda.py
```diff
@@ -45,7 +47,7 @@ def setup():
 
 
 @pytest.mark.parametrize("customApmFlushDeadline", [("-100"), ("10"), ("100"), ("200")])
-@pytest.mark.snapshot()
+@pytest.mark.snapshot
```
mabdinur, line 50: @pytest.mark.snapshot and @pytest.mark.snapshot() are interchangeable

ddtrace/contrib/aws_lambda/patch.py
```diff
@@ -121,19 +121,27 @@ def _get_handler_and_module():
     path = os.environ.get("DD_LAMBDA_HANDLER", None)
     _datadog_instrumentation = DatadogInstrumentation()
 
+    handler, handler_module = None, None
+
```
mabdinur, line 125: nit: `handler`  and `handler_module` should never be None. This will cause `unwrap(handler, wrapper)` to fail. 
```suggestion
```

ddtrace/contrib/aws_lambda/patch.py
duncanista: Yes, this is expected. Since we don't know where this class might be coming from, we want to get the direct reference to it. We might encounter something like the AWS Chalice example, where the class `Chalice` is imported from another module. We don't have the path of that module, therefore, it's harder to patch. This way, we get the direct reference to it, and we patch it.

tests/contrib/aws_lambda/test_aws_lambda.py
duncanista: Thanks, I removed it since I realized some of them had it, and some didn't, so I just wanted to keep consistency.

ddtrace/contrib/aws_lambda/patch.py
duncanista: Would it make sense to have two returns here again?

ddtrace/contrib/aws_lambda/patch.py
```diff
@@ -121,19 +121,27 @@ def _get_handler_and_module():
...
     else:
         parts = path.rsplit(".", 1)
         (mod_name, handler_name) = parts
         modified_mod_name = _modify_module_name(mod_name)
         handler_module = import_module(modified_mod_name)
         handler = getattr(handler_module, handler_name)
-        return handler, handler_module, _datadog_instrumentation
+
+        # Check if handler is a class instance that can be called.
+        if not hasattr(handler, "__code__") and hasattr(handler, "__call__"):
```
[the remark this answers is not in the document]
duncanista: You're right. I'm wondering of another way of figuring out if the handler is a class. Do you happen to know? Since all instance classes are objects, I couldn't figure out another one. Also, this will never occur with our use cases for AWS Chalice, which don't create `__code__` in their classes. Another thing, users mostly use functions instead of classes for their AWS Lambda functions. What do you think?

ddtrace/contrib/aws_lambda/patch.py
michaeldimchuk: Just to clarify, with AWS Chalice it'll typically be an object (class instance), not a class reference. For example with this code snippet:
```
# app.py
from chalice.app import Chalice

app = Chalice(__name__)
```
The Lambda handler module would be `app.app`. That being said though, technically all 3 of the following options are valid callables:
```
# app.py
from typing import Any


class StaticMeta(type):
    @staticmethod
    def __call__(event: dict[str, Any], context: Any) -> dict[str, str]:
        return {"hello": f"from {StaticMeta = }"}


class ClassMeta(type):
    @classmethod
    def __call__(cls, event: dict[str, Any], context: Any) -> dict[str, str]:
        return {"hello": f"from {cls = }"}


class StaticHandler(metaclass=StaticMeta):
    pass


class ClassHandler(metaclass=ClassMeta):
    pass


class InstanceHandler:
    @classmethod
    def __call__(self, event: dict[str, Any], context: Any) -> dict[str, str]:
        return {"hello": f"from {self = }"}


handler1 = StaticHandler
handler2 = ClassHandler
handler3 = InstanceHandler()
```
With the Lambda handler modules being `app.handler1`, `app.StaticHandler`, `app.handler2`, `app.ClassHandler` and `app.handler3`.

duncanista: Thanks @michaeldimchuk, I missed when handler can be called from `@classmethod`, the current state of the PR should work for all of them, but the `ClassHandler`. Will add more testing for this and continue to fully support these use cases.

ddtrace/contrib/aws_lambda/patch.py
```diff
@@ -124,16 +124,35 @@ def _get_handler_and_module():
...
+            class_handler = getattr(class_module, "__call__")
+
+            if isinstance(handler, type):
+                # class handler is a metaclass
+                if hasattr(class_handler, "__func__"):
+                    class_handler = class_handler.__func__
+
+            return class_handler, class_module, _datadog_instrumentation
+        else:
+            raise TypeError("Handler is not callable")
```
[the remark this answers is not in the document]
duncanista: This will be caught by the try-catch block in patch/unpatch and logged there. Wdyt?

tests/contrib/aws_lambda/test_aws_lambda.py
```diff
@@ -112,4 +114,28 @@ async def test_module_patching(mocker, context):
...
+def test_class_based_handlers(context, handler):
+    function_name = handler["function_name"]
+    _handler = handler["value"]
+    os.environ.update(
+        {
+            "AWS_LAMBDA_FUNCTION_NAME": function_name,
+            "DD_LAMBDA_HANDLER": "tests.contrib.aws_lambda.handlers." + function_name,
+        }
+    )
```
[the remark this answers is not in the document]
duncanista: This will require me to rewrite all the environment variable handling in our tests, since `override_env` doesn't accepts nests. I'll look into it.

mabdinur (approved):

ddtrace/contrib/aws_lambda/patch.py
```diff
@@ -124,16 +124,35 @@ def _get_handler_and_module():
...
         parts = path.rsplit(".", 1)
         (mod_name, handler_name) = parts
         modified_mod_name = _modify_module_name(mod_name)
         handler_module = import_module(modified_mod_name)
         handler = getattr(handler_module, handler_name)
-        return handler, handler_module, _datadog_instrumentation
+
+        if callable(handler):
+            # handler is a function
+            if not isinstance(handler, type) and hasattr(handler, "__code__"):
```
michaeldimchuk, line 140: Would be worthwhile to add a comment here to indicate that `hasattr(handler, "__code__")` is a "best effort" attempt to detect whether this is a function or a class instance.

Technically adding a `__code__` property to the class is valid (though really strange), so a comment would make it more explicit that we're not accounting for such a use case. Unfortunately I haven't been able to find a way to distinguish between callable class and a function without "reflection" hacks like defining a function, capturing it's type and comparing against that, but even that doesn't cover all cases.

michaeldimchuk (approved):

ddtrace/contrib/aws_lambda/patch.py
duncanista: Ah, you're right! 

Technically, it's valid yeah. I'll add a comment here to address it. 
I haven't been successful to find an example to distinguish it either. 
These won't cover all cases. But we are expecting to go to back to normal and remove the monkey patch once our tracing is done only in this repo.

duncanista: I was thinking, for this use case, I can just add another condition to the statement, like this:
What do you think? 

```python
class_name = type(handler).__name__
is_function = not isinstance(handler, type) and hasattr(handler, "__code__") and class_name == "function"
if is_function:
    return handler, handler_module, _datadog_instrumentation

# class instance with __code__ defined will ignore it, and use __call__ instead here
...
```

michaeldimchuk: That's an option, though it doesn't cover all cases. For example technically (and again, mostly just trying to see how I can break this), `print` is a valid lambda handler, but it's type is `<class 'builtin_function_or_method'>` and would thus miss this check. 

Something like this may be a bit more robust:
```
from types import (
    BuiltinFunctionType,
    BuiltinMethodType,
    ClassMethodDescriptorType,
    FunctionType,
    LambdaType,
    MethodDescriptorType,
    MethodWrapperType,
    WrapperDescriptorType,
)

AnyFunction = (
    BuiltinFunctionType,
    BuiltinMethodType,
    ClassMethodDescriptorType,
    FunctionType,
    LambdaType,
    MethodDescriptorType,
    MethodWrapperType,
    WrapperDescriptorType,
)

def is_function(callable: Any) -> bool:
    return isinstance(callable, AnyFunction)
```
But that's assuming you want to cover all such cases, which may not be the case. If not, then a comment indicating that it's a best effort attempt to check against user defined functions.

duncanista: You're right. I'll leave it as a best effort.
I'll add a comment and a warning log for this, I'll also make sure we only warn on cold starts, in order to avoid logging on every invocation. 

Thank you so much @michaeldimchuk

mabdinur (dismissed):

mabdinur (approved):

PR #2849, opened 2021-09-16
fix(profiling): measure actual CPU time on OSX

P403n1x87:
## Description

WIP. This change makes use of the OSX API to retrieve the total CPU time spent by each thread to yield more accurate CPU time profiles on OSX.

## Checklist
- [ ] Added to the correct milestone.
- [ ] Tests provided or description of manual testing performed is included in the code or PR.
- [ ] Library documentation is updated.
- [ ] [Corp site](https://github.com/DataDog/documentation/) documentation is updated (link to the PR).

PR #5338, opened 2023-03-16
fix(profiling): accounting of time for gevent tasks

P403n1x87:
This change fixes the accounting of wall and CPU time to gevent tasks by only assigning them to the main Python thread, where they are supposed to run under normal circumstances. This chance also ensures that task information does not leak onto any other threads that might be running, most notably the profiler threads themselves.

## Checklist

## Reviewer Checklist

- [ ] Title is accurate.


ddtrace/profiling/_threading.pyx
```diff
@@ -23,42 +21,34 @@ cpdef get_thread_name(thread_id):
...
     try:
         # We prioritize using native ids since we expect them to be surely unique for a program. This is less true
         # for hashes since they are relative to the memory address which can easily be the same across different
         # objects.
-        return thread_obj.native_id
+        return thread.native_id
     except AttributeError:
         # Python < 3.8
-        return hash(thread_obj)
+        return hash(thread)
```
sanchda, line 51: `return hash(thread)` wasn't introduced in this PR, but I wonder what the consequences of trying to do better might be.

`hash(thread)` will return a hash that isn't guaranteed to be stable even when the underlying native thread + python thread are the same, right?  The caller is asking specifically for a quantity that is less volatile than `thread_id`, and we're giving it something more volatile.

We might choose to call `gettid()` (https://man7.org/linux/man-pages/man2/gettid.2.html), but
* We'd need to figure out the corresponding call on Windows
* we add a syscall to every call of `get_thread_native_id()` (and it isn't clear to me how to cache these)


I don't think this needs to be fixed in this PR, but I'm curious to hear your thoughts.  For my part, I vaguely feel like we're doing the wrong thing (although I acknowledge this mostly affects older Pythons, so it isn't critical).

ddtrace/profiling/collector/_task.pyx
```diff
@@ -112,16 +112,22 @@ cpdef list_tasks(thread_id):
...
-    # This should be true as nobody could use a half-monkey-patched gevent
     if _gevent_tracer is not None:
-        tasks.extend([
-            (greenlet_id,
-             _threading.get_thread_name(greenlet_id),
-             greenlet.gr_frame)
-            for greenlet_id, greenlet in list(compat.iteritems(_gevent_tracer.greenlets))
-            if not greenlet.dead
-        ])
+        if type(_threading.get_thread_by_id(thread_id)).__name__.endswith("_MainThread"):
```
sanchda, line 116: I tried to reason about how an antagonistic user might use gevent, but also custom thread names, to cause harm here.  Nothing really came to mind--and if such a user did such a thing, I think we could ask them to stop.

sanchda (dismissed):
LGTM;  left some comments for discussion, but nothing blocking this PR.  Thanks again!

ddtrace/profiling/collector/stack.pyx
```diff
@@ -309,7 +309,11 @@ cdef stack_collect(ignore_profiler, thread_time, max_nframes, interval, wall_tim
     exc_events = []
 
     for thread_id, thread_native_id, thread_name, thread_pyframes, exception, span, cpu_time in running_threads:
-        thread_task_id, thread_task_name, thread_task_frame = _task.get_task(thread_id)
+        if thread_name is None:
```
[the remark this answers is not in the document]
P403n1x87: > does this create an observability gap for applications which pathologically spend a reasonable fraction of CPU-time in thread initialization?

The expectation is that this only happens when a thread is initialised. Under the assumption that an application doesn't spawn threads at very high frequency (it would not do good to Python performance), we would be dealing with long-lived threads on a normal basis, meaning that the initialisation CPU time would disappear in aggregates. So my very optimistic hunch is that we don't pay a huge price in terms of neglected data here.

> Is there any advantage to taking these threads, but giving them one-off names ("unnamed_thread#thread_id" or whatever)?

I think this is the current behaviour, and the picture looks like this

Screenshot 2023-03-20 at 16 19 41

so no much useful initialisation data visible there. So probably still worth filtering here to declutter the thread list?

shatzi (dismissed):
LGTM - kind of new to all gevent issues. but I like the fixes to _threading.pyx

ddtrace/profiling/_threading.pyx
P403n1x87: You are totally right on all your points. The problem with `gettid` is that it would always be called from the main thread, which would then normally return the process ID 🙁 . There is a similar problem for macOS and that's why we don't have true CPU time there. A possible (but perhaps not very clean) way of working around this would be to wrap the `Thread` class to call and store `gettid()` as first thing after the bootstrap, as done in #2849). Another hacky way would be to _guess_ where in the `pthread_t` structure the `tid` field is, and read the value from there (cf. https://github.com/P403n1x87/austin/blob/b48d980f6f3d0808d9221ee611a9c3094e02bb50/src/linux/py_proc.h#L690-L734)

sanchda: Thank you for the thoughts and explanations!  Sounds like you've thought this through :)

ddtrace/__init__.py
```diff
@@ -3,7 +3,14 @@
 
 ModuleWatchdog.install()
 
-from ._logger import configure_ddtrace_logger  # noqa: E402
+# Acquire a reference to the threading module. Some parts of the library (e.g.
+# the profiler) might be enabled programmatically and therefore might end up
+# getting a reference to the tracee's threading module. By storing a reference
```
emmettbutler, line 8:
```suggestion
# getting a reference to the tracer's threading module. By storing a reference
```

setup.cfg
```diff
@@ -12,6 +12,7 @@ exclude=
   .git,__pycache__,
   .eggs,*.egg,
   build,
+  ddtrace/__init__.py,
```
emmettbutler, line 15: Why is this excluded from flake8 linting? I can't see how it's related to the rest of this change.

brettlangdon (commented):
minor nits/questions.

mabdinur (approved):
Aside from Brett's and Emmet's comments/suggestions this change look good to me 👍

ddtrace/__init__.py
P403n1x87: This is indeed the tracee's module. Instead we want the tracer's module.

ddtrace/profiling/collector/stack.pyx
```diff
@@ -344,8 +345,8 @@ cdef stack_collect(ignore_profiler, thread_time, max_nframes, interval, wall_tim
                 thread_id=thread_id,
                 thread_native_id=thread_native_id,
                 thread_name=thread_name,
-                task_id=thread_task_id,
-                task_name=thread_task_name,
+                task_id=None,
+                task_name=None,
```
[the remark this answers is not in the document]
P403n1x87: In the logical abstractions of tasks on top of threads, not all threads will have a task running within them. Besides, since with the module cloning approach we would have the ddtrace threads running as Python threads, we can no longer assign the current task data to all threads, else we would end up reporting e.g. that the stack collector thread is running a gevent task.

emmettbutler (approved):

ddtrace/profiling/collector/stack.pyx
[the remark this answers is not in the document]
brettlangdon: a code comment would be a nice to have since this isn't obvious (at least to me) from the code alone.

setup.cfg
brettlangdon: can we handle in a separate PR so we have a single change/commit/PR description with motivating factors?

your reasoning seems logical, but feels unrelated to this change overall.

tests/profiling/simple_program_gevent.py
```diff
@@ -1,15 +1,15 @@
 from gevent import monkey
 
+from ddtrace.profiling import bootstrap
```
brettlangdon, line 3: does this order now matter? comment?

brettlangdon (commented):
some minor nits, I could see them being non-blocking.

setup.cfg
P403n1x87: In #5428

ddtrace/profiling/collector/stack.pyx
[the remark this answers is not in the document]
P403n1x87: Wall time accounting is part of an ongoing discussion. I think we'll have chances to get back to this with more details. I will postpone this task until we have more defined details if that's ok.

brettlangdon (commented):
just waiting on the `setup.cfg` change, but otherwise lgtm 👍🏻

emmettbutler (approved):

releasenotes/notes/fix-profiler-gevent-tasks-76ed862362210cb7.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    profiling: correct accounting of wall and CPU time for gevent tasks within
+    the main Python thread.
```
mabdinur, line 5: nit: casing and tensing 
```suggestion
    profiling: Corrects accounting of wall and CPU time for gevent tasks within
    the main Python thread.
```

mabdinur (approved):

brettlangdon (approved):

mabdinur (approved):

PR #5339, opened 2023-03-16
fix(tracing): honor sample rate set by agent

emmettbutler:
This change fixes a bug causing sample rate data from the agent not to be honored in the tracer's sampling behavior. The bug is that the sampling decision is made before the `env` tag is set. This tag is a key piece of information necessary to make sampling decisions based on responses from the agent.

The change also includes a regression test.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.

### DIFF
--- a/ddtrace/tracer.py
+++ b/ddtrace/tracer.py
@@ -729,6 +729,9 @@ class Tracer(object):
             span._metrics[PID] = self._pid
 
         # Apply default global tags.
+        if ENV_KEY not in span.get_tags() and self._env:
+            span.set_tag_str(ENV_KEY, self._env)
+
         if self._tags:
             span.set_tags(self._tags)
 
--- a/ddtrace/tracer.py
+++ b/ddtrace/tracer.py
@@ -230,6 +230,7 @@ class Tracer(object):
         self._dogstatsd_url = agent.get_stats_url() if dogstatsd_url is None else dogstatsd_url
         self._compute_stats = config._trace_compute_stats
         self._agent_url = agent.get_trace_url() if url is None else url  # type: str
+        self._env = config.env
         agent.verify_url(self._agent_url)
 
         if self._use_log_writer() and url is None:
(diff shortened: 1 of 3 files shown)


ddtrace/tracer.py
```diff
@@ -694,6 +694,9 @@ def _start_span(
             span._local_root = span
             if config.report_hostname:
                 span.set_tag_str(HOSTNAME_KEY, hostname.get_hostname())
+
+            if config.env:
```
mabdinur, line 698: Nice catch 😄. We also set the env tag on line 743. Can we set this tag once? (maybe on line 662)

mabdinur (commented):
left a minor nit but otherwise LGTM

ddtrace/tracer.py
emmettbutler: I tried that, but doing so breaks a ton of tracer tests. I decided that this duplicating approach was the best balance between safety and fixing this bug.

mabdinur (approved):

tests/integration/test_integration.py
```diff
@@ -721,6 +789,8 @@ def test_call_basic_config(ddtrace_run_python_code_in_subprocess, call_basic_con
     out, err, status, pid = ddtrace_run_python_code_in_subprocess(
         """
 import logging
+
+
```
mabdinur, line 793: nit: we can remove this whitespace

tests/integration/test_integration.py
```diff
@@ -664,6 +730,8 @@ def test_regression_logging_in_context(tmpdir, logs_injection, debug_mode, patch
     f.write(
         """
 import ddtrace
+
```
mabdinur, line 733: ws

emmettbutler:
@mabdinur I had to adjust an expectation constant in a test to fix a failure in py27. If you could re-approve when you get a chance, I'd appreciate it.

ddtrace/tracer.py
Yun-Kim: Your previous [change](https://github.com/DataDog/dd-trace-py/pull/5339/commits/bddad22dd8824ca858869466e24f6a05d13da251) caused 3 test cases to fail, just on multiple versions being tested, so that's why it looks like a ton of tests failed. I think it's worth at least investigating why that change causes it to fail before we can accept the duplicate code 🤔

brettlangdon (approved):

PR #5351, opened 2023-03-20
fix(propagataion): handle whitespace in W3C tracecontext

ZStriker19:
## Checklist

## Reviewer Checklist

- [ ] Title is accurate.

### DIFF
--- a/ddtrace/propagation/http.py
+++ b/ddtrace/propagation/http.py
@@ -687,7 +687,6 @@ class _TraceContext:
 
         ts = _extract_header_value(_POSSIBLE_HTTP_HEADER_TRACESTATE, headers)
         
-        # assert 0
         if ts:
             # strip whitespace
             ts = "".join(ts.split())
--- a/ddtrace/propagation/http.py
+++ b/ddtrace/propagation/http.py
@@ -685,8 +685,10 @@ class _TraceContext:
         origin = None
         meta = {W3C_TRACEPARENT_KEY: tp}  # type: _MetaDictType
 
+        # ts = _extract_header_value(_POSSIBLE_HTTP_HEADER_TRACESTATE, headers)
         ts = _extract_header_value(_POSSIBLE_HTTP_HEADER_TRACESTATE, headers)
-        
+
+        assert 0
         if ts:
             # strip whitespace
             ts = "".join(ts.split())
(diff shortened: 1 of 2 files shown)


emmettbutler (dismissed):

ddtrace/propagation/http.py
```diff
@@ -611,12 +612,11 @@ def _get_traceparent_values(tp):
         return trace_id, span_id, sampling_priority
 
     @staticmethod
-    def _get_tracestate_values(ts):
-        # type: (str) -> Tuple[Optional[int], Dict[str, str], Optional[str]]
+    def _get_tracestate_values(ts_l):
+        # type: (List[str]) -> Tuple[Optional[int], Dict[str, str], Optional[str]]
 
-        # tracestate parsing, example: dd=s:2;o:rum;t.dm:-4;t.usr.id:baz64,congo=t61rcWkgMzE
+        # tracestate list parsing, example: [dd=s:2;o:rum;t.dm:-4;t.usr.id:baz64,congo=t61rcWkgMzE]
```
Yun-Kim, line 618: Is this an example of an input? Isn't the input supposed to be a list of strings?

ZStriker19: ah you're right, need to add the quotes. Good call.

emmettbutler (dismissed):

Kyle-Verhoog (dismissed):
non-blocking: is it worth having a regression unit test for this or are the shared tests sufficient to cover regressions?

ZStriker19:
> non-blocking: is it worth having a regression unit test for this or are the shared tests sufficient to cover regressions?

The system-tests caught this in the first place and I think we can count on them to catch it in the future as well.

emmettbutler (approved):

ddtrace/propagation/http.py
```diff
@@ -611,12 +612,11 @@ def _get_traceparent_values(tp):
         return trace_id, span_id, sampling_priority
 
     @staticmethod
-    def _get_tracestate_values(ts):
-        # type: (str) -> Tuple[Optional[int], Dict[str, str], Optional[str]]
+    def _get_tracestate_values(ts_l):
+        # type: (List[str]) -> Tuple[Optional[int], Dict[str, str], Optional[str]]
 
-        # tracestate parsing, example: dd=s:2;o:rum;t.dm:-4;t.usr.id:baz64,congo=t61rcWkgMzE
+        # tracestate list parsing, example: ["dd=s:2;o:rum;t.dm:-4;t.usr.id:baz64","congo=t61rcWkgMzE"]
```
Yun-Kim, line 618: could we add an example output from this input to show what we're extracting? This would be super helpful.

Yun-Kim (approved):

PR #5386, opened 2023-03-22
chore(asm): deprecate private appsec constants from public API

christophe-papazian:
Mark several private appsec constants from the public API as deprecated.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/constants.py
```diff
@@ -23,23 +23,8 @@
 MULTIPLE_IP_HEADERS = "_dd.multiple-ip-headers"
 
 APPSEC_ENV = "DD_APPSEC_ENABLED"
-APPSEC_ENABLED = "_dd.appsec.enabled"
```
mabdinur, line 26: We can not remove these constants in a minor version release (ex: v1.11.0). This is a breaking change and will go against our version policy. 

Instead we should raise a deprecation warning when these constants are accessed ([ex](https://github.com/DataDog/dd-trace-py/blob/v1.9.4/ddtrace/contrib/grpc/constants.py#L31)), then remove these constants in the next major version bump (ex: v2.0.0).

christophe-papazian: ok, I added them back through the same mechanism as [PEP 562](https://peps.python.org/pep-0562/)

mabdinur: One last nit: Can we add a test case? It should be similar to the this: https://github.com/DataDog/dd-trace-py/blob/v1.9.4/tests/contrib/grpc/test_constants.py#L6

mabdinur (commented):
Small suggestion but otherwise LGTM

ddtrace/constants.py
christophe-papazian: Yes, I'll add that now.

juanjux (approved):

avara1986 (approved):

PR #5395, opened 2023-03-23
feat(sourcecode): add git metadata from tags and from package extraction

sashacmc:
## Description
Add four environment variables:

- DD_GIT_REPOSITORY_URL string, default: empty
- DD_GIT_COMMIT_SHA string, default: empty
- DD_TRACE_GIT_METADATA_ENABLED bool, default: true
- DD_MAIN_PACKAGE string, default empty

Add following steps:

- Check if DD_TRACE_GIT_METADATA_ENABLED is True
- Check if git metadata is already provided by environment variables:
  - Check DD_GIT_REPOSITORY_URL and DD_GIT_COMMIT_SHA, if they exist, use metadata from them.
  - Check DD_TAGS, if it contains git.repository_url and git.commit.sha, use metadata from them.
  - Continue with metadata from artifacts.
- Read package name from DD_MAIN_PACKAGE, if it empty try detect package (will work for direct import)
- Read git metadata from the package metadata field Project-URL: source_code_link, split the value by "#" and use first part as repository URL and second part as commit hash

During trace telemetry sending:

- Add metadata tags to the first span of each chunk (NB: if [V7](https://github.com/DataDog/datadog-agent/blob/3ce2edbd37f3815102e31f59926699930bacdfc2/pkg/trace/api/api.go#L439) payload support already implemented send tags on payload level):
  - _dd.git.repository_url for repository URL
  - _dd.git.commit.sha for commit hash

During profile telemetry sending:
- Add metadata tags to each profile:
  - git.repository_url for repository URL
  - git.commit.sha to commit hash

## Checklist

## Motivation
Add new git metadata tags and extraction support
RFC: [SCI Embedding Git references in built artifacts RFC](https://docs.google.com/document/d/1vxuRUNzHqd6sp1lnF3T383acbLrG0R-xDGi8cdZ4cs8/edit#)
Jira: [SC-1159](https://datadoghq.atlassian.net/browse/SC-1159)

## Testing strategy
1. Build app from the cloned git repo with tracer/profiler with ddgitmetadata module in setup.py
1. Check git.repository_url and git.commit.sha tags on traces/profiles DD pages
1. Same as 1, but with DD_GIT_REPOSITORY_URL and DD_GIT_COMMIT_SHA with different values (values from env variables must present)
1. Same as 1, but with DD_TAGS="git.repository_url:some_repo git.commit.sha:some_sha" (values from env variables must present)
1. Same as 1, but with DD_TRACE_GIT_METADATA_ENABLED=false, tags must not be presented

## Reviewer Checklist
- [x] Title is accurate.

[SC-1159]: https://datadoghq.atlassian.net/browse/SC-1159?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ddtrace/internal/gitmetadata.py
```diff
@@ -0,0 +1,128 @@
...
+    """
+    Cleanup tags from git metadata
+    """
+    tags.pop(TAG_REPOSITORY_URL, None)
+    tags.pop(TAG_COMMIT_SHA, None)
+
+    return tags
+
+
+def update_profiler_tags(tags):
```
P403n1x87, line 117: This helper seems quite generic rather than profiler-specific. Can we perhaps rename it to just `update_tags`? Or `inject_git_metadata_tags`/`add_git_metadata_tags` even?

ddtrace/settings/config.py
```diff
@@ -14,6 +14,7 @@
 from ..internal.constants import PROPAGATION_STYLE_ALL
 from ..internal.constants import PROPAGATION_STYLE_B3
 from ..internal.constants import _PROPAGATION_STYLE_DEFAULT
+from ..internal.gitmetadata import clean_tags
```
P403n1x87, line 17:
```suggestion
from ..internal import gitmetadata
```

ddtrace/settings/config.py
```diff
@@ -198,7 +199,7 @@ def __init__(self):
         self.client_ip_header = os.getenv("DD_TRACE_CLIENT_IP_HEADER")
         self.retrieve_client_ip = asbool(os.getenv("DD_TRACE_CLIENT_IP_ENABLED", default=False))
 
-        self.tags = parse_tags_str(os.getenv("DD_TAGS") or "")
+        self.tags = clean_tags(parse_tags_str(os.getenv("DD_TAGS") or ""))
```
P403n1x87, line 202:
```suggestion
        self.tags = gitmetadata.clean_tags(parse_tags_str(os.getenv("DD_TAGS") or ""))
```

ddtrace/internal/gitmetadata.py
```diff
@@ -0,0 +1,128 @@
...
+_GITMETADATA_TAGS = None  # type: typing.Optional[typing.Dict[str, str]]
+
+TAG_REPOSITORY_URL = "git.repository_url"
+TAG_COMMIT_SHA = "git.commit.sha"
+
+TRACE_TAG_REPOSITORY_URL = "_dd.git.repository_url"
+TRACE_TAG_COMMIT_SHA = "_dd.git.commit.sha"
+
+
+class GitMetadataConfig(Env):
```
P403n1x87, line 17: ❤️

ddtrace/internal/gitmetadata.py
```diff
@@ -0,0 +1,128 @@
...
+    commit_sha = Env.var(str, "git.commit.sha", default="")
+
+    # DD_MAIN_PACKAGE
+    main_package = Env.var(str, "main_package", default="")
+
+    # DD_TAGS
+    tags = Env.var(str, "tags", default="")
+
+
+def __get_tags_from_env(config):
```
P403n1x87, line 36:
```suggestion
def _get_tags_from_env(config):
```

ddtrace/internal/gitmetadata.py
```diff
@@ -0,0 +1,128 @@
...
+    repository_url = Env.var(str, "git.repository_url", default="")
+
+    # DD_GIT_COMMIT_SHA
+    commit_sha = Env.var(str, "git.commit.sha", default="")
+
+    # DD_MAIN_PACKAGE
+    main_package = Env.var(str, "main_package", default="")
+
+    # DD_TAGS
+    tags = Env.var(str, "tags", default="")
```
P403n1x87, line 33: Just a thought: I'd make this a "private" attribute, e.g. `_tags`, and define `tags` as a derivation with
~~~python
tags = Env.der(dict, _get_tags_from_env)
~~~

P403n1x87: In fact, the derivation could be based on the `get_tracer_tags` logic

rpelliard (approved):

ddtrace/internal/gitmetadata.py
```diff
@@ -0,0 +1,128 @@
...
+from ddtrace.internal.utils import formats
+
+
+_GITMETADATA_TAGS = None  # type: typing.Optional[typing.Dict[str, str]]
+
+TAG_REPOSITORY_URL = "git.repository_url"
+TAG_COMMIT_SHA = "git.commit.sha"
+
+TRACE_TAG_REPOSITORY_URL = "_dd.git.repository_url"
+TRACE_TAG_COMMIT_SHA = "_dd.git.commit.sha"
```
mabdinur, line 14: If we plan to import these constants in other modules then we should move them to `ddtrace/ext/git.py`

ddtrace/internal/writer.py
```diff
@@ -382,6 +383,10 @@ def _set_keep_rate(self, trace):
         if trace:
             trace[0].set_metric(KEEP_SPANS_RATE_KEY, 1.0 - self._drop_sma.get())
 
+    def _set_git_metadata(self, trace):
+        if trace:
+            trace[0].set_tags(gitmetadata.get_tracer_tags())
```
mabdinur, line 388: We generally avoid adding trace level tags in the agent writer. We should move `trace[0].set_tags(gitmetadata.get_tracer_tags())` to the [TraceTagsProcessor](https://github.com/DataDog/dd-trace-py/blob/v1.10.1/ddtrace/internal/processor/trace.py#L129).

releasenotes/notes/feature-sci-git-metadata-0b418d0a3529d11f.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    sci: Add git metadata from tags and from package support
```
mabdinur, line 4: Can we reword to this release note to include the new tags we plan to send and the environment variables they should set to use this feature. If this gets too wordy we can include a link to ddtrace library docs or datadog's documentation pages.

ddtrace/internal/gitmetadata.py
```diff
@@ -0,0 +1,128 @@
...
+    tags.pop(TAG_REPOSITORY_URL, None)
+    tags.pop(TAG_COMMIT_SHA, None)
+
+    return tags
+
+
+def update_tags(tags):
+    # type: (typing.Dict[str, str]) -> typing.Dict[str, str]
+    """
+    Update profiler tags with git metadata
```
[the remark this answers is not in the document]
lraucy: Could you describe a bit more what is done in the docstring?
I'm not sure I properly understand the content of this function.
Specifically, it seems weird to me that we are removing tags in all cases, and replace them with other value coming from some place else?

ddtrace/internal/writer.py
sashacmc: We should inject it in the first span of each payload (not root span)
Is it possible int the [TraceTagsProcessor](https://github.com/DataDog/dd-trace-py/blob/v1.10.1/ddtrace/internal/processor/trace.py#L129)?

ddtrace/internal/gitmetadata.py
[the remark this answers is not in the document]
lraucy: Ah after looking into more details I think I got the profiler part.
I would avoid re-using `get_tracer_tags` for profiling, because we are playing weirdly with some dict keys here, which makes the code hard to follow and understand

Here's what I would recommend instead:
- add a `get_git_tags()` function that returns a tuple (repoURL, commitSha)
- add a `set_tracer_tags(repoURL, commitSha)` function
- add a `set_profiler_tags(repoURL, commitSha)` function

I'm not entirely sure the `set_tracer_tags` and `set_profiler_tags` would belong here or in the specific tracer/profiler tags.

ddtrace/profiling/exporter/http.py
```diff
@@ -78,7 +79,7 @@ def __attrs_post_init__(self):
         tags = {
             k: six.ensure_str(v, "utf-8")
             for k, v in itertools.chain(
-                parse_tags_str(os.environ.get("DD_TAGS")).items(),
+                gitmetadata.update_tags(parse_tags_str(os.environ.get("DD_TAGS"))).items(),
```
lraucy, line 82: Could you add comments why we are doing that?
I think I understand why, but I'm not entirely sure and that can be quite confusing for the reader who doesn't have a lot of context.

ddtrace/internal/gitmetadata.py
sashacmc: We have no plans use it outside of this module (should I rename it with "_")?

ddtrace/internal/gitmetadata.py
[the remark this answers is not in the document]
sashacmc: > Not sure I understand why this is profiler tags here, typo?

Because this method was used in the profiler and was named update_profiler_tags before previous reviewer come 😁
But I agree with you suggestion about renaming and I will update the PR.
Thanks!

ddtrace/internal/gitmetadata.py
```diff
@@ -0,0 +1,128 @@
...
+    repository_url = config.repository_url
+    commit_sha = config.commit_sha
+
+    if not (repository_url and commit_sha):
+        tags = formats.parse_tags_str(config.tags)
+        repository_url = tags.get(TAG_REPOSITORY_URL)
+        commit_sha = tags.get(TAG_COMMIT_SHA)
+
+    if not (repository_url and commit_sha):
+        return None
```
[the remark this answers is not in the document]
sashacmc: The original idea was to only report complete data, but if you think having partial data would also be useful, let's add

ddtrace/internal/writer.py
mabdinur: Yes, [TraceTagsProcessor](https://github.com/DataDog/dd-trace-py/blob/v1.10.1/ddtrace/internal/processor/trace.py#L129) adds tags to the first span in a payload not just the root span.

ddtrace/internal/gitmetadata.py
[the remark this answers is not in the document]
lraucy: Generally speaking I think the tracer user would expect each env var to work independently. Otherwise, as a tracer user. I'd expect a single env var containing both info.

ddtrace/internal/gitmetadata.py
mabdinur: Sure but that's not necessary since this is module exists in the internal package.

ddtrace/internal/gitmetadata.py
```diff
@@ -0,0 +1,128 @@
...
+
+from envier import Env
+
+from ddtrace.internal.utils import formats
+
+
+_GITMETADATA_TAGS = None  # type: typing.Optional[typing.Dict[str, str]]
+
+TAG_REPOSITORY_URL = "git.repository_url"
+TAG_COMMIT_SHA = "git.commit.sha"
```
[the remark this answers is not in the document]
mabdinur: Good point. I don't have strong opinions here

releasenotes/notes/feature-sci-git-metadata-0b418d0a3529d11f.yaml
```diff
@@ -0,0 +1,8 @@
+---
+features:
+  - |
+    Extract git metadata from environment variables (DD_GIT_REPOSITORY_URL, DD_GIT_COMMIT_SHA),
+    or from the python package specified by the DD_MAIN_PACKAGE.
+    Send git metadata via _dd.git.repository_url and _dd.git.commit.sha root span tags for tracer,
+    and via git.repository_url and git.commit.sha tags for profiler
+    It can be disabled by the DD_TRACE_GIT_METADATA_ENABLED
```
mabdinur, line 8: I don't think the `_dd.git....` tags are user facing. Correct me if I am wrong but I believe they will be only used internally. Feel free to ignore or edit the suggestion below

```suggestion
    sci: Extracts and sends git metadata from environment variables ``DD_GIT_REPOSITORY_URL``, ``DD_GIT_COMMIT_SHA``,
    or ``DD_MAIN_PACKAGE``.  This feature can be disabled by setting ``DD_TRACE_GIT_METADATA_ENABLED=False``.
```

mabdinur (commented):
Left some nits but aside from that LGTM

releasenotes/notes/feature-sci-git-metadata-0b418d0a3529d11f.yaml
sashacmc: Thanks! Done.
(Just little bit updated, because we extract values not from environment variable DD_MAIN_PACKAGE, but form the package with this name)

mabdinur (dismissed):

ddtrace/internal/gitmetadata.py
```diff
@@ -0,0 +1,114 @@
...
+        if source_code_link != "" and "#" in source_code_link:
+            repository_url, commit_sha = source_code_link.split("#")
+            commit_sha = commit_sha.split("&")[0]
+            return repository_url, commit_sha
+        return "", ""
+    except importlib_metadata.PackageNotFoundError:
+        return "", ""
+
+
+def get_git_tags():
```
lraucy, line 81: is there any kind of logging available in the python tracer that would make sense to use here, so users can understand where the value come from and if it's working properly? At the info or debug loglevel?

I don't know the general policy about the tracer logging, so I don't know if it's possible to do so.

ddtrace/internal/gitmetadata.py
```diff
@@ -0,0 +1,114 @@
...
+
+    if config.enabled:
+        repository_url, commit_sha = _get_tags_from_env(config)
+        if repository_url == "" or commit_sha == "":
+            pkg_repository_url, pkg_commit_sha = _get_tags_from_package(config)
+            if repository_url == "":
+                repository_url = pkg_repository_url
+            if commit_sha == "":
+                commit_sha = pkg_commit_sha
+        _GITMETADATA_TAGS = repository_url, commit_sha
```
lraucy, line 100: Should we wrap this in a try / except to catch any kind of exception? I see that the `_get_tags_from_package` at least is doing some function calls that could possibly fail in some cases (like list access) and which are not protected by a try / except.
I think we should avoid that at all points, failure in extracting the git information should not impact the traced application.

ddtrace/profiling/exporter/http.py
```diff
@@ -65,6 +68,18 @@ class PprofHTTPExporter(pprof.PprofExporter):
 
     endpoint_call_counter_span_processor = attr.ib(default=None, type=EndpointCallCounterProcessor)
 
+    def _update_git_metadata_tags(self, tags):
+        """
+        Update profiler tags with git metadata
+        """
+        gitmetadata.clean_tags(tags)
```
lraucy, line 75: can you maybe add a comment to say why we are removing tags? I think his is because `get_git_tags` will look into the DD_TAGS as well, and here `tags` exactly contains the content of `DD_TAGS`, is that the reason?

lraucy (approved):

mabdinur (approved):

ddtrace/internal/gitmetadata.py
```diff
@@ -0,0 +1,126 @@
...
+    try:
+        global _GITMETADATA_TAGS
+        if _GITMETADATA_TAGS is not None:
+            return _GITMETADATA_TAGS
+
+        config = GitMetadataConfig()
+
+        if config.enabled:
+            repository_url, commit_sha = _get_tags_from_env(config)
+            log.info("git tags from env: %s %s", repository_url, commit_sha)
```
brettlangdon, line 98: does this need to be an info log? would debug be better?

same for all other logs in this function, they feel like debug logs not info.

ddtrace/profiling/exporter/http.py
```diff
@@ -65,6 +68,19 @@ class PprofHTTPExporter(pprof.PprofExporter):
 
     endpoint_call_counter_span_processor = attr.ib(default=None, type=EndpointCallCounterProcessor)
 
+    def _update_git_metadata_tags(self, tags):
+        """
+        Update profiler tags with git metadata
+        """
+        # clean tags, because values will be combined and inserted back in the same way as for tracer
```
brettlangdon, line 75: I am not sure I follow this comment. What is wrong if `REPOSITORY_URL` is in `tags` (`DD_TAGS` I assume from the sounds of it?)

ddtrace/settings/config.py
```diff
@@ -199,7 +200,8 @@ def __init__(self):
         self.client_ip_header = os.getenv("DD_TRACE_CLIENT_IP_HEADER")
         self.retrieve_client_ip = asbool(os.getenv("DD_TRACE_CLIENT_IP_ENABLED", default=False))
 
-        self.tags = parse_tags_str(os.getenv("DD_TAGS") or "")
+        # cleanup DD_TAGS, because values will be inserted back in the optimal way (via _dd.git.* tags)
+        self.tags = gitmetadata.clean_tags(parse_tags_str(os.getenv("DD_TAGS") or ""))
```
brettlangdon, line 204: I am not sure I like this approach.

If someone is going to set them manually via `DD_TAGS` why should they be removed? That feels like unexpected behavior to just implicitly remove them.

Will this lead to broken scenarios for users? If someone is setting them via `DD_TAGS` today, but is running in an environment where they don't have any of the other env variables or package data set, then we are just going to remove the tags without any warning to the user.

releasenotes/notes/feature-sci-git-metadata-0b418d0a3529d11f.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    sci: Extracts and sends git metadata from environment variables ``DD_GIT_REPOSITORY_URL``, ``DD_GIT_COMMIT_SHA``,
```
brettlangdon, line 4: do we have any docs on docs.datadoghq.com we can point people to to learn more about the feature / configuration / etc?

tests/conftest.py
```diff
@@ -281,3 +284,84 @@ def pytest_runtest_protocol(item):
...
+
+# source code fixtures
+
+
+def _run(cmd):
+    return subprocess.check_output(cmd, shell=True)
+
+
+@contextlib.contextmanager
+def create_package(directory, pyproject, setup):
```
brettlangdon, line 297: Does this need to be in the main conftest for all other tests?

Since these seem very test specific, we should move these to be co-located with the tests either in `tests/tracer/conftest.py` or `tests/tracer/test_gitmetadata.py`.

We don't want these fixtures to potentially conflict with others (`temp_dir` or `example` seem like common enough names).

tests/conftest.py
```diff
@@ -281,3 +284,84 @@ def pytest_runtest_protocol(item):
...
+        os.chdir(cwd)
+
+        try:
+            shutil.rmtree(package_dir)
+        except OSError:
+            pass
+
+
+@pytest.fixture
+def temp_dir():
```
brettlangdon, line 336: pytest already has a built-in `tmpdir` fixture.

https://docs.pytest.org/en/6.2.x/tmpdir.html

tests/conftest.py
```diff
@@ -281,3 +284,84 @@ def pytest_runtest_protocol(item):
...
+        _run("git add .")
+        _run("git commit -m init")
+        _run("git remote add origin https://github.com/companydotcom/repo.git")
+
+        yield package_dir
+    finally:
+        os.chdir(cwd)
+
+        try:
+            shutil.rmtree(package_dir)
```
brettlangdon, line 330: we do this rmtree in `temp_dir` as well, so shouldn't be needed here?

tests/tracer/test_gitmetadata.py
```diff
@@ -0,0 +1,114 @@
...
+        os.environ["PYTHONPATH"] = os.getenv("PYTHONPATH", "") + os.pathsep + envdir
+        yield
+    finally:
+        os.chdir(cwd)
+        if python_path is not None:
+            os.environ["PYTHONPATH"] = python_path
+
+
+@pytest.mark.usefixtures("preapare_test_env")
+class GitMetadataTestCase(TracerTestCase):
```
brettlangdon, line 39: do we need any failure case test cases?

e.g. `DD_GIT_COMMIT_SHA` set but not `DD_GIT_REPOSITORY_URL`, or malformed values in the env vars?

also, what about an example that only uses `DD_TAGS`?

tests/profiling/exporter/test_http.py
```diff
@@ -466,3 +466,34 @@ def test_get_tags_legacy(monkeypatch):
     assert tags["mytag"] == "val2"
     assert tags["ddtag"] == "hi"
     assert tags["ddptag"] == "lo"
+
+
+def test_gitmetadata_env(monkeypatch):
```
brettlangdon, line 471: can we have an example that only has the values set in `DD_TAGS`, since we are removing them from the tags.

ddtrace/internal/processor/trace.py
```diff
@@ -115,6 +116,13 @@ def on_span_finish(self, span):
 class TraceTagsProcessor(TraceProcessor):
     """Processor that applies trace-level tags to the trace."""
 
+    def _set_git_metadata(self, chunk_root):
+        repository_url, commit_sha = gitmetadata.get_git_tags()
+        if repository_url != "":
```
brettlangdon, line 121:
```suggestion
        if repository_url:
```

ddtrace/internal/processor/trace.py
```diff
@@ -115,6 +116,13 @@ def on_span_finish(self, span):
...
+            chunk_root.set_tag_str("_dd.git.repository_url", repository_url)
+        if commit_sha != "":
```
brettlangdon, line 123:
```suggestion
        if commit_sha:
```

tests/tracer/test_gitmetadata.py
brettlangdon: What about from a package that doesn't have `Project-URL` set?

brettlangdon (changes requested):
handful of nits, I am worried about how we are handling `DD_TAGS`, and I do not feel like there is enough test coverage (only happy paths are being tested from the looks of it).

ddtrace/settings/config.py
sashacmc: 1. Tagging can be turned off with an environment variable DD_TRACE_GIT_METADATA_ENABLED
2. If we leave the values here, they will go both through `_dd.git....` and through `git...` which will lead to unnecessary duplication and potential conflicts (_dd.git already propagated to all spans on intake)

tests/conftest.py
sashacmc: Yes, it used in `tests/sourcecode/` and in `tests/tracer/`

tests/tracer/test_gitmetadata.py
sashacmc: > e.g. DD_GIT_COMMIT_SHA set but not DD_GIT_REPOSITORY_URL, or malformed values in the env vars?

We do not do these checks in any of the tracers, the values go as they are and are checked by the backend if necessary.

tests/conftest.py
sashacmc: Super! Thanks!

tests/tracer/test_gitmetadata.py
sashacmc: > What about from a package that doesn't have Project-URL set?

Test added

tests/conftest.py
brettlangdon: oh, I think I missed that part. I saw the deletion and assumed it was going away.

ddtrace/internal/gitmetadata.py
```diff
@@ -0,0 +1,126 @@
...
+        for val in importlib_metadata.metadata(config.main_package).get_all("Project-URL"):
+            capt_val = val.split(", ")
+            if capt_val[0] == "source_code_link":
+                source_code_link = capt_val[1].strip()
+                break
+
+        if source_code_link != "" and "#" in source_code_link:
+            repository_url, commit_sha = source_code_link.split("#")
+            commit_sha = commit_sha.split("&")[0]
+            return repository_url, commit_sha
```
brettlangdon, line 78: I would like to see better unit test coverage around this function. We are only testing the paths where `Project-URL` exists and is of the format we expect, or where it is not set. what if there is no `,`? what if `source_code_link` is not there? what if `#` is not in `source_code_link` ? what if the package does not exist?

tests/tracer/test_gitmetadata.py
```diff
@@ -0,0 +1,131 @@
...
+        # must be not present in old tags
+        assert s.get_tag("dd.git.repository_url") is None
+        assert s.get_tag("dd.git.commit.sha") is None
+
+    @run_in_subprocess(
+        env_overrides=dict(
+            DD_TAGS="git.commit.sha:12345,git.repository_url:github.com/user/tag_repo",
+            DD_GIT_COMMIT_SHA="123456",
+            DD_GIT_REPOSITORY_URL="github.com/user/env_repo",
+            DD_MAIN_PACKAGE="gitmetadatapoc",
```
brettlangdon, line 100: is this package name meant to be `mypackage`? (same for the other tests in this file).

if it is meant to be a package that does not exist, can we use a more obvious name like `DD_MAIN_PACKAGE=non_existent_pkg` ("ish")

ddtrace/internal/gitmetadata.py
```diff
@@ -0,0 +1,126 @@
...
+        return "", ""
+
+
+def get_git_tags():
+    # type: () -> typing.Tuple[str, str]
+    """
+    Returns git metadata tags tuple (repository_url, commit_sha)
+    """
+    try:
+        global _GITMETADATA_TAGS
```
brettlangdon, line 90: do we have any tests for this caching?

ddtrace/internal/gitmetadata.py
```diff
@@ -0,0 +1,126 @@
...
+        if _GITMETADATA_TAGS is not None:
+            return _GITMETADATA_TAGS
+
+        config = GitMetadataConfig()
+
+        if config.enabled:
+            repository_url, commit_sha = _get_tags_from_env(config)
+            log.debug("git tags from env: %s %s", repository_url, commit_sha)
+            if repository_url == "" or commit_sha == "":
+                pkg_repository_url, pkg_commit_sha = _get_tags_from_package(config)
```
brettlangdon, line 100: I didn't see any tests that handle the case when both package data is set and env vars are set and which one takes precedence.

ddtrace/internal/gitmetadata.py
```diff
@@ -0,0 +1,126 @@
...
+        global _GITMETADATA_TAGS
+        if _GITMETADATA_TAGS is not None:
+            return _GITMETADATA_TAGS
+
+        config = GitMetadataConfig()
+
+        if config.enabled:
+            repository_url, commit_sha = _get_tags_from_env(config)
+            log.debug("git tags from env: %s %s", repository_url, commit_sha)
+            if repository_url == "" or commit_sha == "":
```
brettlangdon, line 99:
```suggestion
            if not repository_url or not commit_sha:
```

This would be the more pythonic way of checking, a little to read what is happening. It also will help in case the value is even `None` (doesn't seem like it could be, but... just in case)

sashacmc:
@brettlangdon 
> I still don't think we should modify DD_TAGS in any way.

We already discussed it inside of the team and already implemented (and deployed more than month ago) this logic in Go tracer and already implemented the same for Java tracer. I don't think we need separate logic for Python tracer.

ddtrace/internal/gitmetadata.py
```diff
@@ -0,0 +1,126 @@
...
+    Returns tuple (repository_url, commit_sha)
+    """
+    repository_url = config.repository_url
+    commit_sha = config.commit_sha
+
+    tags = formats.parse_tags_str(config.tags)
+    if repository_url == "":
+        repository_url = tags.get(REPOSITORY_URL, "")
+    if commit_sha == "":
+        commit_sha = tags.get(COMMIT_SHA, "")
```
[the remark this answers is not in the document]
sashacmc: Sorry for the misunderstanding! Yes we still support providing git metadata via DD_TAGS but we transfer them to dd in more optimal way.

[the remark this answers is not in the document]
brettlangdon: are there customers today who we know are setting them via `DD_TAGS`, or is this mostly a "just in case they are" scenario?

tests/tracer/test_gitmetadata.py
sashacmc: Yes! Good catch! (it was from the previous tests, we renamed it after refactoring)

tests/profiling/exporter/test_http.py
```diff
@@ -466,3 +466,47 @@ def test_get_tags_legacy(monkeypatch):
...
+    ddtrace.internal.gitmetadata._GITMETADATA_TAGS = None
+
+    monkeypatch.setenv("DD_TAGS", "git.commit.sha:12345,git.repository_url:github.com/user/tag_repo")
+    monkeypatch.setenv("DD_GIT_COMMIT_SHA", "123456")
+    monkeypatch.setenv("DD_GIT_REPOSITORY_URL", "github.com/user/env_repo")
+    monkeypatch.setenv("DD_TRACE_GIT_METADATA_ENABLED", "false")
+    tags = parse_tags_str(http.PprofHTTPExporter(endpoint="")._get_tags("foobar"))
+
+    # must not present
+    assert "git.commit.sha" not in tags
```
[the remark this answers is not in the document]
sashacmc: We would like to be in consistency with other tracers, DD_TRACE_GIT_METADATA_ENABLED=false must disable it even for DD_TAGS

tests/profiling/exporter/test_http.py
```diff
@@ -466,3 +466,47 @@ def test_get_tags_legacy(monkeypatch):
...
+    import ddtrace.internal.gitmetadata
+
+    ddtrace.internal.gitmetadata._GITMETADATA_TAGS = None
+
+    monkeypatch.setenv("DD_TAGS", "git.commit.sha:12345,git.repository_url:github.com/user/tag_repo")
+    tags = parse_tags_str(http.PprofHTTPExporter(endpoint="")._get_tags("foobar"))
+
+    # must be from env variables
+    assert tags["git.commit.sha"] == "12345"
+    assert tags["git.repository_url"] == "github.com/user/tag_repo"
```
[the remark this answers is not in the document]
sashacmc: No, we still support DD_TAGS and for profiler intake we have no optimisation (it is not necessary), we still provide it via normal tags but in combination with other methods (new env variables, metadata from package)

ddtrace/internal/gitmetadata.py
```diff
@@ -0,0 +1,126 @@
...
+                    commit_sha = pkg_commit_sha
+
+            log.debug("git tags: %s %s", repository_url, commit_sha)
+            _GITMETADATA_TAGS = repository_url, commit_sha
+        else:
+            log.debug("git tags disabled")
+            _GITMETADATA_TAGS = ("", "")
+        return _GITMETADATA_TAGS
+    except Exception:
+        log.exception("get_git_tags failed")
```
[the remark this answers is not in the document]
sashacmc: At general we don't expect errors, but if some malformed data comes, or something like it we would like to don't break, but inform that something strange happens.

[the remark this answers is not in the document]
brettlangdon: 👍🏻 agree with the try/except, what I am curious about is the log message.

since an exception log will get shown to the user (likely not long after startup). my question was mostly, if they see that log message (with traceback) then is there anything they can do? is it clear the actions they should take?

which makes me wonder if the message itself should be more actionable like... suggest having them look at the specific env vars / link to product page on how the feature works / ask the to open a support ticket ?

or.... should we just make it a debug log? the flow would likely be, they try to get git metadata setup, it isn't working, they open a support case, we collect debug logs, and then we do the interpreting of the exception (not ideal flow for timeliness, but a valid one).

ddtrace/settings/config.py
sashacmc: We already discussed it inside of the team and already implemented (and deployed more than month ago) this logic in Go tracer and already implemented the same for Java tracer. I don't think we need separate logic for Python tracer.

Just for other one clarification, for old users nothing will changed, they can provide all as before via (DD_TAGS) and after intake it will be in the same place (in all spans) and even if they start embed metadata DD_TAGS will be used as preferred way to don't break anything.

ddtrace/internal/gitmetadata.py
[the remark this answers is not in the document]
sashacmc: Ok, agree with you, lets move it to debug

ddtrace/internal/gitmetadata.py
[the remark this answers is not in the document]
sashacmc: Currently it is backup way for user which don't use docker, I am not sure about amount of python users, but for the Go, we have a lot and transition to new tracer with the same logic was done smoothly.

ddtrace/internal/gitmetadata.py
sashacmc: It in almost all tests there (but yes, the test was slightly incorrect due to an invalid package name, thanks again for finding this mistake)
https://github.com/DataDog/dd-trace-py/pull/5395/files#diff-57e8f64c4323b270c792b2ece79b1a11a6cdad6f51bda5ea938c9c1fa49c7140R54

ddtrace/internal/gitmetadata.py
sashacmc: Unfortunately no, but caching logic trivial and not very critical and to be honest I don't know how properly test it without full refactoring of current tests (and module structure too)

ddtrace/internal/gitmetadata.py
sashacmc: I added test for package witch have no source_code_link, but for malformed url it is difficult to add (without test refactoring) and in the worst case in will just catched by "global" try/catch in get_git_tags method.

ddtrace/internal/gitmetadata.py
sashacmc: Ok, just for get_git_tags without full processing via tracer/profiler it is possible 👍

brettlangdon: yeah, a simple unit test should be more than enough to validate the caching

sashacmc: Done!

ddtrace/internal/gitmetadata.py
```diff
@@ -0,0 +1,126 @@
...
+    try:
+        try:
+            import importlib.metadata as importlib_metadata
+        except ImportError:
+            import importlib_metadata  # type: ignore[no-redef]
+
+        source_code_link = ""
+        for val in importlib_metadata.metadata(config.main_package).get_all("Project-URL"):
+            capt_val = val.split(", ")
+            if capt_val[0] == "source_code_link":
```
mabdinur, line 71: Is this check required? Or will capt_val always contain two elements if `capt_val[0] == "source_code_link"`? 

```suggestion
            if len(capt_val) > 1 and capt_val[0] == "source_code_link":
```

ddtrace/internal/gitmetadata.py
```diff
@@ -0,0 +1,126 @@
...
+                source_code_link = capt_val[1].strip()
+                break
+
+        if source_code_link and "#" in source_code_link:
+            repository_url, commit_sha = source_code_link.split("#")
+            commit_sha = commit_sha.split("&")[0]
```
mabdinur, line 77: Is this format always respected? Is there a scenario where the commit sha is the first value before an `&`?

mabdinur (approved):
Left some non blocking questions

ddtrace/internal/gitmetadata.py
sashacmc: It must contains two elements, but there's nothing wrong with checking

ddtrace/internal/gitmetadata.py
sashacmc: Yes, we generate it in this format

mabdinur (approved):

mabdinur (approved):

mabdinur (approved):
Squashed and signed all commits

PR #5401, opened 2023-03-23
fix(aws_lambda): fix circular dependency when user handler patches modules

duncanista:
Initially discovered in [datadog-lambda-python#313](https://github.com/DataDog/datadog-lambda-python/issues/313).

When patching, code can contain `ddtrace.patch_all()` or `ddtrace.patch(aws_lambda=True)`. Creating a circular dependency, triggering another patch. To avoid this, we skip on `AttributeError` which is the error thrown when patching a second time.

This was only affecting users patching again on their AWS Lambda functions instrumented by Datadog.

## Testing strategy

Tested on AWS Lambda functions. It was not possible to reproduce it in unit testing.

## Risk

There might be a better way to figure out when a circular dependency is happening.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/contrib/aws_lambda/patch.py
```diff
@@ -165,23 +165,37 @@ def patch():
...
-    handler, handler_module, wrapper = _get_handler_and_module()
+    try:
+        handler, handler_module, wrapper = _get_handler_and_module()
 
-    if getattr(handler_module, "_datadog_patch", False):
-        return
-    setattr(handler_module, "_datadog_patch", True)
+        if getattr(handler_module, "_datadog_patch", False):
+            return
+        setattr(handler_module, "_datadog_patch", True)
```
joeyzhao2018, line 173: Should we set the attr after wrap()?

duncanista: I think the order here doesn't matter, but what do you think?

joeyzhao2018: if the `wrap()` fail, it will stil set attr `_datadog_path` to True, right?

duncanista: You're right, i'll switch the spots of the unpatch too.

ddtrace/contrib/aws_lambda/patch.py
```diff
@@ -165,23 +165,37 @@ def patch():
...
         return
 
-    handler, handler_module, wrapper = _get_handler_and_module()
+    try:
+        handler, handler_module, wrapper = _get_handler_and_module()
 
-    if getattr(handler_module, "_datadog_patch", False):
-        return
-    setattr(handler_module, "_datadog_patch", True)
+        wrap(handler, wrapper)
```
joeyzhao2018, line 171: Actually, I was thinking this:
```
if getattr(handler_module, "_datadog_patch", False):
    return
wrap(handler, wrapper)        
setattr(handler_module, "_datadog_patch", True)
```

releasenotes/notes/fix-aws-lambda-circular-dep-when-patching-in-handler-320de20015a8d5cd.yaml
```diff
@@ -0,0 +1,4 @@
+---
+fixes:
+  - |
+    aws_lambda: Fix AttributeError raised when `ddtrace.patch_all()`, or `ddtrace.patch(aws_lambda=True)`, is set on user handler.
```
ZStriker19, line 4: can we add a newline character here?
```suggestion
    aws_lambda: Fix AttributeError raised when `ddtrace.patch_all()`, or `ddtrace.patch(aws_lambda=True)`, is set on user handler.
    
```

P403n1x87:
Is it not possible to break the circular import in the first place?

duncanista:
> Is it not possible to break the circular import in the first place?

Yes, once we stop being a contribution, and patching users code. 

On the current code, I tried marking the code as already patched before doing it, but it didn't show a difference, will try doing it again to see if it changes again, but currently, I couldn't find any other workaround.

joeyzhao2018 (dismissed):
LGTM.

emmettbutler (dismissed):

emmettbutler (approved):

PR #5421, opened 2023-03-28
fix(profiling): memalloc cleanup after final export

P403n1x87:
This change ensures that the _memalloc native module is cleaned up at the very end during the shutdown phase, after any last attempts are made to run the periodic flushing of profiling data. This is achieved by triggering the _memalloc clean-up as part of the on_shutdown callback, which is guaranteed to be invoked only after the last attempt to call the periodic function by the periodic thread.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


brettlangdon (commented):
are we able to reproduce in a test?

ddtrace/profiling/collector/memalloc.py
```diff
@@ -109,10 +109,9 @@ def _start_service(self):
 
         super(MemoryCollector, self)._start_service()
 
-    def _stop_service(self):
-        # type: (...) -> None
-        super(MemoryCollector, self)._stop_service()
-
+    @staticmethod
+    def on_shutdown():
```
Yun-Kim, line 113: Unless I missed something, this PR is not calling the `memalloc.stop()` on `stop()` anymore and only on shutdown - do we expect the `MemoryCollector` to be stopped without being shut down?

P403n1x87: `on_shutdown` is called whenever a periodic service is stopped. So stopping the collector will eventually call `on_shutdown`. The extra guarantee here is that it will be called _after_ the periodic function, which is the one that potentially requires `memalloc` to be in the "started" state.

Yun-Kim (approved):

PR #5427, opened 2023-03-29
chore(service name): initial schema for span attributes (service name)

tabgok:
### Special Note

This is the first of several changes dealing with service naming, gated behind a feature flag.  A future (feat) request will be made when the changes for service naming AND peer.service are complete.

### Introduction

Span service names are the wild west today, both within and across tracers, which has led to a variety of issues:

1. Lack of consistency in naming across tracer libraries in different languages (grpc-client vs grpc.client).
2. Inconsistency with operation names across integrations, even within a tracer. (requests vs httplib)
3. Too-generic service names act as naming 'sinks' (i.e. all databases for 'postgres' are represented as 'postgres')
4. Some clients (i.e. grpc) are represented as full-on services, which is generally unexpected.

The above issues occur because DataDog's features (such as metrics and flame-graphs) require tracer libraries to delineate between services appropriately, and the existing mechanism for doing so was specifying a separate service within an application (either manually or by integration default).  This issue is further compounded by a lack of standardization across DataDog's libraries and selection of some too-board fakes and defaults.

In order to remediate the issues, there are two efforts on-going to update the libraries:

1. Provide (and switch to) a secondary mechanism for specifying calls to outside services ([peer.service](https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/span-general/#general-remote-service-attributes))
2. Standardizing all tracer libraries to using the current application name as a default for existing services (service naming).

(1) allows DataDog's feature set to continue differentiating between internal and external services (using peer.service instead of service) while (2) removes the existing behavior and standardizes the new behavior across all tracers.

This PR deals with (2).

### What's changing in this PR?

This PR introduces the initial changes for "(2) Standardizing all tracer libraries to using the current application name as a default for existing services (service naming)".  The major changes in this revision are:

1. Set a default value for DD_SERVICE of `unnamed-python-service` instead of relying on the agent to do so
2. Create a new concept, 'span attribute schema,' which controls the service naming behavior and control it with a new environment variable `DD_TRACE_SPAN_ATTRIBUTE_SCHEMA`.
3. Create 'v0' (existing behavior) and 'v1' (new behavior)  schemas and behaviors:
    1. v0 (Default) - The existing behavior
    2. v1 - Default integration service names are now set to the value of
       `$DD_SERVICE` (config().service) instead of the prior default values
       (such as 'redis')

Notes:
* These changes only affect the *default* service name, and will not affect overridden values, such as user-defined values or those set by other environment variables such as `DD_REDIS_SERVICE`.
* v0 (current behavior) is also the default. Users will not get the new defaults unless they specify 'v1' for the env var DD_TRACE_SPAN_ATTRIBUTE_SCHEMA.
* $DD_<INTEGRATION>_SERVICE overrides are not affected, only the defaults
* The value of $DD_SERVICE is selected instead of parent.service
* A default value for DD_SERVICE has been added (unnamed-python-service) to prevent fallback for ^

### Risks
The feature flag gate must work appropriately to toggle between behaviors.  If it does not, users may face the following with existing applications due to the way metrics/etc are leveraged by DataDog:
* Alarms firing when they shouldn't
* Alarms NOT firing when they should
* Unexpected phase shifts in metrics
* Disappearance or shifts in services

### Checklist

- [x] Library release note guidelines](https://ddtrace.readthedocs.io/en/stable/contributing.html#Release-Note-Guidelines) are followed.

## Reviewer Checklist

- [x] Title is accurate.

### DIFF
--- a/ddtrace/contrib/redis/patch.py
+++ b/ddtrace/contrib/redis/patch.py
@@ -4,6 +4,7 @@ from six import PY3
 from ddtrace import config
 from ddtrace.vendor import wrapt
 
+from ...internal.schema import schematize_service_name
 from ...internal.utils.formats import stringify_cache_args
 from ...pin import Pin
 from ..trace_utils import unwrap
@@ -11,7 +12,12 @@ from .util import _trace_redis_cmd
 from .util import _trace_redis_execute_pipeline
 
 
-config._add("redis", dict(_default_service="redis"))
+config._add(
+    "redis",
+    {
+        "_default_service": schematize_service_name("redis"),
+    },
+)
 
 
 def patch():
--- a/ddtrace/internal/schema/__init__.py
+++ b/ddtrace/internal/schema/__init__.py
@@ -1,0 +1,24 @@
+import os
+
+from .span_attribute_schema import SPAN_ATTRIBUTE_SCHEMA_VERSIONS
+from .span_attribute_schema import _DEFAULT_SPAN_SERVICE_NAMES
+
+
+# Span attribute schema
+def _validate_schema(version):
+    error_message = (
+        "You have specified an invalid span attribute schema version: '{}'.".format(__schema_version),
+        "Valid options are: {}. You can change the specified value by updating".format(SPAN_ATTRIBUTE_SCHEMA_VERSIONS),
+        "the value exported in the 'DD_TRACE_SPAN_ATTRIBUTE_SCHEMA' environment variable.",
+    )
+
+    assert version in SPAN_ATTRIBUTE_SCHEMA_VERSIONS, error_message
+
+
+__schema_version = os.getenv("DD_TRACE_SPAN_ATTRIBUTE_SCHEMA", default="v0")
+_validate_schema(__schema_version)
+
+DEFAULT_SPAN_SERVICE_NAME = _DEFAULT_SPAN_SERVICE_NAMES[__schema_version]
+schematize_service_name = SPAN_ATTRIBUTE_SCHEMA_VERSIONS.get(__schema_version)
+
+__all__ = ["schematize_service_name", "DEFAULT_SPAN_SERVICE_NAME"]
--- a/ddtrace/internal/schema/span_attribute_schema.py
+++ b/ddtrace/internal/schema/span_attribute_schema.py
@@ -1,0 +1,23 @@
+SPAN_ATTRIBUTE_SCHEMA_VERSIONS = dict()
+
+_DEFAULT_SPAN_SERVICE_NAMES = {"v0": None, "v1": "unnamed-python-service"}
+
+
+def register_schema_version(version):
+    def _wrap(func):
+        assert version not in SPAN_ATTRIBUTE_SCHEMA_VERSIONS, "duplicate version entry for {}".format(version)
+        SPAN_ATTRIBUTE_SCHEMA_VERSIONS[version] = func
+
+    return _wrap
+
+
+@register_schema_version("v0")
+def service_name_v0(v0_service_name):
+    return v0_service_name
+
+
+@register_schema_version("v1")
+def service_name_v1(*_, **__):
+    from ddtrace import config as dd_config
+
+    return dd_config.service
--- a/tests/contrib/redis/test_redis.py
+++ b/tests/contrib/redis/test_redis.py
@@ -60,6 +60,14 @@ class TestRedisPatch(TracerTestCase):
         assert span.get_tag("span.kind") == "client"
         assert span.get_tag("db.system") == "redis"
 
+    @TracerTestCase.run_in_subprocess(env_overrides=dict(DD_TRACE_SPAN_ATTRIBUTE_SCHEMA="v1"))
+    def test_service_name_v1(self):
+        us = self.r.get("cheese")
+        assert us is None
+        spans = self.get_spans()
+        span = spans[0]
+        assert span.service == "unnamed-python-service"
+
     def test_basics(self):
         us = self.r.get("cheese")
         assert us is None
@@ -226,7 +234,7 @@ class TestRedisPatch(TracerTestCase):
         assert dd_span.resource == "GET cheese"
 
     @TracerTestCase.run_in_subprocess(env_overrides=dict(DD_SERVICE="mysvc"))
-    def test_user_specified_service(self):
+    def test_user_specified_service_default(self):
         from ddtrace import config
 
         assert config.service == "mysvc"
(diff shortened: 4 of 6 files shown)


tabgok:
Here's a diagram showing v0, v1, and high-level decision tree for service name for redis: [PDF](https://github.com/DataDog/dd-trace-py/files/11104905/dd-trace-py.service.selection.pdf)

ddtrace/internal/schema/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+
+# Span attribute schema
+
+__all__ = ["span_attribute_schema"]
+span_attribute_schema = None
+
+__schema_version = config.trace_span_attribute_schema or "v0"
+assert (
+    __schema_version in SPAN_ATTRIBUTE_SCHEMA_VERSIONS
+), f"Invalid span attribute schema version {__schema_version}.  Options are {SCHEMA_VERSIONS}"
```
ZStriker19, line 14: Under what circumstances could you imagine we fail this assert? Should we make the message more helpful for users to fix this?

tests/contrib/redis/test_redis.py
```diff
@@ -260,8 +289,53 @@ def test_env_user_specified_redis_service(self):
...
+        span = self.get_spans()[0]
+        assert span.service == "myredis", span.service
+
+        self.reset()
+
+        # Global config
+        with self.override_config("redis", dict(service="cfg-redis")):
+            from ddtrace import config
+
+            print(config.redis.service)
```
ZStriker19, line 306:
```suggestion
```
should we remove this print?

tests/contrib/redis/test_redis_cluster.py
```diff
@@ -32,6 +32,13 @@ def tearDown(self):
         unpatch()
         super(TestRedisClusterPatch, self).tearDown()
 
+    @TracerTestCase.run_in_subprocess(env_overrides=dict(DD_TRACE_SPAN_ATTRIBUTE_SCHEMA="v1"))
+    def test_span_service_name_v1(self):
+        us = self.r.get("cheese")
+        spans = self.get_spans()
+        span = spans[0]
+        assert span.service == "unnamed-python-service"
```
ZStriker19, line 40: Trying to understand what this will look like down the road once implemented on all integrations. Essentially if you don't set `DD_SERVICE` or any other service naming, and did `DD_TRACE_SPAN_ATTRIBUTE_SCHEMA="v1"` then you'd end up with everything (web app, databases, messaging libraries) reporting data under `service:unnamed-python-service` correct?

tests/contrib/redis/test_redis_cluster.py
```diff
@@ -141,14 +148,56 @@ def test_user_specified_service(self):
...
+
+        spans = self.get_spans()
+        assert len(spans) == 1
+        span = spans[0]
+        assert span.service == "mysvc"
+
+    @TracerTestCase.run_in_subprocess(
+        env_overrides=dict(DD_REDIS_SERVICE="myrediscluster", DD_TRACE_SPAN_ATTRIBUTE_SCHEMA="v0")
+    )
+    def test_env_user_specified_rediscluster_service_v0(self):
```
ZStriker19, line 174: Nit: You could probably parametrize all of these if you wanted to avoid code reuse/clutter, e.g. https://github.com/DataDog/dd-trace-py/blob/1.x/tests/tracer/test_propagation.py#L451-L476

But up to you!

ZStriker19:
Hi @tabgok! For the title, if this is a `feat` we'll want a release note. Unless it's going GA with this PR, or quite soon we may want to switch it to `chore` instead. That's usually what we do for yet unreleased features or those that will need further PRs before release. If you want to keep it `feat` though, please add a release note following the guidance here: https://ddtrace.readthedocs.io/en/stable/releasenotes.html#generating-a-release-note

Is this work for the effort of removing "fake" service names?

ddtrace/internal/schema/__init__.py
tabgok: Updated with longer message

tests/contrib/redis/test_redis.py
tabgok: removed

tests/contrib/redis/test_redis_cluster.py
tabgok: That is correct, but its only half the equation.  These changes will be released 'public announcement' in conjunction with updates for peer.service.

What this means: the current connection code here will report under 'service:unnamed-python-service' but each of the connections will have a child span with 'service:$peer.service'.

tabgok:
Hi @ZStricker19 - Changing to 'chore' for now, as it won't be an official feature until both all changes are made for service naming AND for peer.service.

Also, yes, this is one of the pieces required to remove fake service names!

ddtrace/contrib/redis/patch.py
```diff
@@ -4,14 +4,20 @@
 from ddtrace import config
 from ddtrace.vendor import wrapt
 
+from ...internal.schema import span_attribute_schema
```
tabgok, line 7: This is a bit of magic; I'm importing an instantiated instance of the schema version class.  Is this ok?

ZStriker19: Checked with team: 
From Brett: "If we can avoid global instances let's do it.
Is it possible that we don't need a class/instance for implementing that?"

tabgok: I believe I can make them functions instead of classes/objects

tests/contrib/redis/test_redis_cluster.py
tabgok: I looked into this and didn't find an easy way to parametrize a decorator =/

ddtrace/internal/schema/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+    error_message = (
+        "You have specified an invalid span attribute schema version: '{}'.".format(__schema_version),
+        "Valid options are: {}. You can change the specified value by updating".format(SPAN_ATTRIBUTE_SCHEMA_VERSIONS),
+        "the value exported in the 'DD_TRACE_SPAN_ATTRIBUTE_SCHEMA' environment variable.",
+    )
+
+    assert version in SPAN_ATTRIBUTE_SCHEMA_VERSIONS, error_message
+
+
+__schema_version = os.getenv("DD_TRACE_SPAN_ATTRIBUTE_SCHEMA", default="v0")
```
ZStriker19, line 18: Are v0 and v1 going to be used across tracers?

ddtrace/internal/schema/span_attribute_schema.py
```diff
@@ -0,0 +1,23 @@
+SPAN_ATTRIBUTE_SCHEMA_VERSIONS = dict()
+
+_DEFAULT_SPAN_SERVICE_NAMES = {"v0": None, "v1": "unnamed-python-service"}
+
+
+def register_schema_version(version):
+    def _wrap(func):
+        assert version not in SPAN_ATTRIBUTE_SCHEMA_VERSIONS, "duplicate version entry for {}".format(version)
```
ZStriker19, line 8: Trying to think about this, under what circumstance could we end up with a duplicate version entry? Why do we care?

tests/contrib/redis/test_redis.py
```diff
@@ -260,8 +311,26 @@ def test_env_user_specified_redis_service(self):
         span = self.get_spans()[0]
         assert span.service == "mysvc", span.service
 
-    @TracerTestCase.run_in_subprocess(env_overrides=dict(DD_SERVICE="app-svc", DD_REDIS_SERVICE="env-redis"))
-    def test_service_precedence(self):
+    @TracerTestCase.run_in_subprocess(
+        env_overrides=dict(DD_SERVICE="app-svc", DD_REDIS_SERVICE="env-redis", DD_TRACE_SPAN_ATTRIBUTE_SCHEMA="v0")
```
ZStriker19, line 315: nit: `env-redis` is a pretty strange service name, since `env` is its own tag. We don't want anyone to think they should set a service name like that. Maybe just switch it to some like `redis-custom-serice` ? Not sure why that made it in before. Also this probably from fear instilled in me from old support days and service naming chaos.
```suggestion
        env_overrides=dict(DD_SERVICE="app-svc", DD_REDIS_SERVICE="redis-custom-service", DD_TRACE_SPAN_ATTRIBUTE_SCHEMA="v0")
```

ddtrace/internal/schema/__init__.py
tabgok: yes, 'v0' and 'v1' will be consistent across tracers

ddtrace/internal/schema/span_attribute_schema.py
tabgok: This may be better as a test; the idea is to prevent someone copy+pasting for a new version and forgetting to update the tag

ZStriker19 (approved):

PR #5434, opened 2023-03-29
chore(tracing): adds 128 bit trace id support for b3 and w3c trace headers

mabdinur:
Currently the trace id component of b3 and w3c headers are truncated from 128bits to 64bits. This was done to make b3 and w3c headers compatible with datadog tracers. Moving forward this is not an acceptable solution. 

When ``DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED=True`` the tracer should NOT truncate the trace_id in b3 and w3c headers. The full trace id should be used.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


tests/tracer/test_propagation.py
```diff
@@ -795,27 +803,27 @@ def test_extract_tracestate(caplog, ts_string, expected_tuple, expected_logging,
...
-                "trace_id": 11803532876627986230,
+                "trace_id": TRACE_ID,
                 "span_id": 67667974448284343,
                 "meta": {
                     "tracestate": "dd=s:2;o:rum;t.dm:-4;t.usr.id:baz64,congo=t61rcWkgMzE",
                     "_dd.p.dm": "-4",
                     "_dd.p.usr.id": "baz64",
                     "_dd.origin": "rum",
-                    "traceparent": "00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01",
+                    "traceparent": TRACECONTEXT_HEADERS_VALID_BASIC[_HTTP_HEADER_TRACEPARENT],
```
mabdinur, line 813:
```suggestion
                    "traceparent": TRACECONTEXT_HEADERS_VALID[_HTTP_HEADER_TRACEPARENT],
```

ddtrace/propagation/http.py
```diff
@@ -118,9 +120,11 @@ def _hex_id_to_dd_id(hex_id):
 
 def _dd_id_to_b3_id(dd_id):
     # type: (int) -> str
-    """Helper to convert Datadog trace/span int ids into hex ids"""
-    # DEV: `hex(dd_id)` will give us `0xDEADBEEF`
-    # DEV: this gives us lowercase hex, which is what we want
+    """Helper to convert Datadog trace/span int ids into lower case hex values"""
+    if dd_id > _MAX_UINT_64BITS:
+        # b3 trace ids can have the length of 16 or 32 hex values:
```
mabdinur, line 125:
```suggestion
        # b3 trace ids can have the length of 16 or 32 characters:
```

tests/tracer/test_propagation.py
```diff
@@ -238,25 +258,44 @@ def test_extract_128bit_trace_ids():
...
+        style = os.getenv("DD_TRACE_PROPAGATION_STYLE")
+        if style == "datadog":
+            # Get the hex representation of the 64 most signicant bits
+            trace_id_64bit = trace_id & 2 ** 64 - 1
+            headers = {
+                "x-datadog-trace-id": str(trace_id_64bit),
+                "x-datadog-parent-id": str(span_id),
+                "x-datadog-tags": "=".join([HIGHER_ORDER_TRACE_ID_BITS, trace_id_hex[:16]]),
+            }
+        elif style == "b3multi":
```
ZStriker19, line 275: rather than all of the if elifs to check the style and assert certain headers, could we parametrize the expected values for each header style, like how we do in the other tests? Or does that not work here?

tests/tracer/test_propagation.py
```diff
@@ -456,15 +495,22 @@ def test_get_wsgi_header(tracer):
...
     _HTTP_HEADER_TRACESTATE: "dd=s:2;o:rum",
 }
 
 TRACECONTEXT_HEADERS_VALID = {
-    _HTTP_HEADER_TRACEPARENT: "00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01",
+    _HTTP_HEADER_TRACEPARENT: "00-80f198ee56343ba864fe8b2a57d3eff7-00f067aa0ba902b7-01",
     _HTTP_HEADER_TRACESTATE: "dd=s:2;o:rum;t.dm:-4;t.usr.id:baz64,congo=t61rcWkgMzE",
 }
 
+if asbool(os.getenv("DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED")):
```
ZStriker19, line 507: I think I'm missing something, how does putting this here affect the tests? Doesn't this code only run once since it's not in any method or parametrize?

mabdinur: This is a bit subtle. We run the whole tracer suite with [DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED=True and DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED=False](https://github.com/DataDog/dd-trace-py/blob/1.x/riotfile.py#L280). This ensures all tracer tests are compatible with 128bit trace ids. When 128bit trace ids are enabled by default (hopefully this happens by v1.13.0), I will remove this env var from our riot configurations.

ddtrace/propagation/http.py
```diff
@@ -110,6 +110,8 @@ def _hex_id_to_dd_id(hex_id):
     If the id is > 64 bit then truncate the trailing 64 bit.
```
mabdinur, line 110:
```suggestion
    If the id is > 64 bit and 128bit trace_id generation is disabled then truncate the trailing 64 bit.
    Otherwise convert the full trace id into lower case hex values.
```

tests/tracer/test_propagation.py
ZStriker19: Ah I missed that, understood!

tests/tracer/test_propagation.py
```diff
@@ -51,35 +51,56 @@ def test_inject(tracer):
...
+        with tracer.trace("global_root_span") as span:
+            assert span.trace_id == trace_id
+
+            headers = {}
+            HTTPPropagator.inject(span.context, headers)
+
+            style = os.getenv("DD_TRACE_PROPAGATION_STYLE")
+            trace_id_hex = "{:032x}".format(span.trace_id)
+            span_id_hex = "{:016x}".format(span.span_id)
+            if style == "datadog":
```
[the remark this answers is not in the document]
mabdinur: We could do this if we use the pattern in the other propagation tests (ie: use pytest fixtures and override the `config` object). However this wouldn't be a full integration test. Here I am not directly testing the config object. Instead I am directly testing `DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED` configuration. 

I could use parametrize with subprocess tests but it gets messy (all parameters are set an environment variables, here I would have to convert a dictionary into a string and then back to a dict). Also subprocess tests only support one environment variable so I would need to use a few hacks to get this to work. 

Would you recommend splitting up this test into multiple tests: `test_inject_128bit_trace_id_b3`, `test_inject_128bit_trace_id_b3_multi`, `test_inject_128bit_trace_id_tracecontext`, etc.

tests/tracer/test_propagation.py
mabdinur: Same comment as above

ddtrace/propagation/http.py
```diff
@@ -107,9 +107,12 @@ def _extract_header_value(possible_header_names, headers, default=None):
 def _hex_id_to_dd_id(hex_id):
     # type: (str) -> int
     """Helper to convert hex ids into Datadog compatible ints
-    If the id is > 64 bit then truncate the trailing 64 bit.
+    If the id is > 64 bit and 128bit trace_id generation is disabled then truncate the trailing 64 bit.
+    Otherwise convert the full trace id into lower case hex values.
     "463ac35c9f6413ad48485a3953bb6124" -> "48485a3953bb6124" -> 5208512171318403364
```
mabdinur, line 112:
```suggestion
    If the id is > 64 bit and 128bit trace_id generation is disabled then truncate the trailing 64 bit.
    "463ac35c9f6413ad48485a3953bb6124" -> "48485a3953bb6124" -> 5208512171318403364
    Otherwise convert the full trace id into lower case hex values.
```

tests/tracer/test_propagation.py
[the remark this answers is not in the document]
ZStriker19: Ok that's what I thought. In that case, yeah I think breaking it up is probably preferable.

PR #5437, opened 2023-03-30
chore(debugging): add timeout to context capturing

P403n1x87:
This change adds a timeout to the process of capturing values into snapshots to ensure that it doesn't run for an arbitrary amount of time, thus pausing user code for a potentially long time.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/debugging/_capture/utils.py
```diff
@@ -140,14 +141,20 @@ def capture_exc_info(exc_info):
...
+def capture_value(value, level=MAXLEVEL, maxlen=MAXLEN, maxsize=MAXSIZE, maxfields=MAXFIELDS, hg=None):
+    # type: (Any, int, int, int, int, Optional[HourGlass]) -> Dict[str, Any]
     _type = type(value)
 
     if _type in BUILTIN_SIMPLE_TYPES:
         if _type is NoneType:
             return {"type": "NoneType", "isNull": True}
 
+        if hg is not None and not hg.trickling():
+            return {
```
shatzi, line 153: Maybe there is some room to reduce code duplication.

1. have utility function create_capture_value* that generate that json like create_capture_value_not_captured(_type, "timeout", [size]).
2. make those string as constant NOT_CAPTURED_REASON = "notCapturedReason"

tests/debugging/test_encoding.py
```diff
@@ -265,3 +266,27 @@ def test_format_message(args, expected):
 
 def test_encoding_none():
     assert utils.capture_value(None) == {"isNull": True, "type": "NoneType"}
+
+
+def test_encoding_timeout():
```
shatzi, line 271: While I love the idea here, I think this can be flaky. I suggest create a mocked version hg that return false after X invocations to avoid time flakiness

ddtrace/internal/utils/time.py
```diff
@@ -101,3 +101,37 @@ def stop(self):
...
+        top_0 = self._end_at - self._started_at
+        bottom = self._duration - top_0 + min(t - self._started_at, top_0)
+
+        self._started_at = t
+        self._end_at = t + bottom
+
+    def trickling(self):
+        # type: () -> bool
+        """Check if sand is still trickling."""
+        return compat.monotonic() < self._end_at
```
shatzi, line 129: what is cost of calling `compat.monotonic()` ? we might want to cache trickling as once this is true it should never change.

shatzi (approved):

ddtrace/internal/utils/time.py
P403n1x87: monotonic is implemented in native code, so should be quite cheap

tests/tracer/test_utils.py
```diff
@@ -512,3 +513,25 @@ def _():
...
     for tag in expected_strs:
         assert tag in w3c_get_dd_list_member(context)
+
+
+def test_hourglass_init():
+    """Test that we get the full duration on initialization."""
+    with time.HourGlass(1) as hg, time.StopWatch() as sw:
+        while hg.trickling():
+            sleep(0.1)
+    assert sw.elapsed() > 0.9
```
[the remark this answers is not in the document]
P403n1x87: just a margin of error to reduce potential flakiness

ddtrace/debugging/_capture/utils.py
P403n1x87: I think these structures are simple enough that we can avoid the cost of an extra function call and a global variable lookup. WDYT?

ddtrace/debugging/_capture/utils.py
```diff
@@ -191,7 +205,7 @@ def capture_value(value, level=MAXLEVEL, maxlen=MAXLEN, maxsize=MAXSIZE, maxfiel
             data = {
                 "type": qualname(_type),
                 "elements": [
-                    capture_value(v, level=level - 1, maxlen=maxlen, maxsize=maxsize, maxfields=maxfields)
+                    capture_value(v, level=level - 1, maxlen=maxlen, maxsize=maxsize, maxfields=maxfields, hg=hg)
```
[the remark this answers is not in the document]
P403n1x87: functional programming :)
~~~python
takewhile(lambda _: hg.trickling(), collection)
~~~

tests/debugging/test_encoding.py
P403n1x87: I think that with steps by an order of magnitude we should be fine here. If this becomes an issue we can refactor it. How does that sound?

ddtrace/debugging/_capture/utils.py
shatzi: I don't like magic strings duplicated in multiple places in the code. 
However, this is only file we use those and I don't think this is going to chance any time soon. 
So, I will trust your judgement on what is more maintainable.

ddtrace/debugging/_capture/utils.py
```diff
@@ -140,14 +142,24 @@ def capture_exc_info(exc_info):
     }
 
 
-def capture_value(value, level=MAXLEVEL, maxlen=MAXLEN, maxsize=MAXSIZE, maxfields=MAXFIELDS):
-    # type: (Any, int, int, int, int) -> Dict[str, Any]
+def capture_value(value, level=MAXLEVEL, maxlen=MAXLEN, maxsize=MAXSIZE, maxfields=MAXFIELDS, hg=None):
+    # type: (Any, int, int, int, int, Optional[HourGlass]) -> Dict[str, Any]
+    def time_permits(_=None):
```
shatzi, line 147: why do we need '_=None' ?

tests/debugging/test_encoding.py
shatzi: I get the point of see if this flaky and fix. but the more I think about it - it doesn't test the functionality of the class.

I think we need to have tests to validate:
1. we stop capture in middle of a list.
2. we stop capture in a middle of a dict
3. we stop capture in the middle of an object (fields).

combined with simplified interface for captured value (cancellation function) I think we can change this test to validate those condition specifically and testing random exit point using time.

ddtrace/debugging/_capture/utils.py
```diff
@@ -140,14 +142,24 @@ def capture_exc_info(exc_info):
     }
 
 
-def capture_value(value, level=MAXLEVEL, maxlen=MAXLEN, maxsize=MAXSIZE, maxfields=MAXFIELDS):
-    # type: (Any, int, int, int, int) -> Dict[str, Any]
+def capture_value(value, level=MAXLEVEL, maxlen=MAXLEN, maxsize=MAXSIZE, maxfields=MAXFIELDS, hg=None):
+    # type: (Any, int, int, int, int, Optional[HourGlass]) -> Dict[str, Any]
```
shatzi, line 146: I like the time_permits function - maybe we can simplify this interface and instead of Optional[HourGlass] have a cancellation callback that stops this loop. basically pass time_permits instead of hg.

The problem I have with this function is that cancellation callback not necessarily is a timeout. so that function might return string with the reason.

so, in snapshot it would be: capture_value(..., lambda () => return "timeout" if ! hg.trickling() else None)

WDYT?

shatzi (dismissed):
Awesome changes. few comments but looks great.

ddtrace/debugging/_capture/utils.py
P403n1x87: so that we can call it with arguments (for the `takewhile`) and without (for the other usages in the code).

ddtrace/debugging/_capture/utils.py
P403n1x87: 👍  I'll have a think about this.

tests/debugging/test_encoding.py
```diff
@@ -265,3 +265,87 @@ def test_format_message(args, expected):
...
+                            "size": 1,
+                        },
+                    )
+                ],
+                "size": 1,
+            },
+        ),
+    ],
+)
+def test_encoding_stopping_cond(count, result):
```
shatzi, line 334: Do we want validate all other edge cases (too many fields, to many items, timeout in the middle of a collection) like we did for serialize function?

Maybe another ticket?

shatzi (dismissed):
looks good. few comments

ddtrace/debugging/_capture/snapshot.py
```diff
@@ -32,21 +36,26 @@ def _capture_context(
...
+            }
+            if arguments is not None
+            else {},
+            "locals": {
+                n: utils.capture_value(v, limits.max_level, limits.max_len, limits.max_size, limits.max_fields, timeout)
+                for n, v in _locals
+            }
+            if _locals is not None
+            else {},
+            "throwable": utils.capture_exc_info(throwable),
```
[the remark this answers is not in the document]
P403n1x87: Good question. I'd expect exception arguments to be simple objects, so serialisation shouldn't take long. Perhaps still worth trying to get more info about exceptions so best not to timeout on this?

ddtrace/debugging/_capture/utils.py
```diff
@@ -172,32 +182,65 @@ def capture_value(value, level=MAXLEVEL, maxlen=MAXLEN, maxsize=MAXSIZE, maxfiel
...
-                    capture_value(v, level=level - 1, maxlen=maxlen, maxsize=maxsize, maxfields=maxfields)
-                    for _, v in zip(range(maxsize), value)
-                ],
+                "elements": collection,
                 "size": len(value),
             }
 
-        if len(value) > maxsize:
+        if len(collection) < min(maxsize, len(value)):
+            data["notCapturedReason"] = cond.__name__
```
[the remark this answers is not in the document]
P403n1x87: ah yep, good point!

tests/debugging/test_encoding.py
P403n1x87: adding

majorgreys (dismissed):

shatzi (dismissed):
LGTM - awesome work

majorgreys:
@P403n1x87 Looks like this test is proving to be flaky: https://app.circleci.com/pipelines/github/DataDog/dd-trace-py/32137/workflows/5f60e456-937b-43db-8f4f-7f0af212558b/jobs/2174345

tests/tracer/test_utils.py
```diff
@@ -512,3 +513,29 @@ def _():
...
+
+
+def test_hourglass_turn():
+    with time.HourGlass(1) as hg, time.StopWatch() as sw:
+        # We let 100ms trickle down before turning.
+        sleep(0.1)
+        hg.turn()
+        while hg.trickling():
+            sleep(0.1)
```
P403n1x87, line 532:
```suggestion
            sleep(0.01)
```

majorgreys (approved):

ddtrace/debugging/_capture/snapshot.py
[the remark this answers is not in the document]
shatzi: I agree. if we got an exception we better collect more data. but there are few points I believe we need to think about:
1. timeout is to prevent the program from getting stuck - so I don't see why capture exception doesn't need this protection. I'm open in increasing the timeout if you think it is too restrictive, but 200ms is long enough right?
2. if we believe we want to capture more data. why are using serialize(v) with depth=2? should we capture those as tree so the user can inspect those?

PR #5443, opened 2023-03-30
chore(aws_lambda): log only on cold starts

duncanista:
Update logs to be strictly debug.

We don't want users disabling APM just because a warning/error in the patching happened.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/contrib/aws_lambda/patch.py
```diff
@@ -179,7 +179,7 @@ def patch():
         # which might cause a circular dependency. Skipping.
         return
     except Exception:
-        log.exception("Error patching AWS Lambda")
+        log.debug("Error patching AWS Lambda", exc_info=True)
```
mabdinur, line 182: I don't think the traceback will add a lot of value here. It could be noisy.
```suggestion
        log.debug("Error patching AWS Lambda")
```

ddtrace/contrib/aws_lambda/patch.py
```diff
@@ -199,5 +199,5 @@ def unpatch():
     except AttributeError:
         return
     except Exception:
-        log.exception("Error unpatching AWS Lambda")
+        log.debug("Error patching AWS Lambda", exc_info=True)
```
mabdinur, line 202: Can we add more details to this debug log. It's not clear what the impact is. Is the impact no longer tracing lambdas that timeout?

```suggestion
        log.debug("Error patching AWS Lambda")
```

releasenotes/notes/update-lambda-logs-to-debug-6139084ba50e75bb.yaml
```diff
@@ -0,0 +1,4 @@
+---
+other:
+  - |
+    aws_lambda: change logs to be debug, instead of warning or exception.
```
mabdinur, line 4:
```suggestion
    aws_lambda: Updates the level of warning and exception logs to debug. 
```

duncanista:
> Can we add more context to the PR description. Is there a conversation or github issue we can link to?
https://github.com/DataDog/datadog-lambda-python/issues/308

This is the conversation. I'll add more context to the PR description.
 
> Are there known issues that are causing the uptick in error and warnings logs? If so, I think those issues should be addressed instead of lowering the level of the log messages. This will just hide bugs.

Not known issues. We were able to reproduce it and trying to find a bugfix for this. Surprisingly, it only happens in Python. We are aware of how this could be perceived as hiding bugs. But we have proven that this warn was occurring during invocations who were not close to timeout. Hence, lowering the log level to debug, so users don't think our packages are introducing regressions.

ddtrace/contrib/aws_lambda/patch.py
duncanista: I decided to add it since we might want to have users / support engineers troubleshooting AWS Lambda and knowing what specifically failed.

duncanista:
Updated chore to now log only on cold starts when something unexpected happens.

releasenotes/notes/update-lambda-logs-to-debug-6139084ba50e75bb.yaml
```diff
@@ -0,0 +1,4 @@
+---
+other:
+  - |
+    aws_lambda: Updates log warnings and exceptions on cold start only.
```
mabdinur, line 4:
```suggestion
    aws_lambda: Logs warnings and exceptions on cold start only.
```

ddtrace/contrib/aws_lambda/patch.py
```diff
@@ -71,7 +73,8 @@ def _crash_flush(self, _, __):
             root_span.set_tag_str(ERROR_MSG, "Datadog detected an Impending Timeout")
             root_span.set_tag_str(ERROR_TYPE, "Impending Timeout")
         else:
-            log.warning("An impending timeout was reached, but no root span was found. No error will be tagged.")
+            if is_cold_start():
+                log.warning("An impending timeout was reached, but no root span was found. No error will be tagged.")
```
mabdinur, line 77: non blocking suggestion: To reduce duplication we can can define a LambdaLogger that checks `is_cold_start()` when `log.log(.....)` is called. This would be easier to maintain than making sure every call to `log.warning`, `log.error`, and `log.exception` is nested in a `is_cold_start()` check.  Or at least we can move this check to a helper function:

ex:

```
def lambda_logger(log, level, message):
    if is_cold_start():
        log.(level, message)
```

mabdinur (approved):

ddtrace/contrib/aws_lambda/patch.py
duncanista: Thanks @mabdinur, I wanted to wait until we had tons of places using this, but since we're already past that (3 times), I will definitely take your sugestion.

majorgreys:
@duncanista Can you clarify for me if this is a bug fix or not? It seems like you were able to reproduce an issue for which this is a resolution. If it is a fix, we should use `fix` rather than `chore`, update release note section, and be sure to backport it where necessary.

duncanista:
@majorgreys this is not a bug fix, this is just to avoid logging on every Lambda invocation, but on cold start invocations. 
Sorry if my phrasing was misleading. I meant to say that there is an active bug which we haven't been able to reproduce, but we found some AWS Lambdas in the Serverless self-monitoring app showing lots of logs, which this will reduce, not fix.

We are reducing the amount of logs since a million invocations, will cause a million warn logs. Which for users translates to more money. The serverless team is actively searching how to fix the bug in which we warn that an impending timeout was reached, but in reality, a timeout didn't actually happened.

ddtrace/contrib/aws_lambda/patch.py
duncanista: Will do on another PR after coming back from PTO.

ddtrace/contrib/aws_lambda/patch.py
```diff
@@ -12,7 +12,25 @@
...
+    def exception(self, msg, *args, exc_info=True, **kwargs):
+        if self.is_cold_start:
+            self.logger.error(msg, *args, exc_info=exc_info, **kwargs)
+
+    def warning(self, msg, *args, **kwargs):
+        if self.is_cold_start:
+            self.logger.warning(msg, *args, **kwargs)
+
+
+log = DDLambdaLogger()
```
joeyzhao2018, line 33: I wonder how come this doesn't complain about not finding `datadog_lambda` module....

ddtrace/contrib/aws_lambda/patch.py
```diff
@@ -11,8 +11,27 @@
 from ddtrace.internal.wrapping import unwrap
 from ddtrace.internal.wrapping import wrap
 
+from .cold_start import is_cold_start
+from .cold_start import set_cold_start
 
-log = get_logger(__name__)
+
+class DDLambdaLogger:
```
majorgreys, line 18: If this is not a class users would have to interface with, we should avoid making it part of the public interface. Similarly, the `cold_start` imports above will be in the public interface for this package, and so should be avoided. It might make better sense to move all this code to an internal `_cold_start.py` file.

ddtrace/contrib/aws_lambda/cold_start.py
```diff
@@ -0,0 +1,18 @@
+__cold_start = True
```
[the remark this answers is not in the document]
duncanista: We're trying to avoid global variables from the package `datadog-lambda` which have the name: `_cold_start`, in order to avoid the conflict, I've decided to add an extra underscore. Would it be better to rename this variable to be something like `_ddtrace_cold_start`?

majorgreys (approved):

mabdinur (approved):

PR #5446, opened 2023-03-30
feat(ci-visibility): refactor CI Visibility into its own namespace

Yun-Kim:
### Summary
This PR involves:
- Introducing a new `ddtrace.internal.ci_visibility` module to house CI Visibility-specific code, and  a new `CIVisibility` service to encapsulate CI Visibility information and logic
- Makes the `TraceCiVisibilityFilter` a private filter (and deprecate the public `TraceCiVisibilityFilter` for now until we remove it in 2.0.0) and handle the span processing for test spans (adding CI Visibility-specific tags, `dd_origin`, etc)
- Removes usage of `Pin` objects in the `pytest` and `pytest-bdd` integrations as the `CIVisibility` service keeps its own references to a `tracer` object and `config` dictionary.

### Details

This PR builds on top of #3726  by introducing the `ddtrace.internal.ci_visibility` module, which now contains extracted code enabling and disabling CI Visibility in `ddtrace.contrib.pytest.plugin.py` and `ddtrace.contrib.pytest_bdd.plugin.py`.

This PR also refactors the extracted code into a new `CIVisibility` service, which handles all CI Visibility-specific logic and can be enabled and disabled in the `pytest` and `pytest-bdd` plugins/integrations. The `CIVisibility` service does this by keeping a reference to a tracer and configures a `TraceCiVisibilityFilter` filter into the tracer's trace filters, which handles the span processing for CI Visibility. 

Additionally, as the `CIVisibility` service is no longer tied to a specific integration, the usage of `Pin` objects in the `pytest` and `pytest-bdd` integrations were removed as well as from the `CIVisibility` service, which keeps its own reference to a `tracer` and `config` dictionary. The `Pin` class is traditionally used for integrations, but since CIVisibility is meant to be run only for a test session (execution of a test command) in CI, the corresponding tracer will only be configured for that specific test session (CI Job) and shutdown once the test suite is finished.

*Note*: the introduced `CIVisibility` service and `TraceCiVisibilityFilter` filter are both internal and non-public, meaning that there are no changes to the public API. This PR introduces no new functionality as the `CIVisibility` service is just an abstraction of CI Visibility-specific logic that was already present. 

### CI Visibility Mode Architecture

Now, CI Visibility mode can be enabled and disabled by:
```python
from ddtrace.internal.ci_visibility import CIVisibility
CIVisibility.enable()  # can specify tracer, integration_config, service_name
...
CIVisibility.disable()
```

Behind the scenes, the `CIVisibility` service stores the following:
- tracer instance
- integration config
- CI Visibility-specific tags
- service name
When the `CIVisibility` service is enabled, it reconfigures the stored tracer instance to add a `TraceCiVisibilityFilter`, which then performs the CI Visibility-specific logic (setting tags, dd_origin, sampling_priority, etc). 

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/filters.py
```diff
@@ -75,6 +75,11 @@ def process_trace(self, trace):
         return trace
 
 
+@removals.removed_class(
+    "TraceCiVisibilityFilter",
+    message="TraceCiVisibilityFilter is deprecated and will be removed from the public API.",
+    removal_version="2.0.0",
+)
 class TraceCiVisibilityFilter(TraceFilter):
```
mabdinur, line 83: I am not a big fan of introducing two definitions of `TraceCiVisibilityFilter` even if one has a deprecation warning. 

suggestion: Can we make `ddtrace.ci_visibility._filters.TraceCiVisibilityFilter` backwards compatible and import it in this module. With this approach we might to update the `removals.removed_class` annotation.

ddtrace/filters.py
```diff
@@ -75,6 +75,11 @@ def process_trace(self, trace):
         return trace
 
 
+@removals.removed_class(
+    "TraceCiVisibilityFilter",
+    message="TraceCiVisibilityFilter is deprecated and will be removed from the public API.",
```
mabdinur, line 80: Can we make this deprecation warning actionable? We should document the `CIVisibility` service and how it can be enabled.

ddtrace/ci_visibility/__init__.py
```diff
@@ -0,0 +1,4 @@
+from ddtrace.ci_visibility._recorder import CIVisibility
+
+
+__all__ = ["CIVisibility"]
```
mabdinur, line 4: Should we make `CIVisibility` an internal component? It is enabled by the pytest and pytestbdd plugins and not by users.

If we decide to make this component public we should add documentation

ddtrace/ci_visibility/_filters.py
```diff
@@ -0,0 +1,40 @@
...
+        if not trace:
+            return trace
+
+        local_root = trace[0]._local_root
+        if not local_root or local_root.span_type != SpanTypes.TEST:
+            return None
+
+        local_root.service = self._service
+        local_root.context.dd_origin = ci.CI_APP_TEST_ORIGIN
+        local_root.context.sampling_priority = AUTO_KEEP
```
mabdinur, line 35:
```suggestion
        if local_root.context is not None:
            local_root.context.dd_origin = ci.CI_APP_TEST_ORIGIN
            local_root.context.sampling_priority = AUTO_KEEP
```

ddtrace/ci_visibility/_recorder.py
```diff
@@ -0,0 +1,143 @@
...
+from ddtrace.settings import IntegrationConfig
+
+
+log = get_logger(__name__)
+
+
+def _extract_repository_name_from_url(repository_url):
+    # type: (str) -> str
+    try:
+        return parse.urlparse(repository_url).path.rstrip(".git").rpartition("/")[-1]
```
mabdinur, line 26: There's some duplication between the [utils](https://github.com/DataDog/dd-trace-py/blob/1.x/ddtrace/sourcecode/_utils.py) and [constants](https://github.com/DataDog/dd-trace-py/blob/1.x/ddtrace/ext/git.py) in the sourcecode integration product. Can we consolidate some of these?

emmettbutler (dismissed):

ddtrace/ci_visibility/_recorder.py
```diff
@@ -0,0 +1,143 @@
...
+        # type: () -> None
+        tracer_filters = self.tracer._filters
+        if not any(isinstance(tracer_filter, TraceCiVisibilityFilter) for tracer_filter in tracer_filters):
+            tracer_filters += [TraceCiVisibilityFilter(self._tags, self._service)]  # type: ignore[arg-type]
+            self.tracer.configure(settings={"FILTERS": tracer_filters})
+
+    def _stop_service(self):
+        # type: () -> None
+        try:
+            self.tracer.shutdown()
```
mabdinur, line 110: Do we need to shutdown the tracer here? We'll attempt to shutdown the tracer every time a process forks or shuts down. I'm not a big fan of this tight coupling between the CIVisibility service and the AgentWriter. Is there another workaround?

ddtrace/ci_visibility/_recorder.py
```diff
@@ -0,0 +1,143 @@
...
+        except Exception:
+            log.warning("Failed to shutdown tracer", exc_info=True)
+
+    @classmethod
+    def _restart(cls):
+        # type: () -> None
+        log.info("Restarting %s in child process", cls.__name__)
```
mabdinur, line 117: nit: _reset can be called out of a new process
```suggestion
        log.info("Restarting %s", cls.__name__)
```

ddtrace/ci_visibility/_recorder.py
```diff
@@ -0,0 +1,143 @@
...
+    # type: (str) -> str
+    try:
+        return parse.urlparse(repository_url).path.rstrip(".git").rpartition("/")[-1]
+    except ValueError:
+        # In case of parsing error, default to repository url
+        log.warning("Repository name cannot be parsed from repository_url: %s", repository_url)
+        return repository_url
+
+
+class CIVisibility(Service):
```
mabdinur, line 33: Can we add tests for the `CIVisibility` service. Specifically test cases that involve forking and updating `CIVisibility._instance` across threads

juan-fernandez (dismissed):
OK from CI Visibility's perspective

ddtrace/filters.py
Yun-Kim: `TraceCiVisibilityFilter` should never have been introduced into the public API, as there's no use case for anyone to use it outside of CI Visibility, and the new one has introduced breaking changes (to fit CI-visibility specific logic). 

Essentially with this refactor, `ddtrace.ci_visibility._filters.TraceCiVisibilityFilter` is not backwards compatible with the existing one, so I don't see a problem with keeping the public API unchanged, but deprecating it for removal in 2.0.

ddtrace/filters.py
Yun-Kim: `TraceCiVisibilityFilter` is specific to the CI-Visibility product and should never have been used by users other than internally by us. The point of this deprecation is to make users stop using `TraceCiVisibilityFilter` period.

ddtrace/ci_visibility/__init__.py
Yun-Kim: By this, do you mean moving the `ci_visibility` module to `ddtrace/internal/ci_visibility` module? Or just making the `CIVisibility` a private class?

ddtrace/ci_visibility/_recorder.py
Yun-Kim: I think we should definitely address this duplication as a follow up, but this isn't in the scope of this PR.

ddtrace/ci_visibility/_recorder.py
Yun-Kim: I think we should investigate the implications of shutting the tracer down, but this discussion is out of scope of this PR as this is not new behavior but just extracted from the previous pytest plugin behavior.

FWIW, I personally don't think there's an issue with shutting down the tracer here as this will only happen on `pytest_sessionfinish()` which occurs once a test suite has been completed, i.e. the end of the job for CIVisibility (and therefore the tracer) purposes

ddtrace/ci_visibility/_filters.py
Yun-Kim: `span.context` is never None, it always returns a `Context` object.

ddtrace/ci_visibility/_recorder.py
Yun-Kim: On second thought I removed the `_restart()` implementation for this service, since the CIVisibility isn't meant to be used and restarted on multiple threads/forks. I will add test cases to provide coverage for enabling and disabling the `CIVisibility` service however.

ddtrace/ci_visibility/__init__.py
mabdinur: Either works

ddtrace/ci_visibility/_filters.py
mabdinur: I thought you referencing `local_root._context`. Using `local_root.context` works

mabdinur (approved):

ddtrace/internal/ci_visibility/recorder.py
```diff
@@ -0,0 +1,124 @@
...
+        cls._instance = cls(tracer=tracer, config=config, service=service)
+        cls.enabled = True
+
+        cls._instance.start()
+        atexit.register(cls.disable)
+
+        log.debug("%s enabled", cls.__name__)
+
+    @classmethod
+    def disable(cls):
```
albertvaka, line 82: Do we actually ever need to disable CI Visibility mode once it's enabled?

ddtrace/contrib/pytest/plugin.py
```diff
@@ -86,48 +82,31 @@ def pytest_addoption(parser):
...
 
 def pytest_sessionstart(session):
-    pin = Pin.get_from(session.config)
-    if pin is not None:
-        tracer_filters = pin.tracer._filters
-        if not any(isinstance(tracer_filter, TraceCiVisibilityFilter) for tracer_filter in tracer_filters):
-            tracer_filters += [TraceCiVisibilityFilter()]
-            pin.tracer.configure(settings={"FILTERS": tracer_filters})
+    if is_enabled(session.config):
+        _CIVisibility.enable(config=ddtrace.config.pytest)
```
albertvaka, line 89: From the PR description:

> The Pin class is traditionally used for integrations, but since CIVisibility is meant to be run only for a test suite in CI, the corresponding tracer will only be configured for that specific test suite and shutdown once the test suite is finished.

Note that we also want to support sending events for "modules" (groups of test suites) and "sessions" (the whole execution of a test command), as part of the Test Suite Level Visibility task. From your wording in the PR description it seems it will be harder to do that.

This change of moving the initialization to `pytest_sessionstart` also seems to be something we will need to change then.

Is this something we should take into account now?

albertvaka (approved):
The refactor looks really nice, thanks 🥳

I left a comment to make sure these changes are future-proof, but overall LGTM.

ddtrace/internal/ci_visibility/recorder.py
Yun-Kim: Assuming that CI Visibility will only enabled and run for each CI job, I don't believe we need to disable it. However "disabling" it in this context is just explicitly shutting down the tracer, so this shouldn't matter in the long run.

ddtrace/contrib/pytest/plugin.py
Yun-Kim: Good point, I think I used the wrong wording in that description. I meant that since CIVisibility is meant to run only for the execution of a test session (likely a CI job), the corresponding tracer will only need to be configured for each test session and then be shutdown at the end of the session. Hopefully this clarifies any of the confusion, and I think this shouldn't change our plan of supporting module/session events.

I don't think there's a huge difference between initializing at `pytest_configure` vs `pytest_sessionstart` since the latter is executed almost immediately after the former, but since `pytest_configure` is closer to the start of the actual test command I can move the initialization back to `pytest_configure`.

PR #5464, opened 2023-04-04
fix(rcm): publisher subscriber system

avara1986:
Remote Configuration Publisher-Subscriber system.

# Motivation

A common Python web application uses a WSGI server (e.g., Gunicorn) that employs multiple workers. For each worker, a Remote Config instance polls new configuration from the Agent. In the context of ASM, the typical scenario follows this flow:
- Remote Config enables the ASM_FEATURES product (with 1 click activation). The RC starts polling.
- RC receives a payload to enable ASM_FEATURES. RC activates ASM and registers new products (ASM_DD, ASM_DATA, ASM_DD).
-  RC receives a payload with all products and updates the WAF configuration.

This flow takes around 10-15 seconds (i.e., with a polling interval of 5s, RC needs 2-3 requests to the agent to keep all information updated).

However, there is a problem with the Gunicorn application: sometimes, Gunicorn may restart a worker, which needs to retrieve all RC data again and repeat the aforementioned flow. During this interval, ASM with Remote Config could raise unexpected behaviors, the new worker wouldn’t block it since it may not have yet applied all the new configuration.

## Example
The following example runs a Gunicorn application.

```
ddtrace-run gunicorn -w 4 app:app
```

We use [Locust](https://locust.io/) to request the /block endpoint with a blocked IP. Then, we kill a Gunicorn child process. As a result, we have the following table:

locust_1x

```
/block: 19495 (request) 19325 (request blocked)
```

The `19495 > 19325` means we have 170 requests that return 200 instead of 403 (blocked).

# Description

Remote Configuration needs to keep all workers updated as soon as possible. Therefore, Remote Configuration may start BEFORE the Gunicorn server in sitecustomize.py, it starts to poll information from the RC Agent, and for each new payload, through this Pub-sub system, share this information with all child processes.

In addition to this, there are different Remote Configuration behaviors:

- When the Remote Configuration Client receives a new product target file payload, we need to call a callback.
- When the Remote Configuration Client receives a new product target file payload, we need to aggregate this target file data for each product. After that, call the callback with all aggregated information.
- Remote Configuration may have a callback for each product.
- Remote Configuration may have a callback for one or more products.
- For each payload, Remote Configuration needs to execute specific actions on the main process and a different action on child processes.

To achieve this goal, a Remote Configuration product may register a PubSub instance. A PubSub class contains a publisher that receives the Remote Configuration payload and shares it with Pubsub Subscriber instance. The Subscriber starts a thread on each child process, waiting for a new update of the shared data between the Publisher on the main process and the child process. Remote Configuration creates a thread listening to the main process for each instance of PubSub. To connect this publisher and the child processes subscribers, we need a connector class: Shared Memory or File. Each instance of PubSub works as a singleton when Remote Configuration dispatches the callbacks. That means if we register the same instance of PubSub class on different products, we would have one thread waiting to the main process.

Each DD Product (APM, ASM, DI, CI) may implement its PubSub Class.

Example 1: A callback for one or more Remote Configuration Products

AppSec needs to aggregate different products in the same callback for all child processes.

```
class AppSecRC(PubSubMergeFirst):
    __shared_data = ConnectorSharedMemory()

    def __init__(self, _preprocess_results, callback, name="Default"):
        self._publisher = self.__publisher_class__(self.__shared_data, _preprocess_results)
        self._subscriber = self.__subscriber_class__(self.__shared_data, callback, name)

asm_callback = AppSecRC(preprocess_1click_activation, appsec_callback, "ASM")

remoteconfig_poller.register("ASM_PRODUCT", asm_callback)
remoteconfig_poller.register("ASM_FEATURES_PRODUCT", asm_callback)
```

Example 2: One Callback for each product
DI needs to aggregate different products in the same callback for all child processes.

```
class DynamicInstrumentationRC(PubSub):
    __shared_data = ConnectorSharedMemory()

    def __init__(self, _preprocess_results, callback, name="Default"):
        self._publisher = self.__publisher_class__(self.__shared_data, _preprocess_results)
        self._subscriber = self.__subscriber_class__(self.__shared_data, callback, name)


di_callback_1 = DynamicInstrumentationRC(callback=di_callback_1, name="ASM")
di_callback_2 = DynamicInstrumentationRC(callback=di_callback_2, name="ASM")

remoteconfig_poller.register("DI_1_PRODUCT", di_callback)
remoteconfig_poller.register("DI_2_PRODUCT", di_callback_2)
```

## Results
Following the previous example, if we run a application with gunicorn and we kill some Gunicorn child worker, the result is:

locust_with_refactor
```
/block: 15070 (request) 15070 (request blocked)
```
`19495 == 19325` Hurray!

## Extra improvements of this refactor
if we compare with `docker stats` both branches (1.x and this branch) we see an improvement of memory usage and CPU:

`1.x` branch:
```
CONTAINER ID   NAME                        CPU %     MEM USAGE / LIMIT     MEM %     NET I/O           BLOCK I/O    PIDS
e487b87bc5e0   showcase-flask_showcase_1   34.23%    204.1MiB / 31.07GiB   0.64%     44.8MB / 62.3MB   0B / 451kB   17
```

`avara1986/rc_refactor` branch:
```
CONTAINER ID   NAME                        CPU %     MEM USAGE / LIMIT     MEM %     NET I/O           BLOCK I/O        PIDS
b90ac73a23b6   showcase-flask_showcase_1   14.99%    169.2MiB / 31.07GiB   0.53%     31.3MB / 55.6MB   4.36MB / 303kB   18
```

## PR related

This PR fix this system test problem: https://github.com/DataDog/system-tests/pull/932

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/internal/remoteconfig/worker.py
```diff
@@ -93,3 +82,108 @@ def _online(self):
...
+        # type: () -> RemoteConfigPoller
+        self.enable()
+        return self
+
+    def __exit__(self, *args):
+        # type: (...) -> None
+        self.disable()
+
+
+remoteconfig_poller = RemoteConfigPoller()
```
P403n1x87, line 189: do we have to worry about the case where the main process is only used to spawn another process and then terminates, while the child process keeps doing work? IIUC in this case we wouldn't have a poller running anymore, and a potentially broken pub-sub pipe?

ddtrace/internal/remoteconfig/_pubsub.py
```diff
@@ -0,0 +1,114 @@
...
+        self._subscriber.force_restart()
+
+    def _poll_data(self, test_tracer=None):
+        self._subscriber._get_data_from_connector_and_exec(test_tracer=test_tracer)
+
+    def stop(self):
+        self._subscriber.stop()
+
+
+class PubSub(PubSubBase):
```
P403n1x87, line 100: I wonder if we could simplify this design a little bit. We already have the publisher, the subscriber and the event bus (the connector), so we could potentially do away without this class. I would tend to see the product-specific subclasses more as simply sub-classes of a subscriber than this pub-sub hub. WDYT?

P403n1x87 (commented):
Looks very nice and promising 💪 . A few preliminary comments/questions for now.

ddtrace/debugging/_debugger.py
```diff
@@ -257,8 +257,9 @@ def __init__(self, tracer=None):
             log.info("Disabled Remote Configuration enabled by Dynamic Instrumentation.")
 
         # Register the debugger with the RCM client.
-        RemoteConfig.register("LIVE_DEBUGGING", self.__rc_adapter__(self._on_configuration))
-
+        di_callback = self.__rc_adapter__("", self._on_configuration)
+        di_callback.start_subscriber()
+        remoteconfig_poller.register("LIVE_DEBUGGING", di_callback)
```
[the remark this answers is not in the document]
avara1986: Thanks @P403n1x87 ! AppSec needs to register the same callback for many products:
https://github.com/DataDog/dd-trace-py/pull/5464/files#diff-6c9b0db1351f5cf53547609e99457bf2d1480ba0e8e123ec9be608aeef4712d5R73

In that case, I think the best approach is register all products before to start the subscriber

ddtrace/bootstrap/sitecustomize.py
```diff
@@ -265,6 +266,13 @@ def drop(module_name):
         else:
             log.debug("additional sitecustomize found in: %s", sys.path)
 
+    if asbool(os.environ.get("DD_REMOTE_CONFIGURATION_ENABLED", "true")):
+        from ddtrace.appsec._remoteconfiguration import enable_appsec_rc
+        from ddtrace.internal.remoteconfig.worker import remoteconfig_poller
+
+        remoteconfig_poller.enable()
+        enable_appsec_rc()
+
```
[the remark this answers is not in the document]
avara1986: We need a subscriber in the main process to register dynamically ASM products:
https://github.com/DataDog/dd-trace-py/pull/5464/files#diff-6c9b0db1351f5cf53547609e99457bf2d1480ba0e8e123ec9be608aeef4712d5R119

As we need in the main process, could we move this call to tracer? What do you think is the best approach of thart? thanks in advance! :smile:

ddtrace/internal/remoteconfig/worker.py
avara1986: Could you give an example? In a Gunicorn application, RC worker is polling in main process data and the child process listening, due to the worker is always running in the main process, if it dies all child workers stop too

ddtrace/internal/remoteconfig/_pubsub.py
avara1986: Yeah, it's true, we can simplify this file to use just one class but we need at least one which contains the abstract methods and common methods of the pubsub class and each product-specific implement/inherits from that class, right?

ddtrace/debugging/_debugger.py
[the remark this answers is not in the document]
P403n1x87: Perhaps the `start_subscriber` method could be made idempotent? Would there be any issues with this approach?

ddtrace/bootstrap/sitecustomize.py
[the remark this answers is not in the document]
P403n1x87: I was thinking of having a `if` block just for ASM, like it's done for other products, like the profiler, DI etc.... Functionally it wouldn't change much, it's just to keep the logic tidy by only doing RC enablement in an `if` block guarded by a RC-specific env var. E.g.
~~~python
if asbool(os.environ.get("DD_REMOTE_CONFIGURATION_ENABLED", "true")):
    # Enable RC

...


if <ASM enabling logic here>:
    # Enable ASM stuff
~~~

ddtrace/internal/remoteconfig/worker.py
P403n1x87: I'm thinking of [fork off and die](https://wiki.linuxquestions.org/wiki/Fork_off_and_die) for applications that would start a daemon process. Although rare, I wonder if we should ensure support for this too.

ddtrace/internal/remoteconfig/_pubsub.py
P403n1x87: yep

ddtrace/debugging/_debugger.py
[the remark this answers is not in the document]
avara1986: :+1:

ddtrace/internal/remoteconfig/worker.py
avara1986: I checked that with this example:
```
>>> ps -A | grep gunicorn
 452209 pts/3    00:00:00 gunicorn
 452275 pts/3    00:00:00 gunicorn
 452277 pts/3    00:00:00 gunicorn
 452279 pts/3    00:00:00 gunicorn
 452281 pts/3    00:00:00 gunicorn
>>> sudo kill -9 452209 (main process)
>>> ps -A | grep gunicorn
<empty>
```
What's your thoughts? Do you think it's enough?

avara1986: If I understand correctly, I'm testing that behavior on this test:
https://github.com/DataDog/dd-trace-py/pull/5464/files#diff-213a7c2bf1bb1f526312e4a698d5f7118cf7fe692dd4ebaa3ff7de5209552e79R49

A long time ago (in a commit far, far away...) I had an error on `_stop_service` and that test never finished. It was related to your concern: the main process dies but the child process still running

Do you think this test is enough to validate that? do you think we need a specific test?

P403n1x87: If I understand the test correctly, this is still an example of a server that spawns worker processes and remains in the background. What I would like to validate is the scenario where a process is used to spawn a daemon process which keeps running in the background, while the parent terminates. My understanding is that, with the current design, the poller that only runs in the parent process will die, and the child daemon process will be left with no poller to listen to.

avara1986: Ok, I think I understand now your concerns :smile: 

What do you think about this approach?

Actual behavior of this branch:
```
- Main process -> RC publisher
   - Child 1 -> Subscriber 1
   - ...
   - Child N -> Subscriber N
```

We could create a env var such as `DD_REMOTE_CONFIG_FORK_AND_DIE_SAFE` and if it's True, application runs Remote Config as it is doing now in `1.x`, that means, application runs a Remote Config Worker on each process:
```
- Main process -> RC publisher 1 Subscriber 1
   - Child 1 -> RC publisher 2 Subscriber 2
   - ...
   - Child N -> RC publisher N Subscriber N
```

P403n1x87: Yep this approach would definitely work. How hard would it be to implement a mechanism that would allow the remaining children to detect that the publisher if offline and negotiate who becomes the new publisher?

P403n1x87:
Would it be possible to split this PR into smaller chunks to make it easier to review? I'm thinking we could have a PR that introduces the new classes, and then a follow-up that uses them in the RC sources (and the products that depend on it). WDYT?

avara1986:
> Would it be possible to split this PR into smaller chunks to make it easier to review? I'm thinking we could have a PR that introduces the new classes, and then a follow-up that uses them in the RC sources (and the products that depend on it). WDYT?

Nice!

ddtrace/appsec/_remoteconfiguration.py
```diff
@@ -17,49 +21,57 @@
...
 
     from ddtrace import Tracer
 
 log = get_logger(__name__)
 
 
+class AppSecRC(PubSub):
+    __subscriber_class__ = RemoteConfigSubscriber
+    __publisher_class__ = RemoteConfigPublisherMergeFirst
+    __shared_data = PublisherSubscriberConnector()
```
P403n1x87, line 38:
```suggestion
    __shared_data__ = PublisherSubscriberConnector()
```

ddtrace/debugging/_probe/remoteconfig.py
```diff
@@ -251,19 +251,39 @@ class ProbePollerEvent(object):
...
     """Probe configuration adapter for the RCM client.
 
     This adapter turns configuration events from the RCM client into probe
     events that can be handled easily by the debugger.
     """
 
-    def __init__(self, callback):
-        # type: (Callable[[ProbePollerEventType, Iterable[Probe]], None]) -> None
-        self._callback = callback
+    _status_timestamp = time.time()
```
P403n1x87, line 261: can we make this an instance attribute, please?

ddtrace/debugging/_probe/remoteconfig.py
```diff
@@ -287,27 +307,20 @@ def _update_probes_for_config(self, config_id, config):
...
-            self._callback(ProbePollerEvent.STATUS_UPDATE, probes)
-            self._next_status_update_timestamp()
 
-        if metadata is None:
-            log.debug("no RCM metadata")
-            return
+class ProbeRCAdapter(PubSub):
+    __publisher_class__ = RemoteConfigPublisher
+    __subscriber_class__ = DebuggerRemoteConfigSubscriber
+    __shared_data = PublisherSubscriberConnector()
```
P403n1x87, line 322:
```suggestion
    __shared_data__ = PublisherSubscriberConnector()
```

ddtrace/internal/remoteconfig/_pubsub.py
P403n1x87, line 1: Is this change needed?

ddtrace/debugging/_probe/remoteconfig.py
```diff
@@ -287,27 +307,20 @@ def _update_probes_for_config(self, config_id, config):
...
 
-        self._update_probes_for_config(metadata.id, config)
+    def __init__(self, _preprocess_results, callback):
```
P403n1x87, line 324: what is `_preprocess_results` for?

tests/debugging/probe/test_remoteconfig.py
```diff
@@ -66,9 +68,10 @@ def mock_config():
     ],
 )
 def test_poller_env_version(env, version, expected, mock_config):
+    remoteconfig_poller.disable()
```
P403n1x87, line 71: should we have a fixture to deal with test state preparation and clean-up?

ddtrace/internal/remoteconfig/_pubsub.py
avara1986: I'm not sure, I changed it due to a Mypy error:
```
ddtrace/appsec/_remoteconfiguration.py:37: error: Incompatible types in assignment (expression has type "Type[RemoteConfigPublisherMergeFirst]", base class "PubSub" defined the type as "Type[RemoteConfigPublisherBase]")  [assignment]
```

ddtrace/debugging/_probe/remoteconfig.py
avara1986: It's a function that Publisher instance executes in the main process:
https://github.com/DataDog/dd-trace-py/blob/1.x/ddtrace/internal/remoteconfig/_publishers.py#L106

This function callback has the responsibility of:
- Transforming payloads
- Executing actions ONLY in the main process

Example: ASM needs to register some Remote Config products dynamically in the main process because the child process doesn't care about products.
https://github.com/DataDog/dd-trace-py/pull/5464/files#diff-6c9b0db1351f5cf53547609e99457bf2d1480ba0e8e123ec9be608aeef4712d5R119

P403n1x87: Ah I see. But do I understand it right that for the debugger we are passing an empty string to it?
```
di_callback = self.__rc_adapter__("", self._on_configuration)
```

avara1986: ouch, it works correctly with a empty string, but the correct value should be `None`, good point @P403n1x87 !

juanjux (approved):

juanjux (approved):

juanjux (dismissed):

ddtrace/debugging/_probe/remoteconfig.py
```diff
@@ -251,19 +251,42 @@ class ProbePollerEvent(object):
...
 
+    def _exec_callback(self, data, test_tracer=None):
+        if data:
+            metadatas = data["metadata"]
+            rc_configs = data["config"]
+            # DEV: We emit a status update event here to avoid having to spawn a
+            # separate thread for this.
+            log.debug("[%s][P: %s] Dynamic Instrumentation Updated", os.getpid(), os.getppid())
+            for idx in range(len(rc_configs)):
+                if time.time() > self._status_timestamp:
```
ZStriker19, line 275: When wouldn't this be true? Can you explain it a bit?

avara1986: AFAIK, this condition depends of `config.diagnostics_interval` (1h).

Each time this condition is True, we update `self._status_timestamp` in this function:

https://github.com/DataDog/dd-trace-py/pull/5464/files/970621058ea1214d48946ae19fbe6746d1eb3dbd#diff-ae57ceee6484ce78fa6d58aef59ede86def77b5148eeb3a5a6dde71a56eba9fdR290
```
    def _next_status_update_timestamp(self):
        # type: () -> None
        self._status_timestamp = time.time() + config.diagnostics_interval
```

@P403n1x87, could you confirm this please?

P403n1x87: That's right. We want to re-emit probe status messages periodically, so at every iteration we set the new "deadline" in the future. When we are past that we re-emit the status messages.

juanjux (approved):

P403n1x87 (approved):

ZStriker19 (approved):

ZStriker19 (approved):

ZStriker19 (approved):

juanjux (approved):

P403n1x87 (approved):

PR #5471, opened 2023-04-04
feat(tracing): introduce ci visibility agentless writer/encoder

emmettbutler:
This pull request implements a new "writer" class for agentless CI Visibility reporting. 

This change depends on https://github.com/DataDog/dd-trace-py/pull/5475.

Since the new writer class isn't used anywhere other than tests, there is no significant risk. All applicable existing `AgentWriter`  tests have been adapted to exercise the same functionality in the `CIVisibilityWriter`. 

## Checklist

- [x] update from base branch and standardize "ciapp" naming convention to match
- [x] resolve question about class relationship https://github.com/DataDog/dd-trace-py/pull/3690#issuecomment-1496644571

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/internal/encoding.py
```diff
@@ -156,6 +157,48 @@ def _decode_id_to_hex(hex_id):
...
+    def put(self, spans):
+        self.buffer.append(spans)
+
+    def encode_traces(self, traces):
+        return self._build_payload(traces=traces)
+
+    def encode(self):
+        payload = self._build_payload()
+        self._init_buffer()
+        return payload
```
brettlangdon, line 184: Do we need a lock here? If `put` is called after building the payload, but before resetting the buffer then traces would be lost?

ddtrace/internal/encoding.py
```diff
@@ -156,6 +157,48 @@ def _decode_id_to_hex(hex_id):
...
+
+    def _build_payload(self, traces=None):
+        normalized_spans = [CIAppEncoderV01._convert_span(span) for trace in (traces or self.buffer) for span in trace]
```
brettlangdon, line 187:
```suggestion
    def _build_payload(self, traces):
        normalized_spans = [CIAppEncoderV01._convert_span(span) for trace in traces for span in trace]
```

I think I'd prefer this. Then you can just call `self._build_payload(self.buffer)` in `def encode(self):`. I think it would be a more clear interface rather than hiding that we'll _maybe_ use `self._buffer` (or `traces = []` then it'll use the buffer, which maybe we don't want?)

What do you think?

emmettbutler:
@mabdinur clarified the code-organization question for me in https://github.com/DataDog/dd-trace-py/pull/5475#pullrequestreview-1373274242.

ddtrace/internal/encoding.py
```diff
@@ -156,6 +159,60 @@ def _decode_id_to_hex(hex_id):
...
+    def _build_payload(self, traces):
+        normalized_spans = [CIVisibilityEncoderV01._convert_span(span) for trace in traces for span in trace]
+        self._metadata = {k: v for k, v in self._metadata.items() if k in self.ALLOWED_METADATA_KEYS}
+        return msgpack_packb({"version": 1, "metadata": {"*": self._metadata}, "events": normalized_spans})
+
+    @staticmethod
+    def _convert_span(span):
+        # type: (Span) -> Dict[str, Any]
+        sp = JSONEncoderV2._convert_span(span)
+        sp["trace_id"] = JSONEncoderV2._encode_id_to_hex(span.trace_id)
```
mabdinur, line 204: We should verify that the ci visibility intake supports decoding 128 bit trace ids. If 128 bit trace ids are not supported, we should use this trace id variant:

```suggestion
        sp["trace_id"] = JSONEncoderV2._encode_id_to_hex(span._trace_id_64bits)
```

releasenotes/notes/civisibilitywriter-3fa0ac9af42256c2.yaml
```diff
@@ -0,0 +1,3 @@
+features:
+  - |
+    tracing: Introduces ``CIVisibilityWriter``, an ``HTTPWriter`` subclass that communicates with Datadog's CI Visibility intake endpoint
```
Yun-Kim, line 3:
```suggestion
    tracing: Introduces ``CIVisibilityWriter``, an ``HTTPWriter`` subclass that communicates with Datadog's CI Visibility intake endpoint. This will be automatically configured for the pytest plugin.
```

ddtrace/internal/writer.py
```diff
@@ -706,3 +708,66 @@ def _get_finalized_headers(self, count):
...
+                "library_version": ddtrace.__version__,
+            }
+        )
+
+        headers = headers or dict()
+        headers["dd-api-key"] = os.environ.get("DD_API_KEY", "dummyapikey")
+
+        super(CIVisibilityWriter, self).__init__(
+            intake_url=intake_url,
+            endpoint="v0.5/traces",  # XXX
```
Yun-Kim, line 747: Is this the correct endpoint for CI intake?

ddtrace/internal/writer.py
```diff
@@ -706,3 +708,66 @@ def _get_finalized_headers(self, count):
...
+            priority_sampler=priority_sampler,
+            processing_interval=processing_interval,
+            timeout=timeout,
+            dogstatsd=dogstatsd,
+            sync_mode=sync_mode,
+            reuse_connections=reuse_connections,
+            headers=headers,
+        )
+
+    def stop(self):
```
Yun-Kim, line 759: Is there a reason to implement this? I don't see the `HTTPWriter` class implement this either and this ultimately inherits from `PeriodicService`, which implements its own `stop()`.

ddtrace/internal/encoding.py
Yun-Kim: @juan-fernandez @albertvaka can you confirm if the CI Visibility intake supports 128 bit trace IDs?

ddtrace/internal/encoding.py
```diff
@@ -156,6 +159,59 @@ def _decode_id_to_hex(hex_id):
...
+    def encode(self):
+        with self._lock:
+            payload = self._build_payload(self.buffer)
+            self._init_buffer()
+            return payload
+
+    def _build_payload(self, traces):
+        normalized_spans = [CIVisibilityEncoderV01._convert_span(span) for trace in traces for span in trace]
+        self._metadata = {k: v for k, v in self._metadata.items() if k in self.ALLOWED_METADATA_KEYS}
+        return msgpack_packb({"version": 1, "metadata": {"*": self._metadata}, "events": normalized_spans})
```
Yun-Kim, line 198: Can we store the payload format version as a constant or class attribute to make easier when we need to upgrade?
```suggestion
        return msgpack_packb({"version": PAYLOAD_FORMAT_VERSION, "metadata": {"*": self._metadata}, "events": normalized_spans})
```

ddtrace/internal/encoding.py
```diff
@@ -156,6 +159,59 @@ def _decode_id_to_hex(hex_id):
...
+        sp = JSONEncoderV2._convert_span(span)
+        sp["type"] = span.span_type
+        sp["duration"] = span.duration_ns
+        sp["meta"] = dict(sorted(span._meta.items()))
+        sp["metrics"] = dict(sorted(span._metrics.items()))
+        if span.span_type == "test":
+            event_type = "test"
+        else:
+            event_type = "span"
+        return {"version": 1, "type": event_type, "content": sp}
```
Yun-Kim, line 212: Same here
```suggestion
        return {"version": PAYLOAD_FORMAT_VERSION, "type": event_type, "content": sp}
```

ddtrace/internal/encoding.py
```diff
@@ -156,6 +159,59 @@ def _decode_id_to_hex(hex_id):
...
+        return msgpack_packb({"version": 1, "metadata": {"*": self._metadata}, "events": normalized_spans})
+
+    @staticmethod
+    def _convert_span(span):
+        # type: (Span) -> Dict[str, Any]
+        sp = JSONEncoderV2._convert_span(span)
+        sp["type"] = span.span_type
+        sp["duration"] = span.duration_ns
+        sp["meta"] = dict(sorted(span._meta.items()))
+        sp["metrics"] = dict(sorted(span._metrics.items()))
```
Yun-Kim, line 207: Do these not get set in the span dictionary during `JSONEncoderV2._convert_span`?

releasenotes/notes/civisibilitywriter-3fa0ac9af42256c2.yaml
emmettbutler: Would it be fair to say "An upcoming change will configure the pytest plugin to automatically use this writer."?

Yun-Kim: Yup, that sounds good to me.

ddtrace/internal/encoding.py
emmettbutler: They only get set when the attributes are truthy. The intention here is to include them unconditionally.

ddtrace/internal/writer.py
emmettbutler: It's implemented because `CIVisibilityWriter` defaults to synchronous mode, which doesn't `start()` the service. The conditional in this function avoids trying to stop a service that was never started.

ddtrace/internal/ci_visibility/encoder.py
```diff
@@ -0,0 +1,68 @@
...
+        with self._lock:
+            payload = self._build_payload(self.buffer)
+            self._init_buffer()
+            return payload
+
+    def _build_payload(self, traces):
+        normalized_spans = [CIVisibilityEncoderV01._convert_span(span) for trace in traces for span in trace]
+        self._metadata = {k: v for k, v in self._metadata.items() if k in self.ALLOWED_METADATA_KEYS}
+        return msgpack_packb(
+            {"version": self.PAYLOAD_FORMAT_VERSION, "metadata": {"*": self._metadata}, "events": normalized_spans}
```
albertvaka, line 53: The two versions used here and below shouldn't be the same constant. The events within a v1 payload have their own versioning (eg: test events will be v2 when we add TSLV).

ddtrace/internal/ci_visibility/encoder.py
```diff
@@ -0,0 +1,68 @@
...
+        sp = JSONEncoderV2._convert_span(span)
+        sp["type"] = span.span_type
+        sp["duration"] = span.duration_ns
+        sp["meta"] = dict(sorted(span._meta.items()))
+        sp["metrics"] = dict(sorted(span._metrics.items()))
+        if span.span_type == "test":
+            event_type = "test"
+        else:
+            event_type = "span"
+        return {"version": CIVisibilityEncoderV01.PAYLOAD_FORMAT_VERSION, "type": event_type, "content": sp}
```
albertvaka, line 68:
```suggestion
        return {"version": CIVisibilityEncoderV01.TEST_EVENT_VERSION, "type": event_type, "content": sp}
```

ddtrace/internal/ci_visibility/writer.py
```diff
@@ -0,0 +1,79 @@
...
+            {
+                "language": "python",
+                "env": config.env,
+                "runtime-id": get_runtime_id(),
+                "library_version": ddtrace.__version__,
+            }
+        )
+
+        headers = headers or dict()
+        headers["dd-api-key"] = os.environ.get("DD_API_KEY", "dummyapikey")
```
albertvaka, line 49: If we are using this writer (ie: if Agentless is enabled) but `DD_API_KEY` is not set, we want to error out instead of continuing with a dummy key.

ddtrace/internal/ci_visibility/encoder.py
```diff
@@ -0,0 +1,68 @@
...
+    def encode(self):
+        with self._lock:
+            payload = self._build_payload(self.buffer)
+            self._init_buffer()
+            return payload
+
+    def _build_payload(self, traces):
+        normalized_spans = [CIVisibilityEncoderV01._convert_span(span) for trace in traces for span in trace]
+        self._metadata = {k: v for k, v in self._metadata.items() if k in self.ALLOWED_METADATA_KEYS}
+        return msgpack_packb(
```
albertvaka, line 52:
```suggestion
        # TODO: Split the events in several payloads as needed to avoid hitting the intake's maximum payload size.
        return msgpack_packb(
```

This logic makes builidng the messagepack a bit more convoluted (since you don't really know how many elements you will have in the array ahead of time) so we don't have to fix this now.

albertvaka (commented):
Left some comments but this is looking great 😄 Amazing work, thanks a lot 🙇

ddtrace/internal/ci_visibility/encoder.py
albertvaka: We can also track this internally in a Jira task, if you don't want to merge `TODO` comments.

ddtrace/internal/writer.py
albertvaka: The endpoint is `/api/v2/citestcycle`.

ddtrace/internal/encoding.py
albertvaka: It does not.

tests/utils.py
```diff
@@ -75,6 +70,30 @@ def override_env(env):
...
+    """Decorator for test functions that overrides environment variables passed as a dictionary"""
+
+    def _decorator(fn):
+        def wrapper(fn, *args, **kwargs):
+            with override_env(env):
+                return fn(*args, **kwargs)
+
+        return decorator.decorator(wrapper, fn)
+
+    return _decorator
```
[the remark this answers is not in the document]
emmettbutler: Putting a decorator on the test functions leads to fewer lines of code changed than indenting all of the test function bodies to be under that context manager. I'm actually going to do that, though, because the current implementation requires regenerating every requirement lockfile, which is definitely worse.

releasenotes/notes/civisibilitywriter-3fa0ac9af42256c2.yaml
```diff
@@ -0,0 +1,4 @@
+features:
+  - |
+    tracing: Introduces ``CIVisibilityWriter``, an ``HTTPWriter`` subclass that communicates with Datadog's CI Visibility intake endpoint.
+    An upcoming change will configure the pytest plugin to automatically use this writer.
```
Yun-Kim, line 4:
```suggestion
```
On second thought, I think we should leave this out since this doesn't really pertain to this PR. 
Furthermore, I'm not sure if this should have a changelog since we're not actually changing the public API, right? The only reason I think we should add a changelog is if we want to let users know how to use agentless reporting (to set `DD_CIVISIBILITY_AGENTLESS_ENABLED=true`), but this would be in a future PR anyway.

Yun-Kim (approved):
LGTM! I think one thing I'm curious is to see how much performance improvements we can see if we implement the CIVisibilityEncoder in cython, but definitely happy with this 👍

emmettbutler:
Thanks @Yun-Kim. I put a lot of thought into implementing the encoder in Cython and decided against it because of how much bigger the code would be. The entire `BufferedEncoder` pattern isn't exactly applicable to this encoding scheme, so it will involve a lot of new Cython code. I agree it's worth exploring in the future - for the moment I optimized for working code quickly.

PR #5474, opened 2023-04-05
chore(debugging): try to undecorate functions

P403n1x87:
This change adds the ability to undecorate functions when discovering them from modules. This allows assigning the original undecorated function to the discovered attribute. When the function discovery object is queried for functions to wrap, this ensures that the original function gets wrapped instead of the wrapper function.

Note that, even before this change, it was possible to wrap the original function, but it had to be accessed via its full path through the closure cells. For example

~~~python
def d(f):
    def _(): f()
    return _

@d
def foo(): pass
~~~

the function `foo` would be accessible from `foo.<locals>.foo`, with `foo` pointing to `d`. With this change, both `foo.<locals>.foo` and `foo` point to `foo`.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


shatzi (approved):
very cool

PR #5475, opened 2023-04-05
chore(tracing): encapsulate agent-specific writer functionality

emmettbutler:
This change facilitates the creation of new `*Writer` classes by encapsulating the behavior of the `AgentWriter` that is specific to communication with the Datadog Agent in its own class. This leaves most of the writer's functionality, which could apply to any trace writing workload that uses HTTP, in the new `HTTPWriter` class. This change builds on the approach from https://github.com/DataDog/dd-trace-py/pull/3690.

https://github.com/DataDog/dd-trace-py/pull/5471 builds on this change to implement a new `CIVisibilityWriter` class.

This change only affects code organization, not functionality, so existing tests are sufficient to cover the risk.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.

### DIFF
--- a/ddtrace/internal/writer.py
+++ b/ddtrace/internal/writer.py
@@ -231,13 +231,7 @@ class LogWriter(TraceWriter):
 
 
 class HTTPWriter(periodic.PeriodicService, TraceWriter):
-    """Writer to the Datadog Agent.
-
-    The Datadog Agent supports (at the time of writing this) receiving trace
-    payloads up to 50MB. A trace payload is just a list of traces and the agent
-    expects a trace to be complete. That is, all spans with the same trace_id
-    should be in the same trace.
-    """
+    """Writer to an arbitrary HTTP intake endpoint."""
 
     def __init__(
         self,
@@ -362,9 +356,6 @@ class HTTPWriter(periodic.PeriodicService, TraceWriter):
                 # Reset the connection if reusing connections is disabled.
                 if not self._reuse_connections:
                     self._reset_connection()
-
-    def _downgrade(self, payload, response):
-        raise NotImplementedError
 
     def _get_finalized_headers(self, count):
         return self._headers.copy()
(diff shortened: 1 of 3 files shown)


mabdinur (commented):
An alternative approach is to create an `_HTTPClient` class that encapsulates all logic required to encode and send an http requests containing traces. We can use this client in both the `AgentWriter` and the proposed `CIVisiblityWriter`.  We already use this pattern in the [remote config worker](https://github.com/DataDog/dd-trace-py/blob/f8654e00036bb58bbf5c562ad0178239a043799d/ddtrace/internal/remoteconfig/worker.py#L45) and the [telemetry writer](https://github.com/DataDog/dd-trace-py/blob/v1.11.1/ddtrace/internal/telemetry/writer.py#L132). 

But this pattern is not used in [dynamic instrumentation](https://github.com/DataDog/dd-trace-py/blob/v1.11.1/ddtrace/debugging/_debugger.py#L162), or the [profiler](https://github.com/DataDog/dd-trace-py/blob/f1069ccb0a2a9ae4c78dbfe1a6847da520b8440b/ddtrace/profiling/profiler.py#L181) so I am fine with the current approach.

emmettbutler:
@mabdinur good idea, I'll probably do that in a subsequent pull request. I've addressed all of your suggestions and this change is ready for another review.

mabdinur (approved):

ddtrace/internal/writer.py
```diff
@@ -625,11 +494,215 @@ def _stop_service(
...
+            )
+        buffer_size = buffer_size or get_writer_buffer_size()
+        max_payload_size = max_payload_size or get_writer_max_payload_size()
+        encoder = encoder_cls(
+            max_size=buffer_size,
+            max_item_size=max_payload_size,
+        )
+
+        endpoint = "%s/traces" % self._api_version
+        headers = {
```
Yun-Kim, line 576: @emmettbutler from what I can tell, it looks like we never use the provided `headers` argument and instead always override them to be this, while previously it would update the provided `headers` argument. Is this intentional?

Yun-Kim (commented):
Just one small question, but 💯

ddtrace/internal/writer.py
emmettbutler: Good catch, that was unintentional. I'll fix it and add a regression test.

emmettbutler: https://github.com/DataDog/dd-trace-py/pull/5483

PR #5477, opened 2023-04-05
fix(django): guard against cache truthiness

### LINKED ISSUE #5460: Django patch breaks cache
### Summary of problem
Django cache patch is evaluating the cache result, which might not be valid.

https://github.com/DataDog/dd-trace-py/blob/1.11/ddtrace/contrib/django/patch.py#L182

That line is calling `bool` on `result`, which e.g. is not defined for a pandas DataFrame

1.11

pip 22.0.4

django 3.2.18

<details>
	```
	asgiref==3.6.0
	attrs==22.2.0
	bytecode==0.14.0
	cattrs==22.2.0
	ddsketch==2.0.4
	ddtrace==1.11.0
	Deprecated==1.2.13
	Django==4.2
	envier==0.4.0
	importlib-metadata==6.0.1
	jsonschema==4.17.3
	numpy==1.24.2
	opentelemetry-api==1.17.0
	packaging==23.0
	pandas==2.0.0
	protobuf==4.22.1
	pyrsistent==0.19.3
	python-dateutil==2.8.2
	pytz==2023.3
	six==1.16.0
	sqlparse==0.4.3
	tenacity==8.2.2
	typing_extensions==4.5.0
	tzdata==2023.3
	wrapt==1.15.0
	xmltodict==0.13.0
	zipp==3.15.0
	```
</details>

```python

import os
from ddtrace import patch_all

patch_all()

os.environ.setdefault("DJANGO_SETTINGS_MODULE", "settings")

import django
django.setup()

class NoBool:
    def __bool__(self):
        raise NotImplementedError

from django.core.cache import cache
cache.set(1, NoBool())
cache.get(1)


```

On the example:
```
NotImplementedError
```
If you are using pandas DataFrame:
```
ValueError: The truth value of a DataFrame is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().
```

The saved dataframe without trying to evaluate it

wconti27:
Fixes #5460 

Fix Django cache truthiness bug that occurs when a result has ambiguous truth because the class doesn't implement the __bool__() function. 

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


riotfile.py
```diff
@@ -758,6 +758,7 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
                 "requests": [latest],
                 "redis": ">=2.10,<2.11",
                 "psycopg2-binary": [">=2.8.6"],  # We need <2.9.0 for Python 2.7, and >2.9.0 for 3.9+
+                "pandas": [latest],
```
P403n1x87, line 761: I'd avoid adding this potentially big dependency here if we could just test with a custom class that recreates the issue (e.g. the example provided by the issue reporter). WDYT?

wconti27: good point, will do

nicootto:
Hey @wconti27 ! I couldn't test it locally yet, but I think you are missing this test case

```python
def test_cache_get_rowcount_missing_key_with_default(test_spans):
    # get the default cache
    cache = django.core.cache.caches["default"]
    cache.get("missing_key", default="default_value")
    spans = test_spans.get_spans()
    assert len(spans) == 1
    span = spans[0]
    assert span.service == "django"
    assert span.resource == "django.core.cache.backends.locmem.get"
    assert_dict_issuperset(span.get_metrics(), {"db.row_count": 0})
```

Reading the code I think you will execute `span.set_metric(db.ROWCOUNT, 1)` instead of `0`. Just wanted to mention it. 
Thank you for the fix btw!

tests/contrib/django/test_django.py
```diff
@@ -771,6 +771,46 @@ def test_cache_get_rowcount_missing_key(test_spans):
...
+    # This is the diff with `test_cache_get_rowcount_missing_key`,
+    # we are setting a default value to be returned in case of a cache miss
+    cache.get("missing_key", default="default_value")
+
+    spans = test_spans.get_spans()
+    assert len(spans) == 1
+    span = spans[0]
+    assert span.service == "django"
+    assert span.resource == "django.core.cache.backends.locmem.get"
+    assert_dict_issuperset(span.get_metrics(), {"db.row_count": 1})
```
nicootto, line 811: Aren't you counting this as a cache hit, while it is a cache miss? 
```suggestion
    assert_dict_issuperset(span.get_metrics(), {"db.row_count": 0})
```

nicootto (approved):

ddtrace/contrib/django/patch.py
```diff
@@ -177,10 +177,14 @@ def traced_cache(django, pin, func, instance, args, kwargs):
                 db.ROWCOUNT, sum(1 for doc in result if doc) if result and isinstance(result, Iterable) else 0
             )
         elif command_name == "get":
+            # check if we used default value, and set rowcount to 0 if result is equal to the default
+            if result in args or ("default" in kwargs and kwargs["default"] == result):
```
majorgreys, line 181: I realize this is the row count we might be using in other cache miss cases. In this case, where a default is provided, is there any other metadata on the span that would tell you that the cache get had a default?

wconti27: Not currently.

releasenotes/notes/fix-django-rowcount-bug-52d4f30236049c1b.yaml
```diff
@@ -0,0 +1,4 @@
+---
+fixes:
+  - |
+    Adds fix for bug where Django cache return object throws an error if it does not implement ``__bool__()``.
```
Yun-Kim, line 4:
```suggestion
    django: Adds fix for bug where Django cache return object throws an error if it does not implement ``__bool__()``.
```

Yun-Kim (commented):
This looks great to me, I'd love to get this merged!

tests/contrib/django/test_django.py
Yun-Kim: @wconti27 can you address this? Thanks.

wconti27: @nicootto To keep behavior consistent between all database and other datastore integrations for this tag, any returned data by a database operation such as a `query`, `find`, or `get` in this case will be included in the count.

Yun-Kim (approved):

PR #5480, opened 2023-04-05
feat(pytest): implement test suite visibility

Yun-Kim:
This PR adds additional hooks and fixtures required to implement test suite visibility. Currently, we are aiming to achieve tracing for:
    - Test command/session visibility. This is tied to the execution of the test runner command, i.e. `pytest tests/...`. This is traced through the `pytest_sessionstart()` hook, which occurs immediately after the pytest command invocation, and the `pytest_sessionfinish()` hook, which occurs at the end of the pytest session.
    - Test module/package visibility. This is tied to the execution of a test package, i.e. a directory containing test files `test_contrib/test_*.py`.
    - Test suite/module visibility. This is tied to the execution of a group of tests; for python/pytest, this means a test module, i.e. `test_abc.py`.
    - Test function visibility. This is tied to the execution of each individual test function, i.e. `test_xyz()`. This is already implemented by our current code.

Both test module/package and suite/modules are traced through `pytest_runtest_protocol()`, which is a hook that runs per test. I tried tracing them through respective module/suite scoped fixtures, but pytest internally runs these hooks after `pytest_runtest_protocol`, which messes up the time ordering of spans as the test span would get started first in this scenario. I've adjusted logic to ensure module/suite spans are started and finished at the end of a test module/suite, by storing and extracting the respective span to the corresponding pytest `module/suite` item.

Additionally, test event spans are part of their own trace and are not part of the test session/module/suite traces at tracing time. The backend is what will stitch together the test/suite/module/session events using `test_session_id/test_module_id/test_suite_id`.

Note: CI Visibility and pytest terminology do not match exactly. Here are the analogies between the two terminologies:
- CI Visibility command is equivalent to a pytest session
- CI Visibility module is equivalent to a pytest package
- CI Visibility suite is equivalent to a pytest module
- CI Visibility test event is equivalent to a pytest function

### Reviewing this PR
The main pieces of this PR to focus on for review are:
- `ddtrace.contrib.pytest.plugin.py`, which includes the bulk of the logic for adding test session/module/suite spans.
- `ddtrace.internal.ci_visibility.encoder.py`, which adds filtering/converting test spans to conform to the expected  agentless CIVisibility event payload. This involves removing all of the `test_session_id/test_module_id/test_suite_id` tags from the span meta dictionaries after explicitly adding them to the event payload, removing `span_id/parent_id/trace_id` for the test session/module/suite span payloads.
- `tests.ci_visibility.test_encoder.py` moves CI-Visibility specific encoding tests to this new test file, and ensures that CI Visibility payloads have the expected formats for test suite visibility.
- `tests.contrib.pytest.test_pytest.py` adds test cases to assert proper parenting/tagging for test suite visibility.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/contrib/pytest/plugin.py
```diff
@@ -84,13 +116,84 @@ def pytest_configure(config):
...
+    suite, so we have to use a module-scoped fixture instead.
+    """
+    if not _CIVisibility.enabled:
+        yield
+        return
+    with _CIVisibility._instance.tracer.trace(
+        "pytest.test_module",
+        service=_CIVisibility._instance._service,
+        span_type=SpanTypes.TEST,
+    ) as test_module_span:
```
juan-fernandez, line 158: should this be `test_suite_span`? To follow CI Visibility's convention

ddtrace/contrib/pytest/plugin.py
```diff
@@ -84,13 +116,84 @@ def pytest_configure(config):
...
+        test_module_span.set_tag_str(COMPONENT, "pytest")
+        test_module_span.set_tag_str(SPAN_KIND, KIND)
+        test_module_span.set_tag_str(test.FRAMEWORK, FRAMEWORK)
+        test_module_span.set_tag_str(test.FRAMEWORK_VERSION, pytest.__version__)
+        test_module_span.set_tag_str(_EVENT_TYPE, "test_suite_end")
+        test_module_span.set_tag(_SESSION_ID, test_module_span.trace_id)
+        test_module_span.set_tag(_SUITE_ID, test_module_span.span_id)
```
juan-fernandez, line 165: test suites in CI Visibility should also include `_MODULE_ID`

ddtrace/contrib/pytest/plugin.py
```diff
@@ -84,13 +116,84 @@ def pytest_configure(config):
...
+        test_module_span.set_tag_str(test.SUITE, request.node.name)
+        _store_module_id(request.node, test_module_span.span_id)
```
juan-fernandez, line 167: `_store_suite_id`?

ddtrace/contrib/pytest/plugin.py
```diff
@@ -84,13 +116,84 @@ def pytest_configure(config):
...
+    """
+    Hook to trace test packages in pytest. This fixture runs during the setup/teardown of every pytest test package,
+    which is as close as possible to the start/end of a test package. Pytest doesn't offer hooks at the start of a test
+    package, so we have to use a package-scoped fixture instead.
+    """
+    if not _CIVisibility.enabled:
+        yield
+        return
+    with _CIVisibility._instance.tracer.trace(
+        "pytest.test_package",
```
juan-fernandez, line 182:
```suggestion
        "pytest.test_module",
```

juan-fernandez (commented):
looking good! I left some comments about nomenclature

emmettbutler (dismissed):

ddtrace/contrib/pytest/plugin.py
Yun-Kim: It's possible (thought not usual) in pytest to run a specific test file (suite) only not contained in a test package, meaning sometimes a package might not exist. I've added this, but also with a check to account for this edge case as well.

emmettbutler (dismissed):

ddtrace/contrib/pytest/plugin.py
```diff
@@ -45,15 +51,65 @@ def _store_span(item, span):
...
+        "pytest.test_suite", service=_CIVisibility._instance._service, span_type=SpanTypes.TEST
+    )
+    test_suite_span.set_tag_str(COMPONENT, "pytest")
+    test_suite_span.set_tag_str(SPAN_KIND, KIND)
+    test_suite_span.set_tag_str(test.FRAMEWORK, FRAMEWORK)
+    test_suite_span.set_tag_str(test.FRAMEWORK_VERSION, pytest.__version__)
+    test_suite_span.set_tag_str(test.COMMAND, _get_pytest_command(item.config))
+    test_suite_span.set_tag_str(_EVENT_TYPE, _SUITE_TYPE)
+    test_suite_span.set_tag(_SESSION_ID, test_suite_span.trace_id)
+    test_suite_span.set_tag(_SUITE_ID, test_suite_span.span_id)
```
juan-fernandez, line 107: suites also require the module id `_MODULE_ID`

ddtrace/contrib/pytest/plugin.py
```diff
@@ -116,29 +192,48 @@ def pytest_runtest_protocol(item, nextitem):
...
+        if _extract_span(item.parent.parent) is None:
+            _start_test_module_span(item)
+
+    if _extract_span(item.parent) is None:
+        _start_test_suite_span(item)
+    module_span = _extract_span(item.parent)
+
+    with _CIVisibility._instance.tracer.start_span(
         ddtrace.config.pytest.operation_name,
+        child_of=module_span,
```
juan-fernandez, line 205: ⚠️ We want to keep tests as independent traces, so that the test traces can be navigated without the rest of the session. Does this `child_of` mean that the `trace_id` and `parent_id` of the test will be that of the module?

ddtrace/contrib/pytest/plugin.py
```diff
@@ -116,29 +192,48 @@ def pytest_runtest_protocol(item, nextitem):
...
-        span.set_tag(test.NAME, item.name)
+        span.set_tag_str(SPAN_KIND, KIND)
+        span.set_tag_str(test.FRAMEWORK, FRAMEWORK)
+        span.set_tag_str(_EVENT_TYPE, SpanTypes.TEST)
+        span.set_tag_str(test.NAME, item.name)
+        span.set_tag(_SESSION_ID, span.trace_id)
+        test_suite_span = _extract_span(item.parent)
+        span.set_tag(_SUITE_ID, test_suite_span.span_id)
+
+        if isinstance(item.parent.parent, pytest.Package):
```
juan-fernandez, line 220: are there cases where there'll be no module?

ddtrace/contrib/pytest/plugin.py
```diff
@@ -159,10 +254,16 @@ def pytest_runtest_protocol(item, nextitem):
 
         yield
 
+    test_suite_span = _extract_span(item.parent)
+    if test_suite_span is not None:
+        if nextitem is None or nextitem.parent != item.parent and not test_suite_span.finished:
+            test_suite_span.finish()
```
juan-fernandez, line 260: this means that if it's the last test in the suite, we also finish the suite, right?

juan-fernandez (commented):
looking good! Left some comments

ddtrace/contrib/pytest/plugin.py
Yun-Kim: Yes, exactly!

ddtrace/contrib/pytest/plugin.py
Yun-Kim: Rarely, but yes. For example, if someone doesn't have a test module (no `__init__.py` file in the module where the test file is contained) and runs pytest on a test file directly, i.e. `pytest test_abc.py` there will be no packages recognized by pytest.

ddtrace/contrib/pytest/plugin.py
Yun-Kim: Hmm. My understanding was that test spans part of the same session are all part of the same trace, and all test event spans (in the same file) should be child spans of the test suite (file) span, and all test suite spans (in the same package) should be child spans of the test module span, and etc. Is that not correct?

juan-fernandez: this is correct 👍 

> The test event spans are not children of any spans in the test session trace at tracing time, but the backend will handle that test suite visibility logic.

that's correct. If `test_session_id` and the rest of the ids are present in the test event, the backend will handle the logic.

emmettbutler (dismissed):

ddtrace/contrib/pytest/plugin.py
```diff
@@ -45,15 +51,78 @@ def _store_span(item, span):
...
+    """
+    test_session_span = _extract_span(item.session)
+    parent_span = test_session_span
+    if isinstance(item.parent.parent, pytest.Package):
+        test_module_span = _extract_span(item.parent.parent)
+        parent_span = test_module_span
+    test_suite_span = _CIVisibility._instance.tracer._start_span(
+        "pytest.test_suite",
+        service=_CIVisibility._instance._service,
+        span_type=SpanTypes.TEST,
```
juan-fernandez, line 108: is this the `'type'` tag in the span? This should then be `'test_suite_end'`

ddtrace/contrib/pytest/plugin.py
```diff
@@ -45,15 +51,78 @@ def _store_span(item, span):
...
+    """
+    Starts a test module span at the start of a new pytest test package.
+    Note that ``item`` is a ``pytest.Function`` object referencing the test function being run, so we need to go up
+    to find the ``pytest.Package`` ``item``.
+    """
+    test_session_span = _extract_span(item.session)
+    test_module_span = _CIVisibility._instance.tracer._start_span(
+        "pytest.test_module",
+        service=_CIVisibility._instance._service,
+        span_type=SpanTypes.TEST,
```
juan-fernandez, line 77: is this the `'type'` tag in the span? This should then be `'test_module_end'`

ddtrace/contrib/pytest/plugin.py
```diff
@@ -86,8 +155,28 @@ def pytest_configure(config):
         _CIVisibility.enable(config=ddtrace.config.pytest)
 
 
+def pytest_sessionstart(session):
+    if _CIVisibility.enabled:
+        test_session_span = _CIVisibility._instance.tracer.trace(
+            "pytest.test_session",
+            service=_CIVisibility._instance._service,
+            span_type=SpanTypes.TEST,
```
juan-fernandez, line 163: is this the `'type'` tag in the span? This should then be `'test_session_end'`

ddtrace/contrib/pytest/plugin.py
```diff
@@ -116,29 +205,47 @@ def pytest_runtest_protocol(item, nextitem):
...
-        span.set_tag(test.FRAMEWORK, FRAMEWORK)
-        span.set_tag(test.NAME, item.name)
+        span.set_tag_str(SPAN_KIND, KIND)
+        span.set_tag_str(test.FRAMEWORK, FRAMEWORK)
+        span.set_tag_str(_EVENT_TYPE, SpanTypes.TEST)
+        span.set_tag_str(test.NAME, item.name)
+        test_session_span = _extract_span(item.session)
+        span.set_tag(_SESSION_ID, test_session_span.span_id)
+        test_suite_span = _extract_span(item.parent)
+        span.set_tag(_SUITE_ID, test_suite_span.span_id)
```
juan-fernandez, line 230: it looks as though this tag ends up showing up in `meta` but not in the root of the span. It needs to be in the root

juan-fernandez (commented):
looking good! Some minor comments

ddtrace/contrib/pytest/plugin.py
Yun-Kim: Nope, this is just an identifier so that the tracer knows this is a CI Visibility span. I've updated the CI Visibility-specific encoder to handle that logic you're talking about.

ddtrace/contrib/pytest/plugin.py
Yun-Kim: Same as above

ddtrace/contrib/pytest/plugin.py
Yun-Kim: Same as above

ddtrace/contrib/pytest/plugin.py
Yun-Kim: Unfortunately with our current implementation of converting traces into events, we have to set this tag when creating these spans, but I've added logic into the CI Visibility-specific encoder to remove this when the corresponding event is being created.

ddtrace/contrib/pytest/plugin.py
```diff
@@ -159,10 +359,17 @@ def pytest_runtest_protocol(item, nextitem):
 
         yield
 
+    if test_suite_span is not None and (
+        nextitem is None or nextitem.parent != pytest_module_item and not test_suite_span.finished
+    ):
+        _mark_test_status(item.parent, test_suite_span)
+        test_suite_span.finish()
```
juan-fernandez, line 366: in `flask`, it seems some suites finish before all their tests have run, so this might need some reviewing.

juan-fernandez (approved):
other than the timing issue this looks good :). Great job!

juan-fernandez (approved):

ddtrace/internal/ci_visibility/encoder.py
```diff
@@ -67,24 +73,57 @@ def _convert_span(span, dd_origin):
         # type: (Span, str) -> Dict[str, Any]
         sp = JSONEncoderV2._span_to_dict(span)
         sp = JSONEncoderV2._normalize_span(sp)
-        sp["type"] = span.span_type
+        if dd_origin is not None:
+            sp["meta"].update({"_dd.origin": dd_origin})
```
emmettbutler, line 77:
```suggestion
```

ddtrace/internal/ci_visibility/encoder.py
```diff
@@ -67,24 +73,57 @@ def _convert_span(span, dd_origin):
...
+        sp["type"] = span.get_tag(EVENT_TYPE) or span.span_type
         sp["duration"] = span.duration_ns
         sp["meta"] = dict(sorted(span._meta.items()))
         sp["metrics"] = dict(sorted(span._metrics.items()))
```
emmettbutler, line 81:
```suggestion
        sp["metrics"] = dict(sorted(span._metrics.items()))
        if dd_origin is not None:
            sp["meta"].update({"_dd.origin": dd_origin})
```

emmettbutler: This needs to happen after `"meta"` has been set.

ddtrace/internal/ci_visibility/encoder.py
```diff
@@ -67,24 +73,57 @@ def _convert_span(span, dd_origin):
...
-        sp["trace_id"] = int(sp.get("trace_id") or "1")
-        sp["parent_id"] = int(sp.get("parent_id") or "1")
-        sp["span_id"] = int(sp.get("span_id") or "1")
-        sp["test_suite_id"] = 1  # TODO: populate with real ID
-        sp["test_session_id"] = 1  # TODO: populate with real ID
-        if COVERAGE_TAG_NAME in sp["meta"]:
-            del sp["meta"][COVERAGE_TAG_NAME]
```
emmettbutler, line 80: This condition needs to remain in the diff. It was based on a comment from @juan-fernandez the other day.

ddtrace/internal/ci_visibility/encoder.py
```diff
@@ -67,24 +73,57 @@ def _convert_span(span, dd_origin):
...
+            del sp["span_id"]
+            del sp["parent_id"]
+        else:
+            sp["trace_id"] = int(sp.get("trace_id") or "0")
+            sp["parent_id"] = int(sp.get("parent_id") or "0")
+            sp["span_id"] = int(sp.get("span_id") or "0")
+        if sp["meta"].get(EVENT_TYPE) in [SESSION_TYPE, MODULE_TYPE, SUITE_TYPE, SpanTypes.TEST]:
+            test_session_id = sp["meta"].get("test_session_id")
+            if test_session_id:
+                sp["test_session_id"] = int(test_session_id)
```
emmettbutler, line 112: Using constants for these dictionary keys would make this code more future-proof.

ddtrace/internal/ci_visibility/encoder.py
```diff
@@ -67,24 +73,57 @@ def _convert_span(span, dd_origin):
...
+        Remove trace/span/parent IDs if non-test event, move session/module/suite IDs from meta to outer content layer.
+        """
+        if sp["meta"].get(EVENT_TYPE) in [SESSION_TYPE, MODULE_TYPE, SUITE_TYPE]:
+            del sp["trace_id"]
+            del sp["span_id"]
+            del sp["parent_id"]
+        else:
+            sp["trace_id"] = int(sp.get("trace_id") or "0")
+            sp["parent_id"] = int(sp.get("parent_id") or "0")
+            sp["span_id"] = int(sp.get("span_id") or "0")
```
emmettbutler, line 108: The intake doesn't like when these IDs are 0, so I recommend defaulting them to 1.

emmettbutler (changes requested):
Great work!

ddtrace/internal/ci_visibility/encoder.py
Yun-Kim: I've moved this to `CIVisibilityEncoderV01._filter_ids()` to handle cleaning span meta items.

ddtrace/internal/ci_visibility/encoder.py
Yun-Kim: Do you know what errors you got when you set the IDs to 0? I wonder if there's a difference between an ID of 0 vs 1 (I was under the assumption that a 0 ID is equivalent to `None`)

emmettbutler: I don't have the exact error, but it was a response from the intake endpoint to the effect of "ID has to be greater than zero".

emmettbutler (dismissed):

emmettbutler (approved):

PR #5489, opened 2023-04-06
feat(pytest): use CIVisibilityWriter in CIVisibility module

emmettbutler:
This change configures the CI Visibility recorder to use the CIVisibilityWriter when a new environment variable is set.

The risk of this change is that it could break the pytest integrations. I think this is mitigated by existing tests.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


releasenotes/notes/civisibility-configure-62d5af67c49e75ad.yaml
```diff
@@ -0,0 +1,3 @@
+features:
+  - |
+    ci-visibility: Configures the ``CIVisibility`` service to send traces using the ``CIVisibilityWriter`` by default
```
Yun-Kim, line 3:
```suggestion
    pytest: Adds the ``DD_CIVISIBILITY_AGENTLESS_ENABLED`` environment variable to configure the ``CIVisibility`` service to use an agent-less test reporting ``CIVisibilityWriter``. Note that the ``CIVisibility`` service will use regular agent reporting by default.
```
Currently ci-visibility is an internal module so I think the scope should be pytest for now since that's the only testing framework we support at the moment. Also as a follow up to the environment variable comment, I was thinking of changing this release note to focus on the option to enable agentless reporting. What do you think?

ddtrace/internal/ci_visibility/recorder.py
```diff
@@ -38,6 +41,9 @@ def __init__(self, tracer=None, config=None, service=None):
         super(CIVisibility, self).__init__()
 
         self.tracer = tracer or ddtrace.tracer
+        if ddconfig.ci_visibility_agentless_enabled and not isinstance(self.tracer._writer, CIVisibilityWriter):
+            writer = CIVisibilityWriter()
```
mabdinur, line 45: Does the CIVisitilityWriter support the [ddtrace-patch-all](https://ddtrace.readthedocs.io/en/stable/integrations.html#pytest]) option? (I am not 100% sure what it does but I assume it submits non ci app spans)

ddtrace/internal/ci_visibility/recorder.py
```diff
@@ -38,6 +41,9 @@ def __init__(self, tracer=None, config=None, service=None):
...
+            self.tracer.configure(writer=writer)
```
mabdinur, line 46: If `trace.configure(writer=AgentWriter(...))` is called within a test the `CIVisibilityWriter` will get replaced. Is this a valid edge case?

ddtrace/settings/config.py
```diff
@@ -286,6 +286,8 @@ def __init__(self):
                 log.warning("Invalid obfuscation pattern, disabling query string tracing")
                 self.http_tag_query_string = False  # Disable query string tagging if malformed obfuscation pattern
 
+        self.ci_visibility_agentless_enabledd = asbool(os.getenv("DD_CIVISIBILITY_AGENTLESS_ENABLED", default=False))
```
mabdinur, line 289: If we prefix this configuration with an `_` we will make it "internal". This could give us flexibility if we decide to rename or move this configuration. 

```suggestion
        self._ci_visibility_agentless_enabledd = asbool(os.getenv("DD_CIVISIBILITY_AGENTLESS_ENABLED", default=False))
```

ddtrace/internal/ci_visibility/recorder.py
emmettbutler: As far as I can tell that option operates at the level of the pytest integration and the writer doesn't need to know about it.

ddtrace/internal/ci_visibility/recorder.py
emmettbutler: If someone does that in a test, I'd assume they're testing some part of the AgentWriter's behavior, which I'd say is a fine thing to allow.

mabdinur (dismissed):

ddtrace/settings/config.py
```diff
@@ -286,6 +286,8 @@ def __init__(self):
                 log.warning("Invalid obfuscation pattern, disabling query string tracing")
                 self.http_tag_query_string = False  # Disable query string tagging if malformed obfuscation pattern
 
+        self._ci_visibility_agentless_enabledd = asbool(os.getenv("DD_CIVISIBILITY_AGENTLESS_ENABLED", default=False))
```
Yun-Kim, line 289:
```suggestion
        self._ci_visibility_agentless_enabled = asbool(os.getenv("DD_CIVISIBILITY_AGENTLESS_ENABLED", default=False))
```

ddtrace/internal/ci_visibility/recorder.py
mabdinur: My concern here is that CIVisibility could be enabled on an application that sets `ddconfig.ci_visibility_agentless_enabled =True` and calls `tracer.configure(writer=AgentWriter(...))` to configure tracing for staging/production environments. When the code path configuring the agent writer is hit in test runs, the `CIVisibilityWriter` will be removed and all traces will be sent to a datadog agent. If a datadog agent is not running in the ci environment then all ci app spans will be dropped. 

I am not sure if this a common scenario but it might make sense to have two instances of `ddrace.Tracer()`. One used when testing an application in ci and one used in deployed applications (ex: production/staging environments). What do you think?

emmettbutler: I guess the part of this I don't fully understand is: why would a user set up an `AgentWriter` if they didn't also have an agent running?

Yun-Kim: > CIVisibility could be enabled on an application that sets ddconfig.ci_visibility_agentless_enabled =True and calls tracer.configure(writer=AgentWriter(...)) to configure tracing for staging/production environments.

@mabdinur I'm unsure what you mean by this. Agentless reporting is meant to replace our AgentWriter for CIVisibility, and the latter is meant to be used only if Agentless mode is not enabled. There is no case where the user should specify using the AgentWriter in a CIVisibility context, especially since the tracer is wrapped on a private instance of the `CIVisibility` service.

Yun-Kim (approved):

ddtrace/internal/ci_visibility/recorder.py
mabdinur: The scenario I am describing is if an application manually configures an agent writer (for example this can occur if sampling priority, sampler or statd url is updated) and then an integration test run which executes this code path.

In this scenario the agentless ci visibility writer could be replaced with a datadog agent writer. Since a datadog agent may not be running in the ci environment then spans will not be submitted. 

A fix could be to update the 'Tracer.configure()' method to check the type of the configured writer. Currently it always replaces the current writer with an agent writer.

emmettbutler: I think I understand. You're suggesting a change like this, right? https://github.com/DataDog/dd-trace-py/compare/emmett.butler/conf-writer

PR #5497, opened 2023-04-07
feat(pytest): send payloads to agent EVP proxy endpoint when available

emmettbutler:
This change makes the `CIVisibilityWriter` check the reachability of the Datadog Agent's `citestcycle` endpoint on startup. If it's reachable, the writer starts sending its payloads to that agent endpoint instead of the standard intake URL.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/internal/ci_visibility/writer.py
```diff
@@ -81,3 +88,27 @@ def recreate(self):
...
+    def _set_state(self, new_state):
+        if new_state == self.STATE_AGENTPROXY:
+            self._state = new_state
+            self._endpoint = EVP_PROXY_AGENT_ENDPOINT
+            self.intake_url = agent.get_trace_url()
+            self._headers["X-Datadog-EVP-Subdomain"] = "citestcycle-intake"
+
+    def flush_queue(self, raise_exc=False):
+        if self._state != self.STATE_AGENTPROXY:
+            self._set_state(self._check_agent())
```
albertvaka, line 113: This calls `_check_agent()` each time we call `flush_queue ()` which means doing an additional request. Should we cache the result that `agent.info()` returns for N seconds to avoid making lots of requests? Just an idea for an improvement, not a blocker, since it's a local request so it should be fast.

albertvaka (dismissed):

ddtrace/internal/ci_visibility/writer.py
```diff
@@ -81,3 +88,27 @@ def recreate(self):
...
+            if endpoints and (EVP_PROXY_AGENT_BASE_PATH in endpoints or ("/" + EVP_PROXY_AGENT_BASE_PATH) in endpoints):
+                return self.STATE_AGENTPROXY
+        return self._state
+
+    def _set_state(self, new_state):
+        if new_state == self.STATE_AGENTPROXY:
+            self._state = new_state
+            self._endpoint = EVP_PROXY_AGENT_ENDPOINT
+            self.intake_url = agent.get_trace_url()
+            self._headers["X-Datadog-EVP-Subdomain"] = "citestcycle-intake"
```
Yun-Kim, line 109: can we make `citestcycle-intake` a constant?

ddtrace/internal/ci_visibility/writer.py
Yun-Kim: Do we need to check the agent each flush? I wonder if switching from agentless to EVP proxy via the agent is something that should happen in the middle of a CI job.

emmettbutler: @Yun-Kim I don't know. @albertvaka do you have an answer about whether the agent check should happen once on startup or during execution?

Yun-Kim: cc'ing @juan-fernandez here

juan-fernandez: > Do we need to check the agent each flush?

no, we don't. We can do it once at the beginning of the test session and assume that the agent capabilities will stay fixed until the end of the test session. That's what other tracers are doing.

> I wonder if switching from agentless to EVP proxy via the agent is something that should happen in the middle of a CI job.

We can assume we are not prepared for that. We'll decide configuration at the beginning of the test session.

ddtrace/internal/ci_visibility/writer.py
```diff
@@ -52,9 +59,11 @@ def __init__(
         if not headers["dd-api-key"]:
             raise ValueError("Required environment variable DD_API_KEY not defined")
 
+        self._state = self.STATE_AGENTLESS
```
Yun-Kim, line 62: Are we making agentless the default behavior? I thought agentless is only enabled when `DD_CIVISIBILITY_AGENTLESS_ENABLED=true`.

ddtrace/settings/config.py
```diff
@@ -286,7 +286,7 @@ def __init__(self):
                 log.warning("Invalid obfuscation pattern, disabling query string tracing")
                 self.http_tag_query_string = False  # Disable query string tagging if malformed obfuscation pattern
 
-        self._ci_visibility_agentless_enabled = asbool(os.getenv("DD_CIVISIBILITY_AGENTLESS_ENABLED", default=False))
+        self._ci_visibility_enabled = asbool(os.getenv("DD_CIVISIBILITY_ENABLED", default=False))
```
Yun-Kim, line 289: Does `DD_CIVISIBILITY_ENABLED` match with other tracer client libraries? I'm confused what the purpose of this environment variable is.

ddtrace/internal/ci_visibility/writer.py
```diff
@@ -52,9 +59,11 @@ def __init__(
...
+
         super(CIVisibilityWriter, self).__init__(
             intake_url=intake_url,
-            endpoint="api/v2/citestcycle",
+            endpoint=AGENTLESS_ENDPOINT,
```
Yun-Kim, line 66: If we set this as the default endpoint, what happens when the `_check_agent()` determines that the agent supports EVP proxy? Does the `CIVisibilityWriter` then change this endpoint back to the agent? Or is this endpoint the same for both agent and agentless?

docs/configuration.rst
```diff
@@ -501,10 +501,10 @@ The following environment variables for the tracer are supported:
      version_added:
         v1.9.0:
 
-   DD_CIVISIBILITY_AGENTLESS_ENABLED:
+   DD_CIVISIBILITY_ENABLED:
      type: Boolean
      default: False
-     description: Configures the ``CIVisibility`` service to use an agent-less test reporting ``CIVisibilityWriter``.
+     description: Configures the ``CIVisibility`` service to use a test-reporting ``CIVisibilityWriter``.
```
Yun-Kim, line 507: I think we should clarify either (or both) here or in the PR description what `test-reporting CIVisibilityWriter` means exactly for the user. My understanding so far is that the `CIVisibilityWriter` will send test events to the agentless endpoint by default, or to the agent via EVP proxy if the agent supports it. If `DD_CIVISIBILITY_ENABLED` is false, then we continue reporting test traces directly to the agent. Is that correct?

ddtrace/internal/ci_visibility/recorder.py
```diff
@@ -41,7 +41,7 @@ def __init__(self, tracer=None, config=None, service=None):
         super(CIVisibility, self).__init__()
 
         self.tracer = tracer or ddtrace.tracer
-        if ddconfig._ci_visibility_agentless_enabled and not isinstance(self.tracer._writer, CIVisibilityWriter):
+        if ddconfig._ci_visibility_enabled and not isinstance(self.tracer._writer, CIVisibilityWriter):
```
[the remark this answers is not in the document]
emmettbutler: It's a defensive programming move to guard against the possibility of the writer being set multiple times to the same thing.

ddtrace/internal/ci_visibility/writer.py
emmettbutler: The default behavior is that this `CIVisibilityWriter` class is not used at all. If `_ci_visibility_enabled` is set during `CIVisibility.enable()`, then `CIVisibilityWriter` is used. The choice of whether the `CIVisibilityWriter` will use the proxy or the agentless endpoint is not user-configurable - it depends only on whether the agent's proxy endpoint is reachable.

Keep in mind that this describes my own intent based on my interpretation of the spec. I'm working under the assumption that there are little mismatches between my interpretation and others', and that the beta-testing phase is the moment to fully resolve those mismatches.

ddtrace/internal/ci_visibility/writer.py
emmettbutler: In that situation, the endpoint gets changed [here](https://github.com/DataDog/dd-trace-py/pull/5497/files#diff-8ed1a099c4aaa47a54c8cb4529f67c3462e29a36b171c4fa508605cc644c4282R110).

ddtrace/settings/config.py
emmettbutler: When I made this change, I forgot that this envvar is defined in the spec. I'll revert it.

ddtrace/internal/ci_visibility/writer.py
emmettbutler: My above comment still applies, but the situation might be a little clearer now that I've reverted the envvar name change. My confusion was that that envvar is meant to control whether or not this writer is used, and this writer has an "agent-ful" mode, so using a flag called "agentless" to control it is unintuitive. The diff at this point reflects my best understanding of the spec.

emmettbutler: @juan-fernandez should it be possible for the EVP proxy to be used when the `DD_CIVISIBILITY_AGENTLESS_ENABLED` flag is set? From my read of the specs and RFCs I haven't been able to find a clear answer to this question.

This pull request currently assumes that the answer is "yes".

Yun-Kim: Agreed, I think we would benefit from more clarity on how agentless/proxy configuration is controlled via this environment variable.

juan-fernandez: The rules are the following:

* If `DD_CIVISIBILITY_AGENTLESS_ENABLED` is set _and `DD_API_KEY` is available_, we'll _exclusively_ use agentless mode to report test information. 
* If `DD_CIVISIBILITY_AGENTLESS_ENABLED` is set _and `DD_API_KEY` is **not** available_, we'll log an error and not report anything (no fallback to agentful).
* If `DD_CIVISIBILITY_AGENTLESS_ENABLED` is **not** set, we'll _ask the agent_ whether EVP proxy is available. 
  * If it is, we'll assume EVP proxy is always available until the end of the test session. 
  * If it isn't, we'll assume EVP proxy is never available.

emmettbutler: Thanks @juan-fernandez! I encoded this information in new integration tests.

ddtrace/internal/ci_visibility/recorder.py
```diff
@@ -67,6 +71,35 @@ def __init__(self, tracer=None, config=None, service=None):
         except Exception:
             log.warning("Failed to load CODEOWNERS", exc_info=True)
 
+    def _configure_writer(self):
+        writer = None
+        if ddconfig._ci_visibility_agentless_enabled:
+            site_id = os.environ.get("DD_SITE")
```
Yun-Kim, line 77: https://github.com/DataDog/dd-trace-py/pull/5497#discussion_r1166445998 I think we need to check for `DD_API_KEY` here and not `DD_SITE`?

ddtrace/internal/ci_visibility/recorder.py
```diff
@@ -67,6 +71,35 @@ def __init__(self, tracer=None, config=None, service=None):
...
+            if site_id:
+                writer = CIVisibilityWriter(intake_url="https://citestcycle-intake.%s" % site_id)
+            else:
+                log.error("Environment variable DD_SITE not set - not sending CI Visibility data")
```
Yun-Kim, line 81: If `DD_API_KEY` isn't set, according to https://github.com/DataDog/dd-trace-py/pull/5497#discussion_r1166445998 we shouldn't report anything right? Should we raise an exception here instead?

emmettbutler: This is another area of ambiguity. @juan-fernandez if `DD_API_KEY` is unset and `DD_CIVISIBILITY_AGENTLESS_ENABLED` is set, should ddtrace

* crash with an error message; or
* generate a warning message and let the test suite continue to run, sending non ci-visibility spans from the pytest integration to the datadog agent

ddtrace/internal/ci_visibility/recorder.py
emmettbutler: Good catch, thanks!

ddtrace/internal/ci_visibility/recorder.py
juan-fernandez: other tracers log an error message (or warning, depending on the language) and continue the test execution _without any tracing or ci visibility_, that is, the tracer is not init. Example: https://github.com/DataDog/dd-trace-js/blob/master/ci/init.js#L27-L30

emmettbutler: Sounds good, thank you. I've adjusted the diff to do the same.

ddtrace/contrib/pytest/plugin.py
```diff
@@ -83,7 +83,10 @@ def pytest_addoption(parser):
 def pytest_configure(config):
     config.addinivalue_line("markers", "dd_tags(**kwargs): add tags to current span")
     if is_enabled(config):
-        _CIVisibility.enable(config=ddtrace.config.pytest)
+        try:
+            _CIVisibility.enable(config=ddtrace.config.pytest)
+        except ValueError:
+            log.warning("Failed to initialize ddtrace")
```
Yun-Kim, line 89: What error led to this guard?

ddtrace/internal/ci_visibility/recorder.py
```diff
@@ -72,6 +75,41 @@ def __init__(self, tracer=None, config=None, service=None):
         except Exception:
             log.warning("Failed to load CODEOWNERS", exc_info=True)
 
+    def _configure_writer(self):
+        writer = None
+        if ddconfig._ci_visibility_agentless_enabled:
+            headers = {"dd-api-key": os.environ.get("DD_API_KEY")}
+            if headers["dd-api-key"]:
+                writer = CIVisibilityWriter(
+                    intake_url="https://citestcycle-intake.%s" % os.environ.get("DD_SITE", "datadoghq.com"),
```
Yun-Kim, line 84: Can we make the `https://citestcycle-intake.` a constant?

ddtrace/internal/ci_visibility/writer.py
```diff
@@ -51,16 +59,13 @@ def __init__(
...
         headers=None,  # type: Optional[Dict[str, str]]
+        use_evp=False,  # type: bool
     ):
         if not intake_url:
-            intake_url = "https://citestcycle-intake.%s" % os.environ.get("DD_SITE", "datadoghq.com")
-        headers = headers or dict()
-        headers["dd-api-key"] = os.environ.get("DD_API_KEY") or ""
-        if not headers["dd-api-key"]:
-            raise ValueError("Required environment variable DD_API_KEY not defined")
+            intake_url = "https://citestcycle-intake.datadoghq.com"
```
Yun-Kim, line 65: Could we also make this default intake url a constant?

ddtrace/contrib/pytest/plugin.py
emmettbutler: Catching the exception here allows the test suite to continue running while civisibility is inactive, as opposed to crashing the entire program by letting the exception bubble up. The error is the one added in https://github.com/DataDog/dd-trace-py/pull/5497/commits/3f96d292330369b312eff78faaaabc0381b79177. This change is my interpretation of this recommendation: https://github.com/DataDog/dd-trace-py/pull/5497#discussion_r1170101070

ddtrace/internal/ci_visibility/recorder.py
juan-fernandez: my last comment is wrong, my bad. Javascript is the odd one here. Let's fail the test process if `DD_CIVISIBILITY_AGENTLESS_ENABLED` is set and `DD_API_KEY` is unset.

emmettbutler:
@juan-fernandez @Yun-Kim this is reviewable again. I think I have it matching all of the various discussions and adjustments we've noted in the last few days. Please review when you're able!

ddtrace/internal/ci_visibility/recorder.py
```diff
@@ -72,6 +77,41 @@ def __init__(self, tracer=None, config=None, service=None):
...
+                writer = CIVisibilityWriter(
+                    intake_url="%s.%s" % (AGENTLESS_BASE_URL, os.environ.get("DD_SITE", AGENTLESS_DEFAULT_SITE)),
+                    headers=headers,
+                )
+            else:
+                raise EnvironmentError(
+                    "DD_CIVISIBILITY_AGENTLESS_ENABLED is set, but DD_API_KEY is not set, so ddtrace "
+                    "cannot be initialized."
+                )
+        elif self._agent_evp_proxy_is_available():
```
albertvaka, line 94: Do we need a final `else` case where we fall back to serialising to regular APM traces (ie; the behaviour of the currently released tracer), or will that happen automatically?

albertvaka (approved):
LGTM but I'm not sure there's a fallback to sending traces, left a comment about that.

ddtrace/internal/ci_visibility/recorder.py
emmettbutler: I believe this conflicts with @juan-fernandez's comment here: https://github.com/DataDog/dd-trace-py/pull/5497#discussion_r1171082980

albertvaka: I mean when `DD_CIVISIBILITY_AGENTLESS_ENABLED` is not set. First we should try to use the EVP proxy, but if it is not available, fall back to sending just test traces.

albertvaka: Pseudocode:
```
if DD_CIVISIBILITY_AGENTLESS_ENABLED:
    if not DD_APIKEY:
        throw
    else:
        writer = AgentlessWriter
elif evp_proxy_available:
    writer = EvpProxyWriter
else:
    writer = ApmWriter
```

ddtrace/internal/ci_visibility/writer.py
```diff
@@ -65,8 +66,12 @@ def __init__(
         if not intake_url:
             intake_url = "%s.%s" % (AGENTLESS_BASE_URL, AGENTLESS_DEFAULT_SITE)
 
-        client = CIVisibilityProxiedEventClient() if use_evp else CIVisibilityAgentlessEventClient()
         headers = headers or dict()
+        headers["dd-api-key"] = os.environ.get("DD_API_KEY") or ""
+        if not headers["dd-api-key"]:
+            raise ValueError("Required environment variable DD_API_KEY not defined")
```
albertvaka, line 72: We don't want this in EVP Proxy mode, though.

ddtrace/internal/ci_visibility/recorder.py
```diff
@@ -72,6 +77,41 @@ def __init__(self, tracer=None, config=None, service=None):
...
+
+    def _agent_evp_proxy_is_available(self):
+        try:
+            info = agent.info()
+        except Exception:
+            info = None
+
+        if info:
+            endpoints = info.get("endpoints", [])
+            if endpoints and (EVP_PROXY_AGENT_BASE_PATH in endpoints or ("/" + EVP_PROXY_AGENT_BASE_PATH) in endpoints):
```
juan-fernandez, line 111: nitpick: would this be more robust? I see that some endpoints in the `/info` response include a final `/` and some don't. 
```suggestion
            if endpoints and any(EVP_PROXY_AGENT_BASE_PATH in endpoint for endpoint in endpoints):
```
for this `EVP_PROXY_AGENT_BASE_PATH` would have to remove the trailing slash `/`

ddtrace/internal/ci_visibility/constants.py
```diff
@@ -0,0 +1,7 @@
+EVP_PROXY_AGENT_BASE_PATH = "evp_proxy/v2/"
+EVP_PROXY_AGENT_ENDPOINT = "{}api/v2/citestcycle".format(EVP_PROXY_AGENT_BASE_PATH)
```
juan-fernandez, line 2: nitpick: this is a requirement for the suggestion below. This way you can "forget" about leading and trailing slashes when checking evp proxy compatibility 

```suggestion
EVP_PROXY_AGENT_BASE_PATH = "evp_proxy/v2"
EVP_PROXY_AGENT_ENDPOINT = "{}/api/v2/citestcycle".format(EVP_PROXY_AGENT_BASE_PATH)
```

juan-fernandez (commented):
I have two nitpicks and an important comment: `DD_API_KEY` should _not_ be required if agentless mode is not being used. I believe since `CIVisibilityWriter` is being reused for both agentless and evp proxy, we fail when `DD_API_KEY` is not supplied when EVP proxy is available. 

The rest seems to be working nicely 🎉

ddtrace/internal/ci_visibility/writer.py
juan-fernandez: yeah good catch, I saw this when running e2e tests

albertvaka (approved):
🚀

juan-fernandez (approved):
e2e tests are now passing :)

Yun-Kim (approved):

PR #5498, opened 2023-04-07
chore(profiling): fix flaky test_asyncio test

tabgok:
This change ensures that the MainThread Task with name 'None' is run long enough to be targeted by an event collector during this test.

Prior to this change the test
tests/profiling/collector/test_stack_asyncio.py::test_asyncio failed occasionally due to a race condition between event collectors (which scan threads) and collecting data about the MainThread with name=None. If the collectors didn't run in time to catch the MainThread, no CPU time would be recorded.

After this change, the StackCollector class has been updated to have a 'run count' variable, and sleeps are performed until the run time is incremented at least once, thus guaranteeing a run (for both wall_clock and cpu_time).

Sample error:

```
        assert wall_time_ns[t1_name] > 0
        assert wall_time_ns[t2_name] > 0
        if sys.platform != "win32":
            # Windows seems to get 0 CPU for this
>           assert cpu_time_found
E           assert False

tests/profiling/collector/test_stack_asyncio.py:90: AssertionError
```

Testing

I have tested this change by running the test over 50 times without error.

Risks

None, this is an update to testing only.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.

### DIFF
--- a/tests/profiling/collector/test_stack_asyncio.py
+++ b/tests/profiling/collector/test_stack_asyncio.py
@@ -1,42 +1,69 @@
 import asyncio
 import collections
 import sys
+import types
 
 import pytest
 
 from ddtrace.profiling import _asyncio
 from ddtrace.profiling import profiler
 from ddtrace.profiling.collector import stack_event
+from ddtrace.profiling.collector.stack import StackCollector
 
 from . import _asyncio_compat
+
+
+def patch_stack_collector(stack_collector):
+    """
+    Patch a stack collect so we can count how many times it has run
+    """
+
+    def _collect(self):
+        self.run_count += 1
+        return self._orig_collect()
+
+    stack_collector.run_count = 0
+    orig = stack_collector.collect
+    stack_collector._orig_collect = orig
+    stack_collector.collect = types.MethodType(_collect, stack_collector)
 
 
 @pytest.mark.skipif(not _asyncio_compat.PY36_AND_LATER, reason="Python > 3.5 needed")
 def test_asyncio(tmp_path, monkeypatch) -> None:
     sleep_time = 0.2
-    sleep_times = 5
 
-    async def stuff() -> None:
-        await asyncio.sleep(sleep_time)
+    async def stuff(collector) -> None:
+        count = collector.run_count
+        while collector.run_count == count:
+            await asyncio.sleep(sleep_time)
 
-    async def hello() -> None:
-        t1 = _asyncio_compat.create_task(stuff(), name="sleep 1")
-        t2 = _asyncio_compat.create_task(stuff(), name="sleep 2")
-        for _ in range(sleep_times):
-            await stuff()
+    async def hello(collector) -> None:
+        t1 = _asyncio_compat.create_task(stuff(collector), name="sleep 1")
+        t2 = _asyncio_compat.create_task(stuff(collector), name="sleep 2")
+        await stuff(collector)
         return (t1, t2)
 
     monkeypatch.setenv("DD_PROFILING_CAPTURE_PCT", "100")
     monkeypatch.setenv("DD_PROFILING_OUTPUT_PPROF", str(tmp_path / "pprof"))
     # start a complete profiler so asyncio policy is setup
     p = profiler.Profiler()
+    stack_collector = [collector for collector in p._profiler._collectors if type(collector) == StackCollector][0]
+    patch_stack_collector(stack_collector)
+
     p.start()
     loop = asyncio.new_event_loop()
     asyncio.set_event_loop(loop)
     if _asyncio_compat.PY38_AND_LATER:
-        maintask = loop.create_task(hello(), name="main")
+        maintask = loop.create_task(hello(stack_collector), name="main")
     else:
-        maintask = loop.create_task(hello())
+        maintask = loop.create_task(hello(stack_collector))
+
+    # Wait for the collector to run at least once on this thread, while it is doing something
+    # 2.5+ seconds at times
+    count = stack_collector.run_count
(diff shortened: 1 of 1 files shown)


emmettbutler (dismissed):

tests/profiling/collector/test_stack_asyncio.py
```diff
@@ -37,6 +39,11 @@ async def hello() -> None:
         maintask = loop.create_task(hello(), name="main")
     else:
         maintask = loop.create_task(hello())
+
+    # Make sure we spend enough time in the main task (event.name=None) in MainThread for profile events to be collected
+    # So we wait for twice the collection time to account for small variances in time.sleep()
+    time.sleep(2 * _default_min_interval_time())
```
sanchda, line 45: It makes sense to me that increasing the collection time will fix flakiness caused by inadequate collection, but I'd like to discuss the explanation a bit.  If I say something agreeable, would you update the comment accordingly?

AFAIK, `time.sleep()` as of Python 3.5 will auto-restart if the runtime catches a signal.  It also makes the standard provision that the amount slept will be _at least_ as long as the amount specified, possibly longer.  What this means is, the `time.sleep()` on line 45 will be at _least_ as long as two minimum interval times.  In other words, time-variation of this sleep favors "more collection" than would be a priori expected from just the specified times/frequencies alone.

On the other hand, I think what's actually happening here is that rather than getting short-sleeps from this `time.sleep()`, we're actually getting _long_ collection intervals from `ddtrace/profiling/collector/stack.pyx:StackCollector.collect()`.  `collect()` will exacerbate any case where the scheduling mechanism _also_ provides that the sleep is "at least as long" as the one specified, since it'll only collect if the total waited time exceeds the specified interval (and you'll practically never sleep right _on_ the interval).

If this is accurate, then I think we should specify that we're accounting for variations in the collection sampling interval.

sanchda (approved):
LGTM.  Added a comment which I _think_ should hold true under asyncio, but please feel free to ignore it for the purposes of merging if I'm incorrect.  I'm just trying to record the most accurate cause we have for test flakiness, in case this learning is applied elsewhere.

Thank you!

tests/profiling/collector/test_stack_asyncio.py
tabgok: I will update the comment, you're hi-lighting the correct behavior!

sanchda (approved):
LGTM

Yun-Kim (approved):

emmettbutler (approved):

Yun-Kim (dismissed):
Thanks for the much needed fix to the flaky test @tabgok! I've gone ahead and squashed/signed your commits so that we can merge.

sanchda (dismissed):
This is a pretty good fix.  It addresses the problem of indeterminacy of collection rate without deferring to another collection mechanism (such as allocation sampling).

There might be other tests that ought to adopt this approach--I'll find some time to review what we've got...

LGTM, thank you!

sanchda (approved):

tests/profiling/collector/test_stack_asyncio.py
```diff
@@ -1,42 +1,69 @@
...
 def test_asyncio(tmp_path, monkeypatch) -> None:
     sleep_time = 0.2
-    sleep_times = 5
 
-    async def stuff() -> None:
-        await asyncio.sleep(sleep_time)
+    async def stuff(collector) -> None:
+        count = collector.run_count
+        while collector.run_count == count:
+            await asyncio.sleep(sleep_time)
```
P403n1x87, line 38: Does this approach not come with the potential risk of hanging forever if the collector is not collecting because e.g. a regression? Should we consider adding some form of timeout? Same for the similar coding below.

tests/profiling/collector/test_stack_asyncio.py
```diff
@@ -1,42 +1,69 @@
...
+        t1 = _asyncio_compat.create_task(stuff(collector), name="sleep 1")
+        t2 = _asyncio_compat.create_task(stuff(collector), name="sleep 2")
+        await stuff(collector)
         return (t1, t2)
 
     monkeypatch.setenv("DD_PROFILING_CAPTURE_PCT", "100")
     monkeypatch.setenv("DD_PROFILING_OUTPUT_PPROF", str(tmp_path / "pprof"))
     # start a complete profiler so asyncio policy is setup
     p = profiler.Profiler()
+    stack_collector = [collector for collector in p._profiler._collectors if type(collector) == StackCollector][0]
```
P403n1x87, line 50: nit: this perhaps could have been passed via a closure to the local coroutines

tests/profiling/collector/test_stack_asyncio.py
```diff
@@ -1,42 +1,69 @@
...
+    """
+
+    def _collect(self):
+        self.run_count += 1
+        return self._orig_collect()
+
+    stack_collector.run_count = 0
+    orig = stack_collector.collect
+    stack_collector._orig_collect = orig
+    stack_collector.collect = types.MethodType(_collect, stack_collector)
```
P403n1x87, line 28:
```python
stack_collector.collect = _collect.__get__(stack_collector)
```

P403n1x87: We also have utilities in `ddtrace.internal.wrapping` for wrapping/unwrapping functions/methods of already instantiated stuff. This could perhaps be made into a fixture with setup and teardown steps.

P403n1x87 (commented):
Thanks for taking the time to improve this. I've left a few comments in case you'd be interested in a follow-up.

PR #5499, opened 2023-04-07
feat(profiling): provide libdatadog as collection and export backend

sanchda:
libdatadog is a native library which provides many utilities useful for various Datadog projects.  Notably, it provides a standardized mechanism for collecting and uploading profiling data to the Datadog backend.

This change allows users to specify the libdatadog collector+exporter (or both libdatadog and the "current" Python implementation).

This change intends to be a 1-1 equivalent replacement for the "old" way.  It should support every feature supported by the mainline Python profiler.

## Checklist

- [ ] PR description includes explicit acknowledgement/acceptance of the performance implications of this PR as reported in the benchmarks PR comment.

## Reviewer Checklist

- [ ] Title is accurate.


ddtrace/datadog/ddup.pyx
P403n1x87, line 1: According to our naming conventions, this would make the `ddtrace.datadog` module public. Perhaps we should re-parent this under `ddtrace.internal.datadog`. And maybe, in light of a wider adoption of libdatadog across other products (tracing, DI etc...), we should also consider the use of a sub-module for product-specific stuff, e.g. `ddtrace.internal.datadog.profiling`. We might have the need to define an `exporter` for traces, so this way we won't have source clashes. WDYT?

P403n1x87:
Really awesome stuff! 💪  Just a thought on perhaps splitting this PR into smaller chunks to make it easier to digest. I was thinking a possible way to split it is to have a PR that introduces the new backend (as essentially dead code), and a follow-up that actually wires it in the profiler. WDYT?

sanchda:
> Just a thought on perhaps splitting this PR into smaller chunks to make it easier to digest.
Yeah!  I think that's a great idea, and one I was thinking of doing as well.  I still don't know exactly how to sequence the PRs though, but I think splitting into ~3 PRs would be much better than sending along this huge beast.

ddtrace/datadog/ddup.pyx
sanchda: Ah, good idea.  The naming of `ddup` in particular was tentative.  We have a guild which is considering how to best implement libdatadog into ddtrace for various reasons, so I think it's appropriate to surface this with that group.  Will do that.

ddtrace/datadog/Readme.md
```diff
@@ -0,0 +1,6 @@
+This is a very rough sketch for how to use libdatadog
+
+1. Download the v2 release of libdatadog from https://github.com/DataDog/libdatadog/releases/tag/v2.0.0
+2. untar that release to this directory.  The harness assumes you're using the libc-x86 version.  You'll have to change the toplevel pyrun.sh and build.sh accordingly if you downloaded something else.  Obviously this won't be a feature of the finished version of this...
+3. Execute any Python scripts through the toplevel pyrun.sh, like `./pyrun.sh tests/profiling/collatz.py
```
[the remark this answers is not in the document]
sanchda: Ah, thank you--this Readme is actually defunct now and I should have deleted it...

ddtrace/datadog/src/exporter.cpp
```diff
@@ -0,0 +1,500 @@
...
+      to_slice(profiler_version),
+      to_slice(family),
+      &tags,
+      ddog_Endpoint_agent(to_slice(url)));
+  ddog_Vec_Tag_drop(tags);
+
+  if (new_exporter.tag == DDOG_PROF_EXPORTER_NEW_RESULT_OK)
+    ptr = new_exporter.ok;
+  else
+    std::cout << "ERROR INITIALIZING LIBDATADOG EXPORTER" << std::endl;
```
r1viollet, line 47: I think we are missing an error drop here

ddtrace/datadog/src/exporter.cpp
```diff
@@ -0,0 +1,500 @@
...
+}
+
+DdogProfExporter::~DdogProfExporter() {
+  ddog_prof_Exporter_drop(ptr);
```
r1viollet, line 51: we could make this a unique ptr with a custom deleter (I'm mostly thinking of how to remove the copy constructors as I'm not sure how this class is used)

ddtrace/datadog/src/exporter.cpp
```diff
@@ -0,0 +1,500 @@
...
+  version(_version),
+  url(_url) {
+    ddog_exporter = std::make_unique<DdogProfExporter>(env, service, version, url);
+}
+
+void DdogProfExporter::add_tag(ddog_Vec_Tag &tags, std::string_view key, std::string_view val) {
+  ddog_Vec_Tag_PushResult res = ddog_Vec_Tag_push(&tags, to_slice(key), to_slice(val));
+  if (res.tag == DDOG_VEC_TAG_PUSH_RESULT_ERR) {
+    std::cout << "Error pushing tag '" << key << "'->'" << val << "'" << std::endl;
+    std::cout << "  err: " << ddog_Error_message(&res.err).ptr << std::endl;
```
r1viollet, line 69: We are missing an error drop in this place.

ddtrace/datadog/src/exporter.cpp
```diff
@@ -0,0 +1,500 @@
...
+
+Uploader::Uploader(std::string_view _env,
+                   std::string_view _service,
+                   std::string_view _version,
+                   std::string_view _url) :
+  env(_env),
+  service(_service),
+  version(_version),
+  url(_url) {
+    ddog_exporter = std::make_unique<DdogProfExporter>(env, service, version, url);
```
r1viollet, line 62: I like the C++ idiom, though I have trouble wrapping my head around the error management

ddtrace/datadog/src/exporter.cpp
```diff
@@ -0,0 +1,500 @@
...
+#include <fcntl.h>
+#include <unistd.h>
+
+bool Uploader::upload(const Profile *profile) {
+  ddog_prof_Profile_SerializeResult result = ddog_prof_Profile_serialize(
+      profile->ddog_profile,
+      nullptr,
+      nullptr);
+  if (result.tag != DDOG_PROF_PROFILE_SERIALIZE_RESULT_OK) {
+    std::cout << "Failure serializing pprof" << std::endl;
```
r1viollet, line 84: Error drop :- )

ddtrace/datadog/src/exporter.cpp
```diff
@@ -0,0 +1,500 @@
...
+    return false;
+  }
+
+  ddog_prof_EncodedProfile *encoded = &result.ok;
```
r1viollet, line 88: This should be dropped after usage. I don't think libdatadog claims ownership even in the standard case, so this would leak pretty fast.

ddtrace/datadog/src/exporter.cpp
sanchda: > not sure how this class is used

`DdogProfExporter` is an opaque wrapper class over libdatadog's `ddog_prof_Exporter`.   Every instance of `ddog_prof_Exporter` is under the management of libdatadog, so they are returned as pointers.  I wanted to consolidate the C-isms of the class into RAII, so

1. `DdogProfExporter` calls the appropriate libdatadog functions to create and destroy the `ddog_prof_Exporter` instance.  It also includes a few string-invariants which are only sent to libdatadog during instantiation of `ddog_prof_Exporter`.
2. the `Uploader` class takes a unique_ptr to `DdogProfExporter`.  This is mostly done to defer initialization of the field, so it occurs once inside of the constructor.

The lifetime and scope of `Uploader` and `DdogProfExporter` are comparable, so splitting the wrapper into its own class is a question of aesthetics, as far as I can tell.

ddtrace/datadog/src/exporter.cpp
sanchda: Ah, this whole section is actually problematic.

ddtrace/datadog/src/exporter.cpp
```diff
@@ -0,0 +1,500 @@
...
+  // Close out request
+
+  if (res.tag == DDOG_PROF_EXPORTER_SEND_RESULT_ERR) {
+    std::cout << "Failed to send result to backend" << std::endl;
+    std::cout << "  url: " <<  url << std::endl;
+    std::cout << "  " << ddog_Error_message(&res.err).ptr << std::endl;
+  }
+
+  // We're done exporting, so reset the profile
+  if (!ddog_prof_Profile_reset(profile->ddog_profile, nullptr)) {
```
r1viollet, line 146: minor - resetting within an upload function might be seem strange

ddtrace/datadog/src/exporter.cpp
```diff
@@ -0,0 +1,500 @@
...
+  }
+
+//  ddog_prof_Profile_SerializeResult_drop(result);
+//  ddog_prof_Exporter_Request_drop(req);
+//  ddog_prof_Exporter_SendResult_drop(res);
+
+  return true;
+}
+
+Profile::Profile() {
```
r1viollet, line 158: Just stating the obvious: We will need some state to know what profile types are active ?

sanchda: Indeed!  I originally had that state, but then took it out to mirror what the Python profiler did (which is to populate *every* available profiling type).  This was primarily so that my overhead figures weren't overblown due to trivial effects.

Lately, there's been some interest in fixing the original Python exporter so that it only populates the pprof entries that it needs to.  I think the correct thing to do is catch the configured profiling types and only provide storage for those.

ddtrace/datadog/include/exporter.hpp
```diff
@@ -0,0 +1,163 @@
...
+           std::string_view _url = "http://localhost:8126");
+
+  bool upload(const Profile *profile);
+
+};
+
+class Profile {
+  bool is_valid = false;
+
+  std::vector<ddog_prof_Location> locations;
```
r1viollet, line 60: I think we are using these mostly as standard arrays as we expect stable addressing.

ddtrace/datadog/src/exporter.cpp
```diff
@@ -0,0 +1,500 @@
...
+       address,
+       {&lines.back(), 1},
+       false,
+  });
+
+  ++frames;
+}
+
+void Profile::push_label(const std::string_view &key, const std::string_view &val) {
+  labels[cur_label].key = to_slice(key);
```
r1viollet, line 233: You would need to guarantee the static lifetime of the key or store it.

r1viollet:
Super happy to see this shape up! I'd be happy to go through it one more time once we figure out the error management parts.

ddtrace/datadog/Readme.md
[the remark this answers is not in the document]
sanchda: Fixed in 429e72865

sanchda:
Consolidated some fixups from a different branch; feel free to review, but I'll be pushing this out to a stacked PR soon.

ddtrace/datadog/ddup.pyx
sanchda: FYI, moved things into ddtrace.internal.datadog.profiling in e39dfe0db.

ddtrace/internal/datadog/profiling/src/exporter.cpp
```diff
@@ -0,0 +1,560 @@
...
+  return true;
+}
+
+void Profile::push_frame_impl(std::string_view name, std::string_view filename,
+                              uint64_t address, int64_t line) {
+
+  // Ensure strings are stored.
+  // Slightly wasteful since it requires allocating another string.
+  auto insert_or_get = [&](std::string_view sv) -> std::string_view {
+    std::string str(sv);
```
r1viollet, line 357: To avoid copy, you can use a custom hash function in the set. We have this in `ddprof` in the runtime_symbol_lookup.hpp file.

ddtrace/internal/datadog/profiling/src/exporter.cpp
```diff
@@ -0,0 +1,560 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0. This product includes software
+// developed at Datadog (https://www.datadoghq.com/). Copyright 2021-Present
+// Datadog, Inc.
+
+// High-level skip for invalid architectures
+#ifndef __linux__
+#elif __aarch64__
+#elif __i386__
+#else
```
r1viollet, line 10: This is probably to scope out some of the testing and build type changes.
The code seems to be OK for all these platforms / archs. 
I'm fine with leaving this. I'm just trying to figure out where the triggers should be.

ddtrace/internal/datadog/profiling/include/exporter.hpp
```diff
@@ -0,0 +1,261 @@
...
+  unsigned int nframes;
+
+  std::string errmsg;
+
+  // Keeps temporary buffer of frames in the stack
+  std::vector<ddog_prof_Location> locations;
+  std::vector<ddog_prof_Line> lines;
+
+  // Storage for strings
+  std::unordered_set<std::string> strings;
```
r1viollet, line 169: Minor: having class members prefixed with `_` really makes life easier, though I agree this is a matter of taste.

ddtrace/internal/datadog/profiling/src/exporter.cpp
```diff
@@ -0,0 +1,560 @@
...
+  constexpr std::array<std::string_view,
+                       static_cast<size_t>(ExportLabelKey::_Length)>
+      keys = {EXPORTER_LABELS(X_STR)};
+  std::string_view key_sv = keys[static_cast<size_t>(key)];
+  labels[cur_label].key = to_slice(key_sv);
+
+  // Label may not persist, so it needs to be saved
+  auto [it, _] = strings.insert(std::string{val});
+  labels[cur_label].str = to_slice(*it);
+  cur_label++;
```
r1viollet, line 401: It is probably best to add some safety bound check around this.

ddtrace/internal/datadog/profiling/src/exporter.cpp
```diff
@@ -0,0 +1,560 @@
...
+}
+
+void Profile::push_label(const ExportLabelKey key, int64_t val) {
+  constexpr std::array<std::string_view,
+                       static_cast<size_t>(ExportLabelKey::_Length)>
+      keys = {EXPORTER_LABELS(X_STR)};
+  std::string_view key_sv = keys[static_cast<size_t>(key)];
+  labels[cur_label].key = to_slice(key_sv);
+  labels[cur_label].num = val;
+  cur_label++;
```
r1viollet, line 411: same

ddtrace/internal/datadog/profiling/include/exporter.hpp
```diff
@@ -0,0 +1,261 @@
...
+
+private:
+  unsigned int type_mask;
+  unsigned int max_nframes;
+  unsigned int nframes;
+
+  std::string errmsg;
+
+  // Keeps temporary buffer of frames in the stack
+  std::vector<ddog_prof_Location> locations;
```
r1viollet, line 165: I think previous comments around using std::array remain valid. wdyt ?

ddtrace/internal/datadog/profiling/src/exporter.cpp
```diff
@@ -0,0 +1,560 @@
...
+  ddog_profile = ddog_prof_Profile_new({&samplers[0], samplers.size()},
+                                       &default_sampler, nullptr);
+
+  // Prepare for use
+  reset();
+
+  // Initialize the size for buffers
+  locations.reserve(2040);
+  lines.reserve(2048);
+  strings.reserve(8192);
```
r1viollet, line 331: Minor: The reserve around the string storage is no longer useful.

ddtrace/internal/datadog/profiling/src/exporter.cpp
```diff
@@ -0,0 +1,560 @@
...
+  // Slightly wasteful since it requires allocating another string.
+  auto insert_or_get = [&](std::string_view sv) -> std::string_view {
+    std::string str(sv);
+    auto [it, _] = strings.insert(std::move(str));
+    return *it;
+  };
+  name = insert_or_get(name);
+  filename = insert_or_get(filename);
+
+  lines.push_back({
```
r1viollet, line 364: Bound checks are important, though they will come with the change in storage container I assume.

ddtrace/internal/datadog/profiling/src/exporter.cpp
```diff
@@ -0,0 +1,560 @@
...
+    values[val_idx.lock_release_time] += release_time;
+    values[val_idx.lock_release_count] += count;
+  }
+  return true;
+}
+
+bool Datadog::Profile::push_alloc(uint64_t size, uint64_t count) {
+  if (type_mask & ProfileType::Allocation) {
+    values[val_idx.alloc_space] += size;
+    values[val_idx.alloc_count] += count;
```
r1viollet, line 500: minor: returning true when we are not pushing anything might seem strange.

ddtrace/internal/datadog/profiling/src/exporter.cpp
```diff
@@ -0,0 +1,560 @@
...
+
+  ddog_prof_Exporter_NewResult new_exporter = ddog_prof_Exporter_new(
+      to_slice("dd-trace-py"), to_slice(profiler_version), to_slice(family),
+      &tags, ddog_Endpoint_agent(to_slice(url)));
+  ddog_Vec_Tag_drop(tags);
+
+  if (new_exporter.tag == DDOG_PROF_EXPORTER_NEW_RESULT_OK) {
+    ptr = new_exporter.ok;
+  } else {
+    // TODO consolidate errors
```
r1viollet, line 54: I think the builder provides an easy way to at least stop profiing if things go wrong in the exporter creation.

ddtrace/internal/datadog/profiling/src/exporter.cpp
```diff
@@ -0,0 +1,560 @@
...
+
+#include "exporter.hpp"
+#include <iostream>
+
+#include <fcntl.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+using namespace Datadog;
```
r1viollet, line 20: minor, I think putting everything in the Datadog namespace is slightly better (and avoids some amount of Datadog::)

ddtrace/internal/datadog/profiling/src/exporter.cpp
```diff
@@ -0,0 +1,560 @@
...
+  auto get_value_idx = [&samplers](std::string_view value,
+                                   std::string_view unit) {
+    size_t idx = samplers.size();
+    samplers.push_back({to_slice(value), to_slice(unit)});
+    return idx;
+  };
+
+  // Check which samplers were enabled by the user
+  if (type_mask & ProfileType::CPU) {
+    val_idx.cpu_time = get_value_idx("cpu-time", "nanoseconds");
```
r1viollet, line 289: nice, we now have the dynamic configuration of types.

ddtrace/internal/datadog/profiling/include/exporter.hpp
```diff
@@ -0,0 +1,261 @@
...
+
+  std::string errmsg;
+
+public:
+  Uploader(std::string_view _env, std::string_view _service,
+           std::string_view _version, std::string_view _runtime,
+           std::string_view _runtime_version,
+           std::string_view _profiler_version, std::string_view _url,
+           DdogProfExporter::ExporterTagset &user_tags);
+  bool set_runtime_id(const std::string &id);
```
r1viollet, line 109: minor: could be a string view

ddtrace/internal/datadog/profiling/include/exporter.hpp
```diff
@@ -0,0 +1,261 @@
...
+  bool flush_sample();
+
+  // Clears temporary things
+  void clear_buffers();
+
+  // Zero out stats
+  void zero_stats();
+
+  bool reset();
+  Profile(ProfileType type, unsigned int _max_nframes = 64);
```
r1viollet, line 242: Minor: I'm not sure 64 is what you meant ?
Also aligning that with values above might be nice ? I'm not sure this will be a parameter (vs fixed value) if we use it as a storage size for std::arrays.

ddtrace/internal/datadog/profiling/include/interface.hpp
```diff
@@ -0,0 +1,47 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0. This product includes software
+// developed at Datadog (https://www.datadoghq.com/). Copyright 2021-Present
+// Datadog, Inc.
+
+#pragma once
+
+#include "exporter.hpp"
+
+extern "C" {
```
r1viollet, line 10: I think pyx has support for c++ ?

ddtrace/internal/datadog/profiling/src/interface.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+#include "interface.hpp"
+#include "exporter.hpp"
+
+#include <iostream>
+#include <thread>
+
+// State
+bool is_initialized = false;
+Datadog::Uploader *g_uploader;
```
r1viollet, line 20: Minor: I think we can wrap the c++ classes in cython objects
https://cython.readthedocs.io/en/latest/src/userguide/wrapping_CPlusPlus.html

r1viollet: This is probably something we can do in a latter PR.

ddtrace/internal/datadog/profiling/src/exporter.cpp
sanchda: Ah, good point.  The label definition was wrong and should have been set to the converted `_Length` of the enum.  Fixed in 61f10cc50

ddtrace/internal/datadog/profiling/include/exporter.hpp
sanchda: Indeed, fixed in 61f10cc50

ddtrace/internal/datadog/profiling/src/exporter.cpp
sanchda: This was changed in 61f10cc50 and subsequently had to be made more strict.

ddtrace/internal/datadog/profiling/src/exporter.cpp
sanchda: Very, very true.  Fixed in 61f10cc50, although I don't know whether I'll consume the return in the PR.

ddtrace/internal/datadog/profiling/src/exporter.cpp
sanchda: Very true.  Fixed in 61f10cc50 by

1. getting rid of the wrapper class over ddog_profile_Export
2. checking subsequent initialization in the builder
3. adding a custom deleter to the unique_ptr, which is now in the uploader

still need to consume the check, but I think the immediate part is better now.

ddtrace/internal/datadog/profiling/include/exporter.hpp
sanchda: Yeah, overspecifying defaults is kind of hard not to do.

The max_nframes is a user-defined maximum that defines the depth at which unwinding will stop.  It's actually enforced on the Python side, but we use the max_nframes parameter in order to compute how many frames were omitted (since we get the true stack depth).

I think I slightly fixed this in 61f10cc50 by changing parameter defaults and leaning more heavily into the builders.

ddtrace/internal/datadog/profiling/include/interface.hpp
sanchda: Indeed, it was just annoying to refactor everything in this PR.  It's definitely my plan to fix this.

ddtrace/internal/datadog/profiling/src/interface.cpp
sanchda: Yup, indeed.  Doing so will be cleaner and generate better code.  I'd like to hold it off for a later PR, like you say, though.  👍

ddtrace/internal/datadog/profiling/src/exporter.cpp
sanchda: Just realized that I inverted the logic--fixing now in case someone is looking at this and scratching their heads.

ddtrace/internal/datadog/profiling/src/exporter.cpp
```diff
@@ -0,0 +1,560 @@
...
+      address,
+      {&lines.back(), 1},
+      false,
+  });
+}
+
+void Profile::push_frame(std::string_view name, std::string_view filename,
+                         uint64_t address, int64_t line) {
+
+  push_frame_impl(name, filename, address, line);
```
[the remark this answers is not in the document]
sanchda: The python profiler has a user-configurable maximum stack depth, and when it exceeds that depth it appends a single "omitted frames" frame to the top of the stack (err... bottom of the flamegraph, err... icicle chart).  I used to implement this on the Python side, but it required putting the same logic in multiple places.  Instead, I consolidated it in the native code since your last review.

That check (at the top of flush_sample) reuses the same `push_frame_impl()` logic, but doesn't check for the `max_nframes`.

Ah; I see I elided the check in `push_frames()` in this version.  :)

This is actually enforced by the Python side, but I wanted to be more strict.  I fixed the missing check. Oops!

ddtrace/internal/datadog/profiling/src/exporter.cpp
sanchda: Good suggestion, made a note of this in that comment.  Thank you.

ddtrace/internal/datadog/profiling/src/exporter.cpp
sanchda: The immediate code should build, but libdatadog.a is a link-time requirement.

This is mostly just to spare me from having to diagnose any issues, should they arise, on build system configurations I don't care about yet (e.g., I can't control CI compiler versions at the moment).

ddtrace/profiling/collector/memalloc.py
```diff
@@ -100,48 +103,87 @@ def _get_thread_id_ignore_set(self):
...
-                    sample_size=self.heap_sample_size,
+
+        stacks = [
+            ((stack, nframes, thread_id), size)
+            for (stack, nframes, thread_id), size in _memalloc.heap()
+            if not self.ignore_profiler or thread_id not in thread_id_ignore_set
+        ]
+        if self.export_libdd_enabled:
+            for (frames, nframes, thread_id), size in stacks:
+                ddup.start_sample(nframes)
```
P403n1x87, line 114: do you reckon that we could make this look like a context manager from the Python side, so that we could write something like
~~~
with sample as ddup.sample():
    sample.push_heap(size)
    ...
~~~

ddtrace/settings/profiling.py
```diff
@@ -37,6 +39,11 @@ def _derive_default_heap_sample_size(heap_config, default_heap_sample_size=1024
     return int(max(math.ceil(total_mem / max_samples), default_heap_sample_size))
 
 
+def _is_glibc_linux_x86_64():
```
P403n1x87, line 42: it looks like we need this logic in multiple places so perhaps we can make it a util helper?

sanchda: Indeed; we do need it in two places (profiler.py settings and in setup.py) (although this will go away once the universe has submitted to the wonders of libdatadog).  Could you propose a sensible place to consolidate?

ddtrace/profiling/collector/memalloc.py
sanchda: Ah, yes--that's an interesting idea.

There are several ways of doing this and I think we can make it quite ergonomic, but

1. The native code actually hardcodes precisely two such contexts and switch between them.  This is a poor man's way of implementing async uploads.  So the context we send would be more like a view into a context, rather than the creation of a new context.

2. I'm thinking more and more of moving as much of the profiler into native code.  If that's an agreeable goal, then I would have a slight preference not to invest too much in ergonomics here at this point

With that in mind, could I offer you some à la carte options, sir?
1.  Ignore for now.  Either we fix in a different PR or we never fix it, TBD
2.  Fix for this PR (since when else are we gonna do it?)

P403n1x87: I'm fine with making this potential follow-up work and keep the current approach for now.

ddtrace/settings/profiling.py
P403n1x87: Ah if we need it in the setup.py script, we would have to duplicate the code there unfortunately. I see this helper occurring also in `ddtrace/profiling/profiler.py`. Perhaps we could make it an `En.d` (a derivation) as part of the profiler configuration
~~~ python
class ProfilerConfig(En):
    ...
    _is_glibc_linux_x86_64 = En.d(
        bool,
        lambda c: _is_glibc_linux_x86_64(),
    )
~~~
and then retrieve it from the `config` object when needed?

sanchda: Ah; apologies--the profiler.py thing is a mistake.  It indeed should all be consolidated within `config` whenever possible.  Will fix--thank you!

sanchda: Fixed the profiler.py thing in d4181b151.  In general, the config should be a single source of truth for this stuff when possible imho.

sanchda:
Closing this PR as it is reimplemented as a stack
#5940
  #5967
    #5969 
    
For folks who were watching this thread with concern about overhead, please see #5968

PR #5509, opened 2023-04-10
fix(tracing): avoid truncating b3 and w3c trace ids to 64bits

mabdinur:
Builds on: https://github.com/DataDog/dd-trace-py/pull/5434

Previously the trace id in b3 and w3c trace headers were truncated from 128 bits to 64 bits if `DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED=False`. This behavior is incorrect. 
The trace ids in b3 and w3c headers should NEVER be truncated. 

Note - This PR does not require a release note. It is a fix for an [unreleased](https://github.com/DataDog/dd-trace-py/commit/a1e27bb1dca865bfdc480639cc148f4fff580cea) change. 

Note - A regression test is not required. This issue was detected by system-tests: [ex](https://github.com/DataDog/system-tests/blob/6168fa7e6f485e2655f81801a677356624409620/parametric/test_128_bit_traceids.py#L230).

### Background

The implementation of 128 bit trace ids is compatible with all versions of the agent. Trace ids are always sent to the agent as 64 bits (ex: [v05 encoder](https://github.com/DataDog/dd-trace-py/blob/v1.11.2/ddtrace/internal/_encoding.pyx#L721)). The higher order bits are stored in a tag in the [tracetagsprocessor](https://github.com/DataDog/dd-trace-py/blob/v1.11.2/ddtrace/internal/processor/trace.py#L136). 
  - If the datadog agent only support 64bit trace ids, the agent will ignore the tag and the original 128 bit trace id will be truncated to 64bits. This preserves the current behavior.
  - If the datadog agent supports 128bit trace ids the agent will reconstruct the full 128 bit trace id using the encoded trace_id field and the _dd.p.tid tag. 

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/propagation/http.py
```diff
@@ -106,14 +106,8 @@ def _extract_header_value(possible_header_names, headers, default=None):
...
 def _hex_id_to_dd_id(hex_id):
     # type: (str) -> int
-    """Helper to convert hex ids into Datadog compatible ints
-    If the id is > 64 bit and 128bit trace_id generation is disabled then truncate the trailing 64 bit.
-    "463ac35c9f6413ad48485a3953bb6124" -> "48485a3953bb6124" -> 5208512171318403364
-    Otherwise convert the full trace id into lower case hex values.
-    """
-    if config._128_bit_trace_id_enabled:
-        return int(hex_id, 16)
-    return int(hex_id[-16:], 16)
```
Yun-Kim, line 116: So by no longer truncating it here, are we not keeping them as 128 bit trace IDs? I thought we have to ensure trace IDs sent to the agent are 64 bits.

mabdinur: The w3c and b3 trace ids will be 128 bits within a process. This 128 bit trace id will be used in sampling decisions and to parent spans. However when the trace id leaves the process (ie: encoded then sent to the agent or in a distributed trace) then it will be truncated to 64bits. I updated the PR description with more details.

This change should be backwards will all agent versions.

Yun-Kim: where is the truncating to 64 bits happening if not here? If I'm understanding this correctly, this change is taking out the responsibility of truncating to 64 bits out of the http propagation module?

mabdinur: Yup truncation should only happen during encoding. It should never happen in the http module.

Yun-Kim (approved):

PR #5536, opened 2023-04-12
feat(serverless): spawn serverless mini agent child process

thedavl:
## Description

**Goal: Traces + Trace Metrics in Google Cloud Functions** 

This PR adds a function `maybe_start_serverless_mini_agent` to be called during tracer initialization.

The Serverless Mini Agent is a small, focused version of our Agent for use in Serverless contexts written in rust.

This added function looks for the presence of one of two pairs of env vars, which indicates a Google Cloud Function environment. If one of pair of the cloud function env vars are present, the function will spawn a child process that executes the rust binary to start our mini agent.

This tracer-spawned agent approach is required because of the constraints of a Google Cloud Function environment and is the product of thorough research & POCs.

The plan is for the rust mini agent binary to be published to PyPi (and npm), rather than the file be included in the tracer itself.

For the beta, customers will have to specify this additional import in their GCP Function's requirements.txt.

Afterwards, the mini agent package will become a dependency of the tracer itself

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


releasenotes/notes/spawn-serverless-trace-mini-agent-5c91a600d9c51441.yaml
```diff
@@ -0,0 +1,5 @@
+---
+features:
+  - |
+    Adds support for traces and trace metrics in Google Cloud Function environments when used in conjunction with
```
thedavl, line 4: public facing documentation will be added once the serverless trace mini agent package is published to PyPi

tests/internal/test_serverless.py
```diff
@@ -0,0 +1,52 @@
+import os
+
+import mock
+
+from ddtrace.internal.serverless.mini_agent import maybe_start_serverless_mini_agent
+
+
+@mock.patch("ddtrace.internal.serverless.mini_agent.Popen")
+def test_dont_spawn_mini_agent_if_not_cloud_function(mock_popen):
+    os.environ["DD_MINI_AGENT_PATH"] = "fake_path"
```
thedavl, line 10: I tried to use `@pytest.mark.subprocess(env=dict())` and `@run_in_subprocess(env_overrides=dict())` instead of manually setting & unsetting environment variables. 

with `@run_in_subprocess`, the environment variables weren't accessible within the function. with `@pytest.mark.subprocess`, the mock was "undefined" whenever I tried to access it within the function.

Let me know if I'm missing something with using either of the run test in subprocess annotations!

ddtrace/internal/serverless/__init__.py
```diff
@@ -17,3 +17,14 @@ def has_aws_lambda_agent_extension():
...
+
+
+def in_gcp_function():
+    # type: () -> bool
+    """Returns whether the environment is a GCP Function.
+    This is accomplished by checking for the presence of one of two pairs of environment variables,
+    with one pair being set by deprecated GCP Function runtimes, and the other set by newer runtimes.
+    """
+    is_deprecated_gcp_function = bool(environ.get("FUNCTION_NAME", False)) and bool(environ.get("GCP_PROJECT", False))
+    is_newer_gcp_function = bool(environ.get("K_SERVICE", False)) and bool(environ.get("FUNCTION_TARGET", False))
```
mabdinur, line 29: nit: we don't need to convert to bool, we can just check if the environment variable does not contain an empty string.
```suggestion
    is_deprecated_gcp_function = environ.get("FUNCTION_NAME", "")  != "" and environ.get("GCP_PROJECT",  "")  != ""
    is_newer_gcp_function = environ.get("K_SERVICE", "")  != "" and environ.get("FUNCTION_TARGET", "")  != ""
```

ddtrace/internal/serverless/mini_agent.py
```diff
@@ -0,0 +1,40 @@
...
+
+    rust_binary_path = os.getenv("DD_MINI_AGENT_PATH")
+    if rust_binary_path is None:
+        (major, minor, _) = python_version_tuple()
+        rust_binary_path = (
+            "/layers/google.python.pip/pip/lib/python"
+            + major
+            + "."
+            + minor
+            + "/site-packages/datadog-serverless-agent-linux-amd64/datadog-serverless-trace-mini-agent"
```
mabdinur, line 26: Can we avoid hardcoding the default rust binary path? 

Is linux amd64 the only supported architecture for gcp lambas? Should we disable this feature if another platform is being used?

ddtrace/internal/serverless/mini_agent.py
```diff
@@ -0,0 +1,40 @@
...
+        )
+
+    st = os.stat(rust_binary_path)
+    os.chmod(rust_binary_path, st.st_mode | stat.S_IEXEC)
```
mabdinur, line 30: Should we exception handling here? This operation will fail if the path is not found or a lambda does not have the right permissions

ddtrace/tracer.py
```diff
@@ -211,6 +213,9 @@ def __init__(
         :param url: The Datadog agent URL.
         :param dogstatsd_url: The DogStatsD URL.
         """
+
+        maybe_start_serverless_mini_agent()
```
mabdinur, line 217: We should avoid starting the mini agent if it can't be used to submit traces. We can defer this to a future PR

ddtrace/internal/serverless/mini_agent.py
thedavl: removed os.stat + os.chmod (they were unecessary)

releasenotes/notes/spawn-serverless-trace-mini-agent-5c91a600d9c51441.yaml
thedavl: removed release note

ddtrace/propagation/http.py
```diff
@@ -710,6 +710,9 @@ def _get_sampling_priority(traceparent_sampled, tracestate_sampling_priority):
     def _extract(headers):
         # type: (Dict[str, str]) -> Optional[Context]
 
+        if "http_x_cloud_trace_context" in headers:
```
thedavl, line 713: Google sets `x-cloud-trace-context` and `traceparent` headers when our cloud function is triggered, used for Google Cloud Tracing. Because of the is, the tracer sets the `parent_id` of our cloud function trace's root span, thereby causing the span to not show up under the `Trace Root Spans` in Datadog.

This change prevents the traceparent context from being used to set the span context, so cloud function trace root spans show up properly as trace root spans.

ddtrace/propagation/http.py
```diff
@@ -750,6 +751,12 @@ def _extract(headers):
                 else:
                     log.debug("no dd list member in tracestate from incoming request: %r", ts)
 
+        if (
```
thedavl, line 754: Google sets `traceparent` and `x-cloud-trace-context` headers when our cloud function is triggered, used for Google Cloud Tracing. Because of the is, the tracer sets the `parent_id` of our cloud function trace's root span, thereby causing the span to not show up under the `Trace Root Spans` in Datadog.

I've talked with my team, and we thought it would make sense to drop the span context span_id, in the case that datadog headers aren't set and we are in a google cloud environment.

ddtrace/propagation/http.py
```diff
@@ -859,7 +866,15 @@ def my_controller(url, headers):
 
             # loop through the extract propagation styles specified in order
             for prop_style in config._propagation_style_extract:
-                propagator = _PROP_STYLES[prop_style]
+                if (
```
thedavl, line 869: google cloud function traces will have both the `x-cloud-trace-context` and `traceparent` headers set. If the cloud function was not called by an upstream datadog instrumented service (meaning no `x-datadog-trace-id` is set), we want to enter the regular traceparent extractor (where the parent_id will be intentionally left empty).

Otherwise (datadog trace context headers are set, but `traceparent is also set`), use the datadog extractor.

ddtrace/internal/constants.py
```diff
@@ -3,7 +3,7 @@
 PROPAGATION_STYLE_B3_SINGLE_HEADER = "b3 single header"
 _PROPAGATION_STYLE_W3C_TRACECONTEXT = "tracecontext"
 _PROPAGATION_STYLE_NONE = "none"
-_PROPAGATION_STYLE_DEFAULT = "tracecontext,datadog"
+_PROPAGATION_STYLE_DEFAULT = "datadog,tracecontext"
```
thedavl, line 6: with google cloud functions, requests can include both tracecontext & datadog headers. this is because google cloud functions will always inject a `tracecontext`, even if a function is directly invoked.

In the even that a function is directly invoked by another datadog instrumented service upstream, we want datadog headers to be extracted (not tracecontext)

ddtrace/propagation/http.py
```diff
@@ -750,6 +752,9 @@ def _extract(headers):
                 else:
                     log.debug("no dd list member in tracestate from incoming request: %r", ts)
 
+        if in_gcp_function():
```
thedavl, line 755: directly invoked google cloud functions (that weren't invoked by some datadog instrumented service) will have a `tracecontext` header. This causes the function to not show up as a Trace Root Span (as it's parent_id is being set to the id specified in the tracecontext header which google uses for their own tracing solution.

Because of this, we want to drop the span_id context for cloud functions

ddtrace/internal/constants.py
```diff
@@ -1,9 +1,12 @@
+from ddtrace.internal.serverless import in_gcp_function
+
+
 PROPAGATION_STYLE_DATADOG = "datadog"
 PROPAGATION_STYLE_B3 = "b3multi"
 PROPAGATION_STYLE_B3_SINGLE_HEADER = "b3 single header"
 _PROPAGATION_STYLE_W3C_TRACECONTEXT = "tracecontext"
 _PROPAGATION_STYLE_NONE = "none"
-_PROPAGATION_STYLE_DEFAULT = "tracecontext,datadog"
+_PROPAGATION_STYLE_DEFAULT = "datadog,tracecontext" if in_gcp_function() else "tracecontext,datadog"
```
thedavl, line 9: with google cloud functions, requests can include both tracecontext & datadog headers. this is because google cloud functions will always inject a `tracecontext`, even if a function is directly invoked.

In the even that a function is directly invoked by another datadog instrumented service upstream, we want datadog headers to be extracted (not tracecontext), so try to extract datadog headers first.

Is there currently any reason we prioritize tracecontext over datadog propagation?

thedavl:
There are 2 main problems (outside of spinning up the mini agent) that this PR tries to solve:

Trace propagation doesn't work when a cloud function is invoked via it's trigger URL
_Solution:_
- For GCP cloud functions only, try to extract datadog headers first before w3c headers (gcp functions called by another datadog instrumented service will always have both w3c headers and datadog headers set. The current behavior in node & py tracers is try to extract s3c first)

2. Trickier: How can we get GCP functions that are directly invoked by the user (not an upstream datadog service) to show up as the Trace Root Span in the UI? This is particularly useful when users setup a demo cloud function and trigger it manually for testing.
_Possible solutions:_
- Within the W3C extractor, don't set context's span_id if we are in a GCP function. 
   - This has the side effect of breaking W3C header support for all GCP functions
- Somehow detect that the GCP function was not invoked by another service, but by the user directly
   - I don't think this is possible, as GCP HTTP functions are invoked by making an HTTP request to their exposed endpoint. You can't really tell if the HTTP request is from a user or service reliably
- Just accept that GCP function spans will never be the trace root span

This PR (in it's current state as of this comment) changes the default header extraction order (only when in GCP functions), and deliberately doesn't set span_id in the W3C extractor if we are in a GCP function.

ddtrace/internal/writer/writer.py
```diff
@@ -510,7 +511,7 @@ def __init__(
         #      as a safety precaution.
         #      https://docs.python.org/3/library/sys.html#sys.platform
         is_windows = sys.platform.startswith("win") or sys.platform.startswith("cygwin")
-        default_api_version = "v0.4" if is_windows else "v0.5"
+        default_api_version = "v0.4" if (is_windows or in_gcp_function()) else "v0.5"
```
thedavl, line 514: the mini agent (used for gcp functions) uses v0.4

ddtrace/settings/config.py
```diff
@@ -264,7 +269,7 @@ def __init__(self):
         # Raise certain errors only if in testing raise mode to prevent crashing in production with non-critical errors
         self._raise = asbool(os.getenv("DD_TESTING_RAISE", False))
         self._trace_compute_stats = asbool(
-            os.getenv("DD_TRACE_COMPUTE_STATS", os.getenv("DD_TRACE_STATS_COMPUTATION_ENABLED", False))
+            os.getenv("DD_TRACE_COMPUTE_STATS", os.getenv("DD_TRACE_STATS_COMPUTATION_ENABLED", in_gcp_function()))
```
thedavl, line 272: the mini agent doesn't compute any trace stats itself, but exposes the trace stats endpoint to the tracers.

ddtrace/internal/serverless/mini_agent.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+log = get_logger(__name__)
+
+
+def maybe_start_serverless_mini_agent():
+    if not in_gcp_function():
+        return
+
+    (major, minor, _) = python_version_tuple()
```
mabdinur, line 17: Nit 
```suggestion
    major, minor, _ = python_version_tuple()
```

ddtrace/internal/serverless/mini_agent.py
```diff
@@ -0,0 +1,42 @@
+import logging
+import os
+from platform import python_version_tuple
```
mabdinur, line 3: Very minor nit: 

We can import the python from this internal module. There is a (small) performance benefit of accessing platform information once. 

https://github.com/DataDog/dd-trace-py/blob/1.x/ddtrace/internal/compat.py#L52

ddtrace/internal/serverless/mini_agent.py
```diff
@@ -0,0 +1,42 @@
...
+        + "."
+        + minor
+        + "/site-packages/datadog-serverless-agent-linux-amd64/datadog-serverless-trace-mini-agent"
+    )
+
+    if os.getenv("DD_MINI_AGENT_PATH") is not None:
+        rust_binary_path = os.getenv("DD_MINI_AGENT_PATH")
+
+    log.log(
+        logging.DEBUG,
```
mabdinur, line 30: nit:
```suggestion
    log.debug(
```

ddtrace/internal/serverless/mini_agent.py
```diff
@@ -0,0 +1,42 @@
...
+        "rust_binary_path: %s",
+        rust_binary_path,
+    )
+
+    try:
+        Popen(rust_binary_path)
+    except Exception as e:
+        log.log(
+            logging.ERROR,
```
mabdinur, line 39:
```suggestion
        log.error(
```

mabdinur (commented):
Left some nit but overall looks good to me.

Suggestion: can we try to consolidate some of the is_gcp checks? We're making changes to  quite a few modules. It would be nice to isolate this logic

ddtrace/internal/constants.py
thedavl: removed

ddtrace/internal/serverless/mini_agent.py
```diff
@@ -0,0 +1,38 @@
+import os
+from subprocess import Popen
+
+from . import in_gcp_function
+from ..compat import PYTHON_VERSION
```
mabdinur, line 5:
```suggestion
from ..compat import PYTHON_VERSION_INFO
```

ddtrace/internal/serverless/mini_agent.py
```diff
@@ -0,0 +1,38 @@
...
+            "/layers/google.python.pip/pip/lib/python"
+            + python_version_list[0]
+            + "."
+            + python_version_list[1]
+            + "/site-packages/datadog-serverless-agent-linux-amd64/datadog-serverless-trace-mini-agent"
+        )
+
+        if os.getenv("DD_MINI_AGENT_PATH") is not None:
+            rust_binary_path = os.getenv("DD_MINI_AGENT_PATH")
+
```
mabdinur, line 28: nit: We can simplify this code block a bit (this probably needs formatting)
```suggestion
        rust_binary_path = os.getenv("DD_MINI_AGENT_PATH", "/layers/google.python.pip/pip/lib/python%d.%d/site-packages/datadog-serverless-agent-linux-amd64/datadog-serverless-trace-mini-agent" % PYTHON_VERSION_INFO[:2])
```

ddtrace/internal/serverless/mini_agent.py
```diff
@@ -0,0 +1,38 @@
...
+        log.debug(
+            "rust_binary_path: %s",
+            rust_binary_path,
+        )
```
mabdinur, line 32: nit: this can be one line
```suggestion
        log.debug("rust_binary_path: %s", rust_binary_path)
```

ddtrace/internal/serverless/mini_agent.py
```diff
@@ -0,0 +1,38 @@
...
+        Popen(rust_binary_path)
+    except Exception as e:
+        log.error(
+            "Error spawning Serverless Mini Agent process: %s",
+            repr(e),
+        )
```
mabdinur, line 38: This can be one line too
```suggestion
        log.error("Error spawning Serverless Mini Agent process: %s", repr(e))
```

ddtrace/tracer.py
mabdinur: Thoughts? @thedavl

mabdinur (approved):
I left some non blocking suggestions. Thanks for pushing this through 😄

tests/internal/test_serverless.py
```diff
@@ -0,0 +1,40 @@
...
+    # do not set any GCP environment variables
+
+    maybe_start_serverless_mini_agent()
+
+    mock_popen.assert_not_called()
+
+
+@mock.patch("ddtrace.internal.serverless.mini_agent.Popen")
+def test_spawn_mini_agent_if_deprecated_gcp_function_runtime(mock_popen):
+    os.environ["FUNCTION_NAME"] = "test_function"
```
mabdinur, line 19: Instead of setting and removing environment variables we can use tests.utils.override_config(). This will better isolate tests. Here's an example: https://github.com/DataDog/dd-trace-py/blob/1.x/tests/appsec/test_appsec_utils.py#L23

ddtrace/internal/serverless/mini_agent.py
thedavl: fixed

tests/internal/test_serverless.py
thedavl: fixed

mabdinur (approved):

mabdinur (approved):

mabdinur (approved):

PR #5578, opened 2023-04-18
feat(pytest): upload git packfiles from civisibility service

emmettbutler:
This change implements the git packfile uploading protocol that's part of the Intelligent Test Runner product.

### Changes

* Allows data to be passed to the command's stdin in `ddtrace.ext.git._git_subprocess_cmd`.
* Adds some git utility functions used by civisibility
* Moves some utility functions from `ddtrace/internal/agent.py` and `ddtrace/internal/writer/writer.py` to more semantically meaningful shared locations
* Implements and tests `CIVisibilityGitClient`, which runs the Intelligent Test Runner's packfile protocol in a subprocess
* Moves the `DD_API_KEY` check to the `CIVisibility` class to share it between multiple modules that need it. Adds a `CIVisibilityGitClient` instance member to this class.
* Adds `ddtrace.internal.compat.TemporaryDirectory`, which implements a temporary directory context manager in a manner compatible with Py2 and Py3

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/internal/ci_visibility/git_client.py
```diff
@@ -0,0 +1,187 @@
...
+        # type: (str, list[str], CIVisibilityGitClientSerDeV1) -> list[str]
+        payload = serde.search_commits_encode(repo_url, latest_commits)
+        response = cls.retry_request()(cls._do_request, "/search_commits", payload, serde)
+        result = serde.search_commits_decode(response.body)
+        return result
+
+    @classmethod
+    def _do_request(cls, endpoint, payload, serde, headers=None):
+        # type: (str, str, CIVisibilityGitClientSerDeV1, Optional[dict]) -> Response
+        url = "https://git-api-ci-app-backend.us1.staging.dog/repository%s" % endpoint
```
juan-fernandez, line 83: is this URL correct?

emmettbutler: @juan-fernandez No, it's a staging URL. Do you happen to know the correct production URL?

juan-fernandez: the base URL is the same as the one for ITR config and skippable calls: `https://api.${DD_SITE}`

Yun-Kim (commented):
Initial round of review, just added a few comments. I'll give a more thorough review once I've read up on the packfile specification docs 👍

ddtrace/internal/ci_visibility/recorder.py
```diff
@@ -46,14 +48,26 @@ def __init__(self, tracer=None, config=None, service=None):
...
         self.config = config  # type: Optional[IntegrationConfig]
         self._tags = ci.tags(cwd=_get_git_repo())  # type: Dict[str, str]
         self._service = service
         self._codeowners = None
 
+        self._git_client = None
+        app_key = os.environ.get("DD_APPLICATION_KEY")
+        if app_key is None:
+            log.warning(
+                "Environment variable DD_APPLICATION_KEY not set, which is required for CI Visibility reporting"
```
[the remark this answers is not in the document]
juan-fernandez: it's not a requirement if neither agentless nor evp proxy is available. In that case we should stick to reporting test traces and nothing else

ddtrace/ext/git.py
```diff
@@ -87,48 +87,64 @@ def extract_user_info(cwd=None):
...
     }
 
 
+def extract_remote_url(cwd=None):
+    remote_url = git_subprocess_cmd("config --get remote.origin.url", cwd=cwd)
+    return remote_url
+
+
+def extract_latest_commits(cwd=None):
+    latest_commits = git_subprocess_cmd("log --format=%H -n 1000", cwd=cwd)
```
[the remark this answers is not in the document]
juan-fernandez: good catch. See js for reference: https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/plugins/util/git.js#L43

ddtrace/ext/git.py
```diff
@@ -87,48 +87,64 @@ def extract_user_info(cwd=None):
...
+    return latest_commits.split("\n") if latest_commits else []
+
+
+def get_rev_list_excluding_commits(commit_shas, cwd=None):
+    exclusions = " ".join(["^%s" % sha for sha in commit_shas])
+    commits = git_subprocess_cmd("rev-list --objects --filter=blob:none HEAD %s" % exclusions, cwd=cwd)
```
juan-fernandez, line 110: In addition to Yun's comment, we need to pass `--no-object-names`. JS implementation for reference: https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/plugins/util/git.js#L55 
```suggestion
    commits = git_subprocess_cmd("rev-list --objects --no-object-names --filter=blob:none --since="1 month ago" HEAD %s" % exclusions, cwd=cwd)
```

ddtrace/internal/ci_visibility/git_client.py
```diff
@@ -0,0 +1,198 @@
...
+        # type: (list[str], Optional[str]) -> list[str]
+        return get_rev_list_excluding_commits(excluded_commits, cwd=cwd)
+
+    @classmethod
+    @contextlib.contextmanager
+    def _build_packfiles(cls, revisions, cwd=None):
+        basename = str(random.randint(1, 1000000))
+        with TemporaryDirectory() as tempdir:
+            path = "{tempdir}/{basename}".format(tempdir=tempdir, basename=basename)
+            git_subprocess_cmd("pack-objects --compression=9 --max-pack-size=3m %s" % path, cwd=cwd, std_in=revisions)
```
juan-fernandez, line 119: just wondering, shouldn't this `git` command also be in `git.py`?

ddtrace/internal/ci_visibility/git_client.py
```diff
@@ -0,0 +1,198 @@
...
+            ]
+        )
+        file_name = os.path.basename(file_path)
+        f = open(file_path, "rb")
+        file_content = f.read()
+        f.close()
+        body.extend(
+            [
+                b"--" + BOUNDARY,
+                b'Content-Disposition: form-data; name="file"; filename="%s"' % file_name.encode("utf-8"),
```
juan-fernandez, line 191: not sure that it matters, but this is `packfile` for other tracers 
```suggestion
                b'Content-Disposition: form-data; name="packfile"; filename="%s"' % file_name.encode("utf-8"),
```

ddtrace/internal/ci_visibility/recorder.py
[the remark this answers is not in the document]
juan-fernandez: `DD_APPLICATION_KEY` is a requirement for git upload (and therefore ITr), but it isn't a requirement for CI Visibility, so the error message should read something like 
```
Environment variable DD_APPLICATION_KEY is not set, so no git metadata will be uploaded. 
```

juan-fernandez (commented):
Left some comments. Functionality looks good, but let's hide git upload functionality behind a feature flag for the moment. Since it's a part of the ITR, let's hide it behind `DD_CIVISIBILITY_ITR_ENABLED`

ddtrace/ext/git.py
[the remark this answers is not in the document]
emmettbutler: I tried this `git log` command on my local git and it didn't like the `--since` specifier. Are either of you able to run it locally?

juan-fernandez (commented):
LGTM

ddtrace/ext/git.py
```diff
@@ -185,3 +220,12 @@ def extract_user_git_metadata(env=None):
...
     return tags
+
+
+@contextlib.contextmanager
+def build_git_packfiles(revisions, cwd=None):
+    basename = str(random.randint(1, 1000000))
+    with TemporaryDirectory() as tempdir:
+        path = "{tempdir}/{basename}".format(tempdir=tempdir, basename=basename)
+        _git_subprocess_cmd("pack-objects --compression=9 --max-pack-size=3m %s" % path, cwd=cwd, std_in=revisions)
+        yield path
```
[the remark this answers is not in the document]
emmettbutler: No, it's more like a "prefix" than a full path.

ddtrace/internal/compat.py
```diff
@@ -365,3 +368,83 @@ def ip_is_global(ip):
         return parsed_ip.is_global
 
     return not (parsed_ip.is_loopback or parsed_ip.is_private)
+
+
+# https://stackoverflow.com/a/19299884
+class TemporaryDirectory(object):
```
Yun-Kim, line 374: If we are only enabling code coverage for Python 3, should we just do the same for git packfile uploading instead of adding this code to make it PY2 compatible? I don't see much value in adding this if ultimately ITR is dependent on both code coverage and git packfile uploading and therefore won't work in Python 2.

docs/configuration.rst
```diff
@@ -521,6 +521,15 @@ The following environment variables for the tracer are supported:
      version_added:
         v1.13.0:
 
+   DD_CIVISIBILITY_ITR_ENABLED:
```
Yun-Kim, line 524: Is this the right name for this env var? The description makes it seem more of a pre-requisite for ITR rather than a flag for enabling ITR

tests/ci_visibility/test_ci_visibility.py
```diff
@@ -6,15 +10,31 @@
...
+
+@pytest.fixture
+def git_repo_empty(tmpdir):
+    yield utils.git_repo_empty(tmpdir)
+
+
+@pytest.fixture
+def git_repo(git_repo_empty):
+    yield utils.git_repo(git_repo_empty)
+
```
Yun-Kim, line 36: Can we make the same fixtures as defined in `tests.contrib.pytest.test_pytest` import from here rather than define the same fixture twice?

tests/ci_visibility/test_ci_visibility.py
```diff
@@ -121,6 +149,85 @@ def test_repository_name_not_extracted_warning():
...
+    assert not os.path.isdir(directory)
+
+
+def test_git_client_upload_packfiles(git_repo):
+    serializer = CIVisibilityGitClientSerializerV1("foo", "bar")
+    remote_url = "git@github.com:test-repo-url.git"
+    with CIVisibilityGitClient._build_packfiles(b"%s\n" % TEST_SHA.encode("utf-8"), cwd=git_repo) as packfiles_path:
+        CIVisibilityGitClient._upload_packfiles(
+            "", remote_url, packfiles_path, serializer, DUMMY_RESPONSE, cwd=git_repo
+        )
```
Yun-Kim, line 228: Can we add some assertions here?

ddtrace/ext/git.py
```diff
@@ -95,6 +102,34 @@ def extract_user_info(cwd=None):
...
+
+
+def extract_latest_commits(cwd=None):
+    latest_commits = _git_subprocess_cmd(["log", "--format=%H", "-n", "1000" '--since="1 month ago"'], cwd=cwd)
+    return latest_commits.split("\n") if latest_commits else []
+
+
+def get_rev_list_excluding_commits(commit_shas, cwd=None):
+    command = ["rev-list", "--objects", "--filter=blob:none"]
+    if extract_git_version(cwd=cwd) >= (2, 23, 0):
```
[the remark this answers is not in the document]
emmettbutler: Yes

ddtrace/ext/git.py
```diff
@@ -95,6 +102,34 @@ def extract_user_info(cwd=None):
...
+    return version_info
+
+
+def extract_remote_url(cwd=None):
+    remote_url = _git_subprocess_cmd("config --get remote.origin.url", cwd=cwd)
+    return remote_url
+
+
+def extract_latest_commits(cwd=None):
+    latest_commits = _git_subprocess_cmd(["log", "--format=%H", "-n", "1000" '--since="1 month ago"'], cwd=cwd)
```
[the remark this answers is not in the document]
emmettbutler: It appears only to apply to `rev-list`.

ddtrace/internal/ci_visibility/git_client.py
```diff
@@ -0,0 +1,189 @@
...
+    @classmethod
+    def _get_latest_commits(cls, cwd=None):
+        # type: (Optional[str]) -> list[str]
+        return extract_latest_commits(cwd=cwd)
+
+    @classmethod
+    def _search_commits(cls, base_url, repo_url, latest_commits, serializer, _response):
+        # type: (str, str, list[str], CIVisibilityGitClientSerializerV1, Optional[Response]) -> list[str]
+        payload = serializer.search_commits_encode(repo_url, latest_commits)
+        response = _response or cls.retry_request(base_url, "/search_commits", payload, serializer)
```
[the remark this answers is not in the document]
emmettbutler: It's used by tests to generate a dummy response. It has to be done this way as opposed to mocking because some of the tests run this function in a subprocess, and Mock objects are not picklable.

ddtrace/internal/ci_visibility/git_client.py
```diff
@@ -0,0 +1,189 @@
...
+        directory = "/".join(parts[:-1])
+        rand = parts[-1]
+        for filename in os.listdir(directory):
+            if not filename.startswith(rand):
+                continue
+            file_path = os.path.join(directory, filename)
+            content_type, payload = serializer.upload_packfile_encode(repo_url, sha, file_path)
+            headers = {"Content-Type": content_type}
+            response = _response or cls.retry_request(base_url, "/packfile", payload, serializer, headers=headers)
+            return response.status == 204
```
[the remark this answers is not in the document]
emmettbutler: That's right, good catch.

docs/configuration.rst
emmettbutler: It's a name I got from @juan-fernandez. The description says what it actually does on this branch, and I expect we'll update the description later on when it does more stuff.

ddtrace/internal/compat.py
emmettbutler: Personally I don't think so, because keeping Py2 support in this change makes life easier for our future selves when we decide to figure out intelligent test running on Py2.

tests/ci_visibility/test_ci_visibility.py
emmettbutler: I tried to set that up, but the fact that they're fixtures makes that a bit tricky. Do you know how to put fixtures into shared library code?

ddtrace/internal/compat.py
Yun-Kim: Fair point.

Yun-Kim (commented):
Is the increased RSS usage due to the packfile building? I can't tell why only the span metrics scenario is reporting this performance regression.

emmettbutler:
@Yun-Kim I don't know, I haven't looked into it either. Where's the best place to start that investigation?

docs/configuration.rst
juan-fernandez: to give more context: the only thing we use packfiles for is ITR right now. While we could be more granular with feature flags, we don't want to clutter things prematurely. If at some point packfiles are used for something else and we want them regardless of ITR, we can create _another_ feature flag for customers who want to opt out.

ddtrace/ext/git.py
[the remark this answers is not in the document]
juan-fernandez: this might be an issue for large repos. If `--since` and `--no-object-names` are not supported for <2.23 we might settle with not being compatible with older versions of git. 2.23 seems to have been released more than 4 years ago (https://github.com/git/git/blob/master/Documentation/RelNotes/2.23.0.txt) so I think it's fine. 

I don't think we need to do that in this PR though

ddtrace/internal/ci_visibility/recorder.py
```diff
@@ -70,7 +71,7 @@ def __init__(self, tracer=None, config=None, service=None):
         super(CIVisibility, self).__init__()
 
         self.tracer = tracer or ddtrace.tracer
-        self._app_key = os.getenv("DD_APP_KEY")
+        self._app_key = os.getenv("DD_APPLICATION_KEY")
```
juan-fernandez, line 74: nitpick: 

If it's easy, I think we should be compatible with these three names for application keys: 
`DD_APP_KEY`, `DD_APPLICATION_KEY` and `DATADOG_APPLICATION_KEY`

juan-fernandez (approved):

Yun-Kim:
> @Yun-Kim I don't know, I haven't looked into it either. Where's the best place to start that investigation?
> 
> I guess this is the code that runs the failing benchmark? https://github.com/DataDog/dd-trace-py/blob/1.x/benchmarks/span/scenario.py On a first read through that code I don't see any way that it exercises any of the code touched in this change.

@emmettbutler I have a hunch it might be a flaky test result as I'm seeing a couple other flaky performance regression/improvement outputs from the comment (check comment edit history). Looks like it first started reporting a RSS regression on April 26, ~12PM, which means these two commits `4e059867a5986b3f784d75924413a6061072d78b` and `c66ac0d0ecc04fb9e1caa45e05306e1379f80e46` might be the source of this (i don't see how a comment change could cause it so further makes me think it's a flaky result).

Yun-Kim (approved):

PR #5589, opened 2023-04-19
feat(asm): automatic user login events for Django

juanjux:
Implement the RFC "Automated User Events Tracking" which adds some automatic traces that will be set automatically on login events if ASM is enabled.

Note: also has #6064 since the fix is needed for CI to work here.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


christophe-papazian (dismissed):

ddtrace/contrib/django/patch.py
```diff
@@ -666,14 +674,188 @@ def django_asgi_modifier(span, scope):
     return TraceMiddleware(func(*args, **kwargs), integration_config=config.django, span_modifier=django_asgi_modifier)
 
 
+_POSSIBLE_USER_ID_FIELDS = ["pk", "id", "uid", "userid", "user_id", "PK", "ID", "UID", "USERID"]
+_POSSIBLE_LOGIN_FIELDS = ["username", "user", "login", "USERNAME", "USER", "LOGIN"]
+_POSSIBLE_EMAIL_FIELDS = ["email", "mail", "address", "EMAIL", "MAIL", "ADDRESS"]
+_POSSIBLE_NAME_FIELDS = ["name", "fullname", "full_name", "NAME", "FULLNAME", "FULL_NAME"]
+
+
+def _find_in_user_model(user, possible_fields):
```
ZStriker19, line 683: I like the way this is setup, very intuitive

ddtrace/contrib/django/patch.py
```diff
@@ -666,14 +674,188 @@ def django_asgi_modifier(span, scope):
...
+
+
+def _get_user_info(user):
+    """
+    In safe mode, try to get the user id from the user object.
+    In extended mode, try to also get the username (which will be the returned user_id),
+    email and name.
+
+    Since the Django Authentication model supports pluggable users models
+    there is some heuristic involved in extracted this field, through it should
```
ZStriker19, line 744:
```suggestion
    there is some heuristic involved in extracting this field, though it should
```

ddtrace/contrib/django/patch.py
```diff
@@ -666,14 +674,188 @@ def django_asgi_modifier(span, scope):
...
+    work well for the most common case with the default User model.
+    """
+    safe_mode = config._automatic_login_events_mode == "safe"
+    user_extra_info = {}
+
+    if safe_mode:
+        user_id = _get_userid(user)
+    else:
```
ZStriker19, line 752: So anything other than `safe`, e.g. `extended` and we'll get extended behavior, correct?

ddtrace/appsec/trace_utils.py
```diff
@@ -105,6 +117,30 @@ def track_user_login_failure_event(tracer, user_id, exists, metadata=None, login
     span.set_tag_str("%s.failure.%s" % (APPSEC.USER_LOGIN_EVENT_PREFIX, user.EXISTS), exists_str)
 
 
+def track_user_signup_event(tracer, user_id, success, login_events_mode=LOGIN_EVENTS_MODE.SDK):
```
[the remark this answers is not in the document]
juanjux: Right now is for the users, we don't automatically call it (but probably will in the future).

ddtrace/contrib/trace_utils.py
```diff
@@ -18,6 +18,7 @@
 from typing import cast
 
 from ddtrace import Pin
+from ddtrace import Span
```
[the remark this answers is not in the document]
juanjux: No idea. I think PyCharm changed the import and I missed this on my self-review, good catch!

docs/configuration.rst
```diff
@@ -478,6 +478,52 @@ The following environment variables for the tracer are supported:
...
 
+   DD_APPSEC_AUTOMATED_USER_EVENTS_TRACKING:
+      type: String
+      default: "safe"
+      description: |
+         Sets the mode for the automated user login events tracking feature which sets some traces on each user login event. The
+         supported modes are ``safe`` which will only stores the user id or primary key, ``extended`` which will also store
+         the username, email and full name and ``disabled``. Note that this feature requires ``DD_APPSEC_ENABLED`` to be 
+         set to ``true`` to work. For apps using the normal framework authentication models you shouldn't need to configure
+         anything else but in some cases with custom user models you could need to use once of the ``DD_USERMODEL_SOMETHING_FIELD``
```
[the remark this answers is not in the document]
juanjux: The docs were outdated (the RFC changed to only-automatic inference of these fields), removed this, thanks!

docs/configuration.rst
```diff
@@ -478,6 +478,52 @@ The following environment variables for the tracer are supported:
...
+         v1.15.0:
+
+   DD_USERMODEL_LOGIN_FIELD:
+      type: String
+      default: ""
+      description: |
+         Sets the field of the authentication user model to use for retrieving the user id or username. This will be used in the
+         automatic user login events feature (see ``DD_APPSEC_AUTOMATED_USER_EVENTS_TRACKING`` above). Use it only if the automatic
+         inference doesn't work. Please note that if you set this variable, the tracer will try to use whatever field you set
+         even if ``safe`` mode is enabled!
```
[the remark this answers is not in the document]
juanjux: Ditto. Outdated.

ddtrace/contrib/django/patch.py
juanjux: That is actually a nasty bug you got. I will fix this and also add a test checking that safe is always the default.

ddtrace/appsec/trace_utils.py
```diff
@@ -23,8 +23,10 @@
 log = get_logger(__name__)
 
 
-def _track_user_login_common(tracer, user_id, success, metadata=None, login_events_mode=LOGIN_EVENTS_MODE.SDK):
-    # type: (Tracer, str, bool, Optional[dict], str) -> Optional[Span]
+def _track_user_login_common(
+    tracer, user_id, success, metadata=None, login_events_mode=LOGIN_EVENTS_MODE.SDK, login=None, name=None, email=None
```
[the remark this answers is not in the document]
juanjux: Utility lost in a previous refactor it seems (also unused on 1.x), I removed it, thanks!

ddtrace/contrib/django/patch.py
```diff
@@ -666,14 +674,187 @@ def django_asgi_modifier(span, scope):
...
+    return _find_in_user_model(user, _POSSIBLE_USER_ID_FIELDS)
+
+
+def _get_username(user):
+    username = getattr(user, config._user_model_name_field, None)
+    if username:
+        return username
+
+    if hasattr(user, "get_username"):
+        return user.get_username()
```
ZStriker19, line 706: Since we're running an unknown function, should we wrap this in a try except? Not sure if this is being overly cautious. It looks like we do this in some other places too e.g. `_get_name()`

ZStriker19: Technically I guess we don't even know if it's a function right? Or that it returns a string?

ddtrace/contrib/django/patch.py
```diff
@@ -666,14 +674,187 @@ def django_asgi_modifier(span, scope):
...
+                        user_id=user_id,
+                        session_id=session_key,
+                        propagate=True,
+                        login_events_mode=mode,
+                        **user_extra
+                    )
+                    return
+
+                # Login failed but the user exists
+                track_user_login_failure_event(pin.tracer, user_id=user_id, exists=True, login_events_mode=mode)
```
ZStriker19, line 805: Would moving the above return to below `track_user_login_failure_event()`  be helpful here? I just worry about modifying this function in the future and getting confusing behavior if you added code outside of the else statement since it would run, when we would have wanted to return already.
```suggestion
                else:
                        track_user_login_failure_event(pin.tracer, user_id=user_id, exists=True, login_events_mode=mode)
                return
```

ddtrace/settings/config.py
```diff
@@ -273,6 +274,10 @@ def __init__(self):
         )
         self._data_streams_enabled = asbool(os.getenv("DD_DATA_STREAMS_ENABLED", False))
         self._appsec_enabled = asbool(os.getenv(APPSEC_ENV, False))
+        self._automatic_login_events_mode = os.getenv(APPSEC.AUTOMATIC_USER_EVENTS_TRACKING, "extended")
```
ZStriker19, line 277: why is this defaulting to extended when we say the default is safe? Am I getting something mixed up? I couldn't find a test demonstrating what we wanted.

ddtrace/contrib/django/patch.py
juanjux: It's part of the base User model of django (as a method) but you're right that it could be anything else on user code. All the login tracing stuff is wrapped in a try... except (with logging) in `traced_login` so I think that could cover these corner cases pretty well however in this case I think it merits an additional check to continue trying to extract the username (same for `get_full_name`). Fixed on next commit.

ddtrace/settings/config.py
juanjux: Bug. I must have messed on some rebase :( There are some tests, but they test that safe it's the default when there is not value (which won't happen here) or has any other value, so this one escaped. Fixed on next commit.

ddtrace/contrib/django/patch.py
juanjux: Yes, being explicit is probably better in this case, commited plus another explicit return on the other failure track call below that one.

releasenotes/notes/automatic-user-login-events-ded104398ff85a1d.yaml
```diff
@@ -0,0 +1,13 @@
+---
+prelude: >
+  - Datadog has added support for automatically creating login success or failure events when a configured Django authentication backend is used. 
+    configured Django authentication backend are used. This will automatically fill the following tags in these cases:
```
ZStriker19, line 4: I'm not sure this makes sense to me.

juanjux: Just noticed the repeated sentence, fixed.

ddtrace/settings/config.py
```diff
@@ -273,6 +274,10 @@ def __init__(self):
         )
         self._data_streams_enabled = asbool(os.getenv("DD_DATA_STREAMS_ENABLED", False))
         self._appsec_enabled = asbool(os.getenv(APPSEC_ENV, False))
+        self._automatic_login_events_mode = os.getenv(APPSEC.AUTOMATIC_USER_EVENTS_TRACKING, "safe")
+        self._user_model_login_field = os.getenv(APPSEC.USER_MODEL_LOGIN_FIELD, default="")
```
ZStriker19, line 278: Do we want to add these envars to our configuration.rst?

tests/contrib/django/test_django_appsec.py
```diff
@@ -729,3 +730,146 @@ def test_request_suspicious_request_block_match_response_headers(client, test_sp
...
+
+    with override_global_config(dict(_appsec_enabled=True, _automatic_login_events_mode="safe")):
+        test_user = User.objects.create(username="fred2")
+        test_user.set_password("secret")
+        test_user.save()
+        assert not get_user(client).is_authenticated
+        client.login(username="fred2", password="secret")
+        assert get_user(client).is_authenticated
+        login_span = test_spans.find_span(name="django.contrib.auth.login")
+        assert login_span
```
ZStriker19, line 806: should we assert that the "extended" fields aren't present?

tests/contrib/django/test_django_appsec.py
```diff
@@ -729,3 +730,146 @@ def test_request_suspicious_request_block_match_response_headers(client, test_sp
...
+        assert login_span.get_tag(APPSEC.USER_LOGIN_EVENT_PREFIX + ".success.track") == "true"
+        assert login_span.get_tag(APPSEC.USER_LOGIN_EVENT_PREFIX + ".success.auto.mode") == "safe"
+
+
+@pytest.mark.django_db
+def test_django_login_sucess_safe_is_default_if_wrong(client, test_spans, tracer):
+    from django.contrib.auth import get_user
+    from django.contrib.auth.models import User
+
+    with override_global_config(dict(_appsec_enabled=True, _automatic_login_events_mode="foobar")):
```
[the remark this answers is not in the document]
juanjux: Yes, we are testing that any value for the mode should be treated as `safe`, thus the `foobar` value.

tests/contrib/django/test_django_appsec.py
```diff
@@ -729,3 +730,146 @@ def test_request_suspicious_request_block_match_response_headers(client, test_sp
...
+        assert not get_user(client).is_authenticated
+        login_span = test_spans.find_span(name="django.contrib.auth.login")
+        assert login_span.get_tag(APPSEC.USER_LOGIN_EVENT_PREFIX + ".failure.track") == "true"
+        assert login_span.get_tag(APPSEC.USER_LOGIN_EVENT_PREFIX + ".failure." + user.ID) == "missing"
+        assert login_span.get_tag(APPSEC.USER_LOGIN_EVENT_PREFIX + ".failure." + user.EXISTS) == "false"
+        assert login_span.get_tag(APPSEC.USER_LOGIN_EVENT_PREFIX + ".failure.auto.mode") == "extended"
+
+
+@pytest.mark.django_db
+def test_django_login_sucess_safe_but_user_set_login(client, test_spans, tracer):
```
ZStriker19, line 858: Ah this demonstrates some interesting behavior, which I think I noted in the code earlier. Maybe we should document this somewhere for users?

juanjux: Yup. Will note this also on the new change on the `configuration.rst`.

ddtrace/settings/config.py
juanjux: Done!

tests/contrib/django/test_django_appsec.py
juanjux: Done

ddtrace/contrib/django/patch.py
```diff
@@ -667,14 +675,195 @@ def django_asgi_modifier(span, scope):
...
+        user_id = _get_username(user)
+        if not user_id:
+            user_id = _find_in_user_model(user, _POSSIBLE_USER_ID_FIELDS)
+
+        user_extra_info = {
+            "login": user_id,
+            "email": _get_user_email(user),
+            "name": _get_name(user),
+        }
+    else:  # safe mode, default
```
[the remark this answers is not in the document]
ZStriker19: What if someone set the envar to `false` or something like that? I'm fine with this though.

docs/configuration.rst
```diff
@@ -478,6 +478,45 @@ The following environment variables for the tracer are supported:
...
+         Please note that, if set, this field will be user to retrieve the user login even if ``DD_APPSEC_AUTOMATED_USER_EVENTS_TRACKING`` is set to ``safe`` and, 
+         in some cases, the selected field could hold potentially private information.
+      version_added:
+         v1.15.0
+
+   DD_USER_MODEL_EMAIL_FIELD:
+      type: String
+      default: ""
+      description: |
+         Field to be used to read the user email when using a custom ``User`` model for the automatic login events. Use only if the automatic inference doesn't work.
```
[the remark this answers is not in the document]
juanjux: If the env var is set it will be used before automatic inference.

ddtrace/contrib/django/patch.py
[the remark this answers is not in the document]
juanjux: Then it will be "safe" and won't evaluate to a boolean since we're not doing the `asbool`.

docs/configuration.rst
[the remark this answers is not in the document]
ZStriker19:
```suggestion
         Field to be used to read the user email when using a custom ``User`` model for the automatic login events. This field will take precedence over automatic inference.
```

Oh ok got it, what do you think about this phrasing then for both of the envars?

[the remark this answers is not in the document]
juanjux: Better, thanks!

juanjux:
@ZStriker19 Please take a look when you can, all required CI is green.

avara1986 (dismissed):

docs/configuration.rst
```diff
@@ -478,6 +478,45 @@ The following environment variables for the tracer are supported:
...
+      description: |
+         Field to be used to read the user email when using a custom ``User`` model for the automatic login events. This field will take precedence over automatic inference.
+      version_added:
+         v1.15.0:
+
+   DD_USER_MODEL_NAME_FIELD:
+      type: String
+      default: ""
+      description: |
+         Field to be used to read the user name when using a custom ``User`` model for the automatic login events. Used only if the automatic inference doesn't work.
```
ZStriker19, line 514:
```suggestion
         Field to be used to read the user name when using a custom ``User`` model for the automatic login events. This field will take precedence over automatic inference.
```

ZStriker19 (approved):

avara1986 (approved):

PR #5613, opened 2023-04-20
feat(pytest): fetch and set CIVIsibility feature enablement flags from DD api

emmettbutler:
This change adds and tests logic to `CIVisibility.__init__` issuing a Datadog API request that returns information about enabled CI Visibility features for the current repo/sha/branch combination. In a future change, this information can be used to decide what functionality to enable in the `CIVisibility` service.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/internal/ci_visibility/recorder.py
```diff
@@ -72,6 +93,36 @@ def __init__(self, tracer=None, config=None, service=None):
         except Exception:
             log.warning("Failed to load CODEOWNERS", exc_info=True)
 
+    def _check_enabled_features(self):
+        # type: () -> Tuple[bool, bool]
+        if not self._app_key:
+            return False, False
+        url = "https://api.datadoghq.com/api/v2/libraries/tests/services/setting"
```
juan-fernandez, line 100: this should depend on the configured site `DD_SITE`.

Also: we should be able to do this request via EVP proxy, but this can be done at a later PR

ddtrace/internal/ci_visibility/recorder.py
```diff
@@ -72,6 +93,36 @@ def __init__(self, tracer=None, config=None, service=None):
...
+        _headers = {"dd-api-key": os.getenv("DD_API_KEY"), "dd-application-key": self._app_key}
```
juan-fernandez, line 101: curious: why is one key read from the env and one from `self._app_key` ?

ddtrace/internal/ci_visibility/recorder.py
```diff
@@ -72,6 +93,36 @@ def __init__(self, tracer=None, config=None, service=None):
...
+            "data": {
+                "id": str(uuid4()),
+                "type": "ci_app_test_service_libraries_settings",
+                "attributes": {
+                    "service": self._service,
+                    "env": ddconfig.env,
+                    "repository_url": self._tags.get(ci.git.REPOSITORY_URL),
+                    "sha": self._tags.get(ci.git.COMMIT_SHA),
+                    "branch": self._tags.get(ci.git.BRANCH),
+                },
```
juan-fernandez, line 112: there are some extra configuration parameters here that we might want to add (it's ok to add at a later PR though):
```python
                    "configurations": {
                         "runtime.name": "CPython",
                         "runtime.version": "<version>",
                         # ... some others
                    }
                },
```

juan-fernandez (commented):
looks good! 

Will this request happen before _any test is run_? We need to know this configuration before any test is run, in case we might need to skip any.

ddtrace/internal/ci_visibility/recorder.py
Yun-Kim: Agreed, not sure there's a point to store `self._app_key` when we can just get it directly from `os.getenv(...)` if we're just going to read it once/twice at `CIVisibility.__init__()`.

ddtrace/internal/ci_visibility/recorder.py
Yun-Kim: @juan-fernandez What is the purpose of attaching these extra configuration parameters into the payload? I thought this is just to check the repo/sha/branch combination?

ddtrace/internal/ci_visibility/recorder.py
```diff
@@ -72,6 +93,36 @@ def __init__(self, tracer=None, config=None, service=None):
...
+            }
+        }
+        response = _do_request("POST", url, json.dumps(payload), _headers)
+        try:
+            parsed = json.loads(response.body)
+        except json.JSONDecodeError:
+            return False, False
+        if response.status >= 400 or ("errors" in parsed and parsed["errors"][0] == "Not found"):
+            return False, False
```
Yun-Kim, line 121: Should we add a debug log here to warn users about the status of this request?

ddtrace/internal/ci_visibility/recorder.py
```diff
@@ -53,6 +72,8 @@ def __init__(self, tracer=None, config=None, service=None):
         self._tags = ci.tags(cwd=_get_git_repo())  # type: Dict[str, str]
         self._service = service
         self._codeowners = None
+        self._app_key = os.getenv("DD_APP_KEY")
+        self._code_coverage_enabled_by_api, self._test_skipping_enabled_by_api = self._check_enabled_features()
```
Yun-Kim, line 76: Does this variable naming make more sense?
```suggestion
        self._code_coverage_enabled, self._test_skipping_enabled = self._check_enabled_features()
```

ddtrace/internal/ci_visibility/recorder.py
juan-fernandez: that's a fair question. I know other tracers include it but I forwarded the question to our backend team, because I don't know the reason

juan-fernandez (commented):
other than the hardcoded URL (it should depend on `DD_SITE`), LGTM functionality wise

ddtrace/internal/ci_visibility/recorder.py
emmettbutler: I think the explicitness about the source of the enablement is important in this case, because there will also be environment variables that enable these features.

ddtrace/internal/ci_visibility/recorder.py
emmettbutler: Will the host be `api.<DD_SITE>`?

juan-fernandez: yeah

juan-fernandez (approved):

Yun-Kim (approved):

PR #5632, opened 2023-04-21
chore(service_name): enable service naming schema for multiple contribs

tabgok:
(chore) enable service naming schema for multiple contribs

### Service Naming Summary

The changes introduced in [PR5427](https://github.com/DataDog/dd-trace-py/pull/5427)
introduced a few concepts:
  * `v0`: The existing mechanism for determining service names
  * `v1`: Replacing the 'default' contrib-specific service name with the env var `$DD_SERVICE`

These changes are being in conjunction with changes to `peer.service` updates to bring a
better service-naming experience to customers.  In particular, these changes will help:
* Remove the creation of fake services, such as `grpc.client`
* Standardize service names across and within tracers
* Remove the need to create *as many* fake services to take advantage of Datadog features, such as metrics
* Make more sense to humans

At the moment, these changes are opt-in via feature flag. Users
who do not specify the change will see no changes.

### Changes in this PR
This PR implements the `v1` naming schema for the following contribs:
* aiomysql
* aiopg
* algoliasearch
* asyncpg
* cassandra
* elasticsearch
* mongoengine
* mysql
* mysqldb
* psycopg(2)
* pymemcache
* pymongo
* pymysql
* sqlalchemy
* sqlite3
       

### Performance Implications
The performance impact is minimal both in scope and value.

### Risks

Please see the risks in the [original service naming PR](https://github.com/DataDog/dd-trace-py/pull/5427)

## Checklist

## Reviewer Checklist

- [x] Title is accurate.

### DIFF
--- a/ddtrace/contrib/pymysql/patch.py
+++ b/ddtrace/contrib/pymysql/patch.py
@@ -9,14 +9,14 @@ from ddtrace.vendor import wrapt
 
 from ...ext import db
 from ...ext import net
+from ...internal.schema import schematize_service_name
 from ...internal.utils.formats import asbool
 
 
 config._add(
     "pymysql",
     dict(
-        # TODO[v1.0] this should be "mysql"
-        _default_service="pymysql",
+        _default_service=schematize_service_name("pymysql"),
         _dbapi_span_name_prefix="pymysql",
         trace_fetch_methods=asbool(os.getenv("DD_PYMYSQL_TRACE_FETCH_METHODS", default=False)),
     ),
--- a/tests/contrib/aiomysql/test_aiomysql.py
+++ b/tests/contrib/aiomysql/test_aiomysql.py
@@ -7,7 +7,6 @@ from ddtrace import Pin
 from ddtrace import Tracer
 from ddtrace.contrib.aiomysql import patch
 from ddtrace.contrib.aiomysql import unpatch
-from ddtrace.internal.schema import DEFAULT_SPAN_SERVICE_NAME
 from tests.contrib.config import MYSQL_CONFIG
 
 
@@ -171,7 +170,7 @@ asyncio.run(test())""",
 async def test_unspecified_service_v1(ddtrace_run_python_code_in_subprocess):
     """
     v1: When a user specifies nothing for a service,
-        it should default to DEFAULT_SPAN_SERVICE_NAME
+        it should default to internal.schema.DEFAULT_SPAN_SERVICE_NAME
     """
     env = os.environ.copy()
     env["DD_TRACE_SPAN_ATTRIBUTE_SCHEMA"] = "v1"
--- a/tests/contrib/aiomysql/test_aiomysql.py
+++ b/tests/contrib/aiomysql/test_aiomysql.py
@@ -7,6 +7,7 @@ from ddtrace import Pin
 from ddtrace import Tracer
 from ddtrace.contrib.aiomysql import patch
 from ddtrace.contrib.aiomysql import unpatch
+from ddtrace.internal.schema import DEFAULT_SPAN_SERVICE_NAME
 from tests.contrib.config import MYSQL_CONFIG
 
 
@@ -109,13 +110,14 @@ async def test_patch_unpatch(tracer, test_spans):
 
 @pytest.mark.asyncio
 @pytest.mark.snapshot
-async def test_user_specified_service(ddtrace_run_python_code_in_subprocess):
+async def test_user_specified_service_v0(ddtrace_run_python_code_in_subprocess):
     """
-    When a user specifies a service for the app
+    v0: When a user specifies a service for the app
         The aiomysql integration should not use it.
     """
     env = os.environ.copy()
     env["DD_SERVICE"] = "my-service-name"
+    env["DD_TRACE_SPAN_ATTRIBUTE_SCHEMA"] = "v0"
     out, err, status, pid = ddtrace_run_python_code_in_subprocess(
         """
 import asyncio
@@ -124,6 +126,62 @@ from ddtrace import config
 from tests.contrib.aiomysql.test_aiomysql import AIOMYSQL_CONFIG
 
 assert config.service == "my-service-name"
+async def test():
+    conn = await aiomysql.connect(**AIOMYSQL_CONFIG)
+    await (await conn.cursor()).execute("select 'dba4x4'")
+    conn.close()
+asyncio.run(test())""",
+        env=env,
+    )
+    assert status == 0, err
+    assert out == err == b""
+
+
+@pytest.mark.asyncio
+@pytest.mark.snapshot
+async def test_user_specified_service_v1(ddtrace_run_python_code_in_subprocess):
+    """
+    v1: When a user specifies a service for the app
+        The aiomysql integration should use it.
+    """
+    env = os.environ.copy()
+    env["DD_SERVICE"] = "my-service-name"
+    env["DD_TRACE_SPAN_ATTRIBUTE_SCHEMA"] = "v1"
+    out, err, status, pid = ddtrace_run_python_code_in_subprocess(
+        """
+import asyncio
+import aiomysql
+from ddtrace import config
+from tests.contrib.aiomysql.test_aiomysql import AIOMYSQL_CONFIG
+
+assert config.service == "my-service-name"
+async def test():
+    conn = await aiomysql.connect(**AIOMYSQL_CONFIG)
+    await (await conn.cursor()).execute("select 'dba4x4'")
+    conn.close()
+asyncio.run(test())""",
+        env=env,
+    )
+    assert status == 0, err
+    assert out == err == b""
+
+
+@pytest.mark.asyncio
+@pytest.mark.snapshot
+async def test_unspecified_service_v1(ddtrace_run_python_code_in_subprocess):
+    """
+    v1: When a user specifies nothing for a service,
+        it should default to DEFAULT_SPAN_SERVICE_NAME
+    """
+    env = os.environ.copy()
+    env["DD_TRACE_SPAN_ATTRIBUTE_SCHEMA"] = "v1"
+    out, err, status, pid = ddtrace_run_python_code_in_subprocess(
+        """
+import asyncio
+import aiomysql
+from ddtrace import config
+from tests.contrib.aiomysql.test_aiomysql import AIOMYSQL_CONFIG
+
 async def test():
     conn = await aiomysql.connect(**AIOMYSQL_CONFIG)
     await (await conn.cursor()).execute("select 'dba4x4'")
--- a/tests/contrib/aiopg/test.py
+++ b/tests/contrib/aiopg/test.py
@@ -9,6 +9,7 @@ from ddtrace import Pin
 from ddtrace.constants import ANALYTICS_SAMPLE_RATE_KEY
 from ddtrace.contrib.aiopg.patch import patch
 from ddtrace.contrib.aiopg.patch import unpatch
+from ddtrace.internal.schema import DEFAULT_SPAN_SERVICE_NAME
 from tests.contrib.asyncio.utils import AsyncioTestCase
 from tests.contrib.asyncio.utils import mark_asyncio
 from tests.contrib.config import POSTGRES_CONFIG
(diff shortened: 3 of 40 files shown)


tests/contrib/aiomysql/test_aiomysql.py
```diff
@@ -133,3 +134,59 @@ async def test():
     )
     assert status == 0, err
     assert out == err == b""
+
+
+@pytest.mark.asyncio
+@pytest.mark.snapshot
+async def test_user_specified_service_v1(ddtrace_run_python_code_in_subprocess):
```
tabgok, line 141: I've taken a note to expanding the usage of this pattern, which can be parametrized, in a testing improvement follow-up action item.

tests/contrib/aiopg/test.py
```diff
@@ -211,6 +212,48 @@ def test_user_specified_service(self):
...
+
+        conn = yield from aiopg.connect(**POSTGRES_CONFIG)
+        Pin.get_from(conn).clone(tracer=self.tracer).onto(conn)
+        yield from (yield from conn.cursor()).execute("select 'blah'")
+        conn.close()
+
+        spans = self.get_spans()
+        assert spans, spans
+        assert len(spans) == 1
+        assert spans[0].service == DEFAULT_SPAN_SERVICE_NAME
```
tabgok, line 255: I've gone back and forth on whether to use `DEFAULT_SPAN_SERVICE_NAME` or to use the string it resolves to in v1 to `unnamed-python-service`

tests/contrib/mysql/test_mysql.py
```diff
@@ -44,7 +45,7 @@ def test_simple_query(self):
 
         span = spans[0]
         assert_is_measured(span)
-        assert span.service == "mysql"
+        assert span.service == schematize_service_name("mysql")
```
tabgok, line 48: This is done to handle the case where the user runs the test via riot with `--pass-env` and `DD_TRACE_SPAN_ATTRIBUTE_SCHEMA = v1`.

I am open to removing or expanding this pattern.

tests/contrib/asyncpg/test_asyncpg.py
```diff
@@ -169,7 +169,7 @@ async def test_parenting(patched_conn):
 
 
 @pytest.mark.snapshot(async_mode=False)
-def test_configure_service_name_env(ddtrace_run_python_code_in_subprocess):
+def test_configure_service_name_env_v0(ddtrace_run_python_code_in_subprocess):
```
ZStriker19, line 172: do we need an assert here?

tests/contrib/asyncpg/test_asyncpg.py
```diff
@@ -194,6 +194,101 @@ async def test():
...
     env = os.environ.copy()
     env["DD_ASYNCPG_SERVICE"] = "global-service-name"
+    env["DD_TRACE_SPAN_ATTRIBUTE_SCHEMA"] = "v0"
+    out, err, status, pid = ddtrace_run_python_code_in_subprocess(code, env=env)
+    assert status == 0, err
+    assert err == b""
+
+
+@pytest.mark.snapshot(async_mode=False)
+def test_configure_service_name_env_v1(ddtrace_run_python_code_in_subprocess):
```
ZStriker19, line 204: do we need an assert here?

ddtrace/contrib/pymysql/patch.py
```diff
@@ -9,14 +9,14 @@
...
 from ...ext import db
 from ...ext import net
+from ...internal.schema import schematize_service_name
 from ...internal.utils.formats import asbool
 
 
 config._add(
     "pymysql",
     dict(
-        # TODO[v1.0] this should be "mysql"
```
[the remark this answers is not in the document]
tabgok: Ha!

I dropped this line because:
1. `v1` actually changes this to 'mysql' (via the schematize_service_name call)
2. Service names will be managed via the schema from now on

tests/contrib/aiomysql/test_aiomysql.py
tabgok: The service name is actually being checked via the `@pytest.mark.snapshot` decorator.  

When decorated, the function will gather a span and either compare it to an existing snapshot for the test (if the snapshot exists) or create a new snapshot file.  

In the case of this PR, you can see the snapshots added at the very end.

tests/contrib/aiomysql/test_aiomysql.py
```diff
@@ -133,3 +134,59 @@ async def test():
...
+    """
+    env = os.environ.copy()
+    env["DD_TRACE_SPAN_ATTRIBUTE_SCHEMA"] = "v1"
+    out, err, status, pid = ddtrace_run_python_code_in_subprocess(
+        """
+import asyncio
+import aiomysql
+from ddtrace import config
+from tests.contrib.aiomysql.test_aiomysql import AIOMYSQL_CONFIG
+
```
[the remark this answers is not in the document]
tabgok: See my earlier comment about snapshots.  (also, please correct me if I'm wrong about snapshots.)

tests/contrib/aiopg/test.py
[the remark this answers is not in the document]
tabgok: The reason no changes were required in `aiopg` is because `aiopg` leverages `psycopg` to make changes ([aiopg page](https://aiopg.readthedocs.io/en/stable/) so we kind of get the updates for free.

`_get_vendor()` is a good catch - I'll update that! I'm not sure in what scenario we'd end up with a connection that doesn't have a pin already set, but in case the mechanism ever changes its a good thing to update.  You can see how we patch all connections by looking [here](https://github.com/DataDog/dd-trace-py/blob/1.x/ddtrace/contrib/aiopg/patch.py#L35-L37)

tests/contrib/aiopg/test.py
```diff
@@ -211,6 +212,48 @@ def test_user_specified_service(self):
         assert len(spans) == 1
         assert spans[0].service != "mysvc"
```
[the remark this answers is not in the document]
tabgok: This seems fine to me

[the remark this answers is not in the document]
tabgok: Had to update tests in response to some issues discovered

tests/contrib/aiopg/test.py
tabgok: thanks! updated

tests/contrib/asyncpg/test_asyncpg.py
tabgok: see snapshot comment above

tests/contrib/asyncpg/test_asyncpg.py
tabgok: see comment about snapshots above

tests/contrib/mysql/test_mysql.py
ZStriker19: I think we shouldn't worry about users passing in envars to configure tracer behavior besides obviously needing the configuration to send to the testagent correctly.

tests/contrib/pymemcache/test_client.py
```diff
@@ -376,3 +377,45 @@ def test_user_specified_service(self):
...
+        assert config.service == "mysvc"
+
+        client = self.make_client([b"STORED\r\n", b"VALUE key 0 5\r\nvalue\r\nEND\r\n"])
+        client.set(b"key", b"value", noreply=False)
+
+        pin = Pin.get_from(pymemcache)
+        tracer = pin.tracer
+        spans = tracer.pop()
+
+        assert spans[0].service == "mysvc"
```
ZStriker19, line 421: Should we add a `test_unspecified_service_v1` ?

tests/contrib/pymongo/test.py
```diff
@@ -520,9 +520,84 @@ def test_user_specified_service(self):
...
+        Pin(service="mypymongo", tracer=self.tracer).onto(client)
+        Pin.get_from(client).clone(tracer=tracer).onto(client)
+        # We do not wish to trace tcp spans here
+        Pin.get_from(pymongo.server.Server).remove_from(pymongo.server.Server)
+        client["testdb"].drop_collection("whatever")
+        spans = tracer.pop()
+        assert len(spans) == 1
+        assert spans[0].service == "mypymongo"
 
+    @TracerTestCase.run_in_subprocess(
```
ZStriker19, line 597: General comment: There's some diversity as to which tests you choose to add for each integration. I think we have enough coverage with the mixing, but just wanted to point it out.

tests/snapshots/tests.contrib.aiomysql.test_aiomysql.test_user_specified_service_v0.json
ZStriker19, line 1: I don't think we need to generate a new snapshot file for this one, since all of the attributes that the snapshot agent would check stay the same.

ZStriker19: Oh never mind, it's a file naming change. Got it.

tests/contrib/mysql/test_mysql.py
ZStriker19: Actually, maybe we should remove it for now and then we'll add it everywhere when it becomes default. Kind of confusing to have it be spotty.

tests/contrib/aiopg/test.py
```diff
@@ -190,11 +191,37 @@ def test_patch_unpatch(self):
         assert spans, spans
         assert len(spans) == 1
 
-    @run_in_subprocess(env_overrides=dict(DD_SERVICE="mysvc"))
-    def test_user_specified_service(self):
+    @mark_asyncio
+    @run_in_subprocess(env_overrides=dict(DD_SERVICE="mysvc", DD_TRACE_SPAN_ATTRIBUTE_SCHEMA="v0"))
+    def test_user_specified_service_v0(self):
```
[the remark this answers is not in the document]
tabgok: There are two bugs here:

1. The existing behavior of the code base for `aiopg` is to set service name to `$DD_SERVICE` in the absence of a user-specified name.  This is a bug, because the expected behavior is to use the module name `aiopg` as the default.
1. The original unchanged test `test_user_specified_service` auto-succeeded no matter what asserts were in the function, as long as the asyncio yield code is included.  For example, in the unchanged code, your first line in the test could be `assert False` and it would still pass.

What I've done here is fix the testing bug (2) to document the current behavior and worked with you to open an issue for (1).  If/when (1) is addressed, the v0 test should be updated to once again be `!= "mysvc"`

tests/contrib/pymemcache/test_client.py
tabgok: Yes, thank you!

tests/contrib/mysql/test_mysql.py
tabgok: Ack, I'll remove it for now.

tests/contrib/elasticsearch/test_elasticsearch.py
```diff
@@ -240,6 +241,21 @@ def test_user_specified_service(self):
         assert len(spans) == 1
         assert spans[0].service != "es"
```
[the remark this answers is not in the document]
tabgok: I agree - I am also removing the Pin `Pin(service="es", tracer=self.tracer).onto(self.es.transport)` as it doesn't do anything (in the context of this test).

tests/contrib/aiopg/test.py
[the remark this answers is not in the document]
ZStriker19: Ah, thanks for the re-run on it! Seems good!

tests/contrib/elasticsearch/test_elasticsearch.py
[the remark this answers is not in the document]
ZStriker19: Thanks yeah, that part muddied the waters too.

tests/contrib/pymongo/test.py
ZStriker19: Yeah I think that's enough IMO. We obviously didn't test all of this stuff beforehand, and even with your changes adding behavior, I think you're adding more test coverage than what we originally had.

tests/contrib/elasticsearch/test_elasticsearch.py
ZStriker19, line 1: Should we at least add unspecified v0 and v1 here?

tests/contrib/elasticsearch/test_elasticsearch.py
```diff
@@ -225,20 +225,51 @@ def test_patch_unpatch(self):
...
+
+    @TracerTestCase.run_in_subprocess(env_overrides=dict(DD_TRACE_SPAN_ATTRIBUTE_SCHEMA="v1"))
+    def test_user_specified_service_v1(self):
         self.es.indices.create(index=self.ES_INDEX, ignore=400)
         Pin(service="es", tracer=self.tracer).onto(self.es.transport)
         spans = self.get_spans()
         self.reset()
         assert len(spans) == 1
-        assert spans[0].service != "es"
+        assert spans[0].service == "unnamed-python-service"
```
ZStriker19, line 272: nit: Probably stick to importing `DEFAULT_SPAN_SERVICE_NAME` and using that?

tabgok: will do!

ZStriker19 (dismissed):

ZStriker19 (approved):

ZStriker19 (approved):

PR #5655, opened 2023-04-25
fix(telemetry): collect metrics in gunicorn apps

avara1986:
## Description
Telemetry metrics worker didn't work correctly in Gunicorn applications:

When Gunicorn spawns a child process, the TelemetryLogsMetricsWriter is not running in the child process. As a result, the metrics are collected, as a result, we have a memory leak because any worker sends and flushes this collected data on Gunicorn child processes.

This PR introduce a refactor of `TelemetryLogsMetricsWriter` enabled method:
- Check if the PeriodicService is running before create/append metrics
- Check if `telemetry_metrics_enabled` env var is True before create/append metrics
- Restart worker when Gunicorn spawn a new process
- Refactor of `enable()` method to use common features for TelemetryWriter and TelemetryMetricsLogsWriter
- Renamed attribute `enabled` to `_is_running` to avoid confusions with the method `enable()`

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/internal/telemetry/writer.py
```diff
@@ -220,13 +221,35 @@ def __init__(self):
...
+        return False
+
     def add_log(self, level, message, stack_trace="", tags={}):
         # type: (str, str, str, MetricTagType) -> None
         """
         Queues log. This event is meant to send library logs to Datadog’s backend through the Telemetry intake.
         This will make support cycles easier and ensure we know about potentially silent issues in libraries.
         """
-        if self._enabled:
+        if self.enable():
```
Yun-Kim, line 252: Why are we using `self.enable()` rather than checking the flag `self._enabled` here? I'm not sure I agree with using this logic to both enable the writer and check if it's enabled, which I think `self._enabled` should be used for the latter.

Yun-Kim (commented):
Can we make the PR description more descriptive about what the problem was and how this PR is fixing it? Also cc-ing @mabdinur to review this as well.

ddtrace/internal/telemetry/writer.py
```diff
@@ -152,18 +152,19 @@ def add_event(self, payload, payload_type):
             self._events_queue.append(event)
 
     def enable(self):
-        # type: () -> None
+        # type: () -> bool
```
mabdinur, line 155: The TelemetryBase._enabled should have 3 states (not two). 
 1) TelemetryBase._enabled is None 
      - This occurs during application startup. TelemetryWriter is not yet started but we would like to queue events/logs/metrics. These events will sit in memory until the telemetry writer is started. However there is a risk for memory leaks that needs to be addressed. In a follow up PR we should limit the total size of the telemetry queue.

 2) TelemetryBase._enabled is True
       - This set after the first trace is sent to the agent. Here the Telemetry writer is running and submitting events to the agent.
 3) TelemetryBase._enabled  is False
      - `DD_INSTRUMENTATION_TELEMETRY_ENABLED=False` or `_DD_TELEMETRY_METRICS_ENABLED=False`, the telemetry writer is stopped (or was never started) and no events should be queued or sent. Again there is a risk of memory leaks if we do not handle this case correctly.

There is also an edge case `TelemetryBase._enabled = True` but `TelemetryBase.status is Stopped`. This can occur if the telemetry worker thread is killed. This could occur in gevent and this needs to be handled.

ddtrace/internal/telemetry/writer.py
avara1986: "The `self._enabled` flag works fine with `app-heartbeat` and `app-started` events because the `TelemetryWriter` is running on the main process. However, I found a problem with `self._enabled` in Gunicorn applications. When Gunicorn spawns a child process, the flag `self._enabled` is still true, but the `TelemetryLogsMetricsWriter` is not running in the child process. As a result, the metrics are collected, and we have a memory leak because any worker sends and flushes this collected data on Gunicorn child processes.

In this PR, `TelemetryLogsMetricsWriter` starts correctly when applications spawn a new child process, and we validate `config._telemetry_metrics_enabled` each time we add a metric to avoid leaks. 

I'm thinking the problem is in the naming because I override the `enabled()` method for `TelemetryLogsMetricsWriter`?

ddtrace/internal/telemetry/writer.py
avara1986: >    In a follow up PR we should limit the total size of the telemetry queue.

There is no problem right now. Count, Gauge and Rate metrics increment a value, If we add a Count metric for each request and hit the application 1.000.000 times, the metric will be:
```
{
                "common": True,
                "metric": "test-metric",
                "points": [[1642544540, 1.000.000]],
                "type": "count",
}
```
But it's true that we could have this queue size problem with distribution metrics. For that reason, those types of metrics are disabled in ASM while we discuss the limits.

> There is also an edge case `TelemetryBase._enabled = True` but `TelemetryBase.status is Stopped`. This can occur if the telemetry worker thread is killed. This could occur in gevent and this needs to be handled.

That is the problem we have just now and this PR try to fix :smiley:  (see this comment where I explain it: https://github.com/DataDog/dd-trace-py/pull/5655#discussion_r1177457265)

emmettbutler (commented):
This change makes sense to me. Adding a regression test and a test for the behavior of the environment variables could be a way to resolve the issues from the review discussion.

avara1986:
> This change makes sense to me. Adding a regression test and a test for the behavior of the environment variables could be a way to resolve the issues from the review discussion.

I added regression test @emmettbutler https://github.com/DataDog/dd-trace-py/pull/5655/files#diff-089e1931ac9d84e1975f32926e71d153a49e57e4aa084649ba952c914214e6bcR1 :+1:

ddtrace/internal/telemetry/writer.py
```diff
@@ -152,18 +152,19 @@ def add_event(self, payload, payload_type):
...
         """
         if self.status == ServiceStatus.RUNNING:
-            return
+            return True
 
         self._enabled = True
         self.start()
 
         atexit.register(self.stop)
+        return True
```
mabdinur, line 167:
```suggestion
       if config._telemetry_enabled:  # we need to defined this configuration
            self.start()

            atexit.register(self.stop)
            return True
        return False
```

ddtrace/internal/telemetry/writer.py
```diff
@@ -220,13 +221,35 @@ def __init__(self):
...
+        if config._telemetry_metrics_enabled:
+            if self.status == ServiceStatus.RUNNING:
+                return True
+
+            self._enabled = True
+            self.start()
+
+            atexit.register(self.stop)
+            return True
+        return False
```
mabdinur, line 244:
```suggestion
            return super(TelemetryLogsMetricsWriter, self).enable()
```

mabdinur (commented):
Maybe something like this?

ddtrace/internal/telemetry/writer.py
avara1986: what is `_telemetry_enabled`? the `DD_INSTRUMENTATION_TELEMETRY_ENABLED` env var?
```
_telemetry_enabled = asbool(os.getenv("DD_INSTRUMENTATION_TELEMETRY_ENABLED", True))
```

ddtrace/internal/telemetry/writer.py
avara1986: Yeah, I like it! I think we could remove first `telemetry_metrics_writer._enabled` and then, in a new PR, remove `telemetry_writer._enabled`. If we change both, I think we are mixing features and this PR could be quite huge, what do you think @mabdinur ?

ddtrace/internal/telemetry/writer.py
mabdinur: Yup sorry that was unclear. We can do this in a follow up or make the change here

ddtrace/internal/telemetry/writer.py
mabdinur: Yes this is an enhancement/fix that I tried to sneak into your PR 😅. If DD_INSTRUMENTATION_TELEMETRY_ENABLED is False we should never submit telemetry.

avara1986:
@mabdinur I tried to apply your suggestions but I noticed If we replace `self._enabled` for `self.status` many tests start to fail due to a race condition in the meantime the worker is starting. I'm not sure if the problem is in our telemetry test or it could be a problem in a real scenario. 

BTW, I updated the code in the below commit with:
- Refactor of `enable()` method to use common features for TelemetryWriter and TelemetryMetricsLogsWriter
- Renamed attribute `enabled` to `_is_running` to avoid confusions with the method `enable()`

Take a look of this commit and if it don't work for you I can rollback it :smile: 

https://github.com/DataDog/dd-trace-py/pull/5655/commits/f8e8d7a80461c5a2d43aef2c98b0fa7e05361744

mabdinur (approved):
I will remove `TelemetryBase._is_running` in a follow up PR.

mabdinur (approved):

mabdinur (approved):

PR #5659, opened 2023-04-25
chore(service_name): schemas for operation names

tabgok:
**Service Naming Summary**

The changes introduced in PR5427
introduced a few concepts:

  * v0: The existing mechanism for determining service names
  * v1: Replacing the 'default' contrib-specific service name with the env var $DD_SERVICE

These changes are being in conjunction with changes to peer.service updates to bring a better service-naming experience to customers. In particular, these changes will help:

* Remove the creation of fake services, such as grpc.client
* Standardize service names across and within tracers
* Remove the need to create as many fake services to take advantage of Datadog features, such as metrics
* Make more sense to humans

**Operation Naming Summary**
The service name changes mean that generated spans can no longer rely on `span.service` to differentiate between processes.  A subclass of spans now have a new problem - differentiating between inbound and outbound requests (for the purposes of metrics).  For instance, spans with names `http.request` used to be differentiated by the service they belonged to (incoming requests under `span.service=$DD_SERVICE` and outgoing under `span.service=$DD_CONTRIB_SERVICE_NAME`).  These are now no longer differentiated.

To remedy these issues, operation names are also being changed as part of the service naming effort.  In the above example, `http.request` turns into `http.server.request` (incoming) and `http.client.request` (outgoing).

**This Change**
This change introduces two of the operation name classes, storage and caching, into the service naming schema.

In addition, this change:
* Updates pymongo with operation name changes (database example)
* Updates pylibmc with service and operation name changes (cache example)
* Performs some light refactoring in the `ddtrace.internal.schema` library

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


amarziali (approved):
Naming looks good 👍

ddtrace/internal/schema/span_attribute_schema.py
```diff
@@ -1,23 +1,45 @@
-SPAN_ATTRIBUTE_SCHEMA_VERSIONS = dict()
```
tabgok, line 1: replaced with _SPAN_ATTRIBUTE_TO_FUNCTION.keys() (see bottom of file)

emmettbutler (approved):

tabgok:
> This looks great. I can see how it maps onto the approach we talked about a few weeks ago.
> 
> I think this change probably warrants a release note because it changes some user-visible span attributes.

The current plan is to do release notes once all the changes for service name and peer.service are done.

With the schematization, there should be no changes to existing behavior (behavior without setting the feature flag) for existing users.

tests/contrib/pylibmc/test.py
```diff
@@ -386,3 +391,39 @@ def test_patch_unpatch(self):
         spans = self.pop_spans()
         assert spans, spans
         assert len(spans) == 1
+
+
+class TestPylibmcPatchServiceNames(TestPylibmcPatchDefault):
+    @TracerTestCase.run_in_subprocess(env_overrides=dict(DD_SERVICE="mysvc", DD_TRACE_SPAN_ATTRIBUTE_SCHEMA="v0"))
```
ZStriker19, line 397: Kind of unrelated? But I guess that might just be because I view these as separate to some degree. It all has to get in, so should be fine I think.

ZStriker19 (approved):

ZStriker19:
> > This looks great. I can see how it maps onto the approach we talked about a few weeks ago.
> > I think this change probably warrants a release note because it changes some user-visible span attributes.
> 
> The current plan is to do release notes once all the changes for service name and peer.service are done.
> 
> With the schematization, there should be no changes to existing behavior (behavior without setting the feature flag) for existing users.

I agree, since the operation names won't change by default, and since we have work left to do, holding off on the release note makes sense to me.

tests/contrib/pylibmc/test.py
tabgok: I split up the operation and service name changes originally, but they are all under the same umbrella of "service name" changes.  I'm going to leave this in for now

emmettbutler (approved):

PR #5684, opened 2023-04-27
fix(version_install): do not install cattrs 1.1.0 or above when using python 3.6

ZStriker19:
This fixes a known issue. Previously we'd install the latest `cattrs` version regardless of Python version. cattrs 1.1.0 and above does not support Python 3.6 or below, causing our tracer to crash a customer on 3.6, if they didn't pin the cattrs version themselves.
## Checklist

## Reviewer Checklist

- [x] Title is accurate.


mabdinur:
Addresses: [ink to issue](https://github.com/python-attrs/cattrs/issues/109)

setup.py
```diff
@@ -345,7 +345,8 @@ def get_exts_for(name):
         "attrs>=20; python_version>'2.7'",
         "attrs>=20,<22; python_version=='2.7'",
         "contextlib2<1.0; python_version=='2.7'",
-        "cattrs",
+        "cattrs<=1.0.0; python_version<='3.6'",
```
mabdinur, line 348: very minor nit:
```suggestion
        "cattrs<1.1; python_version<='3.6'",
```

releasenotes/notes/change_cattrs_install_requires-36be09a2fda5e639.yaml
```diff
@@ -0,0 +1,6 @@
+---
+fixes:
+  - |
+    dependencies: This fix resolves an issue where ddtrace would install the latest version of cattrs by default.
+    However, since cattrs ``>=1.1.0`` dropped support for Python 3.6 and below, this would cause applications to crash. 
+    By pinning the cattrs version to ``<1.1.0``, if Python 3.6 and below are detected, we avoid this issue.
```
mabdinur, line 6: We can simplify this a bit
```suggestion
    dependencies: Resolves an issue where ddtrace installs an incompatible version of cattrs when Python 3.6 is used.
```

brettlangdon:
> ddtrace supports pip>=18 ([link](https://ddtrace.readthedocs.io/en/stable/troubleshooting.html#installation-failure)). Instead of making this fix should we just bump the minimum supported version for pip?
@brettlangdon thoughts?

@mabdinur I am confused, where does `pip` come into this problem? `pip` itself doesn't rely on `cattrs`?

mabdinur:
> > ddtrace supports pip>=18 ([link](https://ddtrace.readthedocs.io/en/stable/troubleshooting.html#installation-failure)). Instead of making this fix should we just bump the minimum supported version for pip?
> 
> @brettlangdon thoughts?
> 
> 
> 
> @mabdinur I am confused, where does `pip` come into this problem? `pip` itself doesn't rely on `cattrs`?

Pips dependency resolution appears to be the problem. It appears to ignore the yanked version. This only occurs in older versions of pip. This issue appears to be resolved.

brettlangdon:
Oh, I see what you mean. What is the minimum pip version that supports yanked versions?

I still think this PR is probably the best solution, since if we can allow a large range of `pip` versions would be good.

mabdinur (approved):

P403n1x87 (approved):

brettlangdon:
> > What is the minimum pip version that supports yanked versions
> 
> @mabdinur @brettlangdon It looks version 22? https://pip.pypa.io/en/stable/news/#v22-0 Should I create a backlog card for upgrading to pip 22 so we don't have these issues in the future?

personal opinion, I don't think we need to, since we just resolved an issue for customers using `pip==18`.

we only asked that people use `pip==18` so they can use the `build-requires`/`pyproject.toml` stuff, otherwise installs will just fail. they are free to upgrade to newer versions of pip if they want or require.

PR #5688, opened 2023-04-27
feat(pytest): put real suite and session ID onto encoded coverage spans

emmettbutler:
This change replaces the dummy values for suite ID and session ID in code coverage payloads with real values derived from the pytest integration.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.

### DIFF
--- a/ddtrace/internal/ci_visibility/encoder.py
+++ b/ddtrace/internal/ci_visibility/encoder.py
@@ -155,7 +155,7 @@ class CIVisibilityCoverageEncoderV02(CIVisibilityEncoderV01):
         # type: (Span, str) -> Dict[str, Any]
         return {
             "span_id": span.span_id,
-            "test_session_id": int(span.get_tag(SESSION_ID)),
-            "test_suite_id": int(span.get_tag(SUITE_ID)),
+            "test_session_id": int(span.get_tag(SESSION_ID) or "1"),
+            "test_suite_id": int(span.get_tag(SUITE_ID) or "1"),
             "files": json.loads(span.get_tag(COVERAGE_TAG_NAME))["files"],
         }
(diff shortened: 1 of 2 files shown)


Yun-Kim (approved):

Yun-Kim (approved):

ddtrace/internal/ci_visibility/encoder.py
```diff
@@ -155,7 +155,7 @@ def _convert_span(span, dd_origin):
         # type: (Span, str) -> Dict[str, Any]
         return {
             "span_id": span.span_id,
-            "test_session_id": 1,  # TODO: populate with real IDs
-            "test_suite_id": 1,
+            "test_session_id": int(span.get_tag(SESSION_ID) or "1"),
+            "test_suite_id": int(span.get_tag(SUITE_ID) or "1"),
```
mabdinur, line 159: `span.get_tag(ID) or "1"` will overwrite tags that are set to `0` or `""`. Are these values supported? (maybe not zero since tags are always strings). 

Also if the ids contains a letter int(...) will raise an exception. Should we handle this?

emmettbutler: Those values aren't supported.

I don't think the pytest integration will generate IDs with non-digit characters in them. @Yun-Kim correct me if I'm wrong.

Yun-Kim: The session/suite IDs are copied from the span ID of a given span, so as long as span IDs are integers (I believe they are random 64-bit/128-bit integers?) we should be fine. https://github.com/DataDog/dd-trace-py/blob/1.x/ddtrace/span.py#L154-L160

Yun-Kim: > Those values aren't supported.

👍

PR #5698, opened 2023-04-28
chore(profiling): add native ID to user threads

P403n1x87:
Patch Thread._bootstrap to retrieve and store the native thread ID (the TID) on Linux. This might come handy for determining whether a call stack is on CPU or not.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.

### DIFF
--- a/tests/profiling/collector/test_threading.py
+++ b/tests/profiling/collector/test_threading.py
@@ -1,4 +1,5 @@
 import os
+import sys
 import threading
 import uuid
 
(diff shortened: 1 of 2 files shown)


tests/profiling/collector/test_threading.py
```diff
@@ -312,3 +313,40 @@ def test_lock_acquire_release_speed_patched(benchmark, pct):
...
+    from os import getpid
+    from threading import Thread
+    from threading import _MainThread
+    from threading import current_thread
+    from time import sleep
+
+    main = current_thread()
+    assert isinstance(main, _MainThread)
+    # We expect the main thread to have the same ID as the PID
+    assert main.native_id == getpid(), (main.native_id, getpid())
```
sanchda, line 334: +1 to marking this test Linux-only, as this assumption in particular obviously breaks on other platforms (even macos).

tests/profiling/collector/test_threading.py
```diff
@@ -312,3 +313,40 @@ def test_lock_acquire_release_speed_patched(benchmark, pct):
...
+
+    t = Thread(target=lambda: None)
+    t.start()
+
+    for _ in range(10):
+        try:
+            # The TID should be higher than the PID, but not too high
```
sanchda, line 341: I think it's a good idea to sanity-test pid/tid integrity, but this assumption will break when
* PIDs are just about to wraparound
* The underlying PID namespace has a lot of activity, especially if the current task gets de-scheduled

Neither of these considerations is relevant for CI, so I think it's perfectly fine to take this section--just wanted to callout potential sources of flakiness.

sanchda (approved):
Made some purely editorial comments.  This looks great!  It really makes me want to move more of the Profiler implementation into native code so we can rationalize the behavior through the native profiler.  ❤️

tests/profiling/collector/test_threading.py
P403n1x87: Indeed. I've thought about the first point you made and assumed it should be extremely rare to hit the wraparound boundary in CI. So hopefully we're good 🤞

PR #5723, opened 2023-05-02
fix: guard against django cache result errors when setting row_count

wconti27:
Patch for issue: https://github.com/DataDog/dd-trace-py/issues/5460. Added try except block to catch potential errors that may occur. Added tests to verify patch works as intended to catch errors.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ZStriker19:
Could you run the compile-and-prune script to generate new requirements.txt files since you made a change to riotfile.py? It should be as simple as running the code block there. https://ddtrace.readthedocs.io/en/latest/contributing.html#changing-test-requirements

ddtrace/contrib/django/patch.py
```diff
@@ -189,13 +189,20 @@ def traced_cache(django, pin, func, instance, args, kwargs):
...
-            # if valid result and check for special case for Django~3.0 that returns an empty Sentinel object as
-            # missing key
-            if result is not None and result != getattr(instance, "_missing_key", None):
-                span.set_metric(db.ROWCOUNT, 1)
-            # else result is invalid or None, set row count to 0
-            else:
+            try:
+                # if valid result and check for special case for Django~3.0 that returns an empty Sentinel object as
+                # missing key
+                if hasattr(result, "empty"):
```
P403n1x87, line 195: This I believe is internal knowledge of `pandas`. I would be inclined to use this approach as a last resort. I think it might be more beneficial if we could understand exactly why we still have an issue here and try to address that in its most general form. Otherwise we would end up having to do these sorts of lookups for every framework that exposes objects with a similar behaviour as `DataFrame`. WDYT?

ZStriker19: I've been trying to think about this but can't come up with what the more "general form" approach would be here, although I agree that doing this for every object that has the same behavior as `DataFrame` doesn't seem ideal.

wconti27: Good point. I changed this approach to use the id() function since None is a Python built-in singleton and any value set to None will also return the same id, so that should be safe. For the dataframe bug, checking if the result is an iterable first before using `==` fixed the ambiguity problem.

ddtrace/contrib/django/patch.py
```diff
@@ -189,12 +189,16 @@ def traced_cache(django, pin, func, instance, args, kwargs):
...
-            # missing key
-            if result is not None and result != getattr(instance, "_missing_key", None):
-                span.set_metric(db.ROWCOUNT, 1)
-            # else result is invalid or None, set row count to 0
-            else:
+            try:
+                if result is None:
+                    span.set_metric(db.ROWCOUNT, 0)
+                # check also for special case for Django~3.0 that returns an empty Sentinel object for empty results
+                # also check if result is Iterable first since some iterables return ambiguous truth results with ``==``
```
P403n1x87, line 196: Ah so with `==` we get the same behaviour as `bool`? Do we know if `getattr(instance, "_missing_key")` returns a constant sentinel object, much like `None`? If so then we can just use `is` like we do with `None`.

wconti27: Yes, the `==` was resulting in the same error as `bool`, resulting in an ambiguity error when being compared to some Iterable objects such as a Pandas.Series or Pandas.DataFrame. And I don't think the `getattr(instance, "_missing_key")` returns a sentinel object, [here is where it is instantiated](https://github.com/django/django/blob/fc42edd2e63e89a80e7ca81486291f74359ef8be/django/core/cache/backends/base.py#L55). Instead, I just changed this line to check if the result is an Iterable, and if it is NOT, compare it against the `_missing_key` attribute if found.

ddtrace/contrib/django/patch.py
```diff
@@ -189,12 +189,16 @@ def traced_cache(django, pin, func, instance, args, kwargs):
                 db.ROWCOUNT, sum(1 for doc in result if doc) if result and isinstance(result, Iterable) else 0
             )
         elif command_name == "get":
-            # if valid result and check for special case for Django~3.0 that returns an empty Sentinel object as
-            # missing key
-            if result is not None and result != getattr(instance, "_missing_key", None):
-                span.set_metric(db.ROWCOUNT, 1)
-            # else result is invalid or None, set row count to 0
-            else:
+            try:
```
Yun-Kim, line 192: Just to clarify, the reason we need to make this a `try/except` block is that certain cases of Django results are empty sentinel objects that raise errors when applied to `==` conditionals. Am I understanding correctly?

ddtrace/contrib/django/patch.py
```diff
@@ -189,12 +189,16 @@ def traced_cache(django, pin, func, instance, args, kwargs):
...
+                if result is None:
+                    span.set_metric(db.ROWCOUNT, 0)
+                # check also for special case for Django~3.0 that returns an empty Sentinel object for empty results
+                # also check if result is Iterable first since some iterables return ambiguous truth results with ``==``
+                elif not isinstance(result, Iterable) and result == getattr(instance, "_missing_key", None):
```
Yun-Kim, line 197: can we merge this with the previous if case since we expect to do the same thing in both cases? i.e.
```python
if result is None or (not isinstance(result. Iterable) and result == getattr(instance, "_missing_key", None):
```

wconti27: Yup, good catch

ddtrace/contrib/django/patch.py
wconti27: Yup

Yun-Kim (commented):
LGTM, just need to update the lockfiles via the steps as mentioned in https://github.com/DataDog/dd-trace-py/blob/1.x/docs/contributing.rst#changing-test-requirements.

ZStriker19 (approved):

ZStriker19 (approved):

PR #5724, opened 2023-05-02
chore(ci): integrate test agent to handle all integration test traces

wconti27:
This PR integrates the DD Test Agent to handle all integration test traces. The PR changes the dummy tracer to emit traces to the Test Agent container during CI testing (and local testing if the Test Agent is running). Other required changes were adding the Test Agent to all contrib jobs for CircleCI and adding configuration to the DummyTracer to disable sending traces to the Test Agent when the test agent isn't running, such as tracer and other internal tests. 

## Motivation
We're currently working on integrating the test agent into each tracer's test suites, and for the test agent to handle all traces produced by integrations during testing. Within the Test-Agent, we are then creating checks for Service Naming, Unified Naming Conventions tagging, etc in order to be a centralized location for asserting on integration traces. Allows us to leverage traces produced from integrations during testing.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


.circleci/config.yml
```diff
@@ -160,6 +161,13 @@ commands:
                   - run:
                       name: "Waiting for << parameters.wait >>"
                       command: riot -v run 'wait' << parameters.wait >>
+            # - unless:
```
tabgok, line 164: Should this be removed?  If not, we should have a comment on why its commented out

docker-compose.yml
```diff
@@ -143,7 +143,7 @@ services:
         volumes:
             - ./tests/snapshots:/snapshots
         environment:
-            - LOG_LEVEL=WARNING
+            - LOG_LEVEL=DEBUG
```
tabgok, line 146: was this change done for testing? Should it be moved back to WARNING?

tests/utils.py
```diff
@@ -475,7 +476,17 @@ class DummyWriter(DummyWriterMixin, AgentWriter):
     def __init__(self, *args, **kwargs):
         # original call
         if len(args) == 0 and "agent_url" not in kwargs:
-            kwargs["agent_url"] = "http://localhost:8126"
+            kwargs["agent_url"] = agent.get_trace_url()
+        if "api_version" not in kwargs:
```
tabgok, line 480: nit: `kwargs.get("api_version", "v0.5")` does the same thing but slightly prettier

tests/utils.py
```diff
@@ -475,7 +476,17 @@ class DummyWriter(DummyWriterMixin, AgentWriter):
...
+            kwargs["api_version"] = "v0.5"
+
+        self._trace_flush_enabled = False
+
+        # this is expected to be enabled as long as an agent is running and
+        # this instance was created by a DummyTracer
+        if kwargs.get("trace_flush_enabled", False) is True:
```
tabgok, line 487: This writing is a bit confusing, are we trying to verify that the parameter is supplied and is a bool?  If not, this should be `kwargs.get...` by itself.  If so, then you can use isinstance(...) to check.

tests/utils.py
```diff
@@ -523,20 +543,39 @@ def encoder(self):
...
 
     def configure(self, *args, **kwargs):
         assert "writer" not in kwargs or isinstance(
             kwargs["writer"], DummyWriterMixin
         ), "cannot configure writer of DummyTracer"
+
+        # check test agent status to see if traces should be emitted for additional testing
         if not kwargs.get("writer"):
-            kwargs["writer"] = DummyWriter()
+            trace_flush_enabled = check_test_agent_status()
```
tabgok, line 574: nit: These five lines can be one (or two):

```kwargs[writer] = DummWriter(trace_flush_enabled = check_test_agent_stats()```

File: docker-compose.yml (hunk 2 above)
Comment (wconti27) [reply docker-compose.yml:146]: Yeah, was just changed for testing. [Here](https://app.circleci.com/pipelines/github/DataDog/dd-trace-py/34647/workflows/874f1af1-737c-4f57-bdd2-6e59dcf538b1/jobs/2342911?invite=true#step-107-1306) is an example of the logs when set to debug, they are now available for all integration tests on circleci.

Comment (wconti27) [reply docker-compose.yml:146]: Will switch this back before the final commit

File: tests/utils.py [hunk 6]
```diff
@@ -6,7 +6,7 @@
 import subprocess
 import sys
 import time
-from typing import List
+from typing import List  # noqa
```
ZStriker19, line 9: I don't think we need the # noqa here? flake8 passes without it for me.
```suggestion
from typing import List
```

tests/utils.py
```diff
@@ -487,8 +492,17 @@ def write(self, spans=None):
         if spans:
             traces = [spans]
             self.json_encoder.encode_traces(traces)
-            self.msgpack_encoder.put(spans)
-            self.msgpack_encoder.encode()
+            if self._trace_flush_enabled:
+                AgentWriter.write(self, spans=spans)
+            else:
+                self.msgpack_encoder.put(spans)
```
ZStriker19, line 498: Why did we encode them originally? To make sure performance was as expected, or do we sometimes look at them later, once encoded?

I'm only worried about this because I don't want to accidentally change a performance metric just because we're now writing rather than encoding sometimes. It's fine that we change, it'd just be good to know if we expect to see any performance change.

ZStriker19 (commented):
Looks good! Do we want to add a test that confirms the testagent works with at least one of these integrations, or do other tests still functioning already prove this?

tests/utils.py
wconti27: Those encoded traces are occasionally looked at by some tests like tracer / writer tests. I noticed messing with that encoder caused failures from other tests using it, that's why I went with just writing using an AgentWriter instead.

wconti27:
> Looks good! Do we want to add a test that confirms the testagent works with at least one of these integrations, or do other tests still functioning already prove this?

Yeah, would like to add some small tests today (though the mere passing of integration tests and a quick log check [here](https://app.circleci.com/pipelines/github/DataDog/dd-trace-py/35256/workflows/920ee43a-d23f-4e6a-a87b-72101cb5d32e/jobs/2383191?invite=true#step-107-1359) for an integration confirms that it's running as intended) just to make sure all functionality works

ZStriker19 (approved):

PR #5731, opened 2023-05-02
chore(tests): add testing to ensure patching does not import module

ZStriker19:
Following up on [5455 ](https://github.com/DataDog/dd-trace-py/pull/5455/files#diff-8f30c25b0c036c175363c0913e45dfa2a17cc2fd41f65669202423501b1e60fb). 

We should not be importing the module when we patch. This PR adds an assert to the patch testing to make sure that we do not. It should fail on Django, and likely other integrations currently. From there we'll need to implement fixes.

## Checklist

- [ ] PR description includes explicit acknowledgement/acceptance of the performance implications of this PR as reported in the benchmarks PR comment.

## Reviewer Checklist

- [ ] Title is accurate.


Yun-Kim (approved):
This will probably be blocked until we do the same patch on import change for the other integrations as #5455 did for Django, but love that we have coverage of this now!

tests/contrib/patch.py
```diff
@@ -321,16 +321,20 @@ def test_patch_import(self):
...
 
             an appropriate ``test_patch_import`` would be::
 
                 import redis
                 ddtrace.patch(redis=True)
                 self.assert_module_patched(redis)
             """
-            module = importlib.import_module(self.__module_name__)
             self.__patch_func__()
+            self.assert_not_module_imported(self.__module_name__)
```
P403n1x87, line 336: One important lesson from the Django integration is that whilst we were triggering the patching when `django` was imported, that patching force-imported Django submodules. So checking for just the required module might not be enough. Something better is to check for any modules prefixed with `{self.__module_name__.}`, but even this might not be enough (Django imported the `channels` module if available, for instance)

ZStriker19: Hmmm I can't think of a clever way to do that then, since each integration is loading stuff besides the module it's patching anyways. Is there any way to track what a module was loaded by? I looked around and couldn't find anything. Maybe we go with a `startswith()` to begin?

PR #5743, opened 2023-05-03
refactor(internal): use get_connection in connector

P403n1x87:
We use the get_connection helper in the connector context manager to ensure that we use the same connection construction logic across all internal products.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.

### DIFF
--- a/tests/internal/test_utils_http.py
+++ b/tests/internal/test_utils_http.py
@@ -4,7 +4,7 @@ import pytest
 from ddtrace.internal.utils.http import connector
 
 
-@pytest.mark.parametrize("scheme", ["http", "https", "unix"])
+@pytest.mark.parametrize("scheme", ["http", "https"])
 def test_connector(scheme):
     with httpretty.enabled():
         httpretty.register_uri(httpretty.GET, "%s://localhost:8181/api/test" % scheme, body='{"hello": "world"}')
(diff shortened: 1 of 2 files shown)


brettlangdon:
Failing test: https://app.circleci.com/pipelines/github/DataDog/dd-trace-py/35114/workflows/dd07c0b7-a421-46d1-9f06-46ecc67b7d19/jobs/2374198?invite=true#step-105-2992

tests/internal/test_utils_http.py
```diff
@@ -4,7 +4,7 @@
 from ddtrace.internal.utils.http import connector
 
 
-@pytest.mark.parametrize("scheme", ["http", "https", "unix"])
+@pytest.mark.parametrize("scheme", ["http", "https"])
```
brettlangdon, line 7: wait, are we removing support for unix domain sockets?

can/should we do that?

P403n1x87: We're simply removing a test case. Since we use the common `get_connection`, which is already well-covered in tests, we're now "over-testing" here in fact.

brettlangdon: well, I would have expected it to still work with `unix://` because it was working before your change.

But I think that is because we do extra checks on `unix://` urls in `get_connection()` which the previous code/this test didn't validate (the port in the url?)

So it is less about over testing, and previously we didn't perform similar validations.

P403n1x87: > But I think that is because we do extra checks on `unix://` urls in `get_connection()` which the previous code/this test didn't validate (the port in the url?)

That's right, and probably we don't need that kind of support. That's also why I'm suggesting merging behaviour, instead of having something behaving differently.

PR #5753, opened 2023-05-03
fix(elasticsearch): elasticsearch8 circular import error

ZStriker19:
Currently if using >= elasticsearch 8 the tracer will throw an error as it tries to patch, since we haven't added support for this version. 

This fix version guards patching, so users don't have to disable the integration manually.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


riotfile.py
```diff
@@ -947,6 +947,18 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
...
             ],
         ),
+        Venv(
+            name="elasticsearch8-patch",
+            command="pytest {cmdargs} tests/contrib/elasticsearch/test_es8_patch.py",
+            venvs=[
+                Venv(
+                    pys=select_pys(),
+                    pkgs={
+                        "elasticsearch8": [latest],
```
ZStriker19, line 957: This is wrong, for some reason riot isn't able to find a version greater than 7.x.

emmettbutler: This is happening because `elasticsearch8` [only works](https://pypi.org/project/elasticsearch8/8.0.0/) on 3.6+. You can fix it by being more selective with `pys`.

ddtrace/contrib/elasticsearch/patch.py
```diff
@@ -45,14 +49,21 @@ def _es_modules():
             pass
 
 
+def get_version():
+    import elasticsearch
```
Yun-Kim, line 53: `_patch(elasticsearch)` already takes in an imported `elasticsearch` so we don't need to import it twice.
```suggestion
def get_version(elasticsearch):
```

Yun-Kim (commented):
Other than the two comments, version gating lgtm. On a side note, we will likely need to convert this to registering as patch on import instead of us force importing at patch time (this is unblocking but with the recent patch-on-import changes I figured this was worth bringing up).

ddtrace/contrib/elasticsearch/patch.py
ZStriker19: ah thanks, I got mixed up!

emmettbutler (approved):

ddtrace/contrib/elasticsearch/patch.py
```diff
@@ -1,4 +1,5 @@
 from importlib import import_module
+import logging
```
mabdinur, line 2:
```suggestion
from ddtrace.internal.logger import get_logger
```

ddtrace/contrib/elasticsearch/patch.py
```diff
@@ -20,6 +21,8 @@
 from .quantize import quantize
 
 
+log = logging.Logger(__name__)
```
mabdinur, line 24: We should use the ddtrace logger

```suggestion
log = get_logger(__name__)
```

ddtrace/contrib/elasticsearch/patch.py
```diff
@@ -45,14 +48,19 @@ def _es_modules():
...
 # NB: We are patching the default elasticsearch.transport module
 def patch():
     for elasticsearch in _es_modules():
         _patch(elasticsearch)
 
 
 def _patch(elasticsearch):
-    if getattr(elasticsearch, "_datadog_patch", False):
+    # elasticsearch 8 is not yet supported
+    if getattr(elasticsearch, "_datadog_patch", False) or get_version(elasticsearch) >= (8, 0, 0):
```
mabdinur, line 63: `get_version()` can return a string. It should always return a tuple.

ZStriker19: got it, I'll make sure it can handle either a tuple or str

mabdinur (approved):

PR #5797, opened 2023-05-05
chore(internal): introduce better error logging when we cannot pack a span

brettlangdon:
If we end up with a meta value that is not a string it can raise a cryptic error that doesn't help deubgging.

This change introduces a new error message that includes the `repr(span)` that caused the issue.

This change will *not* include the meta for the span, but having the span name should help narrow things down.

## Checklist

## Reviewer Checklist

- [ ] Title is accurate.


ddtrace/internal/_encoding.pyx
```diff
@@ -477,9 +477,11 @@ cdef class MsgpackEncoderBase(BufferedEncoder):
             dd_origin = self.get_dd_origin_ref(trace[0].context.dd_origin)
 
         for span in trace:
-            ret = self.pack_span(span, dd_origin)
-            if ret != 0:
-                raise RuntimeError("Couldn't pack span")
+            try:
+                ret = self.pack_span(span, dd_origin)
+                if ret != 0: raise RuntimeError("couldn't pack span: {!r}".format(span))
```
brettlangdon, line 482:
```suggestion
                if ret != 0: raise RuntimeError("couldn't pack span")
```

Is this raises redundant?

tests/tracer/test_encoders.py
P403n1x87, line 1: Does this affect only the v03 encoder? What about the v05? Do we not need to test the case where the service name/resource is set to something other than a string too?

brettlangdon: We found the bug originating from V05 encoder. The bug is actually different, since V03 would (pre-this change) would raise a TypeError, but V05 doesn't check the type, so it is just an attribute error calling `len(_GeneratorContextManager())`, but I figured we should centralize the "improved" logging.

Yes, other attributes could have this issue as well. I am not sure how exhaustive we want to be with the tests since we have tests for bad data, we can likely just fill in regressions as we learn about them? (this code should handle _any_ error, so 🤷🏻)

emmettbutler (approved):

emmettbutler (approved):

releasenotes/notes/fix-encoding-exception-ff50914f97554744.yaml
```diff
@@ -0,0 +1,3 @@
+fixes:
+  - |
+    Fixes a cryptic encoding exception message when a span tag is not a string.
```
mabdinur, line 3:
```suggestion
    tracing: Fixes a cryptic encoding exception message when a span tag is not a string.
```

emmettbutler (approved):

PR #5803, opened 2023-05-08
feature(asm): trace subprocess executions

juanjux:
## Description

This PRs implements the internal RFC "Shell Execution Integration" for the Python library. It will add a new span type with information about commands executed by the code either using `os.system`, `os.spawn*` or any of the functions or objects in the `subprocess` module. It will be automatically enabled if ASM is enabled.

## Checklist

## Reviewer Checklist

- [ ] Title is accurate.

### DIFF
--- a/ddtrace/_monkey.py
+++ b/ddtrace/_monkey.py
@@ -87,6 +87,18 @@ PATCH_MODULES = {
     "tornado": False,
 }
 
+# Patch some more modules for tracing subprocess executions if ASM is enabled
+# XXX JJJ get list from the module dict
+if formats.asbool(os.getenv("DD_APPSEC_ENABLED")):
+    # XXX add more as needed
+    PATCH_MODULES.update(
+        {
+            "os": True,
+            "subprocess": True,
+        }
+    )
+
+
 
 # this information would make sense to live in the contrib modules,
 # but that would mean getting it would require importing those modules,
@@ -195,6 +207,7 @@ def patch_all(**patch_modules):
 
     patch(raise_errors=False, **modules)
     patch_iast(**IAST_PATCH)
+    patch_subprocess_executions()
 
 
 def patch_iast(**patch_modules):
@@ -210,6 +223,14 @@ def patch_iast(**patch_modules):
             when_imported("hashlib")(
                 _on_import_factory(module, prefix="ddtrace.appsec.iast.taint_sinks", raise_errors=False)
             )
+
+
+def patch_subprocess_executions():
+    if not config._appsec_enabled:
+        return
+
+    from .appsec._patch_subprocess_executions import _patch
+    _patch()
 
 
 def patch(raise_errors=True, patch_modules_prefix=DEFAULT_MODULES_PREFIX, **patch_modules):
--- a/ddtrace/_monkey.py
+++ b/ddtrace/_monkey.py
@@ -87,18 +87,6 @@ PATCH_MODULES = {
     "tornado": False,
 }
 
-# Patch some more modules for tracing subprocess executions if ASM is enabled
-# XXX JJJ get list from the module dict
-if formats.asbool(os.getenv("DD_APPSEC_ENABLED")):
-    # XXX add more as needed
-    PATCH_MODULES.update(
-        {
-            "os": True,
-            "subprocess": True,
-        }
-    )
-
-
 
 # this information would make sense to live in the contrib modules,
 # but that would mean getting it would require importing those modules,
--- a/ddtrace/_monkey.py
+++ b/ddtrace/_monkey.py
@@ -218,6 +218,7 @@ def patch_subprocess_executions():
         return
 
     from .appsec._patch_subprocess_executions import _patch
+
     _patch()
 
 
--- a/ddtrace/appsec/_patch_subprocess_executions.py
+++ b/ddtrace/appsec/_patch_subprocess_executions.py
@@ -21,11 +21,12 @@ log = get_logger(__name__)
 
 """
 JJJ TODO:
-- add some catching (strings, lists, scrubbed env vars and parameters, et cetera)
-- Truncation
+- add some catching (result for as_string() and as_list(), scrubbed stuff...) 
+- make sure than _unpatch is called at the right time inside the tracer
 - flask snapshot tests from views
 - exception handlers so it never fails and always executes the command
 - constants for the tag names
+- rebase
 """
 
 
@@ -194,7 +195,7 @@ class SubprocessCmdLine(object):
         return str_[0:-(oversize+len(msg))] + msg
 
 
-    def as_list(self):
+    def as_list_and_string(self):
         # type: () -> Tuple[list[str], str]
 
         total_list = self.env_vars + [self.binary] + self.arguments
@@ -203,8 +204,12 @@ class SubprocessCmdLine(object):
         return truncated_list, truncated_str
 
 
+    def as_list(self):
+        return self.as_list_and_string()[0]
+
+
     def as_string(self):
-        return self.as_list()[1]
+        return self.as_list_and_string()[1]
 
 
 def scrub_arg(_arg):
--- a/ddtrace/appsec/_patch_subprocess_executions.py
+++ b/ddtrace/appsec/_patch_subprocess_executions.py
@@ -16,6 +16,7 @@ from ddtrace import config
 from ddtrace.contrib import trace_utils
 from ddtrace.ext import SpanTypes
 from ddtrace.internal import _context
+from ddtrace.internal.compat import shjoin, PY2
 from ddtrace.internal.logger import get_logger
 
 
@@ -43,10 +44,17 @@ def _patch():
 
     Pin().onto(os)
     trace_utils.wrap(os, "system", traced_ossystem(os))
-    # note: popen* uses subprocess, which we already wrap below
 
     # all os.spawn* variants eventually use this one:
     trace_utils.wrap(os, "_spawnvef", traced_osspawn(os))
+
+    if PY2:
+        # note: popen* uses subprocess in Python3, which we already wrap below, but not in
+        # Python2
+        trace_utils.wrap(os, "popen", traced_py2popen(os))
+        trace_utils.wrap(os, "popen2", traced_py2popen(os))
+        trace_utils.wrap(os, "popen3", traced_py2popen(os))
+        trace_utils.wrap(os, "popen4", traced_py2popen(os))
 
     Pin().onto(subprocess)
     # We store the parameters on __init__ in the context and set the tags on wait
@@ -61,9 +69,6 @@ def _unpatch():
     trace_utils.unwrap(os, "_spawnvef")
     trace_utils.unwrap(subprocess.Popen, "__init__")
     trace_utils.unwrap(subprocess.Popen, "wait")
-
-
-_COMPILED_ENV_VAR_REGEXP = re.compile(r"\b[A-Z_]+=\w+")
 
 
 @attr.s(eq=False)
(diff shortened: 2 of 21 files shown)


ddtrace/_monkey.py
```diff
@@ -213,6 +214,15 @@ def patch_iast(**patch_modules):
             )
 
 
+def patch_subprocess_executions():
```
avara1986, line 217: rename to `patch_appsec_subprocess_executions`, WDYT?

ddtrace/appsec/_patch_subprocess_executions.py
```diff
@@ -0,0 +1,400 @@
...
+
+def _patch():
+    # type: () -> None
+    if not config._appsec_enabled:
+        return
+
+    import os
+
+    Pin().onto(os)
+    trace_utils.wrap(os, "system", traced_ossystem(os))
```
P403n1x87, line 40: We know that some frameworks do monkey-patching (e.g. gevent does monkey-patching of the os module). What are the odds that such frameworks would undo our instrumentation?

ddtrace/internal/compat.py
```diff
@@ -454,3 +454,33 @@ def _rmtree(self, path):
             self._rmdir(path)
         except OSError:
             pass
+
+
+try:
+    from shlex import quote as shquote
```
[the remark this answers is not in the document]
juanjux: Python 3.4 up. It's done on other parts of the file that way.

ddtrace/internal/compat.py
```diff
@@ -454,3 +454,33 @@ def _rmtree(self, path):
...
+        if _find_unsafe(s) is None:
+            return s
+
+        # use single quotes, and put single quotes into double quotes
+        # the string $'b is then quoted as '$'"'"'b'
+        return "'" + s.replace("'", "'\"'\"'") + "'"
+
+
+try:
+    from shlex import join as shjoin
```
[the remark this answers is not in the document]
juanjux: ditto2 :)

ddtrace/appsec/_patch_subprocess_executions.py
juanjux: Well, high I guess, specially if the patching is done after we have done ours. This is something that probably should be documented. For gevent, we can detect the patching and at least write a warning, but I'll investigate if there are better options (if you have ideas since you've fought a lot with gevent I'm all ears!).

ddtrace/_monkey.py
juanjux: It's in an appsec module but yeah, more explicit for that like of code :+1:

ddtrace/appsec/_patch_subprocess_executions.py
juanjux: Another update:

Tested with a simple flask app, ran with gevent. Seems to work but of course in this case gevent is run before ddtrace, which matches the case that worked in my first test.

juanjux:
Going to set this as draft again because `test_oce_max_requests_py3` is misteriously failing (only in CI of couse) and I want to make some debug-style changes to the branch to try to get more information about what is happening.

avara1986 (approved):

ddtrace/appsec/_patch_subprocess_executions.py
P403n1x87: > if you have ideas since you've fought a lot with gevent I'm all ears!

I think this is tricky to deal with. Normally we would want to trigger patching only after a module that needs patching is imported. But if we do this here we would still end up with gevent undoing our monkey-patching (we would patch soon after gevent imports the module that it wants to patch, and then gevent would replace things, thus removing our patching).

avara1986 (dismissed):

P403n1x87 (dismissed):

ddtrace/appsec/_patch_subprocess_executions.py
```diff
@@ -0,0 +1,435 @@
...
+
+    TRUNCATE_LIMIT = 4 * 1024
+
+    ENV_VARS_ALLOWLIST = {"LD_PRELOAD", "LD_LIBRARY_PATH", "PATH"}
+
+    BINARIES_DENYLIST = {
+        "md5",
+    }
+
+    SENSITIVE_WORDS_WILDCARDS = (
```
mabdinur, line 125: Can we make this list configurable via environment variable? I can see a scenario where customers would like to add sensitive words. 

This is not a blocking comment, we can do this in a follow up

ddtrace/_monkey.py
```diff
@@ -201,6 +200,11 @@ def patch_all(**patch_modules):
 
     patch(raise_errors=False, **modules)
     patch_iast(**IAST_PATCH)
+    if config._appsec_enabled:
+        patched_modules = patch_appsec_subprocess_executions()
+        with _LOCK:
+            for patched in patched_modules:
+                _PATCHED_MODULES.add(patched)
```
mabdinur, line 207: Suggestion: Right now subprocess executions are only patched if `patch_all()` is used. Can we use a pattern similar to the other integrations?  Ex: define a "subprocess" integration in `PATCH_MODULES` and then add the following on line 252 (in _monket.patch(...)).

```
    if module == "subprocess" and config._appsec_enabled:
        from .appsec._patch_subprocess_executions import _patch
        _patch()
        continue
```

This approach allows customers to opt in or out of subprocess tracing using ``DD_TRACE_SUBPROCESS_ENABLED`` however this is not a typical integration so this code block feels a bit out of place.

An alternative is to define integrations for both os and subprocess.

Again this suggestion is non-blocking the current approach works

releasenotes/notes/asm-subprocess-executions-479ab8fb59d9f312.yaml
```diff
@@ -0,0 +1,12 @@
+---
+prelude: >
+  Application Security Management (ASM) has added support for tracing subprocess executions.
+
+features:
+  - |
+    ASM: add support for tracing subprocess executions (like `os.system`, `os.spawn`, `subprocess.Popen` and others) and adding
```
mabdinur, line 7: We try to avoid using release notes as the primary source for documentation. Ideally a release note should just call out a change and link to a doc for more information. 

Since this is large high impact feature I cam okay with this descriptive release note. In a follow up PR can we start documenting more ASM features and configurations in our ddtrace library docs: 
Ex:
https://github.com/DataDog/dd-trace-py/blob/bb5dccc1405723d103a132475ac00ea693cdafcd/docs/api.rst#api
https://github.com/DataDog/dd-trace-py/blob/bb5dccc1405723d103a132475ac00ea693cdafcd/docs/index.rst
https://github.com/DataDog/dd-trace-py/blob/bb5dccc1405723d103a132475ac00ea693cdafcd/docs/configuration.rst

mabdinur (approved):
I left some non blocking comments. Overall looks good to me.

releasenotes/notes/asm-subprocess-executions-479ab8fb59d9f312.yaml
juanjux: Thanks. Yes, we usually defer to the global docs for ASM features which have Python-specific tabs but improving the dd-trace-py specific documentation would be great, I added a task to our Jira board so we can tackle it on some next sprint.

ddtrace/_monkey.py
juanjux: I still didn't have a clear answer on this (probably the people I contacted didn't know for sure either) but it looks like this will be extended in the future (just like we'll be extending login events soon) to be able to block subprocess executions using RCM so I kept it in the ASM product directory. Regarding the options like the env var you suggested, the current RFC indicated that the feature should be enabled or disabled totally depending only if ASM is enabled.

ddtrace/appsec/_patch_subprocess_executions.py
juanjux: Since I think this is a great suggestion and the current RFC wording doesn't explicitly disallow to add more words (it disallow removing them) I'll implement this.

juanjux: Commited, PTAL when you can.

avara1986 (approved):

mabdinur (dismissed):

avara1986 (approved):

majorgreys (commented):
Can we make this PR introducing a subprocess integration, which is not enabled by default, then a follow-up that includes the logic to enable it for appsec users?

avara1986 (approved):

avara1986 (approved):

juanjux:
> Can we make this PR introducing a subprocess integration, which is not enabled by default, then a follow-up that includes the logic to enable it for appsec users?

Done! Sorry for the delay, was buried in other things. Please take another look when you can.

gnufede (dismissed):

.github/CODEOWNERS
```diff
@@ -32,13 +32,15 @@ tests/internal/test_module.py       @DataDog/debugger-python @DataDog/apm-core-p
 
 # ASM
 ddtrace/appsec/                     @DataDog/asm-python
+ddtrace/contrib/subprocess/         @DataDog/asm-python
```
majorgreys, line 35: I would only make the appsec specific code (which seems to be only the automatic enablement) as owned differently

```suggestion
ddtrace/contrib/subprocess/         @DataDog/asm-python
```

majorgreys (commented):
@juanjux thank you for making this into an integration. My comments here are mainly still cleanup suggestions towards this being a full-fledged integration. I would suggest making the appsec specific behavior a follow-up. Also, this does not yet support distributed tracing. While I don't think it is needed in an initial support for subprocesses, what are your thoughts on it and, specifically, for the purposes of appsec?

ddtrace/contrib/subprocess/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+~~~~~~~~~~~~~
+
+.. py:data:: ddtrace.config.subprocess['sensitive_wildcards']
+
+   Comma separated list of fnmatch-style wildcards Subprocess parameters matching these
+   wildcards will be scrubbed and replaced by a "?".
+
+   Default: ``None`` for the config value but note that there are some wildcards always
+   enabled in this integration that you can check on
+   ```ddtrace.contrib.subprocess.constants.SENSITIVE_WORDS_WILDCARDS```.
```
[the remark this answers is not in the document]
juanjux: I tough about this but this could be confusing, because then the user could thing that explicitly setting the variable to `None` or setting different values would disable them, which is not the case (they'll always be used, even without the environmnent variable set). What do you think?

ddtrace/contrib/subprocess/patch.py
```diff
@@ -0,0 +1,444 @@
...
+from ddtrace.internal.compat import shjoin
+from ddtrace.internal.logger import get_logger
+
+
+log = get_logger(__name__)
+
+config._add(
+    "subprocess",
+    dict(
+        subexec_sensitive_users_wildcars=os.getenv("DD_APPSEC_SUBPROCESS_EXEC_SENSITIVE_WILDCARDS", default="").split(
```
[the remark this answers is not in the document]
juanjux: Done (didn't commit directly your suggestion because there were other places to change it).

ddtrace/contrib/subprocess/patch.py
```diff
@@ -0,0 +1,444 @@
...
+from typing import Tuple
+from typing import Union
+from typing import cast
+
+import attr
+import six
+
+from ddtrace import Pin
+from ddtrace import config
+from ddtrace.appsec._constants import COMMANDS
```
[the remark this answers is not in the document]
juanjux: Done!

.github/CODEOWNERS
juanjux: Right now all the code on that directory is only for implementing an ASM RFC so I think it makes sense for it to have asm ownership, once APM or other team contributes I would make sense to make it multi owned or owned by APM to have a single entry point.

ddtrace/contrib/subprocess/__init__.py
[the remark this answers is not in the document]
majorgreys: Ah, ok - I would then list the wildcards enabled by default. This config option appends to that list.

gnufede (approved):

PR #5808, opened 2023-05-08
chore(service_name): add service schema grpc

divyatagauchan:
What's changing in this PR?
This PR introduces the initial changes for "(2) Standardizing all tracer libraries to using the current application name as a default for existing services (service naming)". The major changes in this revision are:

Create a new concept, 'span attribute schema,' which controls the service naming behavior and control it with a new environment variable DD_TRACE_SPAN_ATTRIBUTE_SCHEMA.
Create 'v0' (existing behavior) and 'v1' (new behavior) schemas and behaviors:
v0 (Default) - The existing behavior
v1 - Default integration service names are now set to the value of
$DD_SERVICE (config().service) instead of the prior default values

Specifically for the GRPC integration. Other PRs are being merged for the other integrations.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


tests/contrib/grpc/test_grpc.py
```diff
@@ -38,6 +42,52 @@ def setUp(self):
...
+        with grpc.insecure_channel("localhost:%d" % (_GRPC_PORT)) as channel:
+            stub = HelloStub(channel)
+            response = stub.SayHello(HelloRequest(name="propogator"))
+
+        spans = self.get_spans_with_sync_and_assert(size=2)
+        client_span, server_span = spans
+        assert spans[0].service != "mysvc"
+
+    @TracerTestCase.run_in_subprocess(env_overrides=dict(DD_GRPC_SERVICE="mygrpc", DD_TRACE_SPAN_ATTRIBUTE_SCHEMA="v0"))
+    def test_unspecified_service_v0(self):
```
ZStriker19, line 83:
```suggestion
    def test_client_service_name_config_env_override_v0(self):
```

ZStriker19: Since we're using `DD_GRPC_SERVICE="mygrpc"` it's not unspecified. 

We should also add a v1 test for this. And probably also v0 and v1 tests for if both `DD_GRPC_SERVICE` and `DD_SERVICE` are specified.

With that I think we should be good.

tests/contrib/grpc/test_grpc.py
```diff
@@ -38,6 +42,52 @@ def setUp(self):
...
+        from ddtrace import config
+
+        assert config.service == "mysvc"
+        with grpc.insecure_channel("localhost:%d" % (_GRPC_PORT)) as channel:
+            stub = HelloStub(channel)
+            response = stub.SayHello(HelloRequest(name="propogator"))
+
+        spans = self.get_spans_with_sync_and_assert(size=2)
+        client_span, server_span = spans
+        assert spans[0].service != "mysvc"
```
ZStriker19, line 80: Rather than assert what the service value is not, let's just assert exactly what we think it should be... which is `grpc-client`.

tests/contrib/grpc/test_grpc.py
```diff
@@ -39,6 +42,93 @@ def setUp(self):
...
+    def test_client_service_name_config_env_override_v0(self):
+        with grpc.insecure_channel("localhost:%d" % (_GRPC_PORT)) as channel:
+            stub = HelloStub(channel)
+            response = stub.SayHello(HelloRequest(name="propogator"))
+
+        spans = self.get_spans_with_sync_and_assert(size=2)
+        client_span, server_span = spans
+        assert spans[0].service == "mygrpc"
+
+    @TracerTestCase.run_in_subprocess(env_overrides=dict(DD_GRPC_SERVICE="mygrpc", DD_TRACE_SPAN_ATTRIBUTE_SCHEMA="v1"))
```
ZStriker19, line 98: This says v0 in test name but sets `DD_TRACE_SPAN_ATTRIBUTE_SCHEMA="v1"`

ZStriker19: It also has the same name as the test on 123

ZStriker19 (approved):
Looks good! Great job!

tabgok (approved):

PR #5818, opened 2023-05-09
fix(asgi): ensure error tags are set on request spans

mabdinur:
If an exception is raised in the ddtrace ASGI middleware the `<asgi_framework_name>.request` span will be finished before the ``error.message`` and ``error.stack`` tags are set on the span. Currently these tags are not submitted to the agent.

Reproduction:

1. Install `fastapi` and `ddtrace==1.9.0`
2. Create fastapi route that raises an Exception. Ex:
    ```python
      from fastapi import FastAPI
      app = FastAPI()

      @app.get("/abc")
      def hello() -> str:
          raise ValueError('error')
          return "ok"
    ```
3. Run fastapi app with ddtrace-run
4. Result:
     - Screen Shot 2023-05-09 at 3 13 05 PM

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/contrib/asgi/middleware.py
```diff
@@ -259,9 +247,6 @@ async def wrapped_blocked_send(message):
                 self.handle_exception_span(exc, span)
                 reraise(exc_type, exc_val, exc_tb)
             finally:
-                try:
-                    del scope["datadog"]["request_span"]
-                except KeyError:
-                    pass
-
+                if span in scope["datadog"]["request_spans"]:
```
ZStriker19, line 250: Could you explain why we need to change this as well?

ZStriker19 (commented):
Looks good, just one question

ddtrace/contrib/asgi/middleware.py
emmettbutler: This change looks to me like a less-verbose way of doing the same `del` logic that was on lines 262-265.

emmettbutler (approved):

ddtrace/contrib/asgi/middleware.py
mabdinur: It's a bit subtle but this is actually a "bug fix". scope["datadog"]["request_span"] is never set, we actually set scope["datadog"]["request_span**s**"].  This line always raises a KeyError. The impact of this bug is pretty low. We'll just hold more spans in memory than we need to. I can move this fix to it's own PR.

juanjux (approved):

victorcelerier:
Hi everyone, following this commit if I'm not wrong, it means we will include background task in web requests span again right ? It was removed inside https://github.com/DataDog/dd-trace-py/commit/5b97489e2b073fa773819904cdef26981a5a28df (asgi, starlette, fastapi: Exclude background tasks duration from web request span )

Yun-Kim:
@mabdinur just for clarity for future reviewers, can we add what the initial problem was and how we fixed it in the PR description? Thanks!

PR #5820, opened 2023-05-09
fix(appsec): fix partial flush encoding error when request context is active

brettlangdon:
#5813 made an attempt to prevent leaking AppSec request contexts as span tags by ensuring we always clean up any that are created when the span finishes.

That fix did not take into account partial flushing. When partial flushing we will copy the trace context's tags onto the first span in the payload. This means if the root span creates an AppSec request context it will be put into the `context._meta`, and it is expected to stay there until the root span is finished. Resulting in the request context still being added as a span tag. This is likely the actual root cause that #5913 was trying to solve.

The solution presented here is to stop using `context._meta` to track AppSec request contexts. Instead we are adding a `WeakKeyDictionary` to the `AppSecSpanProcessor` so it can keep track of the request contexts without the need of the span context at all.

A `WeakKeyDictionary` was used to prevent accidentally holding onto strong references to spans potentially causing a memory leak if `on_span_finish` never gets called.

I tried to create a test case for the possible memory leak and it is really hard. There are a lot of pieces that hold strong references to spans in the tracer/context/processors. I think ultimately a memory leak test is just validating that `WeakKeyDictionary` does what it says it does, so it likely is just over testing (happy to have someone disagree with me).

## Checklist

## Reviewer Checklist

- [ ] Title is accurate.

### DIFF
--- a/ddtrace/appsec/processor.py
+++ b/ddtrace/appsec/processor.py
@@ -5,6 +5,7 @@ import os.path
 import traceback
 from typing import Set
 from typing import TYPE_CHECKING
+from weakref import WeakKeyDictionary
 
 import attr
 from six import ensure_binary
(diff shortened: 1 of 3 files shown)


brettlangdon:
superceded by #5822

PR #5847, opened 2023-05-11
fix(profiling): handle partially loaded threading module

sanchda:
When using `gevent` the application needs to reload some modules that are also used by `ddtrace`. One such modules is `threading`, which plays an important role in the current implementation of the profiler. Since the stack collector needs to access both the `ddtrace` and the application copy of the `threading` module, with the profiler started before the new copy is created by the application, there is a race between the stack sampling and the module import that can result in the stack collector trying to access attributes from the new copy of `threading` before its import is completed.

As a fix for this issue, we guard against expected `AttributeError`s.

## Testing strategy

An existing gunicorn + gevent test has been extended to catch any potential signs of the issue occurring again in the future. Given that this is a race condition, the issue itself is hard to reproduce systematically.

## Risks

We don't expect any risks in guarding for well-known and stable attributes of the `threading` module.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


P403n1x87 (commented):
Just the release note and a test case to catch any potential regressions (if possible), otherwise LGTM!

armcburney (dismissed):
🙏

P403n1x87 (approved):

PR #5849, opened 2023-05-11
chore(asgi): revert regression that removed support for background tasks

mabdinur:
Addresses: https://github.com/DataDog/dd-trace-py/pull/5818#issuecomment-1543604812
Partially reverts: https://github.com/DataDog/dd-trace-py/pull/5818

We will need to backport this fix to 1.9, 1.10, 1.11, 1.12 and 1.13 release lines.

This regression is unreleased so a release note is not required.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


Yun-Kim (commented):
Does this revert mean that the error tags not being set on spans from #5818 is still an issue?

mabdinur:
Nope. This is not a full revert. We only finish the span early if the request span has a "ok" status.

mabdinur:
This should resolve the issues where error messages are not set on asgi spans AND ensure the duration of background tasks are not included in request spans

Yun-Kim (commented):
Thanks for clarifying @mabdinur. Just to confirm, are you still able to reproduce the error from #5818 after this change? I know we have snapshot tests for this now but AFAIK we already had a test case that showed errors being set previous to the original bug report?

mabdinur:
Previously the span would be finished, sent and then the error tags would be sent. So there were tests and those tests were passing but the logic was broken. Using a snapshot test validates that the error tags are actually sent to the agent.

Yun-Kim (approved):

Issue #5722, opened 2023-05-02
Cannot execute Sanic application with ddtrace


EverettDunamu:
### Summary of problem
- I cannot execute my Sanic Web Server with `ddtrace`
```
$ python main.py # works well
$ ddtrace-run python main.py # returns an error!
```

### Which version of dd-trace-py are you using?
1.12.2

### Which version of pip are you using?
23.1.2

### Which libraries and their versions are you using?
sanic == 23.3.0
python = 3.11

<details>
  <summary>`pip freeze`</summary>
```
  aiofiles==23.1.0
aiohttp==3.8.4
aiosignal==1.3.1
async-timeout==4.0.2
attrs==22.2.0
black==23.3.0
boto3==1.26.118
botocore==1.29.118
bytecode==0.14.1
cattrs==22.2.0
certifi==2022.12.7
cfgv==3.3.1
charset-normalizer==3.1.0
click==8.1.3
ddsketch==2.0.4
ddtrace==1.12.2
Deprecated==1.2.13
distlib==0.3.6
envier==0.4.0
filelock==3.10.7
frozenlist==1.3.3
gevent==22.10.2
greenlet==2.0.2
grequests==0.6.0
html5tagger==1.3.0
httptools==0.5.0
identify==2.5.22
idna==3.4
importlib-metadata==6.0.1
jmespath==1.0.1
jsonschema==4.17.3
multidict==6.0.4
mypy-extensions==1.0.0
nodeenv==1.7.0
opentelemetry-api==1.17.0
packaging==23.0
pathspec==0.11.1
platformdirs==3.2.0
pre-commit==3.2.1
protobuf==4.22.3
psycopg2-binary==2.9.6
pyrsistent==0.19.3
python-dateutil==2.8.2
python-dotenv==1.0.0
PyYAML==6.0
requests==2.28.2
s3transfer==0.6.0
sanic==23.3.0
sanic-ext==23.3.0
sanic-routing==22.8.0
six==1.16.0
tenacity==8.2.2
tracerite==1.1.0
typing_extensions==4.5.0
ujson==5.7.0
urllib3==1.26.15
uvloop==0.17.0
virtualenv==20.21.0
websockets==10.4
wrapt==1.15.0
xmltodict==0.13.0
yarl==1.8.2
zipp==3.15.0
zope.event==4.6
zope.interface==6.0
```
</details>

### How can we reproduce your problem?
I created server.py like below
```
from sanic import Sanic
from sanic.response import text

[...]

PR #5867, opened 2023-05-15
fix(sanic): avoids unloading re module when gevent is installed

mabdinur:
## Background
When gevent is installed (or `DD_UNLOAD_MODULES_FROM_SITECUSTOMIZE=auto`), all modules imported  by [ddtrace.bootstrap.sitecustomize](https://github.com/DataDog/dd-trace-py/blob/1.x/ddtrace/bootstrap/sitecustomize.py#L91) are unloaded from sys.modules (except for a small [exclude list](https://github.com/DataDog/dd-trace-py/blob/1.x/ddtrace/bootstrap/sitecustomize.py#L103-L113)).

## Description

Currently the [typing module](https://github.com/DataDog/dd-trace-py/blob/1.x/ddtrace/bootstrap/sitecustomize.py#L110) is excluded from module unloading but re is NOT. This causes [isinstance(some_re_pattern, typing.Pattern)](https://github.com/sanic-org/sanic-routing/blob/main/sanic_routing/router.py#L289) to return False (here typing holds a stale reference to [re.Pattern](https://github.com/python/cpython/blob/main/Lib/typing.py#L3422)).

## Risks

This bug was called out and reproduced in the following github issue: https://github.com/DataDog/dd-trace-py/issues/5722. However this issue could not be reproduced using riot due to how packages are installed. To workaround this limitation a framework test was added for sanic which reproduced this error. This is less than ideal. Module cloning is a vital part of our codebase and should have test coverage for all common use-cases.

### Explanation
Riot uses `pip install -e` to install dependencies. When packages are installed with the `-e` option a `.pth` file is added to `site-packages/` which get run even before `sitecustomize.py`. This is done to extend the Python path to include additional packages. The creation of the `.pth` file has a side-effect of that masks this bug. It loads the re module before importing ddtrace and before LOADED_MODULES can get set, so it is basically the same impact as ignoring/not cleaning up re module. 

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


.github/workflows/test_frameworks.yml
```diff
@@ -52,6 +52,33 @@ jobs:
...
+          path: sanic
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.11"
+      - name: Install sanic and dependencies required to run tests
+        run: pip3 install '.[test]' aioquic gevent
+      - name: Install ddtrace
+        run: pip3 install ../ddtrace
+      - name: Run tests
+        run: ddtrace-run pytest tests -k "not test_add_signal and not test_ode_removes and not test_skip_touchup and not test_dispatch_signal_triggers"
```
mabdinur, line 80:
```suggestion
        run: ddtrace-run pytest -k "not test_add_signal and not test_ode_removes and not test_skip_touchup and not test_dispatch_signal_triggers and not test_keep_alive_connection_context"
```

releasenotes/notes/sanic-fix-is-pattern-check-c6e4109e933bf37f.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    sanic: Resolves ``sanic_routing.exceptions.InvalidUsage`` error raised when
+    gevent is installed and moduling cloning is enabled (``DD_UNLOAD_MODULES_FROM_SITECUSTOMIZE=auto``).
```
mabdinur, line 5:
```suggestion
    gevent is installed or ``DD_UNLOAD_MODULES_FROM_SITECUSTOMIZE`` is set to True.
```

ddtrace/bootstrap/sitecustomize.py
```diff
@@ -108,6 +108,7 @@ def drop(module_name):
             "asyncio",
             "concurrent",
             "typing",
+            "re",
```
P403n1x87, line 111: can we add a comment that explains why we are adding this module here?

P403n1x87:
Nice catch!

brettlangdon (dismissed):

P403n1x87 (approved):

PR #5877, opened 2023-05-16
feat(data_streams): add_data_streams_integration_to_kafka

piochelepiotr:
Add Data Streams integration with Kafka.
The context is right now not propagated inside an application, only between different applications. I will add context propagation inside the application (using trace context) in a follow up PR.

I tested it by writing a small Python application, and submitting traffic to our staging environment.
The code [will be] under a env variable that will act as a feature flag.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


kr-igor (changes requested):

ddtrace/contrib/kafka/patch.py
```diff
@@ -86,6 +87,12 @@ def traced_produce(func, instance, args, kwargs):
         value = None
     message_key = kwargs.get("key", "")
     partition = kwargs.get("partition", -1)
+    if config._data_streams_enabled:
+        # inject data streams context
+        headers = kwargs.get('headers', {})
+        pathway = pin.tracer.data_streams_processor.new_pathway(["type:kafka", "direction:out", "topic:"+topic])
```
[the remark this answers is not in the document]
piochelepiotr: there is a sort function in the new_pathway function, but let me put the order correctly from the start.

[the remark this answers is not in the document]
kr-igor: Yeah, sorting seems like unnecessary overhead, given that the order never changes.

ddtrace/contrib/kafka/patch.py
```diff
@@ -123,6 +130,9 @@ def traced_poll(func, instance, args, kwargs):
         span.set_tag_str(kafkax.RECEIVED_MESSAGE, str(message is not None))
         span.set_tag_str(kafkax.GROUP_ID, instance._group_id)
         if message is not None:
+            headers = {header[0]: header[1] for header in (message.headers() or [])}
+            if config._data_streams_enabled:
+                pin.tracer.data_streams_processor.decode_pathway(headers.get(PROPAGATION_KEY, None), ["type:kafka", "direction:in", "topic:"+message.topic(), "group:"+instance._group_id])
```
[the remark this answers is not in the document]
piochelepiotr: Updated

ddtrace/contrib/kafka/patch.py
```diff
@@ -123,6 +131,10 @@ def traced_poll(func, instance, args, kwargs):
         span.set_tag_str(kafkax.RECEIVED_MESSAGE, str(message is not None))
         span.set_tag_str(kafkax.GROUP_ID, instance._group_id)
         if message is not None:
+            headers = {header[0]: header[1] for header in (message.headers() or [])}
+            if config._data_streams_enabled:
+                ctx = pin.tracer.data_streams_processor.decode_pathway(headers.get(PROPAGATION_KEY, None))
+                ctx.set_checkpoint(["direction:in", "group:"+instance._group_id, "topic:"+message.topic(), "type:kafka"])
```
kr-igor, line 137: It seems like we're not propagating the context anywhere. This will still work, but we not gonna be able to get true end-to-end latency. Should we consider having some sort of  context tracking within `pin.tracer.data_streams_processor`? Maybe using a `ContextVar`?

kr-igor (dismissed):

ddtrace/contrib/kafka/patch.py
piochelepiotr: Added context propagation as discussed.

piochelepiotr:
If I could get a review about potential crashes that could be produced by this PR, that would be great. Should I add everything in try / except for eg? Should I do null checks everywhere? Thanks!

releasenotes/notes/data-streams-integration-bb8077b6d8a5daad.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    Add the Data Streams integration to Python and instrument Kafka with it.
```
majorgreys, line 4: Good to add a link for more information on Data Streams

riotfile.py
```diff
@@ -370,6 +370,50 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
                 ),
             ],
         ),
+        Venv(
+            name="datastreams",
```
majorgreys, line 374: Are there any integrations tests that can be added for kafka?

riotfile.py
```diff
@@ -370,6 +370,50 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
...
+                                # https://www.attrs.org/en/22.2.0/changelog.html#id1
+                                Venv(pys="3.6", pkgs={"attrs": "<22.2.0"}),
+                                Venv(pys="3.7"),
+                            ],
+                        ),
+                        Venv(pys=select_pys(min_version="3.8")),
+                    ],
+                ),
+                Venv(
+                    name="datastreams-snapshot",
```
majorgreys, line 401: Should this be removed as there are no snapshot tests?

piochelepiotr: Done

riotfile.py
piochelepiotr: It looks like right now, there are only snapshots tests for Kafka? https://github.com/DataDog/dd-trace-py/blob/aa3a538e0b39d77a5ca625c5625ecb1c1a32c108/tests/contrib/kafka/test_kafka.py#L67

ddtrace/settings/config.py
```diff
@@ -271,6 +271,7 @@ def __init__(self):
         self._trace_compute_stats = asbool(
             os.getenv("DD_TRACE_COMPUTE_STATS", os.getenv("DD_TRACE_STATS_COMPUTATION_ENABLED", in_gcp_function()))
         )
+        self._data_streams_enabled = asbool(os.getenv("DD_DATA_STREAMS_ENABLED"))
```
ZStriker19, line 274: Follow convention of other checks for envars and make a bit more obvious what default is?

```suggestion
        self._data_streams_enabled = asbool(os.getenv("DD_DATA_STREAMS_ENABLED", False))
```

ddtrace/internal/datastreams/processor.py
```diff
@@ -0,0 +1,270 @@
...
+            else:
+                log.debug("sent %s to %s", _human_size(len(payload)), self._agent_endpoint)
+
+    def periodic(self):
+        # type: () -> None
+
+        with self._lock:
+            serialized_stats = self._serialize_buckets()
+
+        if not serialized_stats:
```
ZStriker19, line 179: I wonder if we should debug log if there's nothing to send and this is enabled? From a solutions pov it'd probably be helpful to see that there's nothing to send, rather than there not being any log line at all, which would bring up the question of if this is enabled/running properly. If it's only every 10 seconds that we flush if there's nothing, I don't think it should be an issue from a performance standpoint.  Thoughts?

ddtrace/internal/datastreams/processor.py
```diff
@@ -0,0 +1,270 @@
...
+
+
+class DataStreamsCtx:
+    def __init__(self, processor, hash_value, pathway_start_sec, current_edge_start_sec):
+        # type: (DataStreamsProcessor, int, float, float) -> None
+        self.processor = processor
+        self.pathway_start_sec = pathway_start_sec
+        self.current_edge_start_sec = current_edge_start_sec
+        self.hash = hash_value
+        self.service = six.ensure_text(config._get_service("unnamed-python-service"))
```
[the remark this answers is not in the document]
ZStriker19: However, will the service name here match the service name set for the Kafka integration? It seems like it would be confusing to have stats submitted under one name and spans under another? Even if they represent different things about that integration.

ZStriker19 (commented):
Looks good so far, still need to review testing.

ddtrace/tracer.py
```diff
@@ -1013,6 +1016,8 @@ def shutdown(self, timeout=None):
             for processor in chain(span_processors, SpanProcessor.__processors__):
                 if hasattr(processor, "shutdown"):
                     processor.shutdown(timeout)
+            if config._data_streams_enabled:
```
ZStriker19, line 1019: Is it ever possible that `config._data_streams_enabled` could be set false during runtime, causing us to incorrectly skip this shutdown? If possible, should we instead just `try except` around the shutdown?

releasenotes/notes/data-streams-integration-bb8077b6d8a5daad.yaml
```diff
@@ -0,0 +1,5 @@
...
```
ZStriker19, line 4:
```suggestion
    kafka: Add the Data Streams integration to Python and instrument Kafka with it.
```

riotfile.py
ZStriker19: Could we add a test producing and consuming and checking the stats generated reliably to some degree? Or at the very least the bucket values?

tests/contrib/kafka/test_kafka.py
```diff
@@ -154,3 +156,66 @@ def test_analytics_without_rate(producer, consumer):
...
+    for i in range(10):
+        x = factory()
+        if x is not None:
+            return x
+        time.sleep(0.1)
+    return None
+
+
+def test_data_streams_kafka():
+    import six
```
ZStriker19, line 171: Do we need to inline these imports? Aren't they already imported at the top of the file?

ddtrace/internal/datastreams/processor.py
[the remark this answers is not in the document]
ZStriker19: spoke offline about this, this makes sense.

ddtrace/tracer.py
```diff
@@ -1013,6 +1016,11 @@ def shutdown(self, timeout=None):
...
```
ZStriker19, line 1019: Wouldn't we instead want to `try except` and then if `if config._data_streams_enabled:` log the warning?

ZStriker19 (approved):

mabdinur (approved):

mabdinur:
> ## Benchmarks
> Comparing candidate commit [b8396a3](https://github.com/DataDog/dd-trace-py/commit/b8396a34acaa2a0d268a82e66e21f1abd70df8c0) in PR branch `piotr-wolski/data-streams-support` with baseline commit [e8a13ca](https://github.com/DataDog/dd-trace-py/commit/e8a13cae3425869845753b34b832afc83d29eaa9) in branch `1.x`.
> 
> Found 0 performance improvements and 1 performance regressions! Performance is the same for 89 cases.
> 
> #### scenario:span-start
> * 🟥 `execution_time` [+0.778ms; +1.122ms] or [+3.280%; +4.729%]

I know @ZStriker19 reviewed/approved this PR but I am just curious if this benchmark has been discussed

PR #5911, opened 2023-05-19
chore(ci-visibility): use agentless or evp proxy for all requests

gnufede:
CI Visibility: All new API requests (except for the regular traces) have to be performed either through agentless mode or evp proxy mode.

This includes git packfile uploading and coverage file uploading.

We do this for the test events (description of the change below), but this also applies to the other requests to the event platform API.

---

The Agent v7.39.0 can now forward citestcycle payloads to the EventPlatform intake directly (ie: the same payloads we send in Agentless mode), instead of having to send APM traces when using the Agent, which contain less information.

To check if the Agent supports this mode, use the Agent API http://localhost:8126/info (where localhost:8126 is the configured Agent host) and check if the endpoints field of the response contains /evp_proxy/v2/.

This new mode requires the following changes to the requests sent from the Tracing Libraries:

The host should be the Agent host (by default http://localhost:8126) instead of https://citestcycle-intake.${site}.

The path should be /evp_proxy/v2/api/v2/citestcycle instead of /api/v2/citestcycle.

The header X-Datadog-EVP-Subdomain should be set to citestcycle-intake

The DD-API-KEY header is no longer needed in the request.

If we know the Container ID, it should be set in the Datadog-Container-ID header.

For now we can leave the code that only uses traces with the Agent in (eg: for customers with old versions of the Agent) but in the future we should be able to remove that.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


Yun-Kim (commented):
Can we add some more detail to the PR description on the current behavior for this issue and how this PR addresses this exactly? I'm having a hard time understanding the context here.

If we want ITR/code coverage/git packfile uploading to be run with agentless or evp proxy, then isn't it simpler to just make that a requirement to enable those features? Or do those features use agentless by default?

gnufede:
> Can we add some more detail to the PR description on the current behavior for this issue and how this PR addresses this exactly? I'm having a hard time understanding the context here.
> 
> If we want ITR/code coverage/git packfile uploading to be run with agentless or evp proxy, then isn't it simpler to just make that a requirement to enable those features? Or do those features use agentless by default?

I added some more detail, it's the same that is already done for citestcycle endpoint, but for all the endpoints (in the RFC, cytestcycle is the example used, but it affects all endpoints)

christophe-papazian (approved):

PR #5935, opened 2023-05-22
chore(service naming): multiple contribs

tabgok:
== Special Note ==
These changes are gated behind a feature flag and are not expected to be
stable until a release note is generated.  Existing customers should be
unaffected.

== Service Naming Summary ==

Service naming introduces a few concepts:

  * v0: The existing mechanism for determining service names
  * v1: Replacing the 'default' contrib-specific service name with the env var $DD_SERVICE

These changes are being in conjunction with changes to peer.service updates to bring a better service-naming experience to customers. In particular, these changes will help:

* Remove the creation of fake services, such as grpc.client * Standardize service names across and within tracers * Remove the need to create as many fake services to take advantage of Datadog features, such as metrics * Make more sense to humans

== Operation Naming Summary == The service name changes mean that generated spans can no longer rely on `span.service` to differentiate between processes. A subclass of spans now have a new problem - differentiating between inbound and outbound requests (for the purposes of metrics). For instance, spans with names `http.request` used to be differentiated by the service they belonged to (incoming requests under `span.service=$DD_SERVICE` and outgoing under `span.service=$DD_CONTRIB_SERVICE_NAME`). These are now no longer differentiated.

To remedy these issues, operation names are also being changed as part of the service naming effort. In the above example, `http.request` turns into `http.server.request` (incoming) and `http.client.request` (outgoing).

== This Change == This change updates the following contribs:
* boto
* grpc
* yaaredis
* starlette
* sqlalchemy
* snowflake
* rq
* pyodbc
* pynamodb
* mariadb
* mako
* kombu
* kafka
* gunicorn
* graphql
* flask_cache
* fastapi
* dogpile
* consul
* aredis
* aioredis
* aiobotocore
* botocore

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/contrib/aiobotocore/patch.py
```diff
@@ -118,7 +120,13 @@ async def _wrapped_api_call(original_func, instance, args, kwargs):
     endpoint_name = deep_getattr(instance, "_endpoint._endpoint_prefix")
 
     service = pin.service if pin.service != "aws" else "{}.{}".format(pin.service, endpoint_name)
-    with pin.tracer.trace("{}.command".format(endpoint_name), service=service, span_type=SpanTypes.HTTP) as span:
+    with pin.tracer.trace(
+        schematize_cloud_api_operation(
```
tabgok, line 124: this is sub-par compared to what happens in botocore, but was discussed with the service name group.

tests/snapshots/tests.contrib.kafka.test_kafka.test_service_override.json
```diff
@@ -15,7 +15,7 @@
       "kafka.message_key": "test_key",
       "kafka.received_message": "True",
       "kafka.tombstone": "False",
-      "kafka.topic": "test_topic",
```
tabgok, line 18: this, and related changes, are to fix a kafka testing bug related to subprocesses, which required starting to use per-test topics

tests/contrib/kafka/test_kafka.py
```diff
@@ -29,14 +34,28 @@ def process_trace(self, trace):
...
+            and trace[0].get_tag("kafka.received_message") == "False"
             else trace
         )
 
 
 dd_tracer.configure(settings={"FILTERS": [KafkaConsumerPollFilter()]})
 
 
+@pytest.fixture()
+def kafka_topic(request):
```
tabgok, line 47: The kafka tests needed to be changed quite a bit to support running with different subprocesses.  Prior to these changes, the subprocess code would flakily segfault due to race conditions with creating subscriptions (confluent_kafka will segfault on some machines if you try to consume from a non-existent topic).

tests/contrib/sqlalchemy/mixins.py
```diff
@@ -24,27 +24,7 @@ class Player(Base):
     name = Column(String(20))
 
 
-class SQLAlchemyTestMixin(object):
```
tabgok, line 27: this isn't removed, its just split into a "Base" (setup/teardown/etc without tests) and a Test Mixin (which has tests) in order to support subclassing of the fomer without the latter

ddtrace/contrib/kafka/patch.py
```diff
@@ -23,7 +25,9 @@
 
 config._add(
     "kafka",
-    dict(_default_service="kafka"),
+    dict(
+        _default_service=schematize_service_name("kafka"),
```
mabdinur, line 29: Can move this logic to IntegrationConfig.service. Again I am worried about maintainbility. We should try to make this change in the fewest places possible or we can end up some integrations with inconsistent service naming.

mabdinur (commented):
Overall this change looks good to me. I am just worried about maintainability. Instead of updating each integration to use a specific helper can we move some of the logic to Pin.__init__() or tracer._start_span(...) or create a new `tracer._start_integration_span(...)`. That way all future integration spans will have to either opt in or out of this functionality.

ddtrace/contrib/grpc/aio_server_interceptor.py
```diff
@@ -181,7 +182,7 @@ def _create_span(pin, handler_call_details, method_kind):
     trace_utils.activate_distributed_headers(tracer, int_config=config.grpc_aio_server, request_headers=headers)
 
     span = tracer.trace(
-        "grpc",
+        schematize_url_operation("grpc", protocol="grpc", direction="inbound"),
         span_type=SpanTypes.GRPC,
         service=trace_utils.int_service(pin, config.grpc_aio_server),
```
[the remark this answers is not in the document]
tabgok: This is part of the wonkiness of the current implementation; services which use the `ext_service` method for service name are automatically updated with a schematized `_default_service_name` when the IntegrationConfig is added.

In this case, the service name is schematized in the contrib's patch.py with this line:

```
default_service=schematize_service_name(constants.GRPC_SERVICE_SERVER),
```

ddtrace/contrib/grpc/server_interceptor.py
```diff
@@ -70,7 +71,7 @@ def _fn(self, method_kind, behavior, args, kwargs):
         trace_utils.activate_distributed_headers(tracer, int_config=config.grpc_server, request_headers=headers)
 
         span = tracer.trace(
-            "grpc",
+            schematize_url_operation("grpc", protocol="grpc", direction="inbound"),
             span_type=SpanTypes.GRPC,
             service=trace_utils.int_service(self._pin, config.grpc_server),
```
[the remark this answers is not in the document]
tabgok: see above comment about patch.py, IntegrationConfig, and _default_service_name

ddtrace/contrib/boto/patch.py
```diff
@@ -85,8 +87,10 @@ def patched_query_request(original_func, instance, args, kwargs):
     endpoint_name = getattr(instance, "host").split(".")[0]
 
     with pin.tracer.trace(
-        "{}.command".format(endpoint_name),
-        service="{}.{}".format(pin.service, endpoint_name),
+        schematize_cloud_api_operation(
+            "{}.command".format(endpoint_name), cloud_provider="aws", cloud_service=endpoint_name
+        ),
+        service=schematize_service_name("{}.{}".format(pin.service, endpoint_name)),
```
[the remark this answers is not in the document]
tabgok: We are schematizing the default service names for integrations here.  In othe words, manual spans and manually specified overrides still take precedence.

The priority order after this change is (generally, not in all cases):

1. Manual service name
2. DD_INTEGRATION_SERVICE env var
3. DD_INTEGRATION_SERVICE_NAME env var
4. DD_SERVICE
5. "unnamed-python-service"

If any level evaluates to "False" in an assert statement, then we go down one.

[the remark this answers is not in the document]
tabgok: > It might be easier if we had one helper in tracer._start_span() that sets the expected service and operation names. 

This is difficult with the current architecture, as every config determines the service name separately, and we want to only schematize a span if its using the default values.  By the time the span is initialized, we've lost all the context about whether or not the passed in service name is a default or if its been provided/overidden by a user.
 
> We will be changes the core api but it will be easier to maintain. I can see contributors creating integrations and forgetting to use these helpers.

I wholeheartedly agree with this, and this is actually a problem across all tracers.  I'd like to decouple this effort with a refactoring effort.  With that said, I am writing this down as a problem

ddtrace/contrib/consul/patch.py
```diff
@@ -24,7 +26,7 @@ def patch():
         return
     setattr(consul, "__datadog_patch", True)
 
-    pin = Pin(service=consulx.SERVICE)
+    pin = Pin(service=schematize_service_name(consulx.SERVICE))
```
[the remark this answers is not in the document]
tabgok: See my earlier comment about wanting to change *only* default values; we know at this point its a default value we're providing, but we lose that information on the other side of the Pin() call

ddtrace/contrib/kafka/patch.py
tabgok: This will work for integrations that use the `ext_service` function, but only a subset of integrations use that, or have this variable defined. For instance, HTTP Servers use int_service and several DB tie-ins (redis, mariadb, etc) use Pin overrides on top of the dbapi module?

If I were to take this course of action, the `config.py:Config` code would look something like:

```
def _add(self, integration, settings, merge=True):
    existing = getattr(self, integration)
    settings = deepcopy(settings)
    for k, v in settings:
         if k == "_default_service":
            settings[k] = schematize_service_name(v)
```

ddtrace/contrib/botocore/patch.py
```diff
@@ -375,18 +384,43 @@ def patched_api_call(original_func, instance, args, kwargs):
                 try:
                     if endpoint_name == "lambda" and operation == "Invoke":
                         inject_trace_to_client_context(params, span)
+                        span.name = schematize_cloud_faas_operation(
+                            trace_operation, cloud_provider="aws", cloud_service="lambda"
+                        )
                     if endpoint_name == "sqs" and operation == "SendMessage":
                         inject_trace_to_sqs_or_sns_message(params, span, endpoint_name)
+                        span.name = schematize_cloud_messaging_operation(
+                            trace_operation, cloud_provider="aws", cloud_service="sqs", direction="outbound"
```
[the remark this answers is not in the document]
tabgok: Will make this an enum

ddtrace/contrib/kafka/patch.py
mabdinur: Yup, I believe we have a long term goal of consolidating these mechanisms. Ideally there should be one api that can create and configuration a span for an integration. We can defer this work for now.

mabdinur (approved):
Thanks for sticking with this 🥳

PR #5943, opened 2023-05-24
Report target host in HTTP clients

jbertran:
## Motivation

For upcoming computation of peer service, we want to have some precursor tags available to all client/producer spans. This is part of that change. 

## Change

Surface http host (i.e. socket information without port) as well as possible given the information available from the traced libraries and report it in the `out.host` span tag.

## Checklist

## Reviewer Checklist

- [ ] Title is accurate.


emmettbutler (dismissed):

ddtrace/contrib/aiohttp/patch.py
```diff
@@ -64,6 +64,19 @@ async def _create_connection(self, req, *args, **kwargs):
             return result
 
 
+def extract_info_from_url(url):
```
jbertran, line 67: This was a rabbit hole of `urlparse` behaviour. Parsing "twice" is the best approach I could come up with, the second parse forces the URL to be absolute so that `netloc` is in the correct position and there's no ambiguity on where the rest of the URL goes in the urlparse 6-tuple - otherwise things go to `scheme` or `path` equivalently with no way of determining which of the 2 is the relevant item.

emmettbutler (approved):
Nice work. The urlparse behavior you dug into is pretty confusing.

ddtrace/contrib/requests/connection.py
```diff
@@ -70,6 +70,7 @@ def _wrap_send(func, instance, args, kwargs):
 
     url = request.url
     hostname = _extract_hostname(url)
+    host_without_port = hostname.split(":")[0] if hostname is not None else None
```
tabgok, line 73: are we guaranteed a port?

jbertran: We're not, but we're guaranteed that if there is one, then `urlparse` puts it in `netloc` (at least, it does if it's an absolute URL, which we make sure of).

PR #5944, opened 2023-05-24
ci(debugging): use full PYTHONPATH in subprocess tests

P403n1x87:
Use the full PYTHONPATH, including the component from the parent process, to ensure that all required modules can be found.

## Checklist

## Reviewer Checklist

- [ ] Title is accurate.

### DIFF
--- a/tests/debugging/exploration/test_bootstrap.py
+++ b/tests/debugging/exploration/test_bootstrap.py
@@ -1,3 +1,4 @@
+import os
 from os.path import dirname
 
 import pytest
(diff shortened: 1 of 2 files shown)


tests/debugging/exploration/test_bootstrap.py
```diff
@@ -44,8 +45,13 @@
 No lines found
 """
 
+if "PYTHONPATH" in os.environ:
```
avara1986, line 48: It looks like this function:
https://github.com/DataDog/dd-trace-py/blob/1.x/tests/internal/test_module.py#L16

Could we use that function in this test to avoid duplicate code? WDYT?

P403n1x87: because we need to pass `__file__` we would have to reimplement that function in this module unfortunately. Perhaps we can do a follow-up where we make a more generic helper?

PR #5945, opened 2023-05-24
docs(tracing): adds DD_TRACE_PARTIAL_FLUSH_MIN_SPANS to configurations

mabdinur:
This PR ensures ``DD_TRACE_PARTIAL_FLUSH_ENABLED`` is documented. 

Partial flushing is an implementation detail of the tracer so I don't think we need to document this configuration in https://github.com/DataDog/documentation (this justification is pretty hand wavy, feel free to push back).

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


docs/configuration.rst
```diff
@@ -368,6 +368,11 @@ The following environment variables for the tracer are supported:
      default: True
      description: Prevents large payloads being sent to APM.
 
+    DD_TRACE_PARTIAL_FLUSH_MIN_SPANS:
+     type: String
```
ZStriker19, line 372: shouldn't type be `int` ?

mabdinur, line 372:
```suggestion
     type: Integer
```

docs/configuration.rst
```diff
@@ -368,6 +368,11 @@ The following environment variables for the tracer are supported:
...
+     default: 500
+     description: Maximum number of spans sent per trace per payload when ``DD_TRACE_PARTIAL_FLUSH_ENABLED=True``
```
mabdinur, line 374:
```suggestion
     description: Maximum number of spans sent per trace per payload when ``DD_TRACE_PARTIAL_FLUSH_ENABLED=True``.
```

ZStriker19:
Should the title be for ` DD_TRACE_PARTIAL_FLUSH_MIN_SPANS` ? I think it's ok to not add this to Datadog documentation, as this seems to follow with what other libraries have done.

ZStriker19 (dismissed):

docs/configuration.rst
```diff
@@ -368,6 +368,11 @@ The following environment variables for the tracer are supported:
      default: True
      description: Prevents large payloads being sent to APM.
 
+    DD_TRACE_PARTIAL_FLUSH_MIN_SPANS:
```
mabdinur, line 371:
```suggestion
   DD_TRACE_PARTIAL_FLUSH_MIN_SPANS:
```

majorgreys:
What is the motivation to add this to our documentation? While we document `DD_TRACE_PARTIAL_FLUSH_ENABLED`, we do communicate when/how to use this feature. 

I also think we should be cautious in documenting configuration options if that impacts our versioning policy. For example, if we were to rewrite the partial flushing with different configuration options, we would have to continue supporting `DD_TRACE_PARTIAL_FLUSH_MIN_SPANS` if it was documented.

mabdinur:
> What is the motivation to add this to our documentation? While we document `DD_TRACE_PARTIAL_FLUSH_ENABLED`, we do communicate when/how to use this feature.
> 
> I also think we should be cautious in documenting configuration options if that impacts our versioning policy. For example, if we were to rewrite the partial flushing with different configuration options, we would have to continue supporting `DD_TRACE_PARTIAL_FLUSH_MIN_SPANS` if it was documented.


A user reached out to our support team about this configuration. They wanted to set this configuration to resolve an issue but were hesitant to set it since it was undocumented. In the past we encouraged users to set this configuration so I think it should be documented. 

I interpreted our version policy for configurations as:
-  All configurations prefixed with an underscore (ex: ``_DD``) are internal. 
- All configurations without the `_` prefix is "public" and can/should be documented.

ZStriker19 (approved):

PR #5946, opened 2023-05-25
chore(tracing): surface rpc.service in grpc contrib

jbertran:
## Motivation

For upcoming computation of peer service, we want to have the `rpc.service` precursor tag available in GRPC client spans. This tag is the combination of the GRPC package name and the GRPC service name (for example, when emitting a span for a call to the full path `/myPackage.myService/myMethod`, the value of `rpc.service` is `myPackage.myService`).
The tag name is defined in the separate effort for consistent span tags across tracers.

## Change

Surface `rpc.service` in GRPC contrib. The package name may be missing in some cases, and in these cases the value of this tag is _only_ the GRPC service name.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


tests/snapshots/tests.contrib.grpc.test_grpc.test_method_service.json
```diff
@@ -19,6 +19,7 @@
       "grpc.status.code": "StatusCode.OK",
       "language": "python",
       "network.destination.port": "58143",
+      "rpc.service": "Servicer",
```
ZStriker19, line 22: could you explain Servicer vs pkg.Servicer and how we gain from these distinct values?

jbertran: This test looks at the trace produced by 2 rpc methods calls: `/Servicer/Handler/` and `/pkg.Servicer/Handler/`.

The span modified here relates to the first call, so we don't have package information, which is expected - there's no duplication of purpose here.

Re: `grpc.method.service` versus `rpc.service`, the purpose of `rpc.service` is to contain package (if available) _and_ service, such that the tag provides a relevant unit of service that can be used by the service map. The end goal is for the new tag to be a precursor of the upcoming `peer.service` tag, which will become the new source of truth for the service map.

tabgok (approved):

tests/snapshots/tests.contrib.grpc.test_grpc.test_method_service.json
ZStriker19: Ah got it, took a slight mental shift. Thanks for the explanation!

releasenotes/notes/report-rpc.service-tag-e6363657f2ccbd5d.yaml
```diff
@@ -0,0 +1,29 @@
+---
+#instructions:
```
ZStriker19, line 2: Let's delete all the commented code here

tabgok (approved):

releasenotes/notes/report-rpc.service-tag-e6363657f2ccbd5d.yaml
jbertran: Done

tabgok (approved):

PR #5968, opened 2023-05-26
feat: use upx to save space on Linux (x86_64/aarch64)

sanchda:
upx is a procedure that uses parts of the ELF standard to implement self-decompressing binaries.  This has the effect of reducing the footprint of native code by a significant factor, with very little overhead (nominally, link-time decompression).

In my tests, I was seeing reduction on the order of 40-60%.

I'm using the `--lzma` parameter, which I've found to be safe and efficient.  This can be tweaked later, especially if some sofiles show better compression in other modes.

Errors should generally arise at link-time, so our current test platform will automatically handle this.

Potential issues:
* May make segfaults a little bit harder to diagnose.  We can recommend using upx to decompress sofiles in order to produce more transparent binaries, or could reverse PC values offline some other way.
* Will add some small amount of link-time overhead.  In my tests it was a fraction of overall startup (runtime + parsing + processing), but if not this should be evident as a regression.  The worst-case scenario is `os.spawn()`, where things will get linked all over again.
* May cause execution to fail on Linux systems where SELinux has disabled the ability for a mapped region to have its executable/writable bits toggled.  I do not believe a user with such a strict SELinux policy will be able to run ddtrace in the first place, but it's worth calling out.

**note**: this functionality has been reduced to only affect the non-cython bundled .so files.  This will have a bigger impact when libdatadog is present (proposed in a parallel PR). 

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


Kyle-Verhoog (dismissed):
🙌 this is great and will help our Kubernetes library injection images big time. Also will mean that we don't have to nag pypi to include our storage limit so often.

Given the risk of the change my suggestion would be to have an env var which can opt-out of the upx feature if it breaks the build.

I think the benefits outweigh the risks here, thanks for outlining them! 👏

majorgreys (dismissed):

mabdinur (commented):
This is incredible. Thanks for this proposal!

Question: UPX appears to support to macos and windows executables. Is there reason why this change is limited to linux platforms? Can we expect similar gains package size improvements for non linux platforms?

setup.py
```diff
@@ -39,6 +40,8 @@
 
 DEBUG_COMPILE = "DD_COMPILE_DEBUG" in os.environ
 
+SKIP_UPX = "DD_SKIP_UPX" in os.environ
```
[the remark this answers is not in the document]
sanchda: > Question: UPX appears to support to macos and windows executables. Is there reason why this change is limited to linux platforms? Can we expect similar gains package size improvements for non linux platforms?

Conceptually the other platforms work, but I preferred to submit an improvement for Linux and see how it does in relenv before investing the time to sanity-check the other platforms.

As far as I can tell, Windows support is pretty good, but I haven't tested on that platform.

macos doesn't have an official release and Apple has introduced several schemes for "protecting execution" between osx 12-13 that may be related to [outstanding issues](https://github.com/upx/upx/issues/612).  Things may work just fine for us, but I'm only able to test on a limited number of configurations and the result seems highly dependent on the version of the distribution and host-level security configuration.

setup.py
```diff
@@ -105,6 +108,90 @@ def run(self):
         CleanCommand.run(self)
 
 
+class BuildExtWithUpx(BuildExtCommand):
+    upx_addresses = {
+        "x86_64": "https://github.com/upx/upx/releases/download/v4.0.2/upx-4.0.2-amd64_linux.tar.xz",
```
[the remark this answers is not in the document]
sanchda: > why are we pinning upx to the latest release
It's the best known release

> Does this release have known issues we should be aware of?
For x86_64/aarch64 Linux, not that I'm aware.  The issues I've read for other platforms seem related to the implementation of binary objects on those platforms, or other OS features (such as execution protection).

> Should we make this version configurable?
Indeed.  We should also check the download against a static checksum to prevent man-in-the-middle attacks.  I propose doing this after #5940 has been accepted and this patch has had a chance to get through relenv.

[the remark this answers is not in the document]
sanchda: #5940 was merged, so I'll make this configurable as part of this PR.

sanchda:
> Nice work, there are a lot of CI failures that need to be resolved though :(

I misspoke in my earlier comment.  There QA parts of the CI pipeline that have to deal with the integrity of the wheels we ship.  One of those steps requires doing a static analysis of the ELF objects.

Might be easy to work around this, but just wanted to clarify that I was incorrect in my statement about being "gold."

setup.py
```diff
@@ -53,6 +54,11 @@
 
 LIBDATADOG_PROF_VERSION = "v2.1.0"
 
+UPX_VERSION = "v4.0.2"
+
+# Disable UPX compression when specified
+UPX_ENABLE = os.getenv("DD_UPX_ENABLE", "yes").lower() in ["false", "no"] and not DEBUG_COMPILE
```
P403n1x87, line 60:
```suggestion
UPX_ENABLE = os.getenv("DD_UPX_ENABLE", "yes").lower() in {"false", "no", "0"} and not DEBUG_COMPILE
```
?

setup.py
```diff
@@ -113,6 +119,42 @@ def is_64_bit_python():
...
+        #                sofiles, so deploying upx early is a cause for failure.  Need
+        #                to align on a strategy.
+        sofiles = glob.glob("**/*.so", recursive=True)
+        sofiles = [s for s in sofiles if "cpython" not in s]
+
+        for so in sofiles:
+            try:
+                # -f because some sofiles may be non-exec at this point
+                subprocess.check_call([BuildExtWithUPX.upx, "--lzma", "-f", "-q", so])
+                print("Compressed " + so + " with upx")
```
P403n1x87, line 148:
```suggestion
                print("Compressed %s with upx" % so)
```

setup.py
```diff
@@ -113,6 +119,42 @@ def is_64_bit_python():
...
+            except subprocess.CalledProcessError:
+                print("Failed to compress " + so + " with upx")
```
P403n1x87, line 150:
```suggestion
                print("Failed to compress %s with upx" % so)
```

setup.py
```diff
@@ -122,12 +164,40 @@ class LibraryDownload:
...
+    @classmethod
+    def untar_gz(cls, filename, suffixes, archive_dir, arch_dir):
+        # Open the tarfile first to get the files needed.
+        # This could be solved with "r:gz" mode, that allows random access
+        # but that approach does not work on Windows
+        with tarfile.open(filename, "r|gz", errorlevel=2) as tar:
+            dynfiles = [c for c in tar.getmembers() if c.name.endswith(suffixes)]
+
+        with tarfile.open(filename, "r|gz", errorlevel=2) as tar:
+            print("extracting files:", [c.name for c in dynfiles])
```
P403n1x87, line 176: just checking that we want to keep these print statements here

P403n1x87 (commented):
Some nits, otherwise LGTM!

setup.py
P403n1x87: Also checking that this is indeed the logic we want to implement. The default is `yes` but then we are assigning `True` to `UPX_ENABLE` if `DD_UPX_ENABLE` is falsy.

majorgreys:
@sanchda Is this a change you still want to push forward?

sanchda:
> @sanchda Is this a change you still want to push forward?

Yes, but not in this form.  I need to make some serious changes for this to be acceptable.  Closing!

PR #5969, opened 2023-05-26
feat(profiling): provide libdatadog as an export method

sanchda:
Builds off of #5967.  Operationally, it might also depend on #5968 because bringing libdatadog as a dependency costs about ~3 megabytes of storage, whereas #5968 reduces overhead enough to account for this (and more!).

This adds support for using libdatadog as an optional backend for the profiler. This decreases memory and increases performance, but it isn't 1-1 complete with the Python profiler yet.

To enable libdatadog mode
DD_PROFILING_EXPORT_LIBDD_ENABLE=true

To disable the Python profiler
DD_PROFILING_EXPORT_PY_ENABLE=false

It is possible for both exporters to be active and when that happens libdd annotates its profile version string accordingly (this is useful for comparison).

The expectation is that One Day libdd will replace the py implementation entirely and we can remove defunct code, but before then we have a significant testing gap to overcome.  I've been using "dual mode" to ensure libdatadog has identical output to py, which is fine as far as manual workflows go, but since libdatadog doesn't provide the degree of introspection the current collector does, we'll take some effort to use a comparable mechanism to reimplement correctness tests.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


emmettbutler (dismissed):

ddtrace/profiling/collector/memalloc.py
```diff
@@ -100,48 +103,87 @@ def _get_thread_id_ignore_set(self):
...
+                ddup.push_heap(size)
+                ddup.push_threadinfo(
+                    thread_id, _threading.get_thread_native_id(thread_id), _threading.get_thread_name(thread_id)
                 )
-                for (stack, nframes, thread_id), size in _memalloc.heap()
-                if not self.ignore_profiler or thread_id not in thread_id_ignore_set
-            ),
-        )
+                ddup.push_class_name(frames[0][3])
+                for frame in frames:
```
P403n1x87, line 120: can we unpack the `frame` collection to make this a bit more readable?

ddtrace/profiling/collector/stack.pyx
```diff
@@ -325,8 +339,22 @@ cdef stack_collect(ignore_profiler, thread_time, max_nframes, interval, wall_tim
             if task_pyframes is None:
                 continue
 
+            if task_id in thread_id_ignore_list:
```
P403n1x87, line 342: What's this for?

ddtrace/profiling/collector/memalloc.py
sanchda: Good call--did the `namedtuple` thing and now these frames look like:
```
  ddup.push_frame(frame.function_name, frame.file_name, 0, frame.lineno)
```

ddtrace/internal/datadog/profiling/ddup.pyx
```diff
@@ -60,31 +60,30 @@ IF UNAME_SYSNAME == "Linux" and UNAME_MACHINE == "x86_64":
...
             env: Optional[str],
             version: Optional[str],
             tags: Optional[typing.Dict[str, str]],
             max_nframes: Optional[int],
             url: Optional[str]) -> None:
 
         # Try to provide a ddtrace-specific default service if one is not given
-        if not service:
-            service = DEFAULT_SERVICE_NAME
+        service = service if service else DEFAULT_SERVICE_NAME
```
P403n1x87, line 71:
```suggestion
        service = service or DEFAULT_SERVICE_NAME
```

P403n1x87 (dismissed):
Just a nit, otherwise LGTM!

ddtrace/internal/datadog/profiling/ddup.pyx
sanchda: Good catch, `x if x` doesn't add any explicitness/precision over the short-circuiting or.  Taken in bd2698b2143261a51d170b55420861dfe5ee3f36

P403n1x87 (approved):

brettlangdon (approved):

PR #6000, opened 2023-06-02
chore(tracer): ensure SpanAggregator runs at the end

P403n1x87:
We defer the SpanAggregator until the very end of the sequence of span processors since this is a special processor that needs to run last.

## Checklist

## Reviewer Checklist

- [ ] Title is accurate.


ddtrace/tracer.py
```diff
@@ -176,15 +176,16 @@ def _default_span_processors_factory(
     if single_span_sampling_rules:
         span_processors.append(SpanSamplingProcessor(single_span_sampling_rules))
 
-    span_processors.append(
+    # These need to run after all the other processors
+    deferred_processors = [
```
mabdinur, line 180: Are we planning to add to the deferred processors list? If not, I think we should avoid adding the span aggregator to a list. The indirection makes this code a bit harder to read.

mabdinur (commented):
Question: Are we currently setting `SpanProcessor.__processors__`? If so, this change is fix and requires a regression test.

emmettbutler (dismissed):

ddtrace/tracer.py
P403n1x87: It's hard to predict what will come in the future. I think it is conceivable that there might be processors that need to run _after_ any other registered processors, so the list gives us a bit of "future-proofness"

P403n1x87:
> Question: Are we currently setting `SpanProcessor.__processors__`? If so, this change is fix and requires a regression test.

We currently don't have products registering their own processors.

mabdinur (approved):

emmettbutler (approved):

PR #6002, opened 2023-06-02
[dbapi] Pass on `exc_...` params to wrapped `__aexit__` calls.

### LINKED ISSUE #6001: psycopg3 integration is broken due to improper implementation of `__aexit__`
### Summary of problem

psycopg3's AsyncConnection requires that the `exc_type`, `exc_val`, and `exc_tb` values be passed:

https://github.com/psycopg/psycopg/blob/580ea8c57b2e57ab1e7e6392e130348ac4d9b146/psycopg/psycopg/connection_async.py#L143-L148

However, the dbapi implementation in this library calls them without passing in those parameters:

https://github.com/DataDog/dd-trace-py/blob/5861a61b063f8cb7f04bdeaa2c9e8141bc25fcd8/ddtrace/contrib/dbapi_async/__init__.py#L220-L224

Not only is this broken, this is likely erasing the error-handling present in libraries where this may work.

v1.14.0

update:

I have verified that v1.13.0-2 are unusable due to #5821 and this bug was present starting with v1.13.3. 

I am able to successfully enter and exit an async context in v1.12.9, however it looks like there is no async support for psycopg in this version, so this is effectively the same an having no telemetry.

n/a

This is with the latest version of psycopg, although the problem likely exists with all versions.

```
❯ poetry show psycopg                                                                                         
 name         : psycopg                                
 version      : 3.1.9                                  
 description  : PostgreSQL database adapter for Python 
```

```python

>>> import ddtrace
>>> ddtrace.patch_all()
>>> import psycopg
>>> async with await psycopg.AsyncConnection.connect("dbname=postgres user=postgres host=127.0.0.1") as conn:
...     await conn.execute("select 1")
...     
  File "/Users/seanstewart/Library/Application Support/JetBrains/Toolbox/apps/PyCharm-P/ch-0/223.8836.34/PyCharm.app/Contents/plugins/python/helpers/pydev/pydevconsole.py", line 367, in runcode
    loop.run_until_complete(coro)
  File "/Users/seanstewart/.pyenv/versions/3.11.0/lib/python3.11/asyncio/base_events.py", line 650, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "<input>", line 1, in <module>
  File "/Users/seanstewart/PycharmProjects/payments/.venv/lib/python3.11/site-packages/ddtrace/contrib/dbapi_async/__init__.py", line 224, in __aexit__
    return await self.__wrapped__.__aexit__()
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: AsyncConnection.__aexit__() missing 3 required positional arguments: 'exc_type', 'exc_val', and 'exc_tb'
```

When tracing is enabled, all successful connections fail on exit from the context.

We should be able to make use of async contexts.

seandstewart:
## Changes

Updates the dbapi_async implementation to pass along the `exc_...` parameters to the `__wrapped__.__aexit__` calls.

This should not introduce any breaking changes or cause any changes to performance.

This resolves #6001 

## Checklist

- [] Risk is outlined (performance impact, potential for breakage, maintainability, etc).

## Reviewer Checklist

- [ ] Title is accurate.

### DIFF
--- a/tests/contrib/psycopg/test_psycopg_async.py
+++ b/tests/contrib/psycopg/test_psycopg_async.py
@@ -320,6 +320,18 @@ class PsycopgCore(AsyncioTestCase):
         assert rows[0][0] == "one"
 
     @mark_asyncio
+    async def test_connection_context_execute(self):
+        """Checks whether connection context manager works as normal."""
+
+        query = SQL("""select 'one' as x""")
+        async with (await psycopg.AsyncConnection.connect(**POSTGRES_CONFIG)) as conn:
+            cur = await conn.execute(query)
+            rows = await cur.fetchall()
+
+        assert len(rows) == 1, rows
+        assert rows[0][0] == "one"
+
+    @mark_asyncio
     async def test_cursor_from_connection_shortcut(self):
         """Checks whether connection execute shortcute method works as normal"""
 
--- a/ddtrace/contrib/dbapi_async/__init__.py
+++ b/ddtrace/contrib/dbapi_async/__init__.py
@@ -33,7 +33,7 @@ class TracedAsyncCursor(TracedCursor):
         # previous versions of the dbapi didn't support context managers. let's
         # reference the func that would be called to ensure that error
         # messages will be the same.
-        return await self.__wrapped__.__aexit__()
+        return await self.__wrapped__.__aexit__(exc_type, exc_val, exc_tb)
 
     async def _trace_method(self, method, name, resource, extra_tags, dbm_propagator, *args, **kwargs):
         """
@@ -221,7 +221,7 @@ class TracedAsyncConnection(TracedConnection):
         # previous versions of the dbapi didn't support context managers. let's
         # reference the func that would be called to ensure that errors
         # messages will be the same.
-        return await self.__wrapped__.__aexit__()
+        return await self.__wrapped__.__aexit__(exc_type, exc_val, exc_tb)
 
     async def _trace_method(self, method, name, extra_tags, *args, **kwargs):
         pin = Pin.get_from(self)


releasenotes/notes/fix-dbapi-async-aexit-0df48d739dfd2a3a.yaml
```diff
@@ -0,0 +1,4 @@
+---
+fixes:
+  - |
+    dbapi-async: Pass `exc_type`, `exc_val`, `exc_tb` down to the wrapped object on `__aexit__` calls.
```
mabdinur, line 4: Currently this bug only impacts the psycopg integration.
```suggestion
    psycopg: Resolves ``TypeError`` raised when an async cursor object is traced. This fix ensures `exc_type`, `exc_val`, and `exc_tb` are passed down to the wrapped object on `__aexit__`.
```

tests/contrib/psycopg/test_psycopg_async.py
```diff
@@ -319,6 +319,18 @@ async def test_connection_execute(self):
         assert len(rows) == 1, rows
         assert rows[0][0] == "one"
 
+    @mark_asyncio
+    async def test_connection_context_execute(self):
```
mabdinur, line 323: Can we also add test coverage for ``TracedAsyncCursor.__aexit__()``

mabdinur (commented):
Left a few nits but overall this fix looks great to me. Thanks for this contribution 😄.

@wconti27 can you also review this fix?

releasenotes/notes/fix-dbapi-async-aexit-0df48d739dfd2a3a.yaml
seandstewart: Should note, the change affected the dbapi_async implementation, which is why I scoped the fix with that category, any async dbapi client could be affected by this, theoretically.

tests/contrib/psycopg/test_psycopg_async.py
wconti27: added.

wconti27:
> Left a few nits but overall this fix looks great to me. Thanks for this contribution 😄.
> 
> @wconti27 can you also review this fix?

@mabdinur Made the required changes as they were minimal. Release note updated and test coverage added. LGTM

tests/contrib/psycopg/test_psycopg_async.py
```diff
@@ -319,6 +319,31 @@ async def test_connection_execute(self):
...
+        """Checks whether connection context manager works as normal."""
+
+        query = SQL("""select 'one' as x""")
+        async with (await psycopg.AsyncConnection.connect(**POSTGRES_CONFIG)) as conn:
+            cur = await conn.execute(query)
+            rows = await cur.fetchall()
+
+        assert len(rows) == 1, rows
+        assert rows[0][0] == "one"
```
mabdinur, line 332:
```suggestion
            assert len(rows) == 1, rows
            assert rows[0][0] == "one"
```

tests/contrib/psycopg/test_psycopg_async.py
```diff
@@ -319,6 +319,31 @@ async def test_connection_execute(self):
...
+        """Checks whether cursor context manager works as normal."""
+
+        query = SQL("""select 'one' as x""")
+        async with (await psycopg.AsyncConnection.connect(**POSTGRES_CONFIG)).cursor() as cur:
+            breakpoint()
+            await cur.execute(query)
+            rows = await cur.fetchall()
+
+        assert len(rows) == 1, rows
+        assert rows[0][0] == "one"
```
mabdinur, line 345:
```suggestion
            assert len(rows) == 1, rows
            assert rows[0][0] == "one"
```

mabdinur (approved):

PR #6024, opened 2023-06-05
chore(debugging): relax safe dict retrieval

P403n1x87:
The existing safety check around the retrieval of `__dict__` are too strict and lead to many objects not being captured down to their fields. This change relaxes the condition on safety by ensuring that we can get an instance of `__dict__` that is of the _exact_ type `dict`.

## Risks

The example that was found in the FastAPI is probably quite exotic as we don't expect actual objects to implement a `__dict__` attribute on a class with potential side-effects.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.

### DIFF
--- a/ddtrace/debugging/safety.py
+++ b/ddtrace/debugging/safety.py
@@ -5,10 +5,8 @@ from typing import Any
 from typing import Dict
 from typing import Iterator
 from typing import Tuple
-from typing import Type
 
 from ddtrace.internal.safety import get_slots
-from ddtrace.internal.utils.cache import cached
 
 
 GetSetDescriptor = type(type.__dict__["__dict__"])  # type: ignore[index]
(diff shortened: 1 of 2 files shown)


ddtrace/debugging/safety.py
```diff
@@ -71,9 +71,14 @@ def _has_safe_dict(_type):
 
 def _safe_dict(o):
     # type: (Any) -> Dict[str, Any]
-    if _has_safe_dict(type(o)):
```
shatzi, line 74: does _has_safe_dict is still used? maybe it can removed?

shatzi (approved):

shatzi (dismissed):

ddtrace/debugging/safety.py
P403n1x87: Good point. I thought of leaving just in case, but I guess we can always retrieve it from the history.

PR #6036, opened 2023-06-06
chore(tracing): surface kafka bootstrap.servers in tag

jbertran:
The underlying C library rdkafka, to which confluent_kafka passes config information more or less as-is, accepts 2 keys for bootstrap list [1]:

* `bootstrap.servers`
* `metadata.broker.list`

Since the Producer object does not expose configuration information to Python, we need to grab this on creation. Luckily, confluent_kafka does not expose dynamic host modification logic, so we need not go further.

[1]: https://github.com/confluentinc/librdkafka/blob/v2.1.1/CONFIGURATION.md

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/contrib/kafka/patch.py
```diff
@@ -112,6 +116,7 @@ def traced_produce(func, instance, args, kwargs):
         span.set_tag(kafkax.PARTITION, partition)
         span.set_tag_str(kafkax.TOMBSTONE, str(value is None))
         span.set_tag(SPAN_MEASURED_KEY)
+        span.set_tag_str(kafkax.HOST_LIST, instance._dd_bootstrap_servers)
```
tabgok, line 119: Do we want to set this if the value is empty or None?

ddtrace/contrib/kafka/patch.py
```diff
@@ -33,6 +33,10 @@
 
 
 class TracedProducer(confluent_kafka.Producer):
+    def __init__(self, config, *args, **kwargs):
+        super(TracedProducer, self).__init__(config, *args, **kwargs)
+        self._dd_bootstrap_servers = config.get("bootstrap.servers") or config.get("metadata.broker.list") or None
```
tabgok, line 38: Do we expect all python statements which evaluate to `False` to be remapped to None?  For instance `[]`?

jbertran: Turns out my recollection of crashes of confluent_kafka upon giving it non-string values was unfounded. I'll make this separate and explicitly not reliant on falsey values.

ddtrace/contrib/kafka/patch.py
jbertran: Nope, wrong tracer for this (in NodeJS null tags are stripped) - fixed

tabgok (dismissed):

releasenotes/notes/report-kafka-bootstrap.servers-value-in-tag-8e6793b89e07bb7c.yaml
```diff
@@ -0,0 +1,5 @@
+---
+features:
+  - |
+    tracing: This adds the confluent-kafka producer configuration value found in `metadata.broker.list` or
+    `bootstrap.servers` to producer spans in the `messaging.kafka.bootstrap.servers` tag
```
mabdinur, line 5:
```suggestion
    kafka: Adds the confluent-kafka producer configuration value found in `metadata.broker.list` or
    `bootstrap.servers` to producer spans in the `messaging.kafka.bootstrap.servers` tag.

```

mabdinur (commented):
minor nit. Overall LGTM

tabgok (dismissed):

ddtrace/contrib/kafka/patch.py
```diff
@@ -33,6 +33,14 @@
 
 
 class TracedProducer(confluent_kafka.Producer):
+    def __init__(self, config, *args, **kwargs):
+        super(TracedProducer, self).__init__(config, *args, **kwargs)
+        self._dd_bootstrap_servers = (
```
mabdinur, line 38: nit: can we set the kafkax.HOST_LIST tag here instead of adding the `_dd_bootstrap_servers` attribute?

mabdinur (approved):

ddtrace/contrib/kafka/patch.py
jbertran: Done via `pin`.

amarziali (approved):

ddtrace/contrib/kafka/patch.py
```diff
@@ -33,6 +33,16 @@
 
 
 class TracedProducer(confluent_kafka.Producer):
+    def __init__(self, config, *args, **kwargs):
+        super(TracedProducer, self).__init__(config, *args, **kwargs)
+        bootstrap_servers = (
+            config.get("bootstrap.servers")
+            if config.get("bootstrap.servers") is not None
+            else config.get("metadata.broker.list")
```
pablomartinezbernardo, line 41: Maybe we should add a test for this as well?

tabgok (dismissed):

ddtrace/contrib/kafka/patch.py
```diff
@@ -104,6 +114,7 @@ def traced_produce(func, instance, args, kwargs):
         service=trace_utils.ext_service(pin, config.kafka),
         span_type=SpanTypes.WORKER,
     ) as span:
+        span.set_tags(pin.tags)
```
majorgreys, line 117: If this is only one additional tag (that we know is a string value), we would be better of using `set_tag_str`.

ddtrace/contrib/kafka/patch.py
```diff
@@ -33,6 +33,16 @@
...
 class TracedProducer(confluent_kafka.Producer):
+    def __init__(self, config, *args, **kwargs):
+        super(TracedProducer, self).__init__(config, *args, **kwargs)
+        bootstrap_servers = (
+            config.get("bootstrap.servers")
+            if config.get("bootstrap.servers") is not None
+            else config.get("metadata.broker.list")
+        )
+        if bootstrap_servers is not None:
+            Pin.override(self, tags={kafkax.HOST_LIST: bootstrap_servers})
```
majorgreys, line 44: Would a user change the config after the `TracerProducer` is initialized?

releasenotes/notes/report-kafka-bootstrap.servers-value-in-tag-8e6793b89e07bb7c.yaml
```diff
@@ -0,0 +1,6 @@
...
```
majorgreys, line 5:
```suggestion
    kafka: Adds `messaging.kafka.bootstrap.servers` tag for the confluent-kafka producer configuration value found in `metadata.broker.list` or
    `bootstrap.servers` 
```

tests/snapshots/tests.contrib.kafka.test_kafka.test_schematized_span_service_and_operation[v1-None].json
```diff
@@ -46,12 +46,15 @@
...
       "_dd.p.dm": "-0",
+      "_dd.peer.service.source": "messaging.kafka.bootstrap.servers",
       "component": "kafka",
       "kafka.message_key": "test_key",
       "kafka.tombstone": "False",
       "kafka.topic": "test_schematized_span_service_and_operation_v1-None",
       "language": "python",
+      "messaging.kafka.bootstrap.servers": "localhost:29092",
       "messaging.system": "kafka",
+      "peer.service": "localhost:29092",
```
Comment (majorgreys) [inline tests/snapshots/tests.contrib.kafka.test_kafka.test_schematized_span_service_and_operation[v1-None].json:57]: Is the addition of `peer.service` here expected?

majorgreys (commented):
Few questions

tests/snapshots/tests.contrib.kafka.test_kafka.test_schematized_span_service_and_operation[v1-None].json
Comment (jbertran) [reply tests/snapshots/tests.contrib.kafka.test_kafka.test_schematized_span_service_and_operation[v1-None].json:57]: Yup, `messaging.kafka.bootstrap.servers` is the only precursor available for Kafka, so we report `peer.service` on these spans as soon as we report the precursor.

ddtrace/contrib/kafka/patch.py
majorgreys: The `set_tag_str` is not overloaded like `set_tag` (which `set_tags` calls), so is the more optimal call if you know the value type is always a string.

I think though your use of pin tags is worth considering. I think pins should be used for user overrides of default configurations of an integration. But this change appears to be how to capture information when the patched producer is initialized so that we can add it to the producer span. Or put differently, this pin tag should not be set by a user nor would we want to override the value in the config argument when the producer is initialized.

jbertran: I was trying to address @mabdinur 's [comment](https://github.com/DataDog/dd-trace-py/pull/6036#discussion_r1232593245) - is there another mechanism I can use to set the tag at `TracedProducer.__init__` before the span gets created?

tests/snapshots/tests.contrib.kafka.test_kafka.test_schematized_span_service_and_operation[v1-None].json
Comment (jbertran) [reply tests/snapshots/tests.contrib.kafka.test_kafka.test_schematized_span_service_and_operation[v1-None].json:57]: This is from `ddtrace.internal.processor.trace:PeerServiceProcessor` added in #6042 - a number of precursor tags are defined that cause the `peer.service` tag to be created from the precursor values. The tag we're adding here is kafka's only precursor, which is why `peer.service` is only now reported.

ddtrace/contrib/kafka/patch.py
mabdinur: Yes, sorry about this. We should avoid using `Span.set_tag` and `Span.set_tags` internally. Unfortunalty we already do this in a few places. 

I would prefer reverting to the previous implementation were we stored the bootstrap_servers in a field and then set it using `set_tag_str()`. Sorry about the back and forth.

mabdinur: Aside from that this PR looks good to me

mabdinur (approved):
Left some bits. Overall LGTM

ddtrace/contrib/kafka/patch.py
```diff
@@ -104,6 +112,7 @@ def traced_produce(func, instance, args, kwargs):
         service=trace_utils.ext_service(pin, config.kafka),
         span_type=SpanTypes.WORKER,
     ) as span:
+        span.set_tags(pin.tags)
```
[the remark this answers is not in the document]
jbertran: 🤦  this is my bad, it shouldn't even be there, since we don't have anything in the pin anymore.

tabgok (approved):

mabdinur (approved):

majorgreys (approved):

PR #751, opened 2018-11-30
[core] add six and replace custom compat functions

brettlangdon:
This PR does a few things:

1. Adds `six` as a dependency and aliases our `ddtrace.compat` properties to the appropriate `six` ones
2. Inlines `compat_async.py` into `compat.py`

PR #6037, opened 2023-06-06
fix: avoid importing asyncio

P403n1x87:
Whilst the library has many integrations and features relying on asyncio, the actual products don't make use of it. The interaction is only aimed at providing observability into Python's native async support. We then avoid importing from asyncio and rely only on import hooks to achieve our goals. This prevents the asyncio module from being loaded, which also means that we no longer need to unload it as part of the better support for frameworks such as gevent.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/internal/compat.py
```diff
@@ -184,7 +184,6 @@ def getrandbits(k):
         textwrap.dedent(
             """
     import functools
-    import asyncio
```
P403n1x87, line 187: @brettlangdon I think this code was introduced by you in https://github.com/DataDog/dd-trace-py/pull/751. Do you reckon we should replace this import with something else (e.g. checking that we can find the loader), or can we just do away with it?

brettlangdon: oh, this function used to do `@asyncio.coroutine`, which is does not anymore, so yes it can likely be removed. it was likely just missed before.

avara1986 (dismissed):

emmettbutler (approved):

releasenotes/notes/fix-gevent-asyncio-13ed3a9411ed8088.yaml
```diff
@@ -0,0 +1,6 @@
+---
```
sanchda, line 1: nice

sanchda (approved):
Hooray!  Looks really great!

tests/conftest.py
```diff
@@ -149,7 +149,7 @@ def unwind_params(params):
         yield None
         return
 
-    for _ in product(*(((k, v) for v in vs) for k, vs in params.items())):
+    for _ in product(*([(k, v) for v in vs] for k, vs in params.items())):
```
P403n1x87, line 152: It's probably best to pull this change out so that we can backport it to other release branches independently

emmettbutler (approved):

sanchda (approved):

majorgreys (approved):

PR #6042, opened 2023-06-06
chore(peer.service) - peer.service logic + flag

tabgok:
== Special Notes ==
These changes are gated behind a feature flag and are not expected to be stable until a release note is generated. Existing customers should be unaffected.

The logic relies on 'precursor' tags, which are being added in other PRs. Logic will not be fully enabled until until the precursors are completed.

These changes are reliant on service naming as well, and will not be fully enabled until that initiative is complete.

== Peer.Service Summary ==
The changes for peer.service will enable new functionality in the service map, when enabled, in conjunction with service naming.

The changes in this PR are:

1. Introduce the peer.service computation, gated behind a feature flag
2. Add in the peer.service and peer.service.source computation logic

When peer service is enabled, the logic is applied as a post-processor on spans, as part of the "on_finish" methods
enabled.

## Checklist

## Reviewer Checklist

- [ ] Title is accurate.


ddtrace/internal/peer_service/__init__.py
```diff
@@ -0,0 +1,15 @@
+import functools
```
tabgok, line 1: I created a new module under internal with the assumption that we're trying to move away from a global Config and to per-product/per-feature modules.  Let me know if this should reside somewhere else

ddtrace/internal/peer_service/processor.py
```diff
@@ -0,0 +1,34 @@
...
+        pass
+
+    def on_span_finish(self, span):
+        if not PEER_SERVICE_ENABLED:
+            return
+
+        if span.get_tag(SPAN_KIND) not in SPAN_KINDS_ENABLED_FOR_PEER_SERVICE:
+            return
+
+        if span.get_tag(PEER_SERVICE_TAG):  # If the tag already exists, assume it is user generated
```
tabgok, line 25: @amarziali  - Is the assumption in the comment correct?

ddtrace/tracer.py
```diff
@@ -136,6 +138,7 @@ def _default_span_processors_factory(
     single_span_sampling_rules,  # type: List[SpanSamplingRule]
     agent_url,  # type: str
     profiling_span_processor,  # type: EndpointCallCounterProcessor
+    peer_service_enabled,  # type: bool
```
tabgok, line 141: I see a few patters for including post-processors here, and I wasn't sure if any where preferred, so I just picked one.  Let me know if I should use a different method!

ddtrace/internal/peer_service/__init__.py
mabdinur: This directory feels a bit off to me. "Peer service" is not really a product. It's more a tracing configuration. I think this configuration should be defined with the the rest of the tracing configurations (in `ddtrace.setting.config.py`). We can refactor this module to split up products in the future 🤞

ddtrace/internal/peer_service/processor.py
```diff
@@ -0,0 +1,34 @@
...
+            span.set_tag(PEER_SERVICE_SOURCE_TAG, PEER_SERVICE_TAG)
+            return
+
+        for data_source in PRIORITIZED_PEER_SERVICE_DATA_SOURCES:
```
mabdinur, line 29: Do we need to set this tag on every span? This can have a significant impact on billing (it increases the size of every span).  Is there a way to set these tags on a smaller set of spans (ex: set this tag only on root spans and spans that have a different service name than their parent span.

mabdinur: I guess this is a better recommendation for the RFC. It might be too late to alter the implementation details.

ddtrace/internal/schema/__init__.py
```diff
@@ -21,6 +21,7 @@ def _validate_schema(version):
 _validate_schema(__schema_version)
 
 DEFAULT_SPAN_SERVICE_NAME = _DEFAULT_SPAN_SERVICE_NAMES[__schema_version]
+SCHEMA_VERSION = __schema_version
```
mabdinur, line 24: nit: Instead of defining `__schema_version` can we use a helper function to retrieve and set the schema version

ddtrace/internal/peer_service/processor.py
```diff
@@ -0,0 +1,34 @@
...
+from ddtrace.ext import SpanKind
+from ddtrace.internal.processor import SpanProcessor
+
+from . import PEER_SERVICE_ENABLED
+
+
+PEER_SERVICE_SOURCE_TAG = "_dd.peer.service.source"
+PEER_SERVICE_TAG = "peer.service"
+SPAN_KINDS_ENABLED_FOR_PEER_SERVICE = {SpanKind.CLIENT, SpanKind.PRODUCER}
+PRIORITIZED_PEER_SERVICE_DATA_SOURCES = ["messaging.kafka.bootstrap.servers", "db.name", "rpc.service", "out.host"]
```
mabdinur, line 11: Should we make this configurable (ex: via an environment variable). Would customers want to add to this list?

ddtrace/internal/peer_service/__init__.py
mabdinur: Can we put this feature in the `schema` directory?

ddtrace/tracer.py
mabdinur: A SpanProcessor feels a bit odd here since these tags will be an important part of the span data model but I don't have a better suggestion. Maybe we can move this logic to [tracer.trace(...)](https://github.com/DataDog/dd-trace-py/blob/1.x/ddtrace/tracer.py#L813) but I am not a big fan of adding more logic to that complex method. 

Just rambling I guess. A SpanProcessor works for me as long as there is not a significant performance impact.

ddtrace/tracer.py
```diff
@@ -262,6 +268,7 @@ def __init__(
         self._appsec_processor = None
         self._iast_enabled = config._iast_enabled
         self._endpoint_call_counter_span_processor = EndpointCallCounterProcessor()
+        self._peer_service_enabled = PEER_SERVICE_ENABLED
```
mabdinur, line 271: nit: was can use the local variable

```suggestion
        self._peer_service_enabled = peer_service_enabled
```

ddtrace/internal/peer_service/processor.py
amarziali: It's a good remark. Short answer is yes it will for sure come as a feature later but right now it has to be specified in order to have some behaviour across tracers

ddtrace/internal/peer_service/processor.py
amarziali: yes it is

amarziali (approved):
looks good

ddtrace/internal/peer_service/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+import os
+
+from ddtrace.internal.schema import SCHEMA_VERSION
+from ddtrace.internal.utils.formats import asbool
+
+
+@functools.lru_cache(maxsize=1)
```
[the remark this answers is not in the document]
tabgok: I can go either way, its also low expense to calculate, so I'll remove the cache

ddtrace/internal/peer_service/processor.py
tabgok: for now its an internal-only API that's somewhat subject to change;  I'd recommend against making it configurable until its something we want to provide as an official option

ddtrace/internal/schema/__init__.py
tabgok: We can use one to retrieve, but not to set.

The issue with setting the variable is the imported values are cached all over the place within the tracer, so even if someone (or something) changes the value after modules are loaded, you'll still get the original behavior.

I'll create a helper method for retrieving, though

ddtrace/tracer.py
tabgok: The other mechanism I came up with for this is updating every contrib to add the tags, which is implementation and testing heavy

mabdinur (dismissed):

mabdinur:
Changes look good but there are some failing tests. Once those are addressed we should be good to go!

tabgok:
> This is really dumb but the telemetry test is failing because it asserts an error was raised on line 226 but now the error is raised on a different line. Can we remove the line number and file name from this test: https://github.com/DataDog/dd-trace-py/blob/1.x/tests/telemetry/test_telemetry.py#L202

I added a fix for this to use a regex

ddtrace/internal/processor/trace.py
```diff
@@ -278,3 +279,33 @@ def on_span_finish(self, span):
...
+        pass
+
+    def on_span_finish(self, span):
+        if not self.enabled:
+            return
+
+        if span.get_tag(SPAN_KIND) not in self._config.enabled_span_kinds:
+            return
+
+        if span.get_tag(self._config.tag_name):  # If the tag already exists, assume it is user generated
```
amarziali, line 302: This case (`peer.service` tag is already there so we set the source) should always be set regardless the span kind

tabgok: Swapping the order to make it so

amarziali (approved):

amarziali (approved):

mabdinur (approved):

PR #6048, opened 2023-06-07
perf(telemetry): Improves the speed of queueing metrics

mabdinur:
## Description
 - Optimizes `Metric.get_point()` by ensuring `time.time()` is called only when required and removes unnecessary casting to float
- Removes `TelemetryBase._lock` from queuing telemetry metrics. Acquiring this lock is expensive. 
    - Here we are sacrificing correctness for performance. This is a small edge case where we can undercount metrics by 1 per interval but I think this is acceptable. 
- Updates `Metric.get_id()` to be a classmethod instead of an instance method. This will allow us to generate a metric id without first initializing a metric object 
   - In this PR I added the metric type and namespace to metric id. This is not a perf improvement. I just wanted to sneak in a bug fix hehe.
   -  Removes `validate_type_metric` from MetricNamespace. This is unnecessary operation that adds performance overhead. 
      - With the removal of this validation we have to update a test. Telemetry metrics now supports sending metrics that have the same name but different types. These metrics will be hashed differently and stored in seperate payloads.

## Testing Strategy 
I benchmarked the performance of queueing count metrics against the latest release (v1.14.0). The results and script that was used is shown below:

### Script
```
from ddtrace.internal.telemetry import telemetry_metrics_writer
import cProfile, pstats


# Start telemetry writer before benchmarks to reduce noise
telemetry_metrics_writer.enable()

profiler = cProfile.Profile()
profiler.enable()
for _ in range(10000):
    telemetry_metrics_writer.add_count_metric("tracers", "datadog.span_finished")
    telemetry_metrics_writer.add_count_metric("tracers", "datadog.span_created")
profiler.disable()

stats = pstats.Stats(profiler).sort_stats('cumtime')
stats.print_stats("ddtrace")
``` 

### Summary

- Queuing 20, 000 telemetry count metrics using ddtracev1.4.0 takes 0.157seconds (7.85us). 
- Queuing 20, 000 telemetry count metrics  using this branch takes 0.031 seconds (1.55us). 
- This is about 5x speed up 🥳 

### Results for v1.4.0

```
#          460001 function calls in 0.157 seconds

#    Ordered by: cumulative time
#    List reduced from 24 to 13 due to restriction <'ddtrace'>

#    ncalls  tottime  percall  cumtime  percall filename:lineno(function)
#     20000    0.006    0.000    0.157    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/writer.py:286(add_count_metric)
#     20000    0.020    0.000    0.151    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/writer.py:300(_add_metric)
#     20000    0.032    0.000    0.077    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/metrics_namespaces.py:64(_add_metric)
#     20000    0.005    0.000    0.038    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/writer.py:46(_get_heartbeat_interval_or_default)
#     20000    0.008    0.000    0.018    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/metrics.py:100(__init__)
#     20000    0.009    0.000    0.012    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/metrics.py:104(add_point)
#     20000    0.008    0.000    0.012    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/writer.py:245(enable)
#     40000    0.011    0.000    0.011    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/metrics.py:49(id)
#     20000    0.009    0.000    0.010    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/metrics.py:31(__init__)
#     20000    0.004    0.000    0.004    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/writer.py:155(enable)
#     20000    0.002    0.000    0.002    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/metrics_namespaces.py:57(validate_type_metric)
#     20000    0.002    0.000    0.002    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/metrics.py:65(set_tags)
```

### Results for  this branch:
```
#          120011 function calls in 0.031 seconds

#    Ordered by: cumulative time
#    List reduced from 11 to 7 due to restriction <'ddtrace'>

#    ncalls  tottime  percall  cumtime  percall filename:lineno(function)
#     20000    0.007    0.000    0.031    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/writer.py:301(add_count_metric)
#     20000    0.012    0.000    0.024    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/metrics_namespaces.py:58(_add_metric)
#     20000    0.007    0.000    0.008    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/metrics.py:48(get_id)
#     20000    0.003    0.000    0.003    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/metrics.py:101(add_point)
#         2    0.000    0.000    0.000    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/metrics.py:30(__init__)
#         2    0.000    0.000    0.000    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/metrics.py:66(set_tags)
```

## Checklist

## Reviewer Checklist

- [ ] Title is accurate.


brettlangdon:
@mabdinur for your benchmark/test I would make sure it runs for a period of time a few times the flush interval you are using. I think given how quickly this script runs you are not hitting a case where the writer is flushing all the existing metrics and resetting the namespace (is this phrase accurate?) ?

ddtrace/internal/telemetry/metrics_namespaces.py
```diff
@@ -42,45 +40,36 @@ class MetricNamespace:
...
-            TELEMETRY_TYPE_GENERATE_METRICS: {TELEMETRY_NAMESPACE_TAG_TRACER: {}, TELEMETRY_NAMESPACE_TAG_APPSEC: {}},
-            TELEMETRY_TYPE_DISTRIBUTION: {TELEMETRY_NAMESPACE_TAG_TRACER: {}, TELEMETRY_NAMESPACE_TAG_APPSEC: {}},
-        }
+            TELEMETRY_TYPE_GENERATE_METRICS: defaultdict(dict),
+            TELEMETRY_TYPE_DISTRIBUTION: defaultdict(dict),
+        }  # type: Dict[str, Dict[str, Dict]]
 
     def get(self):
         # type: () -> Dict[str, Any]
         namespace_metrics = self._metrics_data.copy()
```
brettlangdon, line 55: oof this copy is probably fairly heavy, no ?

.... do we ever remove/reset the existing points on a metric?

it seems like gauge and rate metrics will function ok. count metrics will never update it's timestamp, so we will always send the first timestamp and an absolute count ever growing? and distribution metrics are a memory leak?

also distribution is missing type, common, and timestamp in `to_dict`?

it feels like a missed opportunity to have something like:

```python
def reset(self):
    points, self._points = self._points, []
    return points
```

ddtrace/internal/telemetry/metrics_namespaces.py
```diff
@@ -42,45 +40,36 @@ class MetricNamespace:
...
-        metric = self._metrics_data[metrics_type_payload][namespace].get(metric_check.id, metric_check)
-        if not self.validate_type_metric(metric_check, metric):
-            raise TelemetryTypeError(
-                (
-                    'Error: metric with name "%s" and type "%s" exists. You can\'t create a new metric '
-                    'with this name an type "%s"'
-                )
-                % (name, metric.metric_type, metric_check.metric_type)
+
+        existing_metric = self._metrics_data[metrics_type_payload][namespace].get(metric_id)
```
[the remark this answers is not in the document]
mabdinur: We have the following scenarios where we can undercount metrics:
 1. We under count a metric because we update a metric object that has been copied and sent
 2. A metric is queued from different threads at the same time, in this scenario the last value is stored and the first is overwritten.

Since the critical section executes in ~1.4us and we flush metrics every 10 seconds I don't think this is the biggest issue but I will look in alternative approaches. The metrics will not be 100% correct but it should be approximately accurate. I think this is good enough but can probably do better.

ddtrace/internal/telemetry/metrics_namespaces.py
mabdinur: Yup, I am not benchmarking this. I focused on just optimizing queuing metrics because that is a synchronous operation. I plan to get to this next.

ddtrace/internal/telemetry/metrics_namespaces.py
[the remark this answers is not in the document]
brettlangdon: Looking over how the metrics classes are created and track data, I do think we can get a more thread safe implementation... maybe not entirely lock free, but at least we might be able to reduce the places/scope of what needs a lock.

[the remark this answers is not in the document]
brettlangdon: > We under count a metric because we update a metric object that has been copied and sent

we are copying the main dictionary, but all the values are going to be references, so we actually aren't copying the metric classes themselves. we'd need a deep copy for that.


```python
metrics = {}
metrics["a"] = {"b": []}

c = metrics.copy()
c["a"]["b"].append(1)

print(metrics)

# {'a': {'b': [1]}}
```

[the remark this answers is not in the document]
mabdinur: I removed this copy

ddtrace/internal/telemetry/metrics.py
```diff
@@ -37,43 +36,32 @@ def __init__(self, namespace, name, tags, common, interval=None):
...
         """
-        self.name = name
+        self.name = name.lower()
         self.is_common_to_all_tracers = common
         self.interval = interval
-        self.namespace = namespace
-        self._tags = {}  # type: MetricTagType
-        self.set_tags(tags)
+        self.namespace = namespace.lower()
+        self._tags = {k.lower(): str(v).lower() for k, v in tags.items()}
```
brettlangdon, line 43: since we don't allow mutation of tags, we should consider just storing them as the final list of strings.

```python
self._tags = ["%s:%s" % (k, v) for k, v in tags.items()] if tags else []
```

this is a future improvement and not needed at this time.

ddtrace/internal/telemetry/metrics.py
```diff
@@ -37,43 +36,32 @@ def __init__(self, namespace, name, tags, common, interval=None):
...
-    @property
-    def id(self):
+    @classmethod
+    def get_id(cls, name, namespace, tags, metric_type):
+        # type: (str, str, Dict[str, Any], str) -> str
         """
         https://www.datadoghq.com/blog/the-power-of-tagged-metrics/#whats-a-metric-tag
         """
-        return self.name + str(self._tags)
+        str_tags = str(sorted(tags.items())) if tags else ""
```
brettlangdon, line 53: this isn't doing the same thing as `__init__` for `self._tags`, is that going to end up being a problem?

brettlangdon: e.g. `Metric.get_id(...)` won't be the same as `hash(metric)`

not sure I am asking for a change, just calling out the inconsistency, you are better equipped to decide if it is a problem or not.

brettlangdon (commented):
I still have to finish reviewing

ddtrace/internal/telemetry/metrics.py
mabdinur: Thanks for the call out. This a problem. I think we should just avoid lowercasing tags and make sure the caller sets the correct values. I will update this and tests

mabdinur, line 53:
```suggestion
        str_tags = str(sorted(tags.items())).lower() if tags else ""
```

ddtrace/internal/telemetry/metrics.py
mabdinur: Great idea. Idk why I missed this 😓

tests/telemetry/app.py
```diff
@@ -20,7 +20,7 @@ def metrics_view():
         "test_metric",
         1.0,
     )
-    namespace_metrics = telemetry_metrics_writer._namespace.get()
+    namespace_metrics = telemetry_metrics_writer._namespace._metrics_data
```
brettlangdon, line 23: flush?

ddtrace/internal/telemetry/metrics.py
brettlangdon: especially if any of the tags are static, they can be passed in as `key:value` and save the computation to join them

tests/telemetry/app.py
mabdinur: I didn't want to look into this test further. We could flush but I didn't know if this would break down stream assertions. This has the same behavior of get without exposing that method.

emmettbutler (commented):
I'm going to wait to review this until you've updated the benchmarks per Brett's comment. Let me know if you'd like me to review before then!

mabdinur:
> I'm going to wait to review this until you've updated the benchmarks per Brett's comment. Let me know if you'd like me to review before then!

This PR focuses on the performance of queuing telemetry metrics. Brett suggestion was to benchmark queuing, aggregation, and submission of telemetry metrics to the agent.  I am hesistant to update the PR description because it will change the scope of this PR but here some results from running the benchmark above for 2 minutes on this branch and the 1.4 release branch:


## Summary
 - The benchmarks below queued 100,000 metrics over a 2 minute interval. Over this interval, 12 telemetry payloads  aggregated telemetry metrics and sent this data to the agent. Telemetry metrics are aggregated and sent in a background thread so the profiles do not appear below but time spent submitting this data is captured in the total cumulative time .
- `v1.4.0` results: 100,000 metrics were queued over 1.542 seconds (`15.47us per metric`)
- `munir/telemetry-perf-improvements` results: 100,000 metrics were queued over 0.547 seconds (`5.47us per metric`)
-  This change has a `3x performance improvement` for queuing, aggregating and submitting telemetry metrics 🥳 


## v1.4.0

```
       240001 function calls in 120.857 seconds

   Ordered by: cumulative time
   List reduced from 24 to 12 due to restriction <'ddtrace'>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    10000    0.055    0.000    1.542    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/writer.py:286(add_count_metric)
    10000    0.277    0.000    1.487    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/writer.py:300(_add_metric)
    10000    0.208    0.000    0.560    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/metrics_namespaces.py:64(_add_metric)
    10000    0.079    0.000    0.406    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/writer.py:46(_get_heartbeat_interval_or_default)
    10000    0.130    0.000    0.217    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/writer.py:245(enable)
    10000    0.051    0.000    0.124    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/metrics.py:100(__init__)
    10000    0.073    0.000    0.110    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/metrics.py:104(add_point)
    10000    0.087    0.000    0.087    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/writer.py:155(enable)
    20000    0.085    0.000    0.085    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/metrics.py:49(id)
    10000    0.067    0.000    0.074    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/metrics.py:31(__init__)
    10000    0.019    0.000    0.019    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/metrics_namespaces.py:57(validate_type_metric)
    10000    0.007    0.000    0.007    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/metrics.py:65(set_tags)
```

## munir/telemetry-perf-improvements

```
         80073 function calls in 119.115 seconds

   Ordered by: cumulative time
   List reduced from 13 to 6 due to restriction <'ddtrace'>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    10000    0.137    0.000    0.547    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/writer.py:301(add_count_metric)
    10000    0.226    0.000    0.410    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/metrics_namespaces.py:61(add_metric)
    10000    0.081    0.000    0.094    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/metrics.py:47(get_id)
    10000    0.059    0.000    0.059    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/metrics.py:88(add_point)
       12    0.000    0.000    0.000    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/metrics.py:30(__init__)
       12    0.000    0.000    0.000    0.000 /Users/munirabdinur/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/metrics.py:43(<dictcomp>)
```

emmettbutler (approved):
Very nice work. That's an impressive speedup!

avara1986 (approved):
Great Job @mabdinur !!! nice refactor! :heart: :heart: :heart:

PR #6063, opened 2023-06-08
feat(data_streams): add python data streams integration to sqs

shethhriday29:
Add Data Streams integration for Python applications with SQS.
The context is right now not propagated inside an application, only 
between different applications. The DSM team will add context 
propagation inside the application (using trace context) in a follow 
up PR.

I tested it by writing a small Python application, and submitting
traffic to our staging environment.
The code [will be] under a env variable that will act as a feature flag.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/contrib/botocore/patch.py
```diff
@@ -147,6 +150,8 @@ def inject_trace_to_sqs_or_sns_message(params, span, endpoint=None):
     """
     trace_data = {}
     HTTPPropagator.inject(span.context, trace_data)
+    if data_streams_context is not None:
```
tabgok, line 153: nit: it is sufficient to  write `if data_streams_context`

ddtrace/contrib/botocore/patch.py
```diff
@@ -133,10 +134,12 @@ def inject_trace_to_sqs_or_sns_batch_message(params, span, endpoint=None):
     # or PublishBatchRequestEntries (in case of PublishBatch).
     entries = params.get("Entries", params.get("PublishBatchRequestEntries", []))
     for entry in entries:
+        if data_streams_context is not None:
```
tabgok, line 137: nit: sufficient to write `if data_streams_context`

ddtrace/contrib/botocore/patch.py
```diff
@@ -376,9 +381,48 @@ def patched_api_call(original_func, instance, args, kwargs):
                     if endpoint_name == "lambda" and operation == "Invoke":
                         inject_trace_to_client_context(params, span)
                     if endpoint_name == "sqs" and operation == "SendMessage":
-                        inject_trace_to_sqs_or_sns_message(params, span, endpoint_name)
+                        data_streams_context = None
```
tabgok, line 384: I'd do three things here:

1. Put the core logic into another function
2. Mirror the `inject_trace_to_sqs_or_sns` model and have the batch method call ^ for each entry
3. Rename the `inject_trace_data_to_message_attributes` method to something like `inject_datadog_data_to_message_attributes` and use it for both DSM (what you wrote) and APM (what already exists)

This will get rid of a lot of the repeat code and follow the already-existing pattern.

ddtrace/contrib/botocore/patch.py
```diff
@@ -412,10 +456,65 @@ def patched_api_call(original_func, instance, args, kwargs):
...
+                        ctx = pin.tracer.data_streams_processor.decode_pathway_b64(pathway)
+                        ctx.set_checkpoint(
+                            ["direction:in", "topic:" + queue_name, "type:sqs"]
+                        )
+
+                    except:
+                        return result
+
+                if query_status == "NoMessageAttributeNames":
+                    if 'Messages' in result:
```
tabgok, line 494: nit:

```
messages = result.get("Messages", [])
for message in messages:
...
```

ddtrace/contrib/botocore/patch.py
```diff
@@ -412,10 +456,65 @@ def patched_api_call(original_func, instance, args, kwargs):
...
+
+                        ctx = pin.tracer.data_streams_processor.decode_pathway_b64(pathway)
+                        ctx.set_checkpoint(
+                            ["direction:in", "topic:" + queue_name, "type:sqs"]
+                        )
+
+                    except:
+                        return result
+
+                if query_status == "NoMessageAttributeNames":
```
tabgok, line 493: make this a method with a descriptive name

ddtrace/contrib/botocore/patch.py
```diff
@@ -412,10 +456,65 @@ def patched_api_call(original_func, instance, args, kwargs):
...
+                    if 'Messages' in result:
+                        messages = result['Messages']
+                        for message in messages:
+                            if 'MessageAttributes' in message:
+                                del message['MessageAttributes']
+
+                elif query_status == "NoDatadog":
```
tabgok, line 500: make this another method with a descriptive name.  The method name should be a good summary of what and why we're doing this

tabgok:
Notes:

Its still on me to get a recommendation about the MD5 sum and whether or not to delete parameters

Edge cases:
What happens if we hit the 10-attribute limit and don't attach the attributes to the message?  Will/should we still generate a span? If so, will that span be deleted/closed in the backend?

Testing:
* I'd recommend using Moto/OpenStack (if they supports message attributes) (python library for mocking AWS functions) to send messages to/from an SQS queue.  You can use the model in the tests/ directory with the tracer fixture to build snapshots of the generated spans.  You can then perform asserts on those spans.

You should cover the scenarios and edge cases, such as:
* Sending:
  * sending a message with too many attributes
  * Happy Path
* Receiving:
  * no message attributes requested 
  * Happy Path
  * attributes requested, but all there are are datadog attributes
  * whatever the decision is on the md5sum

releasenotes/notes/botocore_dsm_sqs-02df9afa4b4bb503.yaml
```diff
@@ -0,0 +1,28 @@
+---
+#instructions:
```
tabgok, line 2: remove the comments

releasenotes/notes/botocore_dsm_sqs-02df9afa4b4bb503.yaml
```diff
@@ -0,0 +1,28 @@
...
+#    * Include relevant links (bug issues, upstream issues or release notes, documentation pages)
+#    * Use full sentences with sentence-casing and punctuation.
+#    * Before using Datadog specific acronyms/terminology, a release note must first introduce them with a definition.
+#
+#    Release notes should not:
+#    * Be vague. Example: ``fixes an issue in tracing``.
+#    * Use overly technical language
+#    * Use dynamic links (``stable/latest/1.x`` URLs). Instead, use static links (specific version, commit hash) whenever possible so that they don't break in the future.
+features:
+  - Added data streams monitoring support for SQS
```
tabgok, line 28: something like "added message attributes to support datastreams monitoring" may be a little clearer

ddtrace/contrib/botocore/patch.py
```diff
@@ -108,52 +109,78 @@ def inject_trace_data_to_message_attributes(trace_data, entry, endpoint=None):
...
+    datadog_data = {}
+    HTTPPropagator.inject(span.context, datadog_data)
 
     # An entry here is an SNS or SQS record, and depending on how it was published,
     # it could either show up under Entries (in case of PutRecords),
     # or PublishBatchRequestEntries (in case of PublishBatch).
     entries = params.get("Entries", params.get("PublishBatchRequestEntries", []))
     for entry in entries:
-        inject_trace_data_to_message_attributes(trace_data, entry, endpoint)
+        if data_streams_enabled:
```
tabgok, line 159: Reverse the order here (DSM first, APM second), due to the scarcity of "message attributes."  If we hit the MessageAttribute limit, then enabling DSM will break tracing, which seems backwards since DSM is "Beta" for now

ddtrace/contrib/botocore/patch.py
```diff
@@ -108,52 +109,78 @@ def inject_trace_data_to_message_attributes(trace_data, entry, endpoint=None):
...
+            entry["MessageAttributes"]["_datadog"] = {"DataType": "Binary", "BinaryValue": json.dumps(datadog_data)}
         else:
             log.warning("skipping trace injection, endpoint is not SNS or SQS")
     else:
         # In the event a record has 10 or more msg attributes we cannot add our _datadog msg attribute
         log.warning("skipping trace injection, max number (10) of MessageAttributes exceeded")
 
 
-def inject_trace_to_sqs_or_sns_batch_message(params, span, endpoint=None):
+def set_checkpoint_and_get_datadog_data(pin, queue_name):
```
tabgok, line 125: nit: `...get_pathway(...)`

ddtrace/contrib/botocore/patch.py
```diff
@@ -108,52 +109,78 @@ def inject_trace_data_to_message_attributes(trace_data, entry, endpoint=None):
...
     :endpoint: endpoint of message, "sqs" or "sns"
+    :pin: patch info for the botocore client
+    :data_streams_enabled: boolean for whether data streams monitoring is enabled
 
-    Inject trace headers into MessageAttributes for all SQS or SNS records inside a batch
+    Inject trace headers and DSM info into MessageAttributes for all SQS or SNS records inside a batch
     """
-    trace_data = {}
-    HTTPPropagator.inject(span.context, trace_data)
+    if data_streams_enabled:
```
tabgok, line 147: I'd move this logic into `set_checkpoint_and_get_datadog_data`

ddtrace/contrib/botocore/patch.py
```diff
@@ -419,12 +450,12 @@ def patched_api_call(original_func, instance, args, kwargs):
                             direction=SpanDirection.OUTBOUND,
                         )
                     if endpoint_name == "sns" and operation == "Publish":
-                        inject_trace_to_sqs_or_sns_message(params, span, endpoint_name)
+                        inject_datadog_data_to_sqs_or_sns_message(params, span, endpoint_name)
```
tabgok, line 453: this should make updating for SNS easier, nice

ddtrace/contrib/botocore/patch.py
```diff
@@ -453,10 +484,37 @@ def patched_api_call(original_func, instance, args, kwargs):
...
+                elif "_datadog" not in args[1]["MessageAttributeNames"]:
+                    args[1].update({"MessageAttributeNames": args[1]["MessageAttributeNames"] + ["_datadog"]})
+                    args = (args[0], args[1])
+
+                result = original_func(*args, **kwargs)
+                _set_response_metadata_tags(span, result)
+
+                for message in result["Messages"]:
+                    try:
+                        pathway = json.loads(message["MessageAttributes"]["_datadog"]["StringValue"])["dd-pathway-ctx"]
```
piochelepiotr, line 503: so, in other languages, we add b64 encoded data to the header:
`dd-pathway-ctx-base64`
we can use the same convention here.
Can you check that the encoded data can be decoded in go using the `DecodeStr` function? 
Please, let me know if you need any help for that.
Also, let's put the name `dd-pathway-ctx-base64` as a constant in the core library.
https://github.com/DataDog/data-streams-go/blob/7db3f9c249385c741f5e5e83a01b05a9f4332a81/datastreams/propagator.go#L59

ddtrace/contrib/botocore/patch.py
```diff
@@ -453,10 +484,37 @@ def patched_api_call(original_func, instance, args, kwargs):
...
+            if endpoint_name == "sqs" and operation == "ReceiveMessage" and config._data_streams_enabled:
+                queue_url = params["QueueUrl"]
+                queue_name = queue_url[queue_url.rfind("/") + 1 :]
+
+                if "MessageAttributeNames" not in args[1]:
+                    args[1].update({"MessageAttributeNames": ["_datadog"]})
+                    args = (args[0], args[1])
+                elif "_datadog" not in args[1]["MessageAttributeNames"]:
+                    args[1].update({"MessageAttributeNames": args[1]["MessageAttributeNames"] + ["_datadog"]})
+                    args = (args[0], args[1])
```
piochelepiotr, line 496: what if you have args[2]? it would break the args.

shethhriday29: the way it's structured is that args[0] will be "ReceiveMessage" and args[1] is the actual list of parameters, ie `('ReceiveMessage', {'QueueUrl': <<URL here>>, 'MaxNumberOfMessages': 3, 'WaitTimeSeconds': 2, 'AttributeNames': ['All']})`, so it doesn't break the args because `MessageAttributeNames` is added to args[1]

tabgok (dismissed):

ddtrace/contrib/botocore/patch.py
```diff
@@ -109,52 +110,75 @@ def inject_trace_data_to_message_attributes(trace_data, entry, endpoint=None):
...
         else:
             log.warning("skipping trace injection, endpoint is not SNS or SQS")
     else:
         # In the event a record has 10 or more msg attributes we cannot add our _datadog msg attribute
         log.warning("skipping trace injection, max number (10) of MessageAttributes exceeded")
 
 
-def inject_trace_to_sqs_or_sns_batch_message(params, span, endpoint=None):
+def get_pathway(pin, params):
+    """
```
mabdinur, line 127: Can we add typing to this function
```suggestion
     # type: ( ....) -> ......
     """
```

ddtrace/contrib/botocore/patch.py
```diff
@@ -109,52 +110,75 @@ def inject_trace_data_to_message_attributes(trace_data, entry, endpoint=None):
...
 
-    Inject trace headers into MessageAttributes for the SQS or SNS record
+    Inject trace headers and DSM info into MessageAttributes for the SQS or SNS record
     """
-    trace_data = {}
-    HTTPPropagator.inject(span.context, trace_data)
+    datadog_data = {}
+    HTTPPropagator.inject(span.context, datadog_data)
+
+    if data_streams_enabled:
```
mabdinur, line 178: nit: we don't need the `data_streams_enabled` parameter. We can just access the config object directly.
```suggestion
    if config._data_streams_enabled:
```

ddtrace/contrib/botocore/patch.py
```diff
@@ -467,10 +495,37 @@ def patched_api_call(original_func, instance, args, kwargs):
         span.set_tag(ANALYTICS_SAMPLE_RATE_KEY, config.botocore.get_analytics_sample_rate())
 
         try:
-            result = original_func(*args, **kwargs)
-            _set_response_metadata_tags(span, result)
+            if endpoint_name == "sqs" and operation == "ReceiveMessage" and config._data_streams_enabled:
+                queue_url = params["QueueUrl"]
+                queue_name = queue_url[queue_url.rfind("/") + 1 :]
```
mabdinur, line 500: nit: We should use a helper here. Urls can have many formats. This method of getting the queue name could raise an IndexError or capture the wrong string

ddtrace/contrib/botocore/patch.py
```diff
@@ -467,10 +495,37 @@ def patched_api_call(original_func, instance, args, kwargs):
...
+
+                if "MessageAttributeNames" not in args[1]:
```
mabdinur, line 502: There's no guarantee that args[1] will always exist. The parameter could be passed a key word argument. 

We should use `get_argument_value(args, kwargs, 1, "<insert_kwarg_name_here>")`.

ddtrace/contrib/botocore/patch.py
```diff
@@ -467,10 +495,37 @@ def patched_api_call(original_func, instance, args, kwargs):
...
+                    args[1].update({"MessageAttributeNames": ["_datadog"]})
+                elif "_datadog" not in args[1]["MessageAttributeNames"]:
+                    args[1].update({"MessageAttributeNames": args[1]["MessageAttributeNames"] + ["_datadog"]})
+
+                result = original_func(*args, **kwargs)
+                _set_response_metadata_tags(span, result)
+
+                for message in result["Messages"]:
```
mabdinur, line 510: Will result always contain the key Message? If not this could raise a KeyError

ddtrace/contrib/botocore/patch.py
```diff
@@ -467,10 +495,37 @@ def patched_api_call(original_func, instance, args, kwargs):
...
+                    try:
+                        pathway = json.loads(message["MessageAttributes"]["_datadog"]["StringValue"])[
```
mabdinur, line 512: Same here. Are we guaranteed to have these nested keys?

ddtrace/contrib/botocore/patch.py
```diff
@@ -467,10 +495,37 @@ def patched_api_call(original_func, instance, args, kwargs):
...
+                            PROPAGATION_KEY_BASE_64
+                        ]
+
+                        ctx = pin.tracer.data_streams_processor.decode_pathway_b64(pathway)
+                        ctx.set_checkpoint(["direction:in", "topic:" + queue_name, "type:sqs"])
+
+                    except Exception:
```
mabdinur, line 519: DO we want to log a warning/error here?

Also catching exception is a bit broad. Is there a specific error we are try to handle? This code could always fail silently.

ddtrace/internal/datastreams/processor.py
```diff
@@ -65,6 +66,8 @@ def gzip_compress(payload):
 log = get_logger(__name__)
 
 PROPAGATION_KEY = "dd-pathway-ctx"
+PROPAGATION_KEY_BASE_64 = "dd-pathway-ctx-base64"
```
mabdinur, line 69: nit: This constant is not used in this file. Is there a better place to put this?

riotfile.py
```diff
@@ -1491,6 +1491,9 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
         Venv(
             name="botocore",
             command="pytest {cmdargs} tests/contrib/botocore",
+            env={
+                "DD_DATA_STREAMS_ENABLED": "true",
+            },
             venvs=[
                 Venv(pys=select_pys(min_version="3.8"), pkgs={"moto[all]": latest, "botocore": latest}),
```
mabdinur, line 1498: Instead of setting `DD_DATA_STREAMS_ENABLED=true` for all test runs we should run all the botocore tests with both dsm enabled and disabled.

```suggestion
                Venv(pys=select_pys(min_version="3.8"), pkgs={"moto[all]": latest, "botocore": latest}),
                # Test botocore integration with dsm support
                Venv(pys=[MIN_PYTHON_VERSION, MAX_PYTHON_VERSION], pkgs={"moto[all]": latest, "botocore": latest}, env={
                "DD_DATA_STREAMS_ENABLED": ["false", "true"]},),
```

tests/utils.py
```diff
@@ -528,6 +528,12 @@ def __init__(self, *args, **kwargs):
         super(DummyTracer, self).__init__()
         self._trace_flush_enabled = True
         self.configure(*args, **kwargs)
+        if ddtrace.config._data_streams_enabled:
+            # Inline the import to avoid pulling in ddsketch or protobuf
+            # when importing ddtrace.
+            from ddtrace.internal.datastreams.processor import DataStreamsProcessor
+
+            self.data_streams_processor = DataStreamsProcessor(self._agent_url)
```
mabdinur, line 536: I am not sure if this required. `super(DummyTracer, self).__init__()` should initialize the data stream process if  `ddtrace.config._data_streams_enabled=True`

releasenotes/notes/botocore_dsm_sqs-02df9afa4b4bb503.yaml
```diff
@@ -0,0 +1,3 @@
+---
+features:
+  - Added message attributes to support data streams monitoring
```
mabdinur, line 3: Feel free to update the text. I just wanted to fix the formatting and tensing. Here are the guidelines to writing a release note: https://github.com/DataDog/dd-trace-py/blob/1.x/docs/releasenotes.rst#release-note-sections.

```suggestion
  borocore: Extends data streams monitoring support to sqs botocore.
```

releasenotes/notes/botocore_dsm_sqs-02df9afa4b4bb503.yaml
```diff
@@ -0,0 +1,3 @@
+---
+features:
+botocore: Added message attributes to sqs spans to support data streams monitoring.
```
mabdinur, line 3: builddocs is failing due to formatting issues:
```suggestion
    botocore: Added message attributes to sqs spans to support data streams monitoring.
```

ddtrace/contrib/botocore/patch.py
```diff
@@ -467,10 +496,52 @@ def patched_api_call(original_func, instance, args, kwargs):
...
+                                                message["MessageAttributes"]["_datadog"]["StringValue"]
+                                        ):
+                                            pathway = json.loads(message["MessageAttributes"]["_datadog"]["StringValue"])[
+                                                PROPAGATION_KEY_BASE_64
+                                            ]
+                                            ctx = pin.tracer.data_streams_processor.decode_pathway_b64(pathway)
+                                            ctx.set_checkpoint(["direction:in", "topic:" + queue_name, "type:sqs"])
+
+                                        else:
+                                            log.warning("Unable to find data streams context for the message. Data "
```
tabgok, line 538: I feel like this should be a debug statement; otherwise if a queue has 100s or 1000s of messages, customesr will be spammed with WARNING messages.

Maybe @mabdinur has an opinion

ddtrace/contrib/botocore/patch.py
```diff
@@ -467,10 +505,52 @@ def patched_api_call(original_func, instance, args, kwargs):
...
+                    params.update({"MessageAttributeNames": ["_datadog"]})
+                elif "_datadog" not in params["MessageAttributeNames"]:
+                    params.update({"MessageAttributeNames": params["MessageAttributeNames"] + ["_datadog"]})
+
+                result = original_func(*args, **kwargs)
+                _set_response_metadata_tags(span, result)
+
+                try:
+                    if "Messages" in result:
+                        for message in result["Messages"]:
```
tabgok, line 521: nit: you could rewrite this if stack as a generator, which would make it more readable.

```
for message in messages_with_dsm_header(messages):
    ctx.foo()
```

You can get fancier, too, if you want to handle both matching and unmatching:

```
for message, has_dsm_header in _dsm_messages():
    if has_dsm_header:
        ctx.foo(message)
    else:
         ctx.new_path(message)
```

ddtrace/contrib/botocore/patch.py
```diff
@@ -467,10 +505,51 @@ def patched_api_call(original_func, instance, args, kwargs):
...
+                            max_attributes_flag = True
+                            if "MessageAttributes" in message:
+                                if "_datadog" in message["MessageAttributes"]:
+                                    if "StringValue" in message["MessageAttributes"]["_datadog"]:
```
mabdinur, line 525:
```suggestion
                            if "MessageAttributes" in message and "_datadog" in message["MessageAttributes"] and "StringValue" in message["MessageAttributes"]["_datadog"]:
```

ddtrace/contrib/botocore/patch.py
```diff
@@ -467,10 +505,51 @@ def patched_api_call(original_func, instance, args, kwargs):
...
+
+                                        else:
+                                            log.warning("Unable to find data streams context for the message. Data "
+                                                        "streams monitoring was not enabled by the application "
+                                                        "sending the message at the time of the send.")
+                            if max_attributes_flag:
+                                log.warning("Unable to find trace context for the message.")
+
+                except Exception as ex:
+                    log.exception("Error receiving SQS message with data streams monitoring enabled: %s", ex)
```
mabdinur, line 544:
```suggestion
                    log.debug("Error receiving SQS message with data streams monitoring enabled", exc_info=True)
```

tests/utils.py
```diff
@@ -528,6 +528,14 @@ def __init__(self, *args, **kwargs):
...
         self._trace_flush_enabled = True
         self.configure(*args, **kwargs)
+        """
+        if ddtrace.config._data_streams_enabled:
+            # Inline the import to avoid pulling in ddsketch or protobuf
+            # when importing ddtrace.
+            from ddtrace.internal.datastreams.processor import DataStreamsProcessor
+
+            self.data_streams_processor = DataStreamsProcessor(self._agent_url)
+        """
```
mabdinur, line 538:
```suggestion
```

ddtrace/contrib/botocore/patch.py
mabdinur: Yup debug log would be nice here. In the past we have gotten escalations for noisy error and warning logs.

ddtrace/contrib/botocore/patch.py
```diff
@@ -122,15 +122,43 @@ def inject_trace_data_to_message_attributes(trace_data, entry, endpoint=None):
...
+def get_queue_name(params):
+    # type: (str) -> str
+    """
+    :params: contains the params for the current botocore action
+
+    Return the name of the queue given the params
+    """
+    queue_url = params["QueueUrl"]
+    url = urlparse(queue_url)
+    return url.path.rsplit("/", 1)[-1]
```
mabdinur, line 134: This could be overkill but we can use os.path.split(url.path) to get the queue name

releasenotes/notes/botocore_dsm_sqs-02df9afa4b4bb503.yaml
```diff
@@ -0,0 +1,3 @@
+---
+features:
+    botocore: Added message attributes to sqs spans to support data streams monitoring.
```
mabdinur, line 3:
```suggestion
  - |
    botocore: Added message attributes to sqs spans to support data streams monitoring.
```

mabdinur: adds a new line to release note

ddtrace/contrib/botocore/patch.py
```diff
@@ -467,10 +505,53 @@ def patched_api_call(original_func, instance, args, kwargs):
...
+                                            ctx = pin.tracer.data_streams_processor.decode_pathway_b64(pathway)
+                                            ctx.set_checkpoint(["direction:in", "topic:" + queue_name, "type:sqs"])
+
+                                        else:
+                                            log.debug(
+                                                "Unable to find data streams context for the message. Data "
+                                                "streams monitoring was not enabled by the application "
+                                                "sending the message at the time of the send."
+                                            )
+                            if max_attributes_flag:
```
mabdinur, line 542:
```suggestion
              else:
```

ddtrace/contrib/botocore/patch.py
```diff
@@ -467,10 +505,53 @@ def patched_api_call(original_func, instance, args, kwargs):
...
+                _set_response_metadata_tags(span, result)
+
+                try:
+                    if "Messages" in result:
+                        for message in result["Messages"]:
+                            max_attributes_flag = True
+                            if "MessageAttributes" in message:
+                                if "_datadog" in message["MessageAttributes"]:
+                                    if "StringValue" in message["MessageAttributes"]["_datadog"]:
+                                        max_attributes_flag = False
```
mabdinur, line 526:
```suggestion
```

ddtrace/contrib/botocore/patch.py
```diff
@@ -467,10 +505,53 @@ def patched_api_call(original_func, instance, args, kwargs):
...
+                elif "_datadog" not in params["MessageAttributeNames"]:
+                    params.update({"MessageAttributeNames": params["MessageAttributeNames"] + ["_datadog"]})
+
+                result = original_func(*args, **kwargs)
+                _set_response_metadata_tags(span, result)
+
+                try:
+                    if "Messages" in result:
+                        for message in result["Messages"]:
+                            max_attributes_flag = True
```
mabdinur, line 522:
```suggestion
```

ddtrace/contrib/botocore/patch.py
```diff
@@ -467,10 +505,53 @@ def patched_api_call(original_func, instance, args, kwargs):
...
+                                        ):
+                                            pathway = json.loads(
+                                                message["MessageAttributes"]["_datadog"]["StringValue"]
+                                            )[PROPAGATION_KEY_BASE_64]
+                                            ctx = pin.tracer.data_streams_processor.decode_pathway_b64(pathway)
+                                            ctx.set_checkpoint(["direction:in", "topic:" + queue_name, "type:sqs"])
+
+                                        else:
+                                            log.debug(
+                                                "Unable to find data streams context for the message. Data "
```
mabdinur, line 538: Will this be logged for all botocore users who do not have dsm enabled? If so we should remove it. This could get pretty noisy.

shethhriday29: so i think it might be confusing for a user that has DSM enabled on the receiving side but then doesn't know why it's not showing up on their DSM page on datadog's app, and can't figure out why that's the case

shethhriday29: but i'm not a logging expert by any means — will defer to your judgement here! happy to take out the logs if they're too noisy. there might also be a lot of them, since there's a log for each message that doesn't have the DSM context (and the most we can reduce this to is 1 message/1 receive API call, which I can do but might still be expensive)

mabdinur: Ahhh can we add this context to the log message. This will make the log messages more actionable.

shethhriday29: it currently says "Unable to find data streams context for the message. Data streams monitoring was not enabled by the application sending the message at the time of the send."  — did you want something additional?

shethhriday29: i just updated the other log message to be more actionable as well

mabdinur (dismissed):

ddtrace/contrib/botocore/patch.py
```diff
@@ -14,6 +14,12 @@
 from typing import Tuple
 from typing import Union
 
+
+try:
+    from urllib.parse import urlparse
+except ImportError:
+    from urlparse import urlparse
```
mabdinur, line 21:
```suggestion
from ddtrace.internal.compat import parse 
```

releasenotes/notes/botocore-sqs-queue-nonurl-b11745018c92ea1b.yaml
```diff
@@ -1,5 +0,0 @@
----
-fixes:
-  - |
-    botocore: This fix resolves an issue where ddtrace attempted to parse as URLs SQS QueueUrl attributes that
-    were not well-formed URLs.
```
mabdinur, line 5: We should revert this

mabdinur (approved):

mabdinur (approved):

PR #6121, opened 2023-06-14
core: add DD_TAGS support for runtime metrics

brettlangdon:
This PR introduces a change to collect additional default runtime metrics tags from the tracer/env.

- Collect `DD_VERSION`/`config.version` as `version` tag
- Collect `DD_TAGS`/`tracer.set_tags` tags
- Update/fix how we collect the `env` tag
  - "fix" since it looks for env in `tracer._tags`, but we try to explicitly remove from there
- Populate `tracer._services` with `DD_SERVICE`/`config.service` if we have it on load

One thing to consider, the dogstatsd client supports most of the `DD_*` tags, except it uses `DATADOG_TAGS` instead of `DD_TAGS`, we could/should consider updating the dogstatsd client as well to support `DD_TAGS`. We'd likely still want to make this specific change anyways, since customers can configure these attributes via code instead of env vars.

## Checklist

## Reviewer Checklist

- [ ] Title is accurate.
- [ ] Backport labels are set in a manner that is consistent with the [release branch maintenance policy](../docs/contributing.rst#release-branch-maintenance)


ddtrace/tracer.py
```diff
@@ -230,8 +230,11 @@ def __init__(
         # globally set tags
         self._tags = config.tags.copy()
 
+        # collection of services seen, used for runtime metrics tags
         # a buffer for service info so we don't perpetually send the same things
         self._services = set()  # type: Set[str]
+        if config.service:
```
ZStriker19, line 236: should we add this under line 565 too for after we fork? Or would that be setting the service to config.service which we wouldn't want to assume should happen?

releasenotes/notes/feat-support-global-tags-in-runtime-metrics-58fe01b78b7f9cbe.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    core: Adds support for ``DD_TAGS`` with runtime metrics.
```
ZStriker19, line 4:
```suggestion
    core: Apply ``DD_TAGS`` to runtime metrics.
```
Simpler?

ZStriker19 (commented):
nits

ddtrace/tracer.py
brettlangdon: yes, good catch!

ZStriker19 (approved):

PR #6137, opened 2023-06-14
feat(langchain): add LangChain integration

Yun-Kim:
## Tips for Reviewers

**note: 90% of the line changes in this PR are non-code, in the form of snapshot files, riot lockfiles, and test cassette files used to mock request/responses with the OpenAI/Cohere/HuggingFace/Pinecone API, which stores binary embeddings data in `.yaml` files that Github counts as lines of code.**

The major files that reviewers should focus on are:
- `ddtrace/contrib/langchain/__init__.py` - documentation for the langchain integration
- `ddtrace/contrib/langchain/patch.py` - implementation of each traced langchain function
- `tests/contrib/langchain/test_langchain.py` - shows how each traced langchain function is expected to be used.

## Summary

This PR adds an integration for the [LangChain](https://python.langchain.com/docs/get_started/introduction.html) library. This integration provides tracing for the following (sync and async) interfaces provided by LangChain:
- LLMs: traced via `traced_llm_generate()`, `traced_llm_agenerate()`
- Chat Models: traced via `traced_chat_model_generate()`, `traced_chat_model_agenerate()`
- Embeddings: traced via `traced_embedding()`. Note that we do not yet support asynchronous embedding requests, as it is not yet entirely implemented on LangChain's end.
- Chains: traced via `traced_chain_call()` and `traced_chain_acall()`.
- Vectorstores: traced via `traced_similarity_search()`. Note that we do not yet support other vectorstore operations.

### Cost/Usage tracking

Langchain only provides usage/cost information from the `OpenAI` library, meaning we will only provide such information for OpenAI-specific requests. ([ref](https://python.langchain.com/en/latest/modules/models/llms/examples/token_usage_tracking.html))

## Testing

The testing for this integration involves snapshot testing for each traced function, as well as using `vcrpy` cassettes to mock request/response pairs to each LLM provider API. Due to LangChain's extensive support for a large amount of different LLM/VectorStore providers, we could not test each individual LLM provider in this PR, but tested representatives for each interface (i.e. `openAI, cohere, huggingfacehub, pinecone`).

Note that for Python 3.9 tests required separate test cassette files due to a difference in request/response formatting, which contributed to the large diff size.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/contrib/langchain/patch.py
```diff
@@ -0,0 +1,713 @@
...
+            "langchain.request.api_key:%s" % api_key,
+            "error:%d" % span.error,
+        ]
+        err_type = span.get_tag("error.type")
+        if err_type:
+            tags.append("error_type:%s" % err_type)
+        return tags
+
+    def record_usage(self, span, usage):
+        if not usage or not self._config.metrics_enabled:
```
mabdinur, line 83: Very minor nit: `ddtrace.config.somethingthatdoes_not_exist` will return a truthy value (it will initialize or refer to an IntegrationConfig object). It might be safer here to check if self._config.metrics_enabled is False.

ddtrace/contrib/langchain/patch.py
```diff
@@ -0,0 +1,713 @@
...
+        total_cost = span.get_metric("langchain.tokens.total_cost")
+        if total_cost:
+            self.metric(span, "incr", "tokens.total_cost", total_cost)
+
+
+def _extract_model_name(instance):
+    """Extract model name or ID from llm instance."""
+    for attr in ("model", "model_name", "model_id", "model_key", "repo_id"):
+        if hasattr(instance, attr):
+            return getattr(instance, attr, None)
```
mabdinur, line 99: nit:
```suggestion
            return getattr(instance, attr)
```

ddtrace/contrib/langchain/patch.py
```diff
@@ -0,0 +1,713 @@
...
+    Extract and format LLM-provider API key from instance.
+    Note that langchain's LLM/ChatModel/Embeddings interfaces do not have a
+    standard attribute name for storing the provider-specific API key, so make a
+    best effort here by checking for attributes that end with `api_key/token`.
+    """
+    api_key_attrs = [a for a in dir(instance) if a.endswith(("api_token", "api_key"))]
+    if api_key_attrs and hasattr(instance, str(api_key_attrs[0])):
+        api_key = getattr(instance, api_key_attrs[0], None)
+        if api_key:
+            return "...%s" % api_key[-4:]
```
mabdinur, line 114: nit: Should we add a helper for obfuscating api keys? I think we have similar logic in the openai integration.

ddtrace/contrib/langchain/patch.py
```diff
@@ -0,0 +1,713 @@
...
+            return getattr(instance, attr, None)
+    return None
+
+
+def _extract_api_key(instance):
+    """
+    Extract and format LLM-provider API key from instance.
+    Note that langchain's LLM/ChatModel/Embeddings interfaces do not have a
+    standard attribute name for storing the provider-specific API key, so make a
+    best effort here by checking for attributes that end with `api_key/token`.
```
mabdinur, line 108:
```suggestion
    best effort here by checking for attributes that end with `api_key/api_token`.
```

mabdinur: We make a best effort to extract keys, names and configurations from LLM models. Should we also add debug logging for when our best efforts fail. This could help us debug issues down the line and provide more insight on which models and versions of the api are not actually supported.

ddtrace/contrib/langchain/patch.py
```diff
@@ -0,0 +1,713 @@
...
+                    ],
+                },
+            )
+    except Exception:
+        span.set_exc_info(*sys.exc_info())
+        integration.metric(span, "incr", "request.error", 1)
+        raise
+    finally:
+        span.finish()
+        integration.metric(span, "dist", "request.duration", span.duration_ns)
```
emmettbutler, line 356: There's some duplication in this section of multiple functions in this file. You might consider putting this logic into a function to make future changes easier.

ddtrace/contrib/langchain/patch.py
```diff
@@ -0,0 +1,713 @@
...
+                        integration.trunc(message.content),
+                    )
+                span.set_tag_str(
+                    "langchain.request.messages.%d.%d.message_type" % (message_set_idx, message_idx),
+                    message.__class__.__name__,
+                )
+        model = _extract_model_name(instance)
+        if model is not None:
+            span.set_tag_str("langchain.request.model", model)
+        span.set_tag_str("langchain.request.provider", llm_provider)
```
emmettbutler, line 295: This seems to be a pattern that's repeated a lot. You might consider deduplicating it to make changes easier.

ddtrace/contrib/langchain/patch.py
```diff
@@ -0,0 +1,713 @@
...
+            # If not in langchain's openai model catalog, assume 0 total cost.
+            pass
+    span.set_metric("langchain.tokens.total_cost", propagated_cost + total_cost)
+    if span._parent is not None:
+        _tag_openai_token_usage(span._parent, llm_output, propagated_cost=propagated_cost + total_cost, propagate=True)
+
+
+@with_traced_module
+def traced_llm_generate(langchain, pin, func, instance, args, kwargs):
+    llm_provider = instance._llm_type
```
mabdinur, line 150: Should we first get the pin object and see if it's not None and enabled. We do this in most integrations. It might be best to keep the same pattern here. 

Ex from graphql:

```
pin = Pin.get_from(graphql)
if not pin or not pin.enabled():
    return func(*args, **kwargs)
```

ddtrace/contrib/langchain/patch.py
```diff
@@ -0,0 +1,713 @@
...
+            (config.env or ""),
+            (config.version or ""),
+            (span.get_tag("langchain.request.provider") or ""),
+            (span.get_tag("langchain.request.model") or ""),
+            (span.get_tag("langchain.request.type") or ""),
+            api_key,
+        )
+        return tags
+
+    def _metrics_tags(self, span):
```
emmettbutler, line 64: Does this need to be an instance method? It doesn't use the `self` parameter, so turning it into a class method could arguably make the code easier to read.

emmettbutler (dismissed):
This is awesome. Well done @Yun-Kim.

ddtrace/contrib/langchain/patch.py
```diff
@@ -0,0 +1,713 @@
...
+    integration = langchain._datadog_integration
+    span = integration.trace(pin, "%s.%s" % (instance.__module__, instance.__class__.__name__))
+    try:
+        if integration.is_pc_sampled_span(span):
+            for idx, prompt in enumerate(prompts):
+                span.set_tag_str("langchain.request.prompts.%d" % idx, integration.trunc(str(prompt)))
+        model = _extract_model_name(instance)
+        if model is not None:
+            span.set_tag_str("langchain.request.model", model)
+        span.set_tag_str("langchain.request.provider", llm_provider)
```
mabdinur, line 161: I think I mentioned this about but it think some of these tag names should be constants. It will make this integration easier to maintain.

ddtrace/contrib/langchain/patch.py
```diff
@@ -0,0 +1,713 @@
...
+@with_traced_module
+def traced_llm_generate(langchain, pin, func, instance, args, kwargs):
+    llm_provider = instance._llm_type
+    prompts = args[0]
+    integration = langchain._datadog_integration
+    span = integration.trace(pin, "%s.%s" % (instance.__module__, instance.__class__.__name__))
+    try:
+        if integration.is_pc_sampled_span(span):
+            for idx, prompt in enumerate(prompts):
+                span.set_tag_str("langchain.request.prompts.%d" % idx, integration.trunc(str(prompt)))
```
mabdinur, line 157: Is there an upper bound on how many prompts tags are set on each span. We could end up with some massive traces.

ddtrace/contrib/langchain/patch.py
```diff
@@ -0,0 +1,713 @@
...
+        model = _extract_model_name(instance)
+        if model is not None:
+            span.set_tag_str("langchain.request.model", model)
+        span.set_tag_str("langchain.request.provider", llm_provider)
+        span.set_tag_str("langchain.request.type", "llm")
+        for param, val in getattr(instance, "_identifying_params", {}).items():
+            if isinstance(val, dict):
+                for k, v in val.items():
+                    span.set_tag_str("langchain.request.%s.parameters.%s.%s" % (llm_provider, param, k), str(v))
```
mabdinur, line 166: I am not a big fan of getting tag names dynamically. Can we default to an allowlist for all possible tags. This will be difficult here since `param` and `k` could be anything but maybe we can create a list of supported llm providers?

ddtrace/contrib/langchain/patch.py
```diff
@@ -0,0 +1,713 @@
...
+
+        completions = func(*args, **kwargs)
+        if isinstance(instance, langchain.llms.OpenAI):
+            _tag_openai_token_usage(span, completions.llm_output)
+            integration.record_usage(span, completions.llm_output)
+
+        for idx, completion in enumerate(completions.generations):
+            if integration.is_pc_sampled_span(span):
+                span.set_tag_str("langchain.response.completions.%d.text" % idx, integration.trunc(completion[0].text))
+            if completion[0].generation_info is not None:
```
mabdinur, line 179: Can completion be an empty list?

```suggestion
            if completion and completion[0].generation_info is not None:
```

ddtrace/contrib/langchain/patch.py
```diff
@@ -0,0 +1,713 @@
...
+                "sampled %s.%s" % (instance.__module__, instance.__class__.__name__),
+                attrs={
+                    "prompts": prompts,
+                    "choices": [
+                        [{"text": completion.text} for completion in completions]
+                        for completions in completions.generations
+                    ],
+                },
+            )
+    except Exception:
```
mabdinur, line 201: Should we just use a context manager here:

Ex:

```
with integration.trace(pin, "%s.%s" % (instance.__module__, instance.__class__.__name__)) as span:
  ....... this method .....
```

I guess in this case need a new way to compute error metrics.

ddtrace/contrib/langchain/patch.py
```diff
@@ -0,0 +1,713 @@
...
+    finally:
+        span.finish()
+        integration.metric(span, "dist", "request.duration", span.duration_ns)
+    return completions
+
+
+@with_traced_module
+async def traced_llm_agenerate(langchain, pin, func, instance, args, kwargs):
+    llm_provider = instance._llm_type
+    prompts = args[0]
```
mabdinur, line 214: get_argument_value()?

ddtrace/contrib/langchain/patch.py
```diff
@@ -0,0 +1,713 @@
...
+    integration = langchain._datadog_integration
+    span = integration.trace(pin, "%s.%s" % (instance.__module__, instance.__class__.__name__))
```
mabdinur, line 216: nit: There is some duplication between `traced_llm_agenerate` and `traced_llm_generate`. Would it be possible to pull out some common logic? 

Most of the comments in `traced_llm_generate` apply here as well

ddtrace/contrib/langchain/patch.py
```diff
@@ -0,0 +1,713 @@
...
+                    str(completion[0].generation_info.get("finish_reason")),
+                )
+                span.set_tag_str(
+                    "langchain.response.completions.%d.logprobs" % idx,
+                    str(completion[0].generation_info.get("logprobs")),
+                )
+        if integration.is_pc_sampled_log(span):
+            integration.log(
+                span,
+                "info" if span.error == 0 else "error",
```
mabdinur, line 254: Would this always be an info log since span.error is set on line 265 (10 lines below)

ddtrace/contrib/langchain/patch.py
```diff
@@ -0,0 +1,713 @@
...
+        raise
+    finally:
+        span.finish()
+        integration.metric(span, "dist", "request.duration", span.duration_ns)
+    return chat_completions
+
+
+@with_traced_module
+async def traced_chat_model_agenerate(langchain, pin, func, instance, args, kwargs):
+    llm_provider = instance._llm_type.split("-")[0]
```
mabdinur, line 362: There is a decent amount of duplication here with `traced_chat_model_generate` could we define more helpers to consolidate some of the logic. Is this worth the effort or would it negatively impact readability.

riotfile.py
```diff
@@ -2773,6 +2773,24 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
                 ),
             ],
         ),
+        Venv(
+            name="langchain",
+            command="pytest {cmdargs} tests/contrib/langchain",
+            pys=select_pys(min_version="3.9"),
+            pkgs={
+                "langchain": ["==0.0.192", latest],
+                "openai": latest,
```
mabdinur, line 2782: dumb question: Do we need to specify all these packages? Are some of these packages installed transitively by openai and langchain?

mabdinur (commented):
I did not look over the tests yet but I trust that they have sufficient coverage and address all edge cases. 

I left a lot of comments but most fall under these general themes:
 1) Debugability, we make a best effort to get various tags and then default to certain values. I think we should add logs/metrics to track when default values are used and edge cases that the integration does not support. 
 2) Maintainability: We are setting and getting most tags dynamically. I am a little concerned about the size of the spans that this integration will generate. Also would it be possible to define constants for all LLM tag names. This will make sure we don't end up in the following scenario: A tag is set with a name that has a typo, the tag is then retreived but does not exist so a default value is always used.
3) Cost: Do we have rough estimates on the ingestion costs for this new integration. We are adding complex logs and metrics. I am worried that customers might see their Datadog usage skyrocket due to high volume and high cardinality logs+metrics. 


Overall this integration looks great 😄.

ddtrace/contrib/_trace_utils_llm.py
```diff
@@ -24,6 +24,10 @@ class BaseLLMIntegration:
     _integration_name = "baseLLM"
 
     def __init__(self, config, stats_url, site, api_key):
+        # FIXME: this currently does not consider if the tracer is configured to
```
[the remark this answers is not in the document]
Yun-Kim: Good catch, I'll add a known issue release note.

ddtrace/contrib/langchain/__init__.py
```diff
@@ -0,0 +1,205 @@
...
+
+
+Global Configuration
+~~~~~~~~~~~~~~~~~~~~
+
+.. py:data:: ddtrace.config.langchain["service"]
+
+   The service name reported by default for LangChain requests.
+
+   Alternatively, you can set this option with the ``DD_SERVICE`` or ``DD_OPENAI_SERVICE`` environment
```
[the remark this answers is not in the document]
Yun-Kim: Good catch! This should be `DD_LANGCHAIN_SERVICE`.

ddtrace/contrib/langchain/__init__.py
```diff
@@ -0,0 +1,205 @@
...
+Instance Configuration
+~~~~~~~~~~~~~~~~~~~~~~
+
+To configure the LangChain integration on a per-instance basis use the
+``Pin`` API::
+
+    import langchain
+    from ddtrace import Pin, config
+
+    Pin.override(langchain, service="my-langchain-service")
```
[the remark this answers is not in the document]
Yun-Kim: Yes, you're right 🤦 good catch!

ddtrace/contrib/langchain/patch.py
```diff
@@ -0,0 +1,713 @@
...
+    from ddtrace import Span
+
+
+log = get_logger(__name__)
+
+
+config._add(
+    "langchain",
+    {
+        "logs_enabled": asbool(os.getenv("DD_LANGCHAIN_LOGS_ENABLED", False)),
```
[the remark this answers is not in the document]
Yun-Kim: I don't think so, since these are integration-specific configurations and not necessarily overall/global configurations.

ddtrace/contrib/langchain/patch.py
```diff
@@ -0,0 +1,713 @@
...
+        "metrics_enabled": asbool(os.getenv("DD_LANGCHAIN_METRICS_ENABLED", True)),
+        "span_prompt_completion_sample_rate": float(os.getenv("DD_LANGCHAIN_SPAN_PROMPT_COMPLETION_SAMPLE_RATE", 1.0)),
+        "log_prompt_completion_sample_rate": float(os.getenv("DD_LANGCHAIN_LOG_PROMPT_COMPLETION_SAMPLE_RATE", 0.1)),
+        "span_char_limit": int(os.getenv("DD_LANGCHAIN_SPAN_CHAR_LIMIT", 128)),
+        "_api_key": os.getenv("DD_API_KEY"),
```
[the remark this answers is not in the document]
Yun-Kim: Agreed, we can definitely store it on the global config, but after looking through our library for usages of `DD_API_KEY` I think it's fair enough to do the refactor outside this PR.

ddtrace/contrib/langchain/patch.py
```diff
@@ -0,0 +1,713 @@
...
+        super().__init__(config, stats_url, site, api_key)
+
+    def _set_base_span_tags(self, span):
+        # type: (Span) -> None
+        return None
+
+    def _logs_tags(self, span):
+        # type: (Span) -> str
+        api_key = span.get_tag("langchain.request.api_key") or ""
+        tags = "env:%s,version:%s,langchain.request.provider:%s,langchain.request.model:%s,langchain.request.type:%s,langchain.request.api_key:%s" % (  # noqa: E501
```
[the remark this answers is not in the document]
Yun-Kim: Good question. For conformity purposes, having the tags always be present is helpful, especially for cases where (in the dashboard queries for example) we have to query for specific logs that do not have an empty tag value. I agree your recommendation would reduce the payload size, but I don't think this will effect payload sizes *that* much.

ddtrace/contrib/langchain/patch.py
Yun-Kim: fair point - i'll do that.

ddtrace/contrib/langchain/patch.py
```diff
@@ -0,0 +1,713 @@
...
+            (span.get_tag("langchain.request.model") or ""),
+            (span.get_tag("langchain.request.type") or ""),
+            api_key,
+        )
+        return tags
+
+    def _metrics_tags(self, span):
+        provider = span.get_tag("langchain.request.provider") or ""
+        api_key = span.get_tag("langchain.request.api_key") or ""
+        tags = [
```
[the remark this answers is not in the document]
Yun-Kim: This is expected, the logs tag is stored as a string on the log JSON object while the metrics tags are processed in Dogstatsd as a list. I'll add typing to make it more clear.

ddtrace/contrib/langchain/patch.py
Yun-Kim: Good point.

ddtrace/contrib/langchain/patch.py
Yun-Kim: I agree adding debug logging would be super helpful. This might add more detail to this PR which is already way too big to begin with though, so I'll add that in a future PR.

ddtrace/contrib/langchain/patch.py
```diff
@@ -0,0 +1,713 @@
...
+                span.get_tag("langchain.request.model"),
+                span.get_metric("langchain.tokens.completion_tokens"),
+                is_completion=True,
+            )
+            prompt_cost = get_openai_token_cost_for_model(
+                span.get_tag("langchain.request.model"), span.get_metric("langchain.tokens.prompt_tokens")
+            )
+            total_cost = completion_cost + prompt_cost
+        except ValueError:
+            # If not in langchain's openai model catalog, assume 0 total cost.
```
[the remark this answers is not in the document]
Yun-Kim: Yup, the `ValueError` is explicitly only raised IF the provided model is not part of their model catalog. I'll add a log statement here explaining this: https://github.com/hwchase17/langchain/blob/master/langchain/callbacks/openai_info.py#L94-L97

ddtrace/contrib/langchain/patch.py
```diff
@@ -0,0 +1,713 @@
...
+            pass
+    span.set_metric("langchain.tokens.total_cost", propagated_cost + total_cost)
+    if span._parent is not None:
+        _tag_openai_token_usage(span._parent, llm_output, propagated_cost=propagated_cost + total_cost, propagate=True)
```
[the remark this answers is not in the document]
Yun-Kim: Does partial flushing mean we will no longer have any reference to the parent spans?

ddtrace/contrib/langchain/patch.py
Yun-Kim: we use `@with_traced_module` which does this check internally (otherwise returns the wrapped function without running tracing code).

ddtrace/contrib/langchain/patch.py
Yun-Kim: I don't believe so, but it shouldn't hurt to be defensive.

ddtrace/contrib/langchain/patch.py
Yun-Kim: The two methods are functionally the same, `traced_llm_agenerate()` simply has async syntax. If it were just this method (unfortunately there's also the same issue for `traced_chat_model_a/generate`, `traced_chain_a/call`), I would pull out common logic but with so many methods to do this for I don't see this adding readability.

ddtrace/contrib/langchain/patch.py
```diff
@@ -0,0 +1,713 @@
...
+                    "langchain.response.completions.%d.logprobs" % idx,
+                    str(completion[0].generation_info.get("logprobs")),
+                )
+        if integration.is_pc_sampled_log(span):
+            integration.log(
+                span,
+                "info" if span.error == 0 else "error",
+                "sampled %s.%s" % (instance.__module__, instance.__class__.__name__),
+                attrs={
+                    "prompts": prompts,
```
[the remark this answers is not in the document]
Yun-Kim: Cardinality is a concern, but due to the nature of LLM prompt-completions this will be inevitable. That's why log sampling is disabled by default and 10% when enabled (and must be explicitly configured to have a different rate)

ddtrace/contrib/langchain/patch.py
Yun-Kim: Good point. I've added a helper `_set_base_span_tags()` to set these base level tags for all spans.

ddtrace/contrib/langchain/patch.py
Yun-Kim: There's no explicit upper bound on prompt tags, but with token/request rate limits from each LLM provider we can be confident that users won't be sending insane numbers of prompts.

ddtrace/contrib/langchain/patch.py
Yun-Kim: I can't really see the reason for adding an allowlist of llm providers. Can you elaborate?

ddtrace/contrib/langchain/patch.py
Yun-Kim: That was a great catch - thanks for pointing this out! I'll make a fix.

ddtrace/contrib/langchain/patch.py
Yun-Kim: This might actually negatively impact readability at this point because there are 3 sync/async pairs of methods we need to do this for and this might actually increase even more complexity at this point.

riotfile.py
Yun-Kim: The below packages are all independent (all different LLM provider libraries), though `psutil` and `exceptiongroup` can be removed once riot is upgraded on our end.

ddtrace/contrib/langchain/patch.py
```diff
@@ -0,0 +1,713 @@
...
+
+
+def _tag_openai_token_usage(span, llm_output, propagated_cost=0, propagate=False):
+    """
+    Extract token usage from llm_output, tag on span.
+    Calculate the total cost for each LLM/chat_model, then propagate those values up the trace so that
+    the root span will store the total token_usage/cost of all of its descendants.
+    """
+    for token_type in ("prompt", "completion", "total"):
+        current_metric_value = span.get_metric("langchain.tokens.%s_tokens" % token_type) or 0
```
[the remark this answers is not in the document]
Yun-Kim: I think this is fine for now since this is just 3 metrics. If we do end up needing to add more metrics I think we can consider your suggestion of using constants.

ddtrace/contrib/langchain/patch.py
Yun-Kim: Ahh I did try doing this, but extracting the try/catch for the exception handling to a helper function breaks tracing - I'll add this as a follow up PR task.

emmettbutler (dismissed):

ddtrace/contrib/langchain/patch.py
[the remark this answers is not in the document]
Yun-Kim: Total cost calculations are not dependent on these propagations - propagating here is just a way to show the total cost of the trace as a whole (and root spans in subtraces) in the context of LangChain chains (which themselves don't incur any costs, the individual LLM/chat_model components incur costs). Our dashboard only calculates the total cost from non-chain spans, so we shouldn't have any worries here.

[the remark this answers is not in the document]
mabdinur: Yup with partial flushing we can still have a reference to the span object but the object is no longer in the span aggregator's trace buffer so the updated value will not be sent. So child spans will have update to date cost metrics but the value set on the parent span will be stale.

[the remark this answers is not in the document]
Yun-Kim: No worries, this functionally won't affect cost estimates for the individual child spans which is the most important thing here. We only send cost metrics for child spans, we only slap a cost trace metric (not dogstatsd metric) for parent spans which is only useful for trace visualization purposes.

ddtrace/contrib/langchain/patch.py
mabdinur: This ties into a suggestion I made earlier of listing all possible LangChain name tags in a constants file and then referencing those values in this integration. This way we can have a clear picture on the data that's captured and sent. 

Say if there is an issue with the following tag: `langchain.request.moonchain_llm.parameters.booo.hoooo` it would be very difficult to find where this tag is set in our library. You will need to step through the whole integration to find where this tag is generated dynamically. It's not the biggest concern but something to think about.

ddtrace/contrib/langchain/patch.py
```diff
@@ -0,0 +1,752 @@
...
+    @classmethod
+    def _metrics_tags(cls, span):
+        # type: (Span) -> list
+        provider = span.get_tag(PROVIDER) or ""
+        api_key = span.get_tag("langchain.request.api_key") or ""
+        tags = [
+            "version:%s" % (config.version or ""),
+            "env:%s" % (config.env or ""),
+            "service:%s" % (span.service or ""),
+            "langchain.request.provider:%s" % provider,
```
mabdinur, line 92: nit: We can format some of these strings to use the provider, type and model constants

Ex:
```suggestion
            "%s:%s" % (PROVIDER, provider),
```

ddtrace/contrib/langchain/patch.py
```diff
@@ -0,0 +1,752 @@
...
+    total_cost = span.get_metric("langchain.tokens.total_cost") or 0
+    if not propagate:
+        try:
+            completion_cost = get_openai_token_cost_for_model(
+                span.get_tag("langchain.request.model"),
+                span.get_metric("langchain.tokens.completion_tokens"),
+                is_completion=True,
+            )
+            prompt_cost = get_openai_token_cost_for_model(
+                span.get_tag("langchain.request.model"), span.get_metric("langchain.tokens.prompt_tokens")
```
mabdinur, line 170: We can use the MODEL constant here (this could be overkill but we can add a constant for token too)

ddtrace/contrib/langchain/patch.py
Yun-Kim: Good point - I agree this isn't ideal, but unfortunately with the nature of LangChain (it is an interface for 100+ LLM/embeddings/vectorstores providers) and each different provider has their own unique set of parameter names, meaning it would be pretty infeasible to keep up-to-date with this information by storing them as a constant.

The other alternative is to just stringify the parameters dictionary and store that as one big tag to avoid the cardinality, but that has its own downsides (granularity, visualization). WDYT?

mabdinur (dismissed):
Left some very minor nits about tags. Overall LGTM great work!

mabdinur (dismissed):
:shipit:

mabdinur (approved):

duncanista (approved):
Looks good on the guild side. :shipit:

emmettbutler (approved):

PR #6144, opened 2023-06-15
chore(telemetry): simplify otel configurations and add otel configs to app_started

mabdinur:
## Description

Currently otel support is configured by setting two environment variables. 

- Ensures ``OTEL_PYTHON_CONTEXT`` is set when otel is enabled and ddtrace is imported
   - Previously this envar was only set when ddtrace-run was used.
   - This simplifies otel configurations and allows us to remove ``OTEL_PYTHON_CONTEXT`` from public documentation.
- Sends remaining otel configuration, ``DD_TRACE_OTEL_ENABLED``, to metabase (via instrumentation)
- Tracks whether otel was enabled using ddtrace_run or programatically.

## Testing Strategy 

Manually verify that payloads are received and processed in metabase.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/opentelemetry/__init__.py
```diff
@@ -14,19 +14,19 @@
...
 When using ``ddtrace-run``, OpenTelemetry support can be enabled by setting
-the ``DD_TRACE_OTEL_ENABLED`` environment variable (the default value is ``False``).
+the ``DD_TRACE_OTEL_ENABLED`` environment variable to True (the default value is ``False``).
 
-OpenTelemetry support can be programmatically enabled by setting the ``OTEL_PYTHON_CONTEXT`` environment variable
-to ``ddcontextvars_context`` and setting the ``ddtrace.opentelemetry.TracerProvider``. These configurations
+OpenTelemetry support can be enabled programmatically by setting ``DD_TRACE_OTEL_ENABLED=True``
+and setting the ``ddtrace.opentelemetry.TracerProvider``. These configurations
 must be set before any OpenTelemetry Tracers are initialized::
+    import os
```
mabdinur, line 22:
```suggestion

    import os
```

ddtrace/bootstrap/sitecustomize.py
```diff
@@ -262,6 +259,7 @@ def _(threading):
 
         enable_appsec_rc()
 
+    config._ddtrace_bootstrapped = True
```
brettlangdon, line 262: is this accurate? should we instead emit an env var or something from `ddtrace/commands/ddtrace_run.py` ?

since `ddtrace/auto.py` just imports `ddtrace.bootstrap.sitecustomize`

mabdinur: I don't think it matters if `ddtrace-run` is used or `ddtrace.bootstrap.sitecustomize` or `ddtrace.auto` is imported. I just want to track if this file is executed.

mabdinur: I considered naming the telemetry configuration `ddtrace_was_boostraped` but that name feels a bit cryptic.

tests/telemetry/test_writer.py
```diff
@@ -88,6 +88,16 @@ def test_app_started_event(telemetry_lifecycle_writer, test_agent_session, mock_
                 "origin": "env_var",
                 "value": "['tracecontext', 'datadog']",
             },
+            {
+                "name": "ddtrace_run_used",
```
mabdinur, line 92: ?
```suggestion
                "name": "ddtrace_bootstraped",
```

ddtrace/bootstrap/sitecustomize.py
brettlangdon: 🤔 I _think_ we do want to know the difference?

since with `ddtrace-run` we are controlling imports and patching for a specific ordering, and with `import ddtrace.auto` they can still import that in an incorrect location.... ?

checking `sys.modules` could be "interesting" approach (not convinced this is a good idea, was just curious what it could look like)

```python
import sys


ddtrace_auto_used = "ddtrace.auto" in sys.modules
```

but with `ddtrace-run` we don't get `ddtrace.bootstrap.sitecustomize` since we change the path, so it gets imported as `sitecustomize`, but it could be any `sitecustomize`, not necessarily ours.


(these are random musings, not a request for a change)

mabdinur: Yup, we need to check sys.path to get if `ddtrace_run` was used.

mabdinur: At least for otel it does not matter. Both will configure the tracerprovider. But in general distinguishing between the two can help us resolve escalations.

Maybe we should send another configuration, `ddtrace_auto_used`.

ddtrace/internal/telemetry/writer.py
```diff
@@ -453,6 +454,21 @@ def _app_started_event(self):
                     "origin": "env_var",
                     "value": str(config._propagation_style_extract),
                 },
+                {
+                    "name": "ddtrace_bootstrapped",
```
brettlangdon, line 458: @mabdinur something @XG-xin and I were talking about for https://github.com/DataDog/dd-trace-py/pull/6148

we might want to try and store these values differently, e.g. store them globally or as a class attribute rather than writing statically into the payload like this. since we need to compare against previous values in order to track changes we don't want to have to look through the payload for the previous values.

just calling it out as something that could change on you.

mabdinur: Ahh okay. Ya this isn't the best approach. I just wanted to close out the otel work and then refactor later. I can hold off for now.

brettlangdon: I'll let the two of you decide how to handle. I don't see why order should matter too much, as long as you coordinate.

ddtrace/bootstrap/sitecustomize.py
```diff
@@ -203,15 +203,12 @@ def _(threading):
...
-    if sys.version_info >= (3, 7) and asbool(os.getenv("DD_TRACE_OTEL_ENABLED", False)):
+    if sys.version_info >= (3, 7) and config._otel_enabled:
         from opentelemetry.trace import set_tracer_provider
 
         from ddtrace.opentelemetry import TracerProvider
 
         set_tracer_provider(TracerProvider())
-        # Replaces the default otel api runtime context with DDRuntimeContext
-        # https://github.com/open-telemetry/opentelemetry-python/blob/v1.16.0/opentelemetry-api/src/opentelemetry/context/__init__.py#L53
-        os.environ["OTEL_PYTHON_CONTEXT"] = "ddcontextvars_context"
```
brettlangdon, line 214: does the order of this matter?

brettlangdon: we would be delaying to later, no?

ddtrace/settings/config.py
```diff
@@ -304,6 +304,12 @@ def __init__(self):
         self._ci_visibility_code_coverage_enabled = asbool(
             os.getenv("DD_CIVISIBILITY_CODE_COVERAGE_ENABLED", default=False)
         )
+        self._otel_enabled = asbool(os.getenv("DD_TRACE_OTEL_ENABLED", False))
+        if self._otel_enabled:
+            # Replaces the default otel api runtime context with DDRuntimeContext
+            # https://github.com/open-telemetry/opentelemetry-python/blob/v1.16.0/opentelemetry-api/src/opentelemetry/context/__init__.py#L53
+            os.environ["OTEL_PYTHON_CONTEXT"] = "ddcontextvars_context"
```
brettlangdon, line 311: could this be too late?

ddtrace/bootstrap/sitecustomize.py
mabdinur: Order doesn't really matter as long as this environment variable is set before the first otel span is generated.
This will actually occur earlier since the config object is one of the first things imported by ddtrace.

brettlangdon: what happens if we change this env var after the first span is created?

when does OTel look at this env variable?

ddtrace/internal/telemetry/writer.py
mabdinur: I'll reach out to @XG-xin. I would prefer to get this in as soon as possible.

ddtrace/settings/config.py
mabdinur: see above

emmettbutler (approved):

ddtrace/internal/telemetry/writer.py
XG-xin: @mabdinur I update my PR to store the configuration queue in the class attribute here:
https://github.com/DataDog/dd-trace-py/pull/6148/files#diff-0cb9315daf5ed5208de657467cc10a6b7dda616a161c9b5b1f8cff83c95894deR392
Does this look good to you?

ddtrace/bootstrap/sitecustomize.py
mabdinur: Nope. In that scenario a noop tracer provider is set and spans are not sent. 

A warning log is generated in this scenario. It's easy to debug.

PR #6172, opened 2023-06-20
fix(remoteconfig): handle process fork inside remote config client 

shatzi:
When a process fork, the remote config client need to have a new runtime-id and a client-id.

However, the state of the configuration should remain the same.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


brettlangdon (approved):

brettlangdon:
Can you update the PR title to be a `fix` and include a release note?

avara1986 (approved):

BaptisteFoy (approved):
LGTM for Remote Config. Thanks for this fix!

brettlangdon:
@shatzi can you add a release note, via `reno new <fix-some-slug-name-here>`

https://ddtrace.readthedocs.io/en/stable/releasenotes.html

ddtrace/internal/remoteconfig/worker.py
```diff
@@ -102,6 +102,7 @@ def start_subscribers(self):
         """Subscribers need to be restarted when application forks"""
         self._enable = False
         log.debug("[%s][P: %s] Remote Config Poller fork. Starting Pubsub services", os.getpid(), os.getppid())
+        self._client.renew_id()
```
lloeki, line 105: Watch out: it appears that the RC worker is operating in a thread.

Upon forking, typically all threads except the one that called `fork` are killed in the child. This means that if the thread is killed in the middle of updating the client state when processing an RC update the client state will be inconsistent and cannot be relied upon.

In that case generating a new client id and reusing the client internal state is insufficient. For Ruby we are going to simply create a new client instance (which makes sense since it is a new client): a clean state and new client id will be generated RAII style.

 https://github.com/DataDog/dd-trace-py/blob/e8a4476479a11c2b63d284848de424759be6e80f/ddtrace/internal/periodic.py#LL109C29-L109C29

lloeki: Also I would advise considering that there may be state tracking and stateful versions and what not backend side e.g the same client with a new id reports state+version+whatever that was meant for the previous client id and is tracked as such in the backend.

The backend may very well handle that case gracefully, or it may not, or it may today but not tomorrow. In any case I feel like it's asking for trouble and that it's simply more consistent to just create a new client as it is guaranteed to work by design (after all, it _is_ a new client!).

emmettbutler (commented):
I'm going to hold off on doing a deep review until the existing conversations are resolved.

ddtrace/internal/remoteconfig/worker.py
shatzi: Hi @lloeki, I was thinking about the same issues, but got different conclusions. @BaptisteFoy @arbll - maybe you can help us figure this out.

Lets start with how RC backend handle state. the backend is stateless and respects the state given by the client. compare the given state to the desired state and send back updated configs version if needed. So starting a new client-id with "non-empty" state is ok, as that is how the backend handle any requests.

For DI for example, keeping the state this "right" thing to do as all loaded probes (configurations) are already applied on the the forked process code (wrapped functions). Starting an "empty" state client would mean to unwrap all those functions and then reapply them when the backend send the same configurations. 

Maybe on ASM this is different, and we always need to start with an empty state? for example, IP blocking roles - I think it better to reuse already loaded roles on the forked process, no?

For me, it all about whether the forked process is the "same" services as its parent. if it is, then keep the state the client reduces load on both the backend and forked process as many of those configurations would be needed to apply. If the forked process is a different services than those configurations might no longer be relevant and clearing the state would be the right thing to do. And I'm not familiar enough in Python or Ruby to know which option is more common.

lloeki: While it is interesting to consider this "high level" logic for the client to optimise back and forth and have a ready to use RC client, the problem is deeper and lower level.

Consider this small bit of Ruby code:

```ruby
# fork.rb
class Foo
  attr_reader :foo, :bar

  def initialize
    @foo = []
    @bar = []
  end

  def update(a, b)
    # these two updates should be atomic to have a consistent state
    @mu.lock
    # here state is consistent
    @foo << a
    # here state is now inconsistent
    sleep 1 # artificially inflating time so that breakage reliably happens
    @bar << b
    # here state is consistent again
  ensure
    @mu.unlock
  end
end

foo = Foo.new

t1 = Thread.new do
  foo.update(1, 2)
end.run

sleep 0.5 # artificially controlling time so that breakage reliably happens
fork do
  puts "child"
  p t1
  p foo.instance_eval { @mu.locked? }
  p foo.foo
  p foo.bar
end

t1.join

puts "parent"
p foo.foo
p foo.bar
```

`ruby fork.rb` outputs:

```
child
#<Thread:0x0000000102fd0c98 fork.rb:18 dead>
false
[1]
[]
parent
[1]
[2]
```

Since `fork` is not under our control it can happen anytime, including during a state update, and irrespective of any lock.

The state-updating thread is killed, and the child internal state ends up being inconsistent, thus _cannot_ be relied upon. By internal state this may be the RC repository content, but also any other mutable state of the client; this may involve not just state that we set but language-internal stuff that could be interrupted or altered. Notice that e.g locks are also affected, as the mutex in the child is suddenly unlocked; This is the case for Ruby as it ensures forked mutexes are unlocked to be helpful (otherwise they would never be unlocked since the thread is dead in the child).

The quick and general purpose way out of it is to simply create a fresh new client and dispatch the old one.

To tackle the state carryover improvement, there could be other possibilities:

- resetting the forked client to a known valid state. This means creating some form of overall transaction mechanism over any mutable state, that would be set atomically WRT the `fork` call, and be able to be rolled back atomically as well.
- creating a new client but passing it known-consistent initialisation state. This means being able to save such state atomically WRT the `fork` call so that it can be passed later on.

shatzi: after hashing it out with @lloeki, we narrow the issue into that the that state of RC client and the state or the products could be out of sync after the fork.

For example:
1. a new configuration was in a process of been applied but didn't complete before the fork. so the client will not have on it state but it was applied by the process.
2. a configuration was removed from the state but product instrumentation was not yet removed before the fork.

In order to ensure that the rc-client state is the same the product state we need to one of two solutions:
1. **empty-state after fork** -  the rc-client after fork should notify all products to "remove all applied configurations" and then report an empty state to rc.
2. **ensure-current-state after fork** - the rc-client would notify all process on the "current" state on which configuration are applied. if the product see the configuration already applied, do nothing.

mabdinur: @shatzi, are you planning to make any more changes to this PR? Is this issue something that can be addressed in a follow up PR or is it a hard blocker?

mabdinur (approved):

emmettbutler:
I converted this pull request to a draft to reflect the fact that there are still open questions. Please set it to "ready for review" when everything has been resolved.

ddtrace/internal/remoteconfig/worker.py
shatzi: @mabdinur - I don't think this change add new issues for handling a fork. all those issue already existed before. I would vote for merging this and trying to address state corruption on a future PRs

P403n1x87: I think this change is good as is. The state issues can be tackled in follow-up work.

P403n1x87 (approved):

PR #6181, opened 2023-06-21
chore(tracing): update `rq` to support new `enqueue_many` API

jbertran:
## Changes

`rq` is a job-scheduling module.

In version `1.9.0`, the enqueue_many API was introduced, allowing to batch-queue jobs, which didn’t cause issues because under the hood it called the patched enqueue_job method.

As part of changes to 1.14.0, this enqueue_many API and the original enqueue_job API both use the private _enqueue_job method, which isn’t traced.

This means that:
1. we should test that `enqueue_many` generates traces both in `>=1.9.0` and `latest`
2. we should conditionally patch `_enqueue_job` instead of `enqueue_job` in `>=1.14.0

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


emmettbutler (commented):
There are a bunch of failing CI checks, so I think it makes sense to wait to do a proper review until those are resolved. Let me know if you'd like a review in the current state.

releasenotes/notes/add-enqueue_many-support-to-rq-contrib-1e20380519c153d4.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    tracing: adds support for `enqueue_many` in `rq` versions 1.14.0 and above. The underlying job queuing function had
```
ZStriker19, line 4: I think we want double tick marks for release notes.

```suggestion
    tracing: adds support for ``enqueue_many`` in ``rq`` versions 1.14.0 and above. The underlying job queuing function had
```

tests/snapshots/tests.contrib.rq.test_rq.test_enqueue_many.json
```diff
@@ -0,0 +1,228 @@
+[[
+  {
+    "name": "rq.queue.enqueue_job",
+    "service": "rq",
+    "resource": "tests.contrib.rq.jobs.job_add1",
+    "trace_id": 0,
+    "span_id": 1,
+    "parent_id": 0,
```
ZStriker19, line 8: So we'd expect 3 spans with parent_id 0 for each job that we enqueue, and then the children for each of them while running them through the worker. That makes sense to me I think.

ZStriker19 (dismissed):

ddtrace/contrib/rq/__init__.py
```diff
@@ -78,13 +78,16 @@
 """
 import os
 
+import rq
```
brettlangdon, line 81: we cannot do this import here.

since this integration uses only `contrib/rq/__init__.py` and not the usual `required_modules` + `/patch.py` approach, we cannot import potentially unavailable modules here (hence why we do `import rq` in `def patch():`

brettlangdon (changes requested):

ddtrace/contrib/rq/__init__.py
brettlangdon: This is also why the `build_docs` job is failing.

jbertran: I figured this was why, didn't have much time to get back to it. I'll try splitting it to match the usual approach.

ddtrace/contrib/rq/__init__.py
```diff
@@ -116,6 +119,9 @@
     ),
 )
 
+rq_version_str = getattr(rq, "__version__", "0.0.0")
+rq_version = parse_version(rq_version_str)
```
brettlangdon, line 123:
```suggestion
def _get_version(rq):
    rq_version_str = getattr(rq, "__version__", "0.0.0")
    rq_version = parse_version(rq_version_str)
```

you could also do something like this instead, since the two places where you are trying to use the version you have access to the `rq` module already.

maybe not idea for the `traced_queue_enqueue_job` (to recompute/parse every time), but would fit this setup better.

tabgok (dismissed):

ddtrace/contrib/rq/patch.py
```diff
@@ -0,0 +1,201 @@
...
+config._add(
+    "rq_worker",
+    dict(
+        distributed_tracing_enabled=asbool(os.environ.get("DD_RQ_DISTRIBUTED_TRACING_ENABLED", True)),
+        _default_service=schematize_service_name("rq-worker"),
+    ),
+)
+
+rq_version_str = getattr(rq, "__version__", "0.0.0")
+rq_version = parse_version(rq_version_str)
```
brettlangdon, line 39: Instead of moving everything out of `__init__.py` the alternative could have been to just put this in a function.

Since most of the methods in here accept `rq` as an argument rather than relying on the globally imported module.

```python
def get_rq_version(rq):
    rq_version_str = getattr(rq, "__version__", "0.0.0")
    return rq_version = parse_version(rq_version_str)

... 


rq_version = get_rq_version(rq)
if rq_version >= (1, 14, 0):
    pass
else:
    pass

```


And since we only ever use the rq version from `patch`/`unpatch` we don't have to worry too much about the overhead of `get_rq_version()` calls.

What you have here also works, but if you want an approach that is less changes, this is an approach we could take.

emmettbutler (dismissed):

ddtrace/contrib/rq/patch.py
jbertran: I thought I'd align on the "usual" way to structure the contrib, in the hopes that this would make refactors easier, especially given @emmettbutler 's current work on contribs.

ZStriker19:
Closing this for now, feel free to reopen if we want to continue work.

PR #6190, opened 2023-06-21
refactor: move V2 logwriter into internal module

Yun-Kim:
This PR moves the `V2LogWriter` class into `ddtrace.internal.log_writer` from `ddtrace.contrib.openai._logging` as this will be reused by the LangChain integration, and will no longer be specific to the OpenAI integration.

The test file `tests.contrib.openai.test_logging` will remain in the `contrib.openai` module as this reflects the openai integration.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.

### DIFF
--- a/ddtrace/internal/log_writer.py
+++ b/ddtrace/internal/log_writer.py
@@ -72,7 +72,6 @@ class V2LogWriter(PeriodicService):
             self._buffer.append(log)
 
     def on_shutdown(self):
-        # type: (...) -> None
         self.periodic()
 
     @property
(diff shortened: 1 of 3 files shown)


ddtrace/internal/log_writer.py
```diff
@@ -72,7 +72,6 @@ def enqueue(self, log):
             self._buffer.append(log)
 
     def on_shutdown(self):
-        # type: (...) -> None
```
mabdinur, line 75: Should we revert this change?

mabdinur (approved):

ddtrace/internal/log_writer.py
Yun-Kim: This is a slightly bigger issue than type hinting - mypy (which isn't enabled for contrib files I believe, so it's only flagging it now that we've moved this to `internal`) flags this as the superclass method is actually a static method (so no reference to `self`) and thus has a different function signature `# type: () -> Any`. 

I think we should probably address this through a redesign, but that would be a lot of work and unnecessary for the scope of this PR. I made the decision to ignore this for now and address it as a future PR.

mabdinur (approved):

PR #6196, opened 2023-06-22
chore(peer.service): perf - process traces instead of spans

tabgok:
Prior to this change, the peer.service processor (PeerServiceProcessor) processed invidual spans on_finish.  After this change, the processor now operates at a (partial) trace level.  Whenever a whole or part of a trace is being completed, all spans will be processed at once.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.

### DIFF
--- a/ddtrace/internal/processor/trace.py
+++ b/ddtrace/internal/processor/trace.py
@@ -332,40 +332,41 @@ class SpanSamplingProcessor(SpanProcessor):
                     break
 
 
-class PeerServiceProcessor(SpanProcessor):
+class PeerServiceProcessor(TraceProcessor):
     def __init__(self, peer_service_config):
         self._config = peer_service_config
         self._set_defaults_enabled = self._config.set_defaults_enabled
+        self._mapping = self._config.peer_service_mapping
 
-    def on_span_start(self, span):
-        """
-        We don't do anything on span start
-        """
-        pass
+    def process_trace(self, trace):
+        if not trace:
+            return
 
-    def on_span_finish(self, span):
-        if self._set_defaults_enabled:
-            self._set_span_tag(span)
-        self._remap_peer_service(span)  # Remap regardless of whether defaults are enabled
+        traces_to_process = []
+        if not self._set_defaults_enabled:
+            traces_to_process = filter(lambda x: x.get_tag(self._config.tag_name), trace)
+        else:
+            traces_to_process = filter(
+                lambda x: x.get_tag(self._config.tag_name) or x.get_tag(SPAN_KIND) in self._config.enabled_span_kinds,
+                trace,
+            )
+        any(map(lambda x: self._update_peer_service_tags(x), traces_to_process))
 
-    def _set_span_tag(self, span):
-        if span.get_tag(self._config.tag_name):  # If the tag already exists, assume it is user generated
+        return trace
+
+    def _update_peer_service_tags(self, span):
+        tag = span.get_tag(self._config.tag_name)
+
+        if tag:  # If the tag already exists, assume it is user generated
             span.set_tag_str(self._config.source_tag_name, self._config.tag_name)
-            return
+        else:
+            for data_source in self._config.prioritized_data_sources:
+                tag = span.get_tag(data_source)
+                if tag:
+                    span.set_tag_str(self._config.tag_name, tag)
+                    span.set_tag_str(self._config.source_tag_name, data_source)
+                    break
 
-        if span.get_tag(SPAN_KIND) not in self._config.enabled_span_kinds:
-            return
-
-        for data_source in self._config.prioritized_data_sources:
-            peer_service_definition = span.get_tag(data_source)
-            if peer_service_definition:
-                span.set_tag_str(self._config.tag_name, peer_service_definition)
-                span.set_tag_str(self._config.source_tag_name, data_source)
-                return
-
-    def _remap_peer_service(self, span):
-        current_peer_service = span.get_tag(self._config.tag_name)
-
-        if current_peer_service in self._config.peer_service_mapping:
-            span.set_tag_str(self._config.remap_tag_name, current_peer_service)
-            span.set_tag_str(self._config.tag_name, self._config.peer_service_mapping[current_peer_service])
+        if tag in self._mapping:
+            span.set_tag_str(self._config.remap_tag_name, tag)
+            span.set_tag_str(self._config.tag_name, self._config.peer_service_mapping[tag])
--- a/ddtrace/tracer.py
+++ b/ddtrace/tracer.py
@@ -139,13 +139,12 @@ def _default_span_processors_factory(
     single_span_sampling_rules,  # type: List[SpanSamplingRule]
     agent_url,  # type: str
     profiling_span_processor,  # type: EndpointCallCounterProcessor
-    peer_service_processor,  # type: PeerServiceProcessor
 ):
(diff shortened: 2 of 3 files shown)


ddtrace/internal/processor/trace.py
```diff
@@ -332,21 +332,19 @@ def on_span_finish(self, span):
...
+            return
 
-    def on_span_finish(self, span):
-        if self._set_defaults_enabled:
-            self._set_span_tag(span)
-        self._remap_peer_service(span)  # Remap regardless of whether defaults are enabled
+        for span in trace:
+            if self._set_defaults_enabled:
+                self._set_span_tag(span)
+            self._remap_peer_service(span)  # Remap regardless of whether defaults are enabled
```
brettlangdon, line 347: another possible optimization here:

```python
def _remap_peer_service(self, span):
    if not self._config.peer_service_mapping:
        return
    
    current_peer_service = span.get_tag(self._config.tag_name)

    # I am not sure if this is an optimization vs just doing the `in` check?
    if not current_peer_service:
        return

    if current_peer_service in self._config.peer_service_mapping:
        span.set_tag_str(self._config.remap_tag_name, current_peer_service)
        span.set_tag_str(self._config.tag_name, self._config.peer_service_mapping[current_peer_service])
```

ddtrace/internal/processor/trace.py
```diff
@@ -186,8 +186,8 @@ class _Trace(object):
     _span_api_to_count = attr.ib(
         init=False,
         factory=lambda: {
-            "spans_created": defaultdict(int),
-            "spans_finished": defaultdict(int),
+            "span_created": defaultdict(int),
```
brettlangdon, line 189: bad merge conflict resolution?

ddtrace/tracer.py
```diff
@@ -145,7 +145,7 @@ def _default_span_processors_factory(
     # FIXME: type should be AppsecSpanProcessor but we have a cyclic import here
     """Construct the default list of span processors to use."""
     trace_processors = []  # type: List[TraceProcessor]
-    trace_processors += [TraceTagsProcessor()]
+    trace_processors += [TraceTagsProcessor(), peer_service_processor]
```
mabdinur, line 148: Can we remove the `_ peer_service_processor` property from the tracer and just define it here?

```suggestion
    trace_processors += [TraceTagsProcessor(), PeerServiceProcessor(PeerServiceConfig())]
```

tabgok: 1. Sure! The code has three patterns (pass in a processor, pass in a processor class, and insantiate without passing in), so I just picked one.  If we prefer this pattern, that's fine by me.

2. Already updated, it is now `process_trace` instead of `on_finish`

3. Is the performance improvement worth the loss of clarity?  The TraceTagsProcessor adds trace-level tags, which is different than adding span-level tags for all spans in a trace.  We can combine them and do both, but I don't think the performance change in this case is significant enough to warrant a refactor.  Open to debate or override, though!

mabdinur: We can hold off on 3 for now

tests/internal/peer_service/test_processor.py
```diff
@@ -30,48 +30,57 @@ def test_span():
...
+def test_processing_peer_service_exists(processor, test_trace, span_kind, peer_service_config):
     processor._set_defaults_enabled = True
-    test_span.set_tag(SPAN_KIND, span_kind)
-    test_span.set_tag(peer_service_config.tag_name, "fake_peer_service")
-    test_span.set_tag("out.host", "fake_falue")  # Should not show up
-    processor.on_span_finish(test_span)
+    span = test_trace[0]
+    span.set_tag(SPAN_KIND, span_kind)
+    span.set_tag(peer_service_config.tag_name, "fake_peer_service")
+    span.set_tag("out.host", "fake_falue")  # Should not show up
```
mabdinur, line 44:
```suggestion
    span.set_tag_str(SPAN_KIND, span_kind)
    span.set_tag_str(peer_service_config.tag_name, "fake_peer_service")
    span.set_tag_str("out.host", "fake_falue")  # Should not show up
```

tests/internal/peer_service/test_processor.py
```diff
@@ -30,48 +30,57 @@ def test_span():
...
 @pytest.mark.parametrize("span_kind", [SpanKind.SERVER, SpanKind.CONSUMER])
-def test_nothing_happens_for_server_and_consumer(processor, test_span, span_kind, peer_service_config):
+def test_nothing_happens_for_server_and_consumer(processor, test_trace, span_kind, peer_service_config):
+    span = test_trace[0]
     processor._set_defaults_enabled = True
-    test_span.set_tag(SPAN_KIND, span_kind)
-    test_span.set_tag("out.host", "fake_host")
-    processor.on_span_finish(test_span)
+    span.set_tag(SPAN_KIND, span_kind)
+    span.set_tag("out.host", "fake_host")
```
mabdinur, line 56:
```suggestion
    span.set_tag_str(SPAN_KIND, span_kind)
    span.set_tag_str("out.host", "fake_host")
```

ZStriker19 (approved):

ddtrace/internal/processor/trace.py
```diff
@@ -332,40 +332,41 @@ def on_span_finish(self, span):
...
-            return
+        traces_to_process = []
+        if not self._set_defaults_enabled:
+            traces_to_process = filter(lambda x: x.get_tag(self._config.tag_name), trace)
+        else:
+            traces_to_process = filter(
+                lambda x: x.get_tag(self._config.tag_name) or x.get_tag(SPAN_KIND) in self._config.enabled_span_kinds,
+                trace,
+            )
+        any(map(lambda x: self._update_peer_service_tags(x), traces_to_process))
```
brettlangdon, line 353: I find this code really hard to follow. I don't see why this is easier/better than just:

```python
for span in trace:
    if span_shouldnt_be_processed: continue
    
    # process span
```

It also seems like we are probably calling `get_tag(self._config.tag_name)` one more time than is required (ok, maybe that is being pedantic/optimizing too much, but calling it out)

ddtrace/internal/processor/trace.py
```diff
@@ -332,40 +332,41 @@ def on_span_finish(self, span):
...
-    def _set_span_tag(self, span):
-        if span.get_tag(self._config.tag_name):  # If the tag already exists, assume it is user generated
-            span.set_tag_str(self._config.source_tag_name, self._config.tag_name)
+    def process_trace(self, trace):
+        if not trace:
             return
 
-        if span.get_tag(SPAN_KIND) not in self._config.enabled_span_kinds:
-            return
+        traces_to_process = []
```
brettlangdon, line 345: This should be called `spans_to_process` right? since this function takes a single trace/`List[Span]`

ddtrace/internal/processor/trace.py
brettlangdon: we should remove the usage of `any()` here, it doesn't help legibility.

`any()` is used to iterate over the `map()` because `map()` gives us a `map object` and won't actually iterate over `traces_to_process` until we iterate over the `map object`.

but `any()` will stop after the first successful (truthy) response from `_update_peer_service_tags`. It works here because `_update_peer_service_tags` doesn't return anything (so `None)` which is falsey. But if someone updates `_update_peer_service_tags` to return something truthy in the future, this might start to fail silently without anyone knowing why.

`list()` would be better here instead of `any()` (also comments on why)


I also think this means we are iterating over spans multiple times? Once to filter, and then once again over the spans that need processing?

PR #6199, opened 2023-06-23
chore(asm/iast): update iast installation

avara1986:
# Description
This PR is part of a bigger PR https://github.com/DataDog/dd-trace-py/pull/6108

## Update setup
Extend the `build_ext` command to build C++ binaries with CMake following the [pybind11 recommendations](https://github.com/pybind/).

## Pybind11
We vendorize pybind11 to fix the build wheel files. Why do we use this approach?

- If we add pybind11 as a dependency, we need to find the Python binary in our CMakeList file. With that, the Macos ARM wheel build fails because CMake found the wrong Python version:
  ```
  ld: warning: ignoring file /Library/Frameworks/Python.framework/Versions/3.8/lib/libpython3.8.dylib, building for macOS-arm64 
  but attempting to link with a file built for macOS-x86_64
  ```
- Our first approach was to add pybind11 in `ddtrace/vendor` and link it in `ddtrace/appsec/iast`, but it raised problems with Windows and Alpine.
- Our second approach was to add it as a git submodule, but this change forced us to update all CI steps. If we want to install ddtrace with `pip install git@github...`, it raises errors with submodules.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


juanjux (approved):

riotfile.py
```diff
@@ -172,11 +172,12 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
...
                 Venv(
                     name="codespell",
-                    command='codespell --skip="ddwaf.h" ddtrace/ tests/',
+                    command='codespell --skip="ddwaf.h" '
+                    '--skip="ddtrace/appsec/iast/_taint_tracking/vendor/*" ddtrace/ tests/',
                 ),
                 Venv(
                     name="hook-codespell",
-                    command="codespell {cmdargs}",
+                    command='codespell --skip="ddtrace/appsec/iast/_taint_tracking/vendor/*" {cmdargs}',
```
majorgreys, line 180: This could be moved to the codespell configuration in the `setup.cfg`.

majorgreys (commented):
Can you move rename it to `_vendor` so that it is not in the public interface?

juanjux (dismissed):

juanjux (approved):

tests/smoke_test.py
```diff
@@ -6,11 +6,18 @@
 
 
 if __name__ == "__main__":
+    # ASM IAST smoke test
+    if sys.version_info >= (3, 6, 0) and system() != "Windows":
+        from ddtrace.appsec.iast._taint_tracking._native import ops
+
+        assert ops
```
mabdinur, line 13: Should we log a message when this assertion fails? Is it sufficient to check if ops is a truthy value? Should we add any additional validations?

mabdinur (approved):
Changes to the shared files looks good to me (.github/*, mypy.ini, and pyproject.toml)

tabgok (approved):

tests/smoke_test.py
avara1986: Yes, if IAST is enabled but the SO isn't supported IAST could raise a ImportError

PR #6208, opened 2023-06-26
chore(dsm): handle when dd header not present for consuming service

shethhriday29:
Previously, if a service receiving SQS messages didn't find a datadog header in the message that included the data streams context, it would log a warning message and not set a checkpoint, thereby not relaying any information to the dsm processor. However, other DSM technologies/runtimes handle the edge case by still setting the pathway to `None` and using this pathway to call `set_checkpoint`.

PR was tested by running all the botocore tests, including the DSM tests, to ensure they still work. No new technical risks are introduced by this PR.

## Reviewer Checklist

- [x] Title is accurate.

### DIFF
--- a/ddtrace/contrib/botocore/patch.py
+++ b/ddtrace/contrib/botocore/patch.py
@@ -519,6 +519,7 @@ def patched_api_call(original_func, instance, args, kwargs):
                 try:
                     if "Messages" in result:
                         for message in result["Messages"]:
+                            pathway = None
                             if (
                                 "MessageAttributes" in message
                                 and "_datadog" in message["MessageAttributes"]
@@ -527,9 +528,6 @@ def patched_api_call(original_func, instance, args, kwargs):
                                 pathway = json.loads(message["MessageAttributes"]["_datadog"]["StringValue"]).get(
                                     PROPAGATION_KEY_BASE_64, None
                                 )
-                            else:
-                                pathway = None
-
                             ctx = pin.tracer.data_streams_processor.decode_pathway_b64(pathway)
                             ctx.set_checkpoint(["direction:in", "topic:" + queue_name, "type:sqs"])
 
--- a/ddtrace/contrib/botocore/patch.py
+++ b/ddtrace/contrib/botocore/patch.py
@@ -524,27 +524,14 @@ def patched_api_call(original_func, instance, args, kwargs):
                                 and "_datadog" in message["MessageAttributes"]
                                 and "StringValue" in message["MessageAttributes"]["_datadog"]
                             ):
-                                if PROPAGATION_KEY_BASE_64 in json.loads(
-                                    message["MessageAttributes"]["_datadog"]["StringValue"]
-                                ):
-                                    pathway = json.loads(message["MessageAttributes"]["_datadog"]["StringValue"])[
-                                        PROPAGATION_KEY_BASE_64
-                                    ]
-                                    ctx = pin.tracer.data_streams_processor.decode_pathway_b64(pathway)
-                                    ctx.set_checkpoint(["direction:in", "topic:" + queue_name, "type:sqs"])
-
-                                else:
-                                    log.debug(
-                                        "Unable to find data streams context for the message. Data "
-                                        "streams monitoring was not enabled by the application "
-                                        "sending the message at the time of the send."
-                                    )
-                            else:
-                                log.debug(
-                                    "Unable to find trace context for the message. Please ensure that your Datadog "
-                                    "agent is correctly configured on the application sending the message and that "
-                                    "the SQS messages being sent have 9 or less Message Attributes."
+                                pathway = json.loads(message["MessageAttributes"]["_datadog"]["StringValue"]).get(
+                                    PROPAGATION_KEY_BASE_64, None
                                 )
+                            else:
+                                pathway = None
+
+                            ctx = pin.tracer.data_streams_processor.decode_pathway_b64(pathway)
+                            ctx.set_checkpoint(["direction:in", "topic:" + queue_name, "type:sqs"])
 
                 except Exception:
                     log.debug("Error receiving SQS message with data streams monitoring enabled", exc_info=True)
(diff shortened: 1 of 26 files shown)


ddtrace/contrib/botocore/patch.py
```diff
@@ -524,27 +524,14 @@ def patched_api_call(original_func, instance, args, kwargs):
...
-                                    "Unable to find trace context for the message. Please ensure that your Datadog "
-                                    "agent is correctly configured on the application sending the message and that "
-                                    "the SQS messages being sent have 9 or less Message Attributes."
+                                pathway = json.loads(message["MessageAttributes"]["_datadog"]["StringValue"]).get(
+                                    PROPAGATION_KEY_BASE_64, None
                                 )
+                            else:
+                                pathway = None
+
+                            ctx = pin.tracer.data_streams_processor.decode_pathway_b64(pathway)
```
tabgok, line 533: does this need to be added to the MessageAttributes?

mabdinur: `decode_pathway_b64()` expects a string. Passing None should raise an Error.

releasenotes/notes/botocore_dsm_sqs_no_header-15e92adad81448ea.yaml
```diff
@@ -0,0 +1,7 @@
+---
+other:
```
mabdinur, line 2: If this is a fix for an unreleased change then a release note is not necessary

mabdinur (commented):
Can we add a regression test. Some thing that validates that set_checkpoint sets the expected tags.

ddtrace/contrib/botocore/patch.py
tabgok: discussed offline, the answer is "no", as we are currently only looking at in/out pairs.

ddtrace/contrib/botocore/patch.py
```diff
@@ -524,27 +524,14 @@ def patched_api_call(original_func, instance, args, kwargs):
...
-                            else:
-                                log.debug(
-                                    "Unable to find trace context for the message. Please ensure that your Datadog "
-                                    "agent is correctly configured on the application sending the message and that "
-                                    "the SQS messages being sent have 9 or less Message Attributes."
+                                pathway = json.loads(message["MessageAttributes"]["_datadog"]["StringValue"]).get(
+                                    PROPAGATION_KEY_BASE_64, None
                                 )
+                            else:
+                                pathway = None
```
tabgok, line 531: nit: its a little more readable to set this as None before the if/else

shethhriday29: done!

emmettbutler (approved):

ddtrace/contrib/botocore/patch.py
```diff
@@ -519,32 +519,17 @@ def patched_api_call(original_func, instance, args, kwargs):
...
-                            else:
-                                log.debug(
-                                    "Unable to find trace context for the message. Please ensure that your Datadog "
-                                    "agent is correctly configured on the application sending the message and that "
-                                    "the SQS messages being sent have 9 or less Message Attributes."
+                                pathway = json.loads(message["MessageAttributes"]["_datadog"]["StringValue"]).get(
+                                    PROPAGATION_KEY_BASE_64, None
                                 )
+                            ctx = pin.tracer.data_streams_processor.decode_pathway_b64(pathway)
+                            ctx.set_checkpoint(["direction:in", "topic:" + queue_name, "type:sqs"])
```
mabdinur, line 532: We should avoid calling ``data.encode("utf-8")`` when data is None. Handling an AttributeError here is less than ideal. 

Instead we can just create a new pathway directly. With this suggestion we can revert handling the ``AttributeError`` in `decode_pathway_b64()`

```suggestion
                               ctx = pin.tracer.data_streams_processor.decode_pathway_b64(pathway)
                            else:
                               ctx = pin.tracer.data_streams_processor.new_pathway()
                            ctx.set_checkpoint(["direction:in", "topic:" + queue_name, "type:sqs"])
```

Feel free to experiment with alternative approaches.

ddtrace/contrib/botocore/patch.py
```diff
@@ -519,32 +519,17 @@ def patched_api_call(original_func, instance, args, kwargs):
                 try:
                     if "Messages" in result:
                         for message in result["Messages"]:
+                            pathway = None
```
mabdinur, line 522: Should we move `pathway = None` out of the for loop (ie move it to line 521)

ddtrace/contrib/botocore/patch.py
mabdinur, line 532: We should avoid calling data.encode("utf-8") when data is None. Handling an AttributeError here is less than ideal.

Instead we can just create a new pathway directly. With this suggestion we can revert handling the AttributeError in decode_pathway_b64()

```suggestion
                            if pathway is not None:
                               ctx = pin.tracer.data_streams_processor.decode_pathway_b64(pathway)
                            else:
                               ctx = pin.tracer.data_streams_processor.new_pathway()
                            ctx.set_checkpoint(["direction:in", "topic:" + queue_name, "type:sqs"])
```

ddtrace/internal/datastreams/processor.py
```diff
@@ -262,10 +262,13 @@ def decode_pathway(self, data):
...
-        binary_pathway = data.encode("utf-8")
-        encoded_pathway = base64.b64decode(binary_pathway)
-        data_streams_context = self.decode_pathway(encoded_pathway)
-        return data_streams_context
+        try:
+            binary_pathway = data.encode("utf-8")
+            encoded_pathway = base64.b64decode(binary_pathway)
+            data_streams_context = self.decode_pathway(encoded_pathway)
+            return data_streams_context
+        except AttributeError:
```
mabdinur, line 270: If the change above is accepted we can revert this or at least we should change the method signature to `type: Optional[str] -> DataStreamsCtx ` and then handle the case were data is None without raise an Exception.

ddtrace/contrib/botocore/patch.py
shethhriday29: we need to reset `pathway` to `None` for every message, we could have a case where some messages being received may have the `PROPAGATION_KEY_BASE_64` and others may not, and for messages that don't, we don't want other messages' `PROPAGATION_KEY_BASE_64` being used (rather just `None`)

mabdinur (approved):

PR #6228, opened 2023-06-28
chore(internal): collect runtimemetrics_enabled telemetry configuration

brettlangdon:
This change reports a new telemetry configuration value `runtimemetrics_enabled` which will let us know which customers are using the [runtime metrics](https://docs.datadoghq.com/tracing/metrics/runtime_metrics/python/) feature.

This configuration uses a slightly different approach from other ones we have.

1. We set the initially reported value from the env variable `DD_RUNTIME_METRICS_ENABLED`
2. We have the `RuntimeWorker` update the configuration when it's `enable` and `disable` methods are called.

This way we will likely get the correct value initially from the env variable, but if the runtime worker gets enabled via code or dynamic configuration we'll get the expected value.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


brettlangdon:
The one area that I need to still investigate is how to test the behavior of `RuntimeWorker.enable`/`RuntimeWorker.disable`.

emmettbutler (dismissed):

ddtrace/internal/telemetry/writer.py
```diff
@@ -456,6 +456,7 @@ def _app_started_event(self):
                 ("ddtrace_bootstrapped", config._ddtrace_bootstrapped, "unknown"),
                 ("ddtrace_auto_used", "ddtrace.auto" in sys.modules, "unknown"),
                 ("otel_enabled", config._otel_enabled, "unknown"),
+                ("runtimemetrics_enabled", asbool(os.getenv("DD_RUNTIME_METRICS_ENABLED", default=False)), "unknown"),
```
mabdinur, line 459: This could be beyond the scope of this PR but should we add a `runtime_metrics` property to the global config object. I think we're moving away from reading environment variables directly. Ideally the global config object should be the only interface for setting and updating configurations. This has the added benefit of "locking in" configurations on application start up (these configurations should only change via RC).

mabdinur, line 459: Should we use the constant `TELEMETRY_RUNTIMEMETRICS_ENABLED` constant here? Can define this constant in ddtrace/internal/telemetry/constants.py

mabdinur (approved):
Left some nits. Overall looks good to me

ddtrace/internal/telemetry/writer.py
brettlangdon: I wanted to read from runtime metrics directly but got a circular reference issue.

Adding to `ddtrace.config` would be good, but small issue there is that `runtime_metrics` doesn't use that to check for enabling/disabling, we have the `ddtrace.runtime.RuntimeMetrics.enable()` API. 🤔

mabdinur (approved):

PR #6236, opened 2023-06-29
chore(ci-visibility): test suite skipping using ``pytest_collection_modifyitems``

gnufede:
CI Visibility: Switch Intelligent Test Runner from test skipping mode to test suite skipping mode.

The main reason to use this approach over doing test suite skipping with ``pytest_ignore_collect`` hook or skipping the tests individually inside ``pytest_runtest_protocol`` hook is that this hooks allows us to add the skip marker before ``pytest_runtest_protocol`` is run, so we can decide wether to create the span for it or skip it altogether.

This PR skips the creation of spans for skipped tests, suites and modules, to act as the rest of the languages act while using ITR. 

Main changes it involves:

1. Coverage is now measured and sent by test suite instead of test, hence dropping the ``cover`` context manager that could be used inside the ``pytest_runtest_protocol`` hook to a singleton to be reused until the test suite finishes.
2. By leveraging the ``pytest_collection_modifyitems`` hook, we can add skip markers to the tests right after the test collection step, so we have that information at ``pytest_runtest_protocol`` hook.
3. Some changes in the requests to the backend, to specify we want the skippable tests at test suite level.

Since ITR is not yet beta it doesn't have a release note.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


emmettbutler (approved):

ddtrace/contrib/pytest/plugin.py
```diff
@@ -306,111 +310,132 @@ def _get_test_class_hierarchy(item):
...
-def pytest_runtest_protocol(item, nextitem):
-    if not _CIVisibility.enabled:
-        yield
-        return
-    test_session_span = _extract_span(item.session)
+def get_path(item):
+    if hasattr(item, "path"):
+        return str(item.path)
+    parent = item
+    while not hasattr(parent, "fspath"):
```
romainkomorndatadog, line 317: I was thinking that if we get to a point where `parent.parent == None`, we'll end up with ```AttributeError: 'NoneType' object has no attribute 'parent'```, but I actually think that's okay because it would be a weird enough scenario that raising an exception makes sense.

ddtrace/contrib/pytest/plugin.py
```diff
@@ -306,111 +310,132 @@ def _get_test_class_hierarchy(item):
...
+        pytest_module_item = _find_pytest_item(item, pytest.Module)
+        pytest_package_item = _find_pytest_item(pytest_module_item, pytest.Package)
+
+        test_module_span = _extract_span(pytest_package_item)
+        if pytest_package_item is not None and test_module_span is None:
+            if test_module_span is None:
+                test_module_span = _start_test_module_span(pytest_package_item)
+
+        test_suite_span = _extract_span(pytest_module_item)
+        if pytest_module_item is not None and test_suite_span is None:
```
romainkomorndatadog, line 356: When would `pytest_module_item is None` not be some kind of error state? And should `_extract_span` above be done on `pytest_module_item` before checking that it isn't `None`?

ddtrace/contrib/pytest/plugin.py
```diff
@@ -306,111 +310,132 @@ def _get_test_class_hierarchy(item):
...
+        ) as span:
+            span.set_tag_str(COMPONENT, "pytest")
+            span.set_tag_str(SPAN_KIND, KIND)
+            span.set_tag_str(test.FRAMEWORK, FRAMEWORK)
+            span.set_tag_str(_EVENT_TYPE, SpanTypes.TEST)
+            span.set_tag_str(test.NAME, item.name)
+            span.set_tag_str(test.COMMAND, _get_pytest_command(item.config))
+            span.set_tag_str(_SESSION_ID, str(test_session_span.span_id))
+
+            if test_module_span is not None:
```
romainkomorndatadog, line 378: I'm tempted to say we've already checked for `test_module_span == None` above. Maybe we should let this be some kind of exception (that we potentially catch and log)?

ddtrace/contrib/pytest/plugin.py
```diff
@@ -306,111 +310,132 @@ def _get_test_class_hierarchy(item):
...
-
-    nextitem_pytest_package_item = _find_pytest_item(nextitem, pytest.Package)
-    if test_module_span is not None and (
-        nextitem is None or nextitem_pytest_package_item != pytest_package_item and not test_module_span.finished
-    ):
-        _mark_test_status(pytest_package_item, test_module_span)
-        test_module_span.finish()
+        nextitem_pytest_module_item = _find_pytest_item(nextitem, pytest.Module)
+        if test_suite_span is not None and (
+            nextitem is None or nextitem_pytest_module_item != pytest_module_item and not test_suite_span.finished
```
romainkomorndatadog, line 425: Maybe `nextitem is None` is redundant, because then it should be guaranteed that `nextitem_pytest_module_item != pytest_module_item`.

romainkomorndatadog (approved):
No blocking comments.

ddtrace/contrib/pytest/plugin.py
gnufede: It can happen, maybe for runtime generated tests. I agree that this could raise the exception but I'd rather capture it and log it from the calling function

ddtrace/contrib/pytest/plugin.py
gnufede: 1. For runtime generated tests, I'm not sure a module would be required
2. `_extract_span` would just return `None` if it's passed `None`

ddtrace/contrib/pytest/plugin.py
gnufede: The last we did above is check if is None and then start a test module span:
```
            if test_module_span is None:
                test_module_span = _start_test_module_span(pytest_package_item)
```
I wouldn't consider this a wrong state if it doesn't exist

ddtrace/contrib/pytest/plugin.py
gnufede: Definitely, removing `nextitem is None` doesn't break any test, so it's worth taking a deeper look there, but I consider these changes outside of the scope of this PR (I'm just indenting the code)

ddtrace/contrib/pytest/plugin.py
gnufede: I went with a solution to not reinvent the wheel

romainkomorndatadog (approved):

ddtrace/contrib/pytest/plugin.py
romainkomorndatadog: We looked at this offline and it turns out `test_module_span` is in fact a valid scenario when `pytest` is run in a context that doesn't have a package.

romainkomorndatadog (approved):

emmettbutler (dismissed):

Yun-Kim (approved):

PR #6246, opened 2023-06-30
chore(telemetry): change propagation telemetry event

ZStriker19:
Previously there was no standard for sending propagation telemetry events, either the key or value. We implemented something, but now that there's a standard, this PR is to meet it.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/internal/telemetry/writer.py
```diff
@@ -451,8 +451,8 @@ def _app_started_event(self):
             [
                 ("data_streams_enabled", config._data_streams_enabled, "unknown"),
                 ("appsec_enabled", config._appsec_enabled, "unknown"),
-                ("trace_propagation_style_inject", str(config._propagation_style_inject), "unknown"),
-                ("trace_propagation_style_extract", str(config._propagation_style_extract), "unknown"),
+                ("DD_TRACE_PROPAGATION_STYLE_INJECT", str(",".join(config._propagation_style_inject)), "unknown"),
+                ("DD_TRACE_PROPAGATION_STYLE_EXTRACT", str(",".join(config._propagation_style_extract)), "unknown"),
```
mabdinur, line 455: Do we need to make this change? The current configuration name should be supported by the backend

Yun-Kim (approved):
Does this require docs change?

ZStriker19:
> Does this require docs change?

Talked with Munir, we don't think so.

Yun-Kim (dismissed):

ddtrace/internal/telemetry/writer.py
```diff
@@ -256,8 +256,8 @@ def _app_started_event(self):
...
```
mabdinur, line 260: I would prefer to revert the configuration names for two reasons:

1) `DD_TRACE_PROPAGATION_STYLE_INJECT ` will not show up in metabase. This configuration will be mapped to `trace_propagation_style_inject`. For simplicity I think we should just use the value this configuration is mapped to. 

2) Consistency. Configuration names do not always map to environment variables. It can be an in code configuration (ex: ddtrace_auto_used). Having some refer directly to environment variables and other to the internal mapping could be difficult to maintain. 

```suggestion
                ("trace_propagation_style_inject", str(",".join(config._propagation_style_inject)), "unknown"),
                ("trace_propagation_style_extract", str(",".join(config._propagation_style_extract)), "unknown"),
```

mabdinur (approved):

PR #6299, opened 2023-07-07
chore: move constants to internal module

emmettbutler:
This change moves many of the constants used in IAST taint tracking to `ddtrace.internal.constants`. We're moving toward a more clear separation of concerns between Products like AppSec and Contrib integrations like Flask and Django.

In the case of request blocking, the separation of concerns ideally breaks down as follows. The AppSec Product code in the `ddtrace/appsec` directory knows how to make a block/don't block decision based on communication with libddwaf. The Django code in `ddtrace/contrib` knows how to take that blocking decision into account when processing requests.

According to this separation of concerns, these constants will be used by both the Product and the Contrib. Moving them to `internal.constants` means that the Contrib won't have to talk about AppSec when it imports them.

## Checklist

## Reviewer Checklist

- [ ] Title is accurate.

### DIFF
--- a/ddtrace/appsec/_constants.py
+++ b/ddtrace/appsec/_constants.py
@@ -3,6 +3,19 @@ from typing import TYPE_CHECKING
 
 import six
 
+from ddtrace.internal.constants import HTTP_REQUEST_BLOCKED
+from ddtrace.internal.constants import HTTP_REQUEST_BODY
+from ddtrace.internal.constants import HTTP_REQUEST_COOKIE_NAME
+from ddtrace.internal.constants import HTTP_REQUEST_COOKIE_VALUE
+from ddtrace.internal.constants import HTTP_REQUEST_HEADER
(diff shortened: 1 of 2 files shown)


Yun-Kim (commented):
Unsure if this is unnecessary after #6108 is merged. @avara1986 can you confirm?

emmettbutler:
@Yun-Kim it probably won't be. What do you think about keeping the two pull requests independent, and rebasing whichever one isn't merged first?

mabdinur (approved):

Yun-Kim:
@emmettbutler the little context I have from reviewing #6108 is that this might actually be unnecessary as they are removing these constants entirely (and moving into a `cpp` header file) as part of a larger ASM refactor. I don't have any strong feelings but definitely would like @avara1986 to weigh in before we decide to merge this. Is that okay?

juanjux (approved):

PR #6306, opened 2023-07-10
chore(ci-visibility): test or suite skipping

gnufede:
CI Visibility:

1. Allow for test or suite skipping (controlled by env var)
2. Create spans for ITR-skipped items (as agreed at `ITR skipped tests/suites RFC`)

## Checklist

## Reviewer Checklist
- [x] Title is accurate.

### DIFF
--- a/ddtrace/appsec/_constants.py
+++ b/ddtrace/appsec/_constants.py
@@ -123,6 +123,7 @@ class SPAN_DATA_NAMES(object):
 class API_SECURITY(object):
     """constants related to API Security"""
 
+    ENV_VAR_ENABLED = "DD_EXPERIMENTAL_API_SECURITY_ENABLED"
     REQUEST_HEADERS_NO_COOKIES = "_dd.appsec.s.req.headers"
     REQUEST_QUERY = "_dd.appsec.s.req.query"
     REQUEST_PATH_PARAMS = "_dd.appsec.s.req.params"
--- a/ddtrace/settings/config.py
+++ b/ddtrace/settings/config.py
@@ -5,6 +5,7 @@ from typing import List
 from typing import Optional
 from typing import Tuple
 
+from ddtrace.appsec._constants import API_SECURITY
 from ddtrace.appsec._constants import APPSEC
 from ddtrace.appsec._constants import DEFAULT
 from ddtrace.constants import APPSEC_ENV
@@ -279,7 +280,7 @@ class Config(object):
         self._user_model_email_field = os.getenv(APPSEC.USER_MODEL_EMAIL_FIELD, default="")
         self._user_model_name_field = os.getenv(APPSEC.USER_MODEL_NAME_FIELD, default="")
         self._iast_enabled = asbool(os.getenv(IAST_ENV, False))
-        self._api_security_enabled = asbool(os.getenv("_DD_API_SECURITY_ENABLED", False))
+        self._api_security_enabled = asbool(os.getenv(API_SECURITY.ENV_VAR_ENABLED, False))
         self._waf_timeout = DEFAULT.WAF_TIMEOUT
         try:
             self._waf_timeout = float(os.getenv("DD_APPSEC_WAF_TIMEOUT"))


ddtrace/contrib/pytest/plugin.py
```diff
@@ -331,94 +336,103 @@ def pytest_runtest_protocol(item, nextitem):
...
-            markers = [marker.kwargs for marker in item.iter_markers(name="dd_tags")]
-            for tags in markers:
-                span.set_tags(tags)
-            _store_span(item, span)
-
-            # Run the actual test
-            yield
+        # Finish coverage for the test suite if coverage is enabled
+        if TEST_SKIPPING_LEVEL == TEST and coverage_enabled() and not is_skipped_by_itr:
+            _coverage_end(test_suite_span)
```
romainkomorndatadog, line 427: Why does this end the test suite span's coverage? Wouldn't https://github.com/DataDog/dd-trace-py/blob/abbdd571f48cfddabab298c3efb9600df8516589/ddtrace/internal/ci_visibility/coverage.py#L75 cause the next test to overwrite the coverage tag?

tests/ci_visibility/test_encoder.py
```diff
@@ -1,7 +1,8 @@
-import contextlib
+from contextlib import contextmanager
```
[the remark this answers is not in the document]
gnufede: We're using `contextmanager`, but we're not really using anything else from `contextlib`. It's not really necessary to change this import, but I believe it's better this way

ddtrace/internal/ci_visibility/writer.py
```diff
@@ -29,6 +24,17 @@
...
 
 
+if TYPE_CHECKING:  # pragma: no cover
+    from typing import Dict
+    from typing import List
+    from typing import Optional
+
+    from ddtrace.vendor.dogstatsd import DogStatsd
+
+    from ...sampler import BasePrioritySampler
```
emmettbutler, line 34: This doesn't follow the usual pattern of importing type objects `if TYPE_CHECKING`. Why is this import conditional?

emmettbutler (dismissed):

ddtrace/internal/ci_visibility/writer.py
emmettbutler: Oh, I see! Thanks for clarifying that.

ddtrace/appsec/iast/_taint_tracking/TaintTracking/Source.cpp
```diff
@@ -14,13 +14,15 @@ Source::Source(string name, string value, OriginType origin)
   : name(move(name))
   , value(move(value))
   , origin(origin)
-{}
+{
+}
```
emmettbutler, line 18: Unintended changes?

ddtrace/internal/ci_visibility/recorder.py
```diff
@@ -293,6 +308,8 @@ def _fetch_tests_to_skip(self):
             return
 
         self._test_suites_to_skip = []
+        if TEST_SKIPPING_LEVEL == SUITE:
```
emmettbutler, line 311: Why is this necessary, given that the preceding line does the same thing unconditionally?

emmettbutler (changes requested):

ddtrace/appsec/iast/_taint_tracking/TaintTracking/Source.cpp
gnufede: I believe this was a git hook actually

ddtrace/internal/ci_visibility/recorder.py
gnufede: Indeed, sorry, bad conflict resolution

ddtrace/appsec/iast/_taint_tracking/TaintTracking/Source.cpp
gnufede: Apparently this is related to using a different clang-format version (>=16) than in CI (clang-format==10)

emmettbutler (approved):

ddtrace/internal/ci_visibility/encoder.py
```diff
@@ -195,8 +196,15 @@ def _build_payload(self, traces):
...
     def _convert_span(span, dd_origin):
         # type: (Span, str) -> Dict[str, Any]
-        return {
+        converted_span = {
             "test_session_id": int(span.get_tag(SESSION_ID) or "1"),
             "test_suite_id": int(span.get_tag(SUITE_ID) or "1"),
             "files": json.loads(span.get_tag(COVERAGE_TAG_NAME))["files"],
         }
+
+        from ddtrace.internal.ci_visibility.recorder import _get_test_skipping_level
```
romainkomorndatadog, line 205: Why not import this at the top of the module? This is getting called every time we call `_build_data()`, so it's not really a conditional import.

ddtrace/internal/ci_visibility/git_client.py
```diff
@@ -111,14 +114,18 @@ def _get_repository_url(cls, tags={}, cwd=None):
...
     @classmethod
     def _search_commits(cls, requests_mode, base_url, repo_url, latest_commits, serializer, _response):
-        # type: (int, str, str, list[str], CIVisibilityGitClientSerializerV1, Optional[Response]) -> list[str]
+        # type: (int, str, str, List[str], CIVisibilityGitClientSerializerV1, Optional[Response]) -> Optional[List[str]]
         payload = serializer.search_commits_encode(repo_url, latest_commits)
         response = _response or cls._do_request(requests_mode, base_url, "/search_commits", payload, serializer)
+        if response.status >= 400:
+            log.warning("Response status: %s", response.status)
+            log.warning("Response body: %s", response.body)
+            return None
```
romainkomorndatadog, line 128: Since we're not really using the fact that the return value is `None` in `_run_protocol()`, maybe returning an empty list is more appropriate.

ddtrace/internal/ci_visibility/git_client.py
```diff
@@ -90,6 +90,9 @@ def _run_protocol(cls, serializer, requests_mode, base_url, _tags={}, _response=
 
         latest_commits = cls._get_latest_commits(cwd=cwd)
         backend_commits = cls._search_commits(requests_mode, base_url, repo_url, latest_commits, serializer, _response)
+        if backend_commits is None:
```
romainkomorndatadog, line 93: If this is `True`, should we not still upload packfiles? The worst case scenario is that we accidentally send info about commits that our backend already knows about, right? Is that problematic?

ddtrace/internal/ci_visibility/recorder.py
```diff
@@ -306,12 +321,19 @@ def _fetch_tests_to_skip(self):
...
+
+                if TEST_SKIPPING_LEVEL == SUITE:
+                    self._test_suites_to_skip.append(path)
+                else:
+                    self._tests_to_skip[path].append(item["attributes"]["name"])
 
-    def _should_skip_path(self, path):
-        return os.path.relpath(path) in self._test_suites_to_skip
+    def _should_skip_path(self, path, name):
+        if _get_test_skipping_level() == SUITE:
```
romainkomorndatadog, line 332: Maybe this test-skipping level should be an argument to this function. It arguably shouldn't change at all from the time the plugin starts, let alone between calls to `_should_skip_path()`.

ddtrace/internal/ci_visibility/recorder.py
```diff
@@ -353,7 +375,7 @@ def _start_service(self):
             self.tracer.configure(settings={"FILTERS": tracer_filters})
         if self._git_client is not None:
             self._git_client.start(cwd=_get_git_repo())
-        if self.test_skipping_enabled() and self._test_suites_to_skip is None:
+        if self.test_skipping_enabled() and (not self._tests_to_skip and self._test_suites_to_skip is None):
```
romainkomorndatadog, line 378: maybe
```
if self.test_skipping_enabled() and not (self._tests_to_skip or self._test_suites_to_skip) ?
```

tests/ci_visibility/test_coverage_per_suite.py
```diff
@@ -0,0 +1,111 @@
...
+                        CIVisibility.disable()
+                        CIVisibility.enable(tracer=self.tracer, config=ddtrace.config.pytest)
+
+        with override_env(dict(DD_API_KEY="foobar.baz")):
+            return self.testdir.inline_run(*args, plugins=[CIVisibilityPlugin()])
+
+    @pytest.mark.skipif(compat.PY2, reason="ddtrace does not support coverage on Python 2")
+    def test_pytest_will_report_coverage_by_suite(self):
+        self.testdir.makepyfile(
+            test_ret_false="""
```
romainkomorndatadog, line 42: very small nit: since these aren't meant to be test files, I'd avoid calling them `test_`

romainkomorndatadog (approved):
None of my comments are "blocking", so you can feel free to resolve them without addressing them.

ddtrace/internal/ci_visibility/recorder.py
gnufede: Yes, this is more accurate, but since we change `_test_suites_to_skip` in `_fetch_tests_to_skip()`  from `None` to `[]` it is enough to know wether this initialization function was called

ddtrace/internal/ci_visibility/encoder.py
gnufede: It's due to a circular import, but I can move this function to a different place to avoid it.

PR #6307, opened 2023-07-10
chore(debugging): skip stdlib, third-party frames from exceptions

P403n1x87:
When collection data from an exception attached to a trace in exception debugging, we make the decision to skip frames that are associated to the standard library or third-party libraries, and only capture debugging information for user code. We do so by leveraging the code provenance data, together with the extra information that comes from the DD_MAIN_PACKAGE environment variable of the git metadata embedding support.

Note that when the variable is not set, we treat every non-stdlib path as user code to avoid not capturing any data at all.

## Checklist

## Reviewer Checklist

- [ ] Title is accurate.

### DIFF
--- a/ddtrace/internal/packages.py
+++ b/ddtrace/internal/packages.py
@@ -1,8 +1,10 @@
 import logging
 import os
 import sys
+import sysconfig
 import typing as t
 
+from ddtrace.internal.utils.cache import cached
 from ddtrace.internal.utils.cache import callonce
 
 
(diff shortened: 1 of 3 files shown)


ddtrace/internal/packages.py
```diff
@@ -134,6 +136,34 @@ def filename_to_package(filename):
...
+
+stdlib_path = sysconfig.get_path("stdlib")
+platstdlib_path = sysconfig.get_path("platstdlib")
+
+
+@cached()
+def is_stdlib(filename):
     # type: (str) -> bool
-    return filename_to_package(filename) is not None
+    return filename.startswith(stdlib_path) or filename.startswith(platstdlib_path)
```
P403n1x87, line 169: These paths are not enough when checking with `startswith`: `stdlib` might be a substring of `purelib` (i.e. `purelib.startswith(stdlib) is True`) and therefore everything would be identified as `stdlib`.

PR #6315, opened 2023-07-10
fix(profiling): ignore (another) race condition during process shutdown

sanchda:
This is a followup to #5182, where we prevented a class of race conditions for the allocation profiler.  Unfortunately, the heap profiler suffers from the same defect, which is handled in this PR.

## Checklist

## Reviewer Checklist

- [ ] Title is accurate.

### DIFF
--- a/ddtrace/profiling/collector/memalloc.py
+++ b/ddtrace/profiling/collector/memalloc.py
@@ -100,6 +100,12 @@ class MemoryCollector(collector.PeriodicCollector):
 
     def snapshot(self):
         thread_id_ignore_set = self._get_thread_id_ignore_set()
+        try:
+            events = _memalloc.heap()
+        except RuntimeError:
+            # DEV: This can happen if either _memalloc has not been started or has been stopped.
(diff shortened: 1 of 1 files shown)


P403n1x87 (approved):
Great catch!

emmettbutler (approved):
Really good catch, this error message has confused me a lot in the past. I'd recommend setting some backport labels.

sanchda:
> Really good catch, this error message has confused me a lot in the past. I'd recommend setting some backport labels.

Yeah--although the backport range is somewhat broad.  We've had recent reports from internal users on v1.15, but conceptually we could/should backport it to the same range as #5182.  That'd put us as far back as v1.10?

I'm fine with treating this as a continuation of #5182 and backporting it to the same targets.  @P403n1x87, @emmettbutler do you see any reason to limit the range?

P403n1x87:
Our current backport strategy takes the "lazy" approach of backporting as further back as to where the issue being fixed is reported. In this case I think this would be 1.15. Should we get a report of the same issue on 1.10 we can always extend the backport range then.

P403n1x87 (approved):

PR #6318, opened 2023-07-11
chore(setup): use compatible bytecode version for Python 3.11+

salty-horse:
ddtrace currently requires "any version of `bytecode`" when using Python 3.8+. This is incorrect.

As mentioned in #5512, a specific minimal version of `bytecode` is required when using Python 3.11+ (`BinaryOp` was [introduced in bytecode 0.14.0](https://github.com/MatthieuDartiailh/bytecode/commit/e370b04aa887d8de96876f75e237c5e98eef16ef)).
I just encountered the error in the issue, and this is something the package manager should have taken care of transparently.

This change adds the requirement of a modern version of `bytecode` when using Python 3.11+.

I tested this by installing the package locally with the command `pip install -e .` under Python 3.10 and 3.11, and verifying that it requires different versions of `bytecode`.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.

### DIFF
--- a/ddtrace/internal/utils/__init__.py
+++ b/ddtrace/internal/utils/__init__.py
@@ -77,7 +77,8 @@ def set_argument_value(
 def _get_metas_to_propagate(context):
     # type: (Any) -> List[Tuple[str, str]]
     metas_to_propagate = []
-    for k, v in context._meta.items():
+    # copying context._meta.items() to avoid RuntimeError: dictionary changed size during iteration
+    for k, v in list(context._meta.items()):
         if isinstance(k, six.string_types) and k.startswith("_dd.p."):
             metas_to_propagate.append((k, v))
     return metas_to_propagate
--- a/releasenotes/notes/fix-runtimeerror-on-get-metas-to-propagate-25f9c44ac4137fb9.yaml
+++ b/releasenotes/notes/fix-runtimeerror-on-get-metas-to-propagate-25f9c44ac4137fb9.yaml
@@ -1,0 +1,5 @@
+---
+fixes:
+  - |
+    tracing: This fix resolves an issue where concurrent mutations to the ``context._meta`` dict
+    caused `RuntimeError: dictionary changed size during iteration`.


StasEvseev:
Hello Datadog team!
We faced the issue on python3.11, that `ddtrace==1.20.5` didn't have `bytecode>=14` dependency.

Can we proceed with the PR?
Thank you!

P403n1x87:
I think #7274, which we recently merged into 1.20 as a backport of #7151 should fix the original issue without requiring a version pin. That is not to say that we should not do what it's being proposed in this PR. But at least for now, the issue should be resolved in the latest 2.x releases, and soon on 1.20 once we get the next patch release out.

P403n1x87:
@salty-horse The proposed change makes sense to me, thanks! What surprises me is that the installation of `ddtrace` did not cause the installation of the latest `bytecode` version in your environment. This makes me think that perhaps there are other dependencies that are pulling in a specific version of `bytecode`, which I find interesting and a bit unexpected. Unless there is a hosted PyPI index from which packages are coming from which doesn't necessarily have all the latest package versions?

P403n1x87 (commented):
Since this is fixing an issue we should probably make this a `fix` PR rather than `chore`, and include a release note explaining what is being fixed. A release note can be created with the [`reno`](https://pypi.org/project/reno/) utility.

salty-horse:
@P403n1x87 `bytecode` is only used by `ddtrace` in my project, and I don't have any rule forcing a specific `bytecode` version. I'm using poetry. I think I specifically requested to update `ddtrace` to a newer version, and maybe it didn't see a need to upgrade `bytecode`.

I understand the curiosity but I don't think it's worth investigating further.

P403n1x87:
@salty-horse thanks for the extra details. I agree there is no need to investigate this further. This issue is likely to happen during a very brief transition period around the time of a CPython release, where we are in the process of adding support for it, and if upgrading `ddtrace` in an existing environment. The issue should never occur in a clean environment. Nonetheless, the proposed change won't harm.

emmettbutler (approved):

P403n1x87 (approved):

Issue #4978, opened 2023-01-27
`ddtrace` deadlock in `on_span_start`


nierob:
### Summary of problem
`ddtrace` deadlock in `on_span_start`

### Which version of dd-trace-py are you using?
ddtrace==1.7.2

### Which version of pip are you using?
pip 22.3.1

### Which libraries and their versions are you using?

<details>
  <summary>`pip freeze`</summary>
Brotli==1.0.9
Jinja2==3.1.2
Mako==1.2.2
MarkupSafe==2.1.1
Pillow==9.3.0
PyJWT==2.6.0
PyMySQL==1.0.2
PyNaCl==1.5.0
PyYAML==6.0
Pygments==2.13.0
SQLAlchemy==1.4.46
aenum==3.1.11
aiocontextvars==0.2.2
aiohttp==3.8.3
aiosignal==1.2.0
alembic==1.9.2
amqp==5.1.1
anyio==3.6.1
arq==0.25.0
asttokens==2.0.8
async-cache==1.1.1
async-timeout==4.0.2
asyncpg==0.27.0
attrs==22.2.0
backcall==0.2.0
bcrypt==4.0.1
billiard==4.1.0
bleach==5.0.1
bytecode==0.13.0
cachetools==5.3.0
cattrs==22.1.0
celery==5.3.0b1
certifi==2022.12.7
cffi==1.15.1
charset-normalizer==2.1.1
click-didyoumean==0.3.0
click-plugins==1.1.1
click-repl==0.2.0
click==8.1.3
cryptography==38.0.4
cssselect2==0.6.0
datadog==0.44.0
ddsketch==2.0.4
ddtrace==1.7.2
decorator==5.1.1
defusedxml==0.7.1
dnspython==2.2.1
email-validator==1.3.1
emoji==2.2.0
envier==0.4.0
et-xmlfile==1.1.0
executing==1.0.0
fastapi==0.89.1
fonttools==4.37.1
frozenlist==1.3.1
greenlet==1.1.3
h11==0.12.0
hiredis==2.0.0
html5lib==1.1
idna==3.3
ipython==8.8.0
isodate==0.6.1
jedi==0.18.1
jsonschema==3.2.0
kombu==5.3.0b2
lxml==4.9.2
markdown-it-py==2.1.0
matplotlib-inline==0.1.6
mdurl==0.1.2
multidict==5.2.0
mypy-extensions==0.4.3
netsgiro==2.0.1
numpy==1.23.2
[...]

PR #6325, opened 2023-07-11
fix(tracer): use RLOCK in span aggregator

ZStriker19:
This fix resolves an issue where sqlalchemy rollbacks could intermittently cause deadlocks due SpanAggregator's `on_span_start` using a Lock.   This fix was put behind a flag `DD_TRACE_SPAN_AGGREGATOR_RLOCK` which defaults to false due to possible performance impacts. While our benchmarks don't show anything major, and we suspect this change will not impact performance much at all, we want to be cautious and test on internal services before making it true by default.  

We plan on gathering feedback and testing on internal services, as long as we don't see a negative impact, we plan on making this the new default.

Should fix: https://github.com/DataDog/dd-trace-py/issues/4978

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


majorgreys:
Can we move this to draft?

aetimmes:
Looking at this so far, I'm inclined to say this lock should be a `threading.RLock` either always or by default - the fact that this library causes deadlocks at all is hugely painful, and managing environment variables to prevent them isn't a much better user experience.

docs/configuration.rst
```diff
@@ -412,6 +412,11 @@ The following environment variables for the tracer are supported:
      type: Boolean
      default: True
      description: Send query strings in http.url tag in http server integrations.
+    
+   DD_TRACE_SPAN_AGGREGATOR_RLOCK:
```
brettlangdon, line 416: This might be bike shedding, but I wonder if this should be a private env var?

`_DD_TRACE_SPAN_AGGREGATOR_RLOCK` since the `SpanAggregator` is a private implementation detail.

releasenotes/notes/add_flag_for_span_aggregator_rlock-b08c88e0da85dc66.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    sqlalchemy: This fix resolves an issue where sqlalchemy rollbacks could intermittently cause deadlocks.
```
brettlangdon, line 4: We might want to have this as a known issue instead of a fix?

Since this isn't introducing the fix, this is introducing an opt-in to the possible fix for customer validation. When we change to make the RLock the default, then we can introduce it as a fix?

cc @mabdinur who has been thinking about introducing "known issues" for a bit now.

tests/contrib/gunicorn/test_gunicorn.py
```diff
@@ -175,6 +184,7 @@ def gunicorn_server(gunicorn_server_settings, tmp_path):
         SETTINGS_GEVENT_DDTRACERUN,
         SETTINGS_GEVENT_DDTRACERUN_MODULE_CLONE,
         SETTINGS_GEVENT_DDTRACERUN_DEBUGMODE_MODULE_CLONE,
+        SETTINGS_GEVENT_SPANAGGREGATOR_RLOCK,
```
brettlangdon, line 187: does this reproduce the issue without the RLock?

what relation does this gevent test have to the change? The fix is listed for sqlalchemy, do we have a sqlalchemy test case that replicates this issue?

ZStriker19:
> 

Hi @aetimmes, just updated the description with our plan: We plan on gathering feedback and testing on internal services, as long as we don't see a negative impact, we plan on making this the new default.

tests/contrib/gunicorn/test_gunicorn.py
ZStriker19: No, I haven't been able to recreate it in a test unfortunately. I mostly added it for fear that changing the lock might cause issues with Gunicorn.

docs/configuration.rst
ZStriker19: It seems weird to suggest using a "private" envar to users though doesn't it? Up to you though!

brettlangdon: yeah, not sure personally. I am just thinking if we end up making this the default, easier to phase out if we change the default to RLock.

but this is also an env var so.... .... risks are low if someone is setting and we aren't doing something with it. not sure. would love input.

releasenotes/notes/add_flag_for_span_aggregator_rlock-b08c88e0da85dc66.yaml
mabdinur: Yup a known issue release note + mitigation is what we need here.

tests/contrib/gunicorn/test_gunicorn.py
brettlangdon: I was able to repro manually... but it involved forcing a span creation during `on_span_start` of `SpanAggregator`... but that isn't really something we can recreate in a test?

releasenotes/notes/add_flag_for_span_aggregator_rlock-b08c88e0da85dc66.yaml
ZStriker19: Hmmmm.... should "known issues" just be "issues"? Since we already have that as an option, the release.py script won't be that happy adding "known issues", given it likes to use regex to parse release notes, and the fact that if we're writing about an issue, it implies that it is known?

nierob:
>Looking at this so far, I'm inclined to say this lock should be a threading.RLock either always or by default - the fact that this library causes deadlocks at all is hugely painful, and managing environment variables to prevent them isn't a much better user experience.

+1 from me. As a user I can not make informed decision if `RLock` should be used or not. That is too internal to play with without looking at the code.

emmettbutler (approved):

brettlangdon (approved):

tabgok (approved):
looks good from a contrib standpoint

PR #6343, opened 2023-07-12
fix(pymongo): resolve incompatibility with motor

### LINKED ISSUE #2171: Async: support of Motor
0.47.0

Motor 2.3.0

Can you add the support of Motor or hint on how it could be done?

fergusdixon:
Currently, the [Motor](https://motor.readthedocs.io/en/stable/) library (an async wrapper around pymongo) has poor support for tracing - the only function hit is the [get_socket direct patch](https://github.com/DataDog/dd-trace-py/blob/1.x/ddtrace/contrib/pymongo/patch.py#L73)

Fixes #2171

This is because the Motor library [does not import the pymongo client as expected](https://github.com/mongodb/motor/blob/master/motor/core.py#L102)
(`pymongo.mongo_client.MongoClient` vs the expected `pymongo.MongoClient`)
so the TracedMongoClient is not injected into the async wrapper

After patching this import path and running in a local service, I noticed access failing to some class variables, such as [`_client`](https://github.com/mongodb/mongo-python-driver/blob/master/pymongo/mongo_client.py#L138) and `codec_options` which were fixed by adding the Metaclass for the Traced client to return these values from the wrapped instance.

I then came across an issue when using transactions with [validate_session](https://github.com/mongodb/mongo-python-driver/blob/master/pymongo/pool.py#L993) using `is` to compare equality with the mongo clients, which bypasses the traced client's `__eq__` override.
This validation then fails as it compares the session's client to the traced object, so it needs to be overridden to compare the session's client to the wrapped client object.

This seems to work and generate deeper spans for me `cmd, spec etc.` locally if I inspect the generated trace in my local service

Unsure how to go about adding automated tests for this
This should not be risky
This change is maintainable - and not needed after the client rewrite is complete

Unsure what sort of detail this change requires in the release notes
Unsure if this requires backporting - it is in some ways a fix, in others a feature

## Checklist

## Reviewer Checklist

- [ ] Title is accurate.


ddtrace/contrib/pymongo/client.py
```diff
@@ -285,6 +288,14 @@ def __trace(self, cmd):
             set_address_tags(s, self.address)
         return s
 
+    def validate_session(self, client, session):
```
fergusdixon, line 291: I need to work out how to get his override into `https://github.com/mongodb/mongo-python-driver/blob/master/pymongo/pool.py#L861`

fergusdixon:
Patching the real reference to MongoClient ends up with strange type-errors
I realise this is actually solving 2 issues - motor's non-standard import of pymongo and support for mongo sessions/transactions (introduced in mongo 4.x, tests are running on 3.6)
I believe these should be split into different PRs, and I'll actually try fix the import motor-side first

PR #6369, opened 2023-07-14
Fix langchain template

NarekA:
## Checklist

## Reviewer Checklist

- [ ] Title is accurate.

@Yun-Kim I had some issues with the langchain integration, can you take a look at this fix?

### DIFF
--- a/ddtrace/contrib/langchain/patch.py
+++ b/ddtrace/contrib/langchain/patch.py
@@ -27,6 +27,7 @@ from ddtrace.internal.agent import get_stats_url
 from ddtrace.internal.logger import get_logger
 from ddtrace.internal.utils import get_argument_value
 from ddtrace.internal.utils.formats import asbool
+from ddtrace.internal.utils.formats import deep_getattr
 from ddtrace.pin import Pin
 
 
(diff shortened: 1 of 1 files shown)


Yun-Kim:
Hi @NarekA, thanks for your contribution!

Could you add some details to the PR description in terms of:
- What the issue you were seeing was (and a repro case if possible)
- What this fix involves?

NarekA:
I am getting the following error when tracing my chain:
```
  File "/opt/venv/lib/python3.10/site-packages/ddtrace/contrib/langchain/patch.py", line 549, in traced_chain_call
    span.set_tag_str("langchain.request.prompt", integration.trunc(str(instance.prompt.template)))
AttributeError: 'ChatPromptTemplate' object has no attribute 'template'
```

NarekA:
The fix is just using a safer way of getting nested attributes (the deep_get_attr) method. This seems to be the convention for these kinds of operations in this repo ([see example](https://github.com/DataDog/dd-trace-py/blob/a0803f971085021d974b5049dadc56a4f28f76d8/ddtrace/contrib/pynamodb/patch.py#L86)).

Yun-Kim:
Hi @NarekA, I've looked into this and it looks like not all prompt template classes have a `template` field , specifically chat message prompts which may have multiple messages which are all part of the prompt. 

The problem with using `deep_getattr` is that it only operates within object attributes, so for chat message prompts (which are based on a list of messages, each of which contains a prompt) this won't actually return anything.

I think the best fix here is to just grab the `instance.prompt.template` after checking that it exists, i.e.

```python
if hasattr(instance, "prompt") and getattr(instance.prompt.template, None):
    span.set_tag_str("langchain.request.prompt", integration.trunc(str(instance.prompt.template)))
```
And we can ignore more complex prompt templates for now. 

I can make the fix on my end, is that okay with you?

NarekA:
@Yun-Kim, In other projects, I would use `pydash.get`, but I don't want to add a dependency on pydash.

Yun-Kim:
We can close this for now with my PR is up and being reviewed. Thanks @NarekA for bringing this up!

PR #6391, opened 2023-07-18
fix(ci): break wheel building into multiple jobs

brettlangdon:
In order to combat the ever increasing build/release time we are breaking up the build/publish job into multiple jobs so the work can be better parallelized.

Before this change the build process would take at least 4 hours (sometimes more, or timing out at 6 hours), after this change it completes in about 1.5 hours.

The Python 2.7 + 3.5 build job is kept as-is with no changes.

For Python 3.6+ we have created a reusable workflow and then creates one build step per-python version that we want to support (3.6, 3.7, 3.8, 3.9, 3.10, 3.11).

This change also changes the `workflow_dispatch` trigger to not publish to PyPI so it can be used to validate the build process without releasing.

This change was validated by running manually, and then validating that the number of release files and the release tags created match what was published on PyPI for ddtrace==1.16.1.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


brettlangdon:
We are manually merging this since nothing gets validated by CircleCI, and we don't run this workflow during PRs.

PR #6398, opened 2023-07-18
fix(health-metrics): ensures metrics are aggregated by tags

mabdinur:
## Description

This PR ensures that metrics with the name name but different tags are not combined (ie avoid doing [this](https://github.com/DataDog/dd-trace-py/pull/Dog/dd-trace-py/blob/4601a7181302731e6a35a1533aa1f15ab68b8983/ddtrace/internal/writer/writer.py#L242)).

### Example
With the current implementation if a http.dropped.traces metric is queued with the tag `reason:timeout` and another `http.dropped.tracers` metric is queued with the tag `reason:encoding_err` then self.dogstatsd.distribution(..) will be called with the following arguments: 
  - ("datadog.tracers.http.dropped.traces", 2, ["reason:encoding_err", "reason:timeout"] )
     - This is incorrect.

In the scenario above `self.dogstatsd.distribution(..)` should be called twice with the following arguments:
 - ("datadog.tracers.http.dropped.traces", 1, ["reason:encoding_err"] )
 - ("datadog.tracers.http.dropped.traces", 1, ["reason:timeout"] )

There is no existing test case for this scenario. I will add a regression test to make this more clear.

We also changed tags from a list to tuple because tuples are hashable. This allows us to change the structure of the `_metrics` dictionary from `metric_name -> [tags, count]` to `metric_name -> tags -> count`.

## Motivation

Currently health metrics with the name but different tags are being combined. This is incorrect. Metrics with the same names but different tags should be reported in separate distributions.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/internal/writer/writer.py
```diff
@@ -446,10 +452,9 @@ def _flush_queue_with_client(self, client, raise_exc=False):
                 # This really isn't ideal as now we're going to do a ton of socket calls.
                 self.dogstatsd.distribution("datadog.%s.http.sent.bytes" % namespace, len(encoded))
                 self.dogstatsd.distribution("datadog.%s.http.sent.traces" % namespace, n_traces)
-                for name, metric in self._metrics.items():
-                    self.dogstatsd.distribution(
-                        "datadog.%s.%s" % (namespace, name), metric["count"], tags=metric["tags"]
-                    )
+                for name, metric_tags in self._metrics.items():
+                    for tags, count in metric_tags.items():
+                        self.dogstatsd.distribution("datadog.%s.%s" % (namespace, name), count, tags=list(tags))
```
majorgreys, line 457: I'm having a hard time understanding what the before/after here is. Was the change to the distribution writing necessary or just the using a tuple for tags? Why does the current implementation not work?

mabdinur: This PR ensures that metrics with the name name but different tags are not combined (ie avoid doing [this](Dog/dd-trace-py/blob/4601a7181302731e6a35a1533aa1f15ab68b8983/ddtrace/internal/writer/writer.py#L242)). 

With the current implementation if a `http.dropped.traces` metric is queued with the tag `reason:timeout` and another `http.dropped.tracers` metric is queued with the tag `reason:encoding_err`  then `self.dogstatsd.distribution(..)` will be called with the following arguments: `("datadog.tracers.http.dropped.traces", 2, ["reason:encoding_err", "reason:timeout"] )`. This is incorrect. 


In the scenario above `self.dogstatsd.distribution(..)` should be called twice with the following arguments: 
   - `("datadog.tracers.http.dropped.traces", 1, ["reason:encoding_err"] )`
   - `("datadog.tracers.http.dropped.traces", 1, ["reason:timeout"] )`
 

There is no existing test case for this scenario. I will add a regression test to make this more clear.

mabdinur: I changed tags from a list to tuple because tuples are hashable. 

This allows us to change the structure of the `_metrics` dictionary from `metric_name -> [tags, count]`  to `metric_name -> tags -> count`.

majorgreys (approved):
@mabdinur I would add your excellent scenario description to the PR.

PR #6412, opened 2023-07-20
ci: run the debugger suite only if necessary

P403n1x87:
We introduce the concept of suitespec as a way of describing how sources affect test runs. We use it to ensure that the debugger tests run only if sources that the suite depends on are modified by the current commit.

## Suitespec Implementation Details

The suitespec solution is based on a manual configuration of of test suites. To simplify the declaration of file patterns for test suites, one can make use of _components_, which essentially are a logic collection of patterns. Test suite can then be declared as a list of components to reflect their dependencies on these logic parts, and to DRY the declaration itself by avoiding repetitions.

## Notes

- When the script fails for any reason, tests are run.
- It is important that path patterns are listed correctly, or some tests might not run when they are in fact supposed to.
- Best effort to determine the correct list of changed files via the GitHub REST API. When that fails, we fall back to the less accurate `git diff` against the target branch.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


riotfile.py
```diff
@@ -85,6 +85,20 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
...
+    """Conditional virtualenv
+
+    Run the command on the condition that the needs_testrun.py script returns
+    non-zero exit code.
+    """
+
+    def __init__(self, *args, **kwargs):
+        name = kwargs["name"]
+        command = kwargs["command"]
+        kwargs["command"] = f"scripts/needs_testrun.py {name} || {command}"
```
P403n1x87, line 98: Running the script at this point means that we already create the venv, so ideally we'd want this check to happen earlier.

scripts/needs_testrun.py
```diff
@@ -0,0 +1,96 @@
...
+from argparse import ArgumentParser
+import fnmatch
+import json
+from pathlib import Path
+from subprocess import check_output
+import sys
+import typing as t
+
+
+def get_changed_files() -> t.List[str]:
```
emmettbutler, line 12: Doing this with a subprocess `git diff` call might be slightly less "correct" but it would get around the issue of needing a Github token in CircleCI.

emmettbutler:
I love this idea!

scripts/needs_testrun.py
P403n1x87: The `git diff` would work for single-commit PRs. In general we would need to know the target branch to compute the full diff, but that information is available only at the PR level 🙁 .

brettlangdon (commented):
I know @gnufede was trying to get CI Visibility running for this repo, if we go that route, we might be able to ITR ?

tests/.suitespec.json
```diff
@@ -0,0 +1,43 @@
+{
+    "components": {
```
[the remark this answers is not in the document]
P403n1x87: I think the best way going forward is to reflect the components with the file system organisation. E.g. moving tracing sources under `ddtrace/tracing` would summarise some of the patterns into `ddtrace/tracing/*` thus reducing the risk of forgetting something. But currently forgetting to add a new source is a potential risk. We could have a script that checks for new sources within a commit and emits an error/warning if that file is not listed in any of the components for instance?

emmettbutler (approved):

.circleci/config.yml
```diff
@@ -149,9 +149,14 @@ commands:
...
-                  ./scripts/check-diff ".riot/requirements/" "Changes detected after running riot. Consider deleting changed files, running scripts/compile-and-prune-test-requirements and committing the result."
+                  if scripts/needs_testrun.py -v "$CIRCLE_JOB"
+                  then
+                    # Sort the hashes to ensure a consistent ordering/division between each node
+                    riot list --hash-only '<<parameters.pattern>>' | sort | circleci tests split | xargs -n 1 -I {} ./scripts/ddtest riot -v run --exitfirst --pass-env -s {} $([[ << pipeline.parameters.coverage >> == false ]] && echo '--no-cov' )
+                    ./scripts/check-diff ".riot/requirements/" "Changes detected after running riot. Consider deleting changed files, running scripts/compile-and-prune-test-requirements and committing the result."
+                  else
+                    echo "No changes detected, skipping test run."
+                  fi
```
P403n1x87, line 159: @brettlangdon do you know if we can DRY this up by putting this script somewhere in `scripts` and running that instead? Would that work?

scripts/needs_testrun.py
P403n1x87: I've added calls to the REST API to get accurate results and used the `git diff` as a fallback

tests/.suitespec.json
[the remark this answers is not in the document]
P403n1x87: Some unrelated ideas for future work perhaps:
- Write a script that reports which sources are not part of any component. Once we reach full "source coverage" such a script could run as a pre-check to ensure that every source is in one (and maybe only one?) component.
- Do a similar thing for the CODEOWNERS file to ensure every source is assigned to at least one entity
- Perhaps the suitespec and the CODEOWNERS files might have some links that can be used for further validation (e.g. a team owning a component must also be an owner of the sources?)

.circleci/config.yml
brettlangdon: yeah, it _should_ work. I am game for trying it!

I am ok with that since we can try to format it a little better too rather than trying to do it as a massive oneliner.

scripts/needs_testrun.py
brettlangdon: nice, then we can at least see what kind of API limits we are hitting.

.circleci/config.yml
P403n1x87: I've given this a try 🤞

emmettbutler (dismissed):

emmettbutler (dismissed):

emmettbutler:
I'm really excited about this change, I want to write suite specs for everything immediately.

.circleci/run_test.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
brettlangdon, line 1:
```suggestion
#!/bin/bash
set -ex
```

?

scripts/needs_testrun.py
```diff
@@ -0,0 +1,132 @@
+#!/usr/bin/env python3
```
brettlangdon, line 1: should we add some doctests?

would be interesting to add one for `get_base_branch` since if we can use a static `PR_NUMBER`, then as long as the HTML/CSS doesn't change, the expected result should be consistent. could be a good litmus test to catch if something changes/stops working.

same idea for `get_changd_files`?

this isn't a hard requirement, just a suggestion, feel free to resolve if you think it is not needed.

brettlangdon (commented):
some minor nits/suggestions, but otherwise lgtm

scripts/needs_testrun.py
P403n1x87: good idea!

.circleci/config.yml
P403n1x87: Attempt failed. It seems that some jobs end up with a messed up environment when running with a script, whether it's bash or sh. I've reverted the DRY changes, perhaps we can get back to this after.

.circleci/config.yml
```diff
@@ -176,9 +181,14 @@ commands:
                 environment:
                   RIOT_RUN_RECOMPILE_REQS: "<< pipeline.parameters.riot_run_latest >>"
                 command: |
-                  # Sort the hashes to ensure a consistent ordering/division between each node
-                  riot list --hash-only '<<parameters.pattern>>' | sort | circleci tests split | xargs -n 1 -I {} riot -v run --exitfirst --pass-env -s {} $([[ << pipeline.parameters.coverage >> == false ]] && echo '--no-cov' )
-                  ./scripts/check-diff ".riot/requirements/" "Changes detected after running riot. Consider deleting changed files, running scripts/compile-and-prune-test-requirements and committing the result."
+                  if scripts/needs_testrun.py -v "$CIRCLE_JOB"
```
brettlangdon, line 184: I wonder if we can use an anchor to DRY this? 🤔 

but I am not sure when parameters are replaced... but I think anchors get applied by YAML during loading, so should be basically equivalent....

brettlangdon (approved):
possible suggestion to DRY, and I'd still love to see doctests added, but feel free to resolve if you want to handle as a follow-up.

.circleci/config.yml
P403n1x87: 👍  I'd see we keep this PR on point for now, and attempt to DRY this up later, as it doesn't seem to be too trivial

gnufede:
> > I know @gnufede was trying to get CI Visibility running for this repo, if we go that route, we might be able to ITR ?
> 
> My understanding is that ITR is a per-test rather than per-test-suite. So I see ITR improving this even further rather than an alternative?

I'm working towards supporting either of them, so we could use whichever suits us best, but there are some interferences when we auto-instrument ourselves

.circleci/config.templ.yml
```diff
@@ -0,0 +1,1153 @@
...
+  requires:
+    - pre_check
+    - ccheck
+
+requires_base_venvs: &requires_base_venvs
+  requires:
+    - pre_check
+    - ccheck
+    - build_base_venvs
+
```
P403n1x87, line 1128: Up to this point this is the same as the old `config.yml` file

.circleci/config.templ.yml
```diff
@@ -0,0 +1,1153 @@
...
+requires_tests: &requires_tests
+  requires:
+#[[[cog from needs_testrun import for_each_testrun_needed as fetn;fetn(lambda job: print(f"    - {job}")) ]]]
```
P403n1x87, line 1131: This generates the list of required jobs for `requires_test`. All the jobs are listed in the `needs_testrun.py` script now. This helps DRY this config source a bit since we can list them once, instead of repeating them in the workflow jobs below.

.circleci/config.templ.yml
```diff
@@ -0,0 +1,1153 @@
...
+      - ccheck
+
+      # Build necessary base venvs for integration tests
+      - build_base_venvs
+
+      # Docs
+      - build_docs: *requires_pre_check
+
+      # Integration test suites
+#[[[cog from needs_testrun import for_each_testrun_needed as fetn;fetn(lambda job: print(f"      - {job}: *requires_base_venvs")) ]]]
```
P403n1x87, line 1150: Here we auto-generate the list of jobs that we want to run based on the suitespec. This adds entries of the form
```
      - job_name: *requires_base_venvs
```

emmettbutler (dismissed):

.circleci/config.templ.yml
```diff
@@ -0,0 +1,1153 @@
...
+
+  ccheck:
+    executor: cimg_base
+    steps:
+      - checkout
+      - run: sudo apt-get update
+      - run: sudo apt-get install --yes clang-format gcc-10 g++-10 python3 python3-setuptools python3-pip cppcheck
+      - run: scripts/cformat.sh
+      - run: scripts/cppcheck.sh
+      - run: DD_COMPILE_DEBUG=1 DD_TESTING_RAISE=1 CC=gcc-10 CXX=g++-10 pip -vvv install .
```
brettlangdon, line 294:
```suggestion
```

Removed in another PR

brettlangdon (approved):

emmettbutler (approved):

brettlangdon (approved):

PR #6419, opened 2023-07-20
chore(datastreams): kinesis checkpoints

tabgok:
This change adds checkpoints to Kinesis for the DataStreams product.

In particular, this change supports single-edge applications (i.e. app->kinesis->app) by:

* Checkpointing messages sent into the queue
* Checkpoint messages received from a queue, using "ApproximateArrivalTimestamp" to compute latency.

## Checklist

## Reviewer Checklist

- [ ] Title is accurate.


ddtrace/contrib/botocore/patch.py
```diff
@@ -328,16 +348,36 @@ def inject_trace_to_kinesis_stream_data(record, span):
...
+    stream_arn = params.get("StreamARN")
+
+    if not stream_arn:
+        log.debug("Unable to determine StreamARN for request with params: ", params)
+        return
+
+    processor = pin.tracer.data_streams_processor
+    for record in results.get("Records", []):
+        time_estimate = record.get("ApproximateArrivalTimestamp", datetime.now()).timestamp()
+        ctx = DataStreamsCtx(processor, 0, time_estimate, time_estimate)
```
piochelepiotr, line 361: I don't think this will work, since set_checkpoint won't set parent_hash to 0 here. I will make a PR to be able to pass time_estimate directly to `set_checkpoint`

piochelepiotr: Oh! I see you made that change actually, so here, no need to create this context right?
You can call `set_checkpoint` with time_estimate directly?

tabgok: I added in the capability to `set_checkpoint` but its not really needed.

What I believe my code does:

1. Create CTX with the arrival timestamp ([link](https://github.com/DataDog/dd-trace-py/blob/1.x/ddtrace/internal/datastreams/processor.py#L293)) which sets the "current edge start" to the estimated arrival and current hash to "0"
2.  Call set checkpoint which:
  a. skips [this section](https://github.com/DataDog/dd-trace-py/blob/1.x/ddtrace/internal/datastreams/processor.py#L343-L351)
  b. sets parent hash to 0 ([link](https://github.com/DataDog/dd-trace-py/blob/1.x/ddtrace/internal/datastreams/processor.py#L343-L351))
  c. computes path latencies using the values passed to create the DataStreamsCtx and time.time() ([link](https://github.com/DataDog/dd-trace-py/blob/1.x/ddtrace/internal/datastreams/processor.py#L358-L359))

majorgreys (commented):
Should this have a release note?

ddtrace/contrib/botocore/patch.py
```diff
@@ -328,16 +347,37 @@ def inject_trace_to_kinesis_stream_data(record, span):
...
+    stream_arn = params.get("StreamARN")
+
+    if not stream_arn:
+        log.debug("Unable to determine StreamARN for request with params: ", params)
+        return
+
+    processor = pin.tracer.data_streams_processor
+    for record in results.get("Records", []):
+        time_estimate = record.get("ApproximateArrivalTimestamp", datetime.now()).timestamp()
+        processor.new_pathway(now_sec=time_estimate).set_checkpoint(
```
kr-igor, line 360: I don't think we need to create a separate pathway for every message. Since they all came from the same destination it would be better to create a single pathway object and set checkpoints for every message. This means we'd need to pass the `time_estimate` to `set_checkpoint` (The pathway object has logic to deal with repeating checkpoints correctly - it will disambiguate baed on the direction tag)

kr-igor (dismissed):

majorgreys (commented):
Can the test be updated to mock time as was done in #6532?

majorgreys:
This should be a `feat` PR.

releasenotes/notes/enable-data-stream-monitoring-for-kinesis-d5685ec1ac504cc3.yaml
```diff
@@ -0,0 +1,9 @@
+---
+prelude: >
+  Data Streams Monitoring (DSM) has added support for AWS Kinesis
+
+features:
+  - |
+     DSM: Adds DSM support for AWS Kinesis.  For information about DSM, see the [official documentation](https://docs.datadoghq.com/data_streams/).
```
mabdinur, line 7: sphinx does not support links with the format above. Here's a doc on how external links should be formatted

mabdinur (commented):
Left a small nit otherwise LGTM

releasenotes/notes/enable-data-stream-monitoring-for-kinesis-d5685ec1ac504cc3.yaml
mabdinur, line 7: something like this should work

```suggestion
     DSM: Adds DSM support for AWS Kinesis.  For information about DSM, see the `official documentation <https://docs.datadoghq.com/data_streams/>`_ . 
```

tabgok: done, thank you!

mabdinur (dismissed):

kr-igor (dismissed):

tabgok:
> Can the test be updated to mock time as was done in #6532?

yes

tabgok:
> This should be a `feat` PR.

whoops, fixed (thanks mabdunir)

majorgreys (approved):

mabdinur (dismissed):

mabdinur (approved):

PR #6459, opened 2023-07-25
ci: add all test suites to suitespec

emmettbutler:
This change updates `tests/.suitespec.json` to include file listings for all test suites. This makes CircleCI only run a given test suite when it can't be certain that no related files changed in a given pull request. The method I used to find dependencies between contribs was `grep -R contrib ddtrace/contrib | grep import`.

It also makes a change that finds the pull request number via a different environment variable as a fallback.

There's a risk with this change that I missed something. If that's the case there may be some test suites that don't run in cases where they should. The main reason this is an acceptable risk is that running the entire test suite on 1.x means that oversights from pull requests will be caught. Merging a pull request that breaks the 1.x build isn't ideal, but it's unlikely enough and easy enough to fix by editing `.suitespec.json` that I believe it's worth the risk.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


tests/.suitespec.json
```diff
@@ -26,18 +26,785 @@
...
+            "@core",
+            "@tracing",
+            "@algoliasearch",
+        ],
+        "langchain": [
+            "@bootstrap",
+            "@core",
+            "@tracing",
+            "@langchain",
+        ],
```
P403n1x87, line 808:
```suggestion
        ]
```

tests/.suitespec.json
```diff
@@ -26,18 +26,785 @@
...
+        "mako": [
+            "ddtrace/contrib/mako/*",
+        ],
+        "algoliasearch": [
+            "ddtrace/contrib/algoliasearch/*",
+        ],
+        "langchain": [
+            "ddtrace/contrib/langchain/*",
+        ],
```
P403n1x87, line 282:
```suggestion
        ]
```

tests/.suitespec.json
```diff
@@ -26,18 +26,785 @@
...
             "@bootstrap",
             "@core",
             "@tracing"
-        ]
+        ],
+        "kafka": [
+            "@bootstrap",
+            "@core",
+            "@tracing",
+            "@kafka"
```
P403n1x87, line 295: This approach is totally fine, but in cases where we can expect to have a single pattern, like for most integrations, perhaps we can just put that pattern directly here instead of creating a dedicated component. WDYT?

P403n1x87 (commented):
Thanks for being so proactive on this 💪! I can't wait to see this merged!

emmettbutler:
@P403n1x87 looks like all of the contrib tests are still being run. Do you have any guesses about why that is?

emmettbutler:
I think I figured it out. [This test run](https://app.circleci.com/pipelines/github/DataDog/dd-trace-py/41363/workflows/25afc6dd-0fd1-4a88-9e50-aee16a10f809) seems to be a pretty clear example of the rate limit being hit after going through a bunch of the suites in alphanumeric order, which tells me that the skips added in this change are working.

scripts/needs_testrun.py
```diff
@@ -122,6 +123,7 @@ def get_base_branch(pr_number: int) -> str:
     return BASE_BRANCH_PATTERN.search(pr_page_content).group(1)
 
 
+@functools.cache
```
P403n1x87, line 126: 👍 👍

scripts/needs_testrun.py
```diff
@@ -240,10 +244,21 @@ def needs_testrun(suite: str, pr_number: int) -> bool:
     return bool(matches)
 
 
+def _get_pr_number():
+    number = os.environ.get("CIRCLE_PR_NUMBER")
+    if not number:
+        pr_url = os.environ.get("CIRCLE_PULL_REQUEST", "")
```
P403n1x87, line 250: ah interesting! so there are cases where the PR number is not set?

P403n1x87 (dismissed):
🙌 awesome!

scripts/needs_testrun.py
emmettbutler: yeah, it's only set on forked PRs https://support.circleci.com/hc/en-us/articles/360047521451-Why-is-CIRCLE-PR-NUMBER-empty-

P403n1x87: ah good catch!

gnufede:
Is it possible to make it so that `pytest` changes also trigger `ci_visibility` tests, and viceversa? Since they are not fully-dependent but kind of related

gnufede (dismissed):

brettlangdon (approved):

PR #6469, opened 2023-07-25
ci(integrations): make test_span_tags less flaky

mabdinur:
Hypothesis: `test_span_tags` appears to be run while traces from another test is being flushed in the background. This causes an assertion in test_span_tags to fail. 

Fix: Convert `test_span_tags` to a snapshot test. With this approach we no longer need to assert logs were generated.

https://app.circleci.com/pipelines/github/DataDog/dd-trace-py/41326/workflows/deaf010e-53e7-4bba-a7ba-6c2317fe02ee/jobs/2769455
## Checklist

## Reviewer Checklist

- [x] Title is accurate.


tests/integration/test_integration.py
```diff
@@ -745,21 +746,29 @@ def test_downgrade(encoding, monkeypatch):
     log.error.assert_not_called()
 
 
-@allencodings
-def test_span_tags(encoding, monkeypatch):
-    monkeypatch.setenv("DD_TRACE_API_VERSION", encoding)
+def test_span_tags(run_python_code_in_subprocess):
```
brettlangdon, line 749: you could keep this with `@allencodings`, and then just use it to set the `DD_TRACE_API_VERSION` env var for the subprocess call.

tests/integration/test_integration.py
```diff
@@ -745,21 +746,29 @@ def test_downgrade(encoding, monkeypatch):
...
+    code = """
+import mock
+from ddtrace import tracer as t
+
+with mock.patch("ddtrace.internal.writer.writer.log") as log:
```
brettlangdon, line 754: we don't need to mock this anymore do we?

can we just check that stdout and stderr are empty from the subprocess?

brettlangdon (commented):
looks good, I think this would make sense to have as a snapshot test.

tests/integration/test_integration.py
```diff
@@ -747,19 +748,14 @@ def test_downgrade(encoding, monkeypatch):
 
 @allencodings
 def test_span_tags(encoding, monkeypatch):
-    monkeypatch.setenv("DD_TRACE_API_VERSION", encoding)
-
-    t = Tracer()
-    with mock.patch("ddtrace.internal.writer.writer.log") as log:
-        s = t.trace("operation", service="my-svc")
+    with snapshot_context(token="tests.integration.test_integration.test_span_tags"):
```
mabdinur, line 751: Using `snapshot_context` allows us to generate one snapshot file for all encodings. Using `pytest.mark.snapshot(token=....)` did not work.

brettlangdon: ah, because otherwise we get one per-encoding... yeah... smart using `snapshot_context`

tests/integration/test_integration.py
```diff
@@ -747,19 +748,14 @@ def test_downgrade(encoding, monkeypatch):
 
 @allencodings
 def test_span_tags(encoding, monkeypatch):
-    monkeypatch.setenv("DD_TRACE_API_VERSION", encoding)
-
-    t = Tracer()
```
brettlangdon, line 752: I think we still need to create this tracer instance, no?

Because we need init/config to happen to pick up on `DD_TRACE_API_VERSION`, otherwise this test is going to test only the default encoding only.

tests/integration/test_integration.py
mabdinur, line 751:
```suggestion
    monkeypatch.setenv("DD_TRACE_API_VERSION", encoding)

    t = Tracer()
    with snapshot_context(token="tests.integration.test_integration.test_span_tags"):
```

tests/integration/test_integration.py
```diff
@@ -750,16 +751,14 @@ def test_span_tags(encoding, monkeypatch):
     monkeypatch.setenv("DD_TRACE_API_VERSION", encoding)
 
     t = Tracer()
-    with mock.patch("ddtrace.internal.writer.writer.log") as log:
-        s = t.trace("operation", service="my-svc")
+    with snapshot_context(token="tests.integration.test_integration.test_span_tags"):
+        monkeypatch.setenv("DD_TRACE_API_VERSION", encoding)
```
mabdinur, line 755:
```suggestion
```

tests/integration/test_integration.py
```diff
@@ -750,16 +751,13 @@ def test_span_tags(encoding, monkeypatch):
...
     t = Tracer()
-    with mock.patch("ddtrace.internal.writer.writer.log") as log:
-        s = t.trace("operation", service="my-svc")
+    with snapshot_context(token="tests.integration.test_integration.test_span_tags"):
+        s = ddtrace.tracer.trace("operation", service="my-svc")
         s.set_tag("env", "my-env")
         s.set_metric("number", 123)
         s.set_metric("number", 12.0)
         s.set_metric("number", "1")
         s.finish()
```
mabdinur, line 760:
```suggestion
        s.finish()
        t.shutdown()
```

tests/integration/test_integration.py
```diff
@@ -750,16 +751,14 @@ def test_span_tags(encoding, monkeypatch):
     monkeypatch.setenv("DD_TRACE_API_VERSION", encoding)
 
     t = Tracer()
-    with mock.patch("ddtrace.internal.writer.writer.log") as log:
+    with snapshot_context(token="tests.integration.test_integration.test_span_tags"):
         s = t.trace("operation", service="my-svc")
         s.set_tag("env", "my-env")
         s.set_metric("number", 123)
         s.set_metric("number", 12.0)
         s.set_metric("number", "1")
```
brettlangdon, line 759: wait, this test doesn't make any sense... we use the same metric name for all three.....

can you update this?

tests/snapshots/tests.integration.test_integration.test_span_tags.json
```diff
@@ -0,0 +1,27 @@
...
+      "env": "my-env",
+      "language": "python",
+      "runtime-id": "074495c7a08d40f09556fce6dfeca081"
+    },
+    "metrics": {
+      "_dd.agent_psr": 1.0,
+      "_dd.top_level": 1,
+      "_dd.tracer_kr": 1.0,
+      "_sampling_priority_v1": 1,
+      "number": 1.0,
```
brettlangdon, line 22: this would need to get updated too

tests/integration/test_integration.py
```diff
@@ -747,19 +748,14 @@ def test_downgrade(encoding, monkeypatch):
...
 @allencodings
 def test_span_tags(encoding, monkeypatch):
-    monkeypatch.setenv("DD_TRACE_API_VERSION", encoding)
-
-    t = Tracer()
-    with mock.patch("ddtrace.internal.writer.writer.log") as log:
-        s = t.trace("operation", service="my-svc")
+    with snapshot_context(token="tests.integration.test_integration.test_span_tags"):
+        monkeypatch.setenv("DD_TRACE_API_VERSION", encoding)
+        s = ddtrace.tracer.trace("operation", service="my-svc")
```
brettlangdon, line 753: this still needs to be updated to use `t = Tracer()` again, otherwise the env var isn't being used at all.

mabdinur: The snapshot tests only capture traces generated by the global tracer. I had to convert this test to a subprocess test.

Also, traces generated by v0.4 and v0.5 encodings generate different snapshots. The v0.5 encoding adds a `type` field to all spans. This kinda surprised me.

brettlangdon: oh.... makes sense on both accounts

tests/integration/test_integration.py
```diff
@@ -746,20 +747,27 @@ def test_downgrade(encoding, monkeypatch):
 
 
 @allencodings
-def test_span_tags(encoding, monkeypatch):
-    monkeypatch.setenv("DD_TRACE_API_VERSION", encoding)
+@pytest.mark.snapshot()
+def test_span_tags(encoding, ddtrace_run_python_code_in_subprocess):
+    import os
```
brettlangdon, line 752: move to top of file, no reason it needs to be here.

brettlangdon (approved):
lgtm

tests/integration/test_integration.py
mabdinur, line 752:
```suggestion
```

mabdinur: It's already at the top. I don't know why I left it in. 😓

mabdinur:
Something wonky is happening in the test agent. A trace with the name `huge` is being received ([failure](https://app.circleci.com/pipelines/github/DataDog/dd-trace-py/41745/workflows/2748d8e6-9dee-429c-9f48-899c49d79591/jobs/2790914)) while `test_span_tags` is running. This is likely what made this test flaky in the first place. 

I believe this test is culprit: https://github.com/DataDog/dd-trace-py/blob/1.x/tests/integration/test_integration.py#L338. I will add `tracer.shutdown()` to this test. This should resolve the issue.

tests/integration/test_integration.py
```diff
@@ -546,6 +548,7 @@ def test_civisibility_intake_with_evp_available():
                 == EVP_SUBDOMAIN_HEADER_EVENT_VALUE
             )
             CIVisibility.disable()
+            t.shutdown()
```
mabdinur, line 551:
```suggestion
```

tests/integration/test_integration.py
```diff
@@ -398,6 +399,7 @@ def test_trace_bad_url(encoding, monkeypatch):
         )
     ]
     log.error.assert_has_calls(calls)
+    t.shutdown()
```
mabdinur, line 402:
```suggestion
```

brettlangdon (approved):
I wonder if it is worth moving `test_span_tags` to `test_integration_snapshot.py`?

not a blocker, lgtm

PR #6475, opened 2023-07-25
fix(langchain): add patch test cases, fix double patch issue for embeddings/vectorstores

Yun-Kim:
This PR adds patch test cases for the langchain integration, and fixes a bug where we were double patching `langchain.embeddings.HuggingFaceEmbeddings.embed_query/documents` and `langchain.vectorstores.Milvus.similarity_search`. This issue stemmed from the two classes being reused/inherited by other classes (`SentenceTransformerEmbeddings`, `Zilliz` respectively), meaning that when we wrapped the latter two classes' methods, we unintentionally wrapped the former two classes' methods twice.

The fix involves checking that the function to wrap isn't already a wrapped method (specifically a `wrapt.ObjectProxy` which we use in our integrations). This could potentially be problematic if a user wraps this before running our patching code as we would ignore this, but at the moment of writing I don't see a huge risk nor a cleaner solution to avoid double patching.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


mabdinur (commented):
Should we also update: test_langchain_patch.py?

Yun-Kim:
> Should we also update: test_langchain_patch.py?

@mabdinur what do you mean by that? `test_langchain_patch.py` was added in this PR which uncovered the double patch issue.

mabdinur:
> > Should we also update: test_langchain_patch.py?
> 
> @mabdinur what do you mean by that? `test_langchain_patch.py` was added in this PR which uncovered the double patch issue.

I missed that. Nevermind.

mabdinur (approved):

PR #6478, opened 2023-07-26
ci: run framework tests only when necessary

P403n1x87:
We filter on paths within the repository that require the framework tests to run to avoid running them when not necessary. To do this we introduce an extra check job that we make a dependency for the test suites. If the check job determines that no tests are required, the steps within each framework test is not run. This way we can still effectively run the job, which is required by PRs, but turn them into no-op for a quicker and cheaper feedback.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.

### DIFF
--- a/.github/workflows/test_frameworks.yml
+++ b/.github/workflows/test_frameworks.yml
@@ -13,8 +13,23 @@ concurrency:
     cancel-in-progress: true
 
 jobs:
+  needs-run:
+    runs-on: ubuntu-latest
+    outputs:
+      outcome: ${{ steps.run_needed.outcome }}
+    steps:
+      - uses: actions/checkout@v3
+      - id: run_needed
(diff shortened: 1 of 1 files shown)


brettlangdon:
> @brettlangdon the framework tests still appear as required even though they are not going to run on this PR. Do you know if there is a way to "fix" this?

We can either make them not required (let's be sure we have a feedback mechanism to ensure they don't break and we ignore them)... or.... move the check on whether to run or not into the job themselves.

Like, add a `when` clause (or w/e it is for GitHub Actions) or some kind of bail out.

Similar to what we do for the code coverage job in CircleCI, if there is nothing to do, just pass.

This way we can keep them as required, just most of the times they do nothing.

Doing it this way means you might be able to get more granular as well.

majorgreys (approved):

PR #6519, opened 2023-07-28
chore(ci): itr on ddtrace

gnufede:
Enables Intelligent Test Runner on ddtrace, allowing to skip non-required tests in the CI.

Depends on:
- [x] #6543 
- [x] #6545 
- [x] #6546  
- [x] #6547 
- [x] #6548 
- [x] #6565 

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


.circleci/config.templ.yml
```diff
@@ -36,6 +36,7 @@ machine_executor: &machine_executor
     # https://support.circleci.com/hc/en-us/articles/360045268074-Build-Fails-with-Too-long-with-no-output-exceeded-10m0s-context-deadline-exceeded-
     - PYTHONUNBUFFERED: 1
   steps:
+    - run: sudo apt-get update && sudo DEBIAN_FRONTEND=noninteractive apt-get install -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" --force-yes -y --no-install-recommends openssh-client git
```
brettlangdon, line 39: should we be using a different base image? we are using ubuntu-2004, wonder if we should be using a more recent version?

this is likely an aside and not a requirement to change.

.coveragerc
```diff
@@ -1,5 +1,6 @@
 [run]
 omit = ddtrace/vendor/*
+parallel = true
```
brettlangdon, line 3: what does this do?

riotfile.py
```diff
@@ -129,7 +132,7 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
         Venv(
             name="appsec",
             pys=select_pys(),
-            command="pytest {cmdargs} tests/appsec",
+            command="pytest --ddtrace {cmdargs} tests/appsec",
```
brettlangdon, line 135: should we just add `--ddtrace` to the pytest config in `setup.cfg` ?

riotfile.py
```diff
@@ -202,7 +205,7 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
         Venv(
             name="integration",
             # Enabling coverage for integration tests breaks certain tests in CI
-            command="pytest --no-cov {cmdargs} tests/integration/",
+            command="pytest --no-cov --ddtrace --ignore-glob='*civisibility*' {cmdargs} tests/integration/ && pytest --no-cov {cmdargs} tests/integration/test_integration_civisibility.py",  # noqa: E501
```
brettlangdon, line 208: comment here would be nice, why do we need to break them up? I assume we cannot run with `--ddtrace` for civisibility integration tests?

riotfile.py
brettlangdon: I see below we are not using it for one scenario, are we able to selectively disable for a single run vs needing to enable everywhere?

P403n1x87, line 135: Are there benefits/downsides for running this locally as well? If not, perhaps we can pass this option only via the CircleCI config

gnufede: Yes, there are a few cases where it doesn't work (mainly pytest tests and similar): pytest, ci_visibility, pytest-bdd and asynctest

I can explore the option to create a `--no-ddtrace` option and see if it works.

Is it OK if I do that in a follow-up PR?

riotfile.py
gnufede: Yes, that's the reason, I can add a comment

riotfile.py
gnufede: We can run it locally as well, but it may not work properly for changes that are not committed, so maybe it's worth it to try to run it only in CircleCI

.circleci/config.templ.yml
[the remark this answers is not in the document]
gnufede: Yeah, it definitely should be better, since this is time consuming, but I wanted to keep the changes to the minimum here

.coveragerc
gnufede: This resolves an issue with coverage, since we're tracking the coverage per test with ITR, and there is another coverage instance running, there are conflicts while trying to persist the coverage file if we don't use this option.

riotfile.py
gnufede: How about passing a `--no-ddtrace` option locally to disable it?

P403n1x87: > but it may not work properly for changes that are not committed

So the best benefit of ITR is in CI then? If so, personally I'd prefer to have the CLI option passed only in via the circleci config to minimise the impact on contributors wanting to run tests locally.

romainkomorndatadog: There should be some benefit locally, too, though it requires `DD_API_KEY` and `DD_APPLICATION_KEY` env vars, and either setting up the service in on https://app.datadoghq.com/ci/settings/test-service to be enabled for the `DD_SERVICE` and `DD_ENV` default .

I think passing the CLI option through the CircleCI is the safe option but, on the other hand, it might be nice to get a bit of "forced" adoption.

setup.cfg
```diff
@@ -40,6 +40,7 @@ rst-directives = py:data
 [tool:pytest]
 # --cov-report is intentionally empty else pytest-cov will default to generating a report
 addopts =
+  --ddtrace
```
gnufede, line 43: @brettlangdon @P403n1x87 Should we do some trick to have this line only in the CI? Something like a separate setup.cfg for CI than for local development?

gnufede: OK, fixed by changing the command in the CI template file

ddtrace/contrib/pytest/plugin.py
```diff
@@ -62,7 +62,7 @@ def encode_test_parameter(parameter):
 
 def is_enabled(config):
     """Check if the ddtrace plugin is enabled."""
-    return config.getoption("ddtrace") or config.getini("ddtrace")
+    return (config.getoption("ddtrace") or config.getini("ddtrace")) and not config.getoption("no-ddtrace")
```
brettlangdon, line 65: sorry to do it to you, but would you be willing to make this it's own PR.

we probably want to add a release note as a feature here for `--no-ddtrace` since we are modifying the pytest plugin that customers also use.

riotfile.py
```diff
@@ -1368,12 +1373,12 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
...
         Venv(
             name="boto",
-            command="pytest {cmdargs} tests/contrib/boto",
+            command="pytest --no-ddtrace {cmdargs} tests/contrib/boto",
             venvs=[Venv(pys=select_pys(max_version="3.6"), pkgs={"boto": latest, "moto": "<1.0.0"})],
         ),
         Venv(
             name="botocore",
-            command="pytest {cmdargs} tests/contrib/botocore",
+            command="pytest --no-ddtrace {cmdargs} tests/contrib/botocore",
```
brettlangdon, line 1381: can we document these exceptions (code comments are fine), like, will this never work with `--ddtrace`, or is this something we are just saving as a todo to fix later?

gnufede: Actually I think this one should work, but I got some errors that I wasn't sure if they were failing in other branches, like, but maybe it's actually flaky?:
https://app.circleci.com/pipelines/github/DataDog/dd-trace-py/42049/workflows/7cf2183b-9f81-47eb-ab70-78a7886fa2dd/jobs/2805599/parallel-runs/3?filterBy=FAILED&invite=true#step-108-166
https://app.circleci.com/pipelines/github/DataDog/dd-trace-py/41887/workflows/c1231f57-cb36-4639-b42e-9d599b83d9cc/jobs/2797712?invite=true#step-108-141

Yun-Kim: I believe those were flaky tests that should be fixed by now? Can you try re-running without ITR disabled to verify?

mabdinur (commented):
A few tests suites are failing: https://app.circleci.com/pipelines/github/DataDog/dd-trace-py/42484/workflows/2e21ac1d-c61e-44bb-bfed-2a91fff41bfc

mabdinur (approved):

PR #6529, opened 2023-07-30
fix(profiling): ddup support arm, fix import on osx/windows

sanchda:
The recent libdatadog merge introduced a few issues on various platforms

* arm Linux
* Windows
* macos

This patch
* adds better guards for not-supported platforms
* completes the support for arm-linux
* fixes a bug in the memory profiler when ddup is enabled
* rationalizes the return type of various profiling interfaces to using a common DDFrame namedtuple (legacy-exporter is not modified for purposes of stability)
* eliminates the auto-backtrace capability of the ddup exporter; this was causing issues on some musl distributions and was also interfering with other tools for native backtraces

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/profiling/collector/_memalloc_tb.c
```diff
@@ -182,19 +229,17 @@ traceback_to_tuple(traceback_t* tb)
...
 
-        PyTuple_SET_ITEM(frame_tuple, 0, frame->filename);
-        Py_INCREF(frame->filename);
-        PyTuple_SET_ITEM(frame_tuple, 1, PyLong_FromUnsignedLong(frame->lineno));
-        PyTuple_SET_ITEM(frame_tuple, 2, frame->name);
-        Py_INCREF(frame->name);
-        /* Class name */
-        PyTuple_SET_ITEM(frame_tuple, 3, empty_string);
-        Py_INCREF(empty_string);
+        PyObject* frame_tuple = PyObject_CallFunctionObjArgs(ddframe_class,
```
P403n1x87, line 234: Would we be able to grab the definition of `DDFrame` from the event submodule and call that instead? This way we
- might save creating the namedtuple using the C API
- ensure that we use the same namedtuple class everywhere.

ddtrace/profiling/collector/_memalloc_tb.c
```diff
@@ -18,6 +18,53 @@ static PyObject* empty_string = NULL;
...
+    PyObject* class_name = PyUnicode_FromString("DDFrame");
+    PyObject* field_names = Py_BuildValue("(ssss)", "file_name", "lineno", "function_name", "class_name");
+
+    PyObject* args = PyTuple_Pack(2, class_name, field_names);
+    Py_DECREF(class_name);
+    Py_DECREF(field_names);
+
+    if (args == NULL) {
+        PyErr_Print();
+        return false;
```
P403n1x87, line 54: we don't DECREF `namedtuple_fun` before returning here I believe

ddtrace/profiling/collector/_memalloc_tb.c
```diff
@@ -18,6 +18,53 @@ static PyObject* empty_string = NULL;
...
+    PyObject* namedtuple_fun = PyObject_GetAttrString(collections, "namedtuple");
+    Py_DECREF(collections);
+    if (namedtuple_fun == NULL) {
+        PyErr_Print();
+        return false;
+    }
+
+    // Very important that this matches the definition of DDFrame in event.py
+    PyObject* class_name = PyUnicode_FromString("DDFrame");
+    PyObject* field_names = Py_BuildValue("(ssss)", "file_name", "lineno", "function_name", "class_name");
```
P403n1x87, line 46: Perhaps worth checking that we get valid Python objects here?

sanchda: Got rid of this entirely; now it's pulling from event.py

ddtrace/profiling/collector/_memalloc_tb.c
```diff
@@ -195,14 +228,21 @@ traceback_to_tuple(traceback_t* tb)
         PyTuple_SET_ITEM(frame_tuple, 3, empty_string);
         Py_INCREF(empty_string);
 
+        // Set the class
+#if PY_VERSION_HEX >= 0x03090000
+        Py_SET_TYPE(frame_tuple, (PyTypeObject*)ddframe_class);
```
P403n1x87, line 233: should we check that `ddframe_class` is not `NULL` on entry here for sanity?

sanchda: I'm not 100% sure how we might arrive to such a state, but there's no harm in doing it, so I added the check.  The check converts a potential segfault into an exception, so there's still an error we could probably catch.

I suppose it could occur from either of the two memalloc interfaces, since they both call this `traceback_to_tuple` in one way or another.  I'll take a look at that.

P403n1x87: Yep with the race conditions of the likes of `the memalloc module is already started` I thought it might be better to be safe than sorry here.

P403n1x87 (approved):

majorgreys (approved):

PR #6548, opened 2023-08-01
chore(ci-visibility): decouple from tracer

gnufede:
CI Visibility: decouple from regular Tracer and ContextProvider to avoid issues when self-instrumenting ddtrace.

Also decouple env vars so we can have 2 separate sets of DD_API_KEY and DD_APP_KEY

## Checklist

## Reviewer Checklist

- [x] Title is accurate.

### DIFF
--- a/ddtrace/internal/ci_visibility/recorder.py
+++ b/ddtrace/internal/ci_visibility/recorder.py
@@ -104,10 +104,10 @@ class CIVisibility(Service):
             self.tracer = CITracer(context_provider=CIContextProvider())
 
         self._app_key = os.getenv(
-            "CI_DD_APP_KEY",
+            "_CI_DD_APP_KEY",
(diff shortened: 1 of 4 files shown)


gnufede:
> Duplicating the environment variables seems like it could be confusing for users. I'm also not understanding the issue that `CITracer` solves - maybe a regression test could help with that.
> 
> Could you provide a little more rationale for this change? What problems is it solving?

Sure, let me explain, this change allows us to split things as much as possible when we instrument our tests, so we can avoid mixing the spans and the information gathered by the "CI Visibility tracer" with the "APM tracer" while running our pytest tests with the `--ddtrace` plugin. If we don't do this, it affects for example the snapshot tests, where the number of spans collected doesn't match, etc.

emmettbutler (approved):

romainkomorndatadog (approved):

PR #6559, opened 2023-08-02
feat(unittest): add unittest support for test visibility

ericlaz:
## Motivation
This PR aims to add unittest support for `dd-tracer-py` which is a built-in Python testing library widely used.

So far, this PR is for test visibility only. It is planned to add full CI Visibility to unittest in the future.

## Changes
Added test visibility support for unittests behind a feature flag. The `DD_CIVISIBILITY_UNITTEST_ENABLED` environment variable must be enabled in order to use unittest test visibility.

## Example

You can find an example of the trace JSON in this [span](https://dd.datad0g.com/ci/test-runs?query=test_level%3Atest%20%40test.service%3Aericnavarro-test%20&currentTab=overview&eventStack=AgAAAYoI6g9uHpHMiwAAAAAAAAAYAAAAAEFZb0k2aEVjQUFEaEhyTWEyU1AwZkRoYgAAACQAAAAAMDE4YTA5MWUtNjZiNy00MTM1LWJmM2YtYTgxY2U5YWRhZDgz&graphType=flamegraph&index=citest&sort=time&spanID=9606617518799500644&testId=AgAAAYoI6g9uHpHMiwAAAAAAAAAYAAAAAEFZb0k2aEVjQUFEaEhyTWEyU1AwZkRoYgAAACQAAAAAMDE4YTA5MWUtNjZiNy00MTM1LWJmM2YtYTgxY2U5YWRhZDgz&trace=AgAAAYoI6g9uHpHMiwAAAAAAAAAYAAAAAEFZb0k2aEVjQUFEaEhyTWEyU1AwZkRoYgAAACQAAAAAMDE4YTA5MWUtNjZiNy00MTM1LWJmM2YtYTgxY2U5YWRhZDgz&start=1690988854106&end=1691593654106&paused=false)

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


emmettbutler:
@ericlaz would you like a review while this is in "draft" state, or would you prefer that I wait until it's "ready for review"?

ericlaz:
> @ericlaz would you like a review while this is in "draft" state, or would you prefer that I wait until it's "ready for review"?

I would appreciate some reviews now if possible, some CI tests are not yet passing so that is why it is in draft state for now still but suggestions are always very appreciated.

Thanks in advance @emmettbutler!

riotfile.py
```diff
@@ -638,7 +638,7 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
         ),
         Venv(
             name="cherrypy",
-            command="python -m pytest {cmdargs} tests/contrib/cherrypy",
+            command="pytest --no-ddtrace {cmdargs} tests/contrib/cherrypy",
```
emmettbutler, line 641: Pulling these `--no-ddtrace` changes out into a separate pull request would clarify to reviewers why they're necessary and clearly prove that they don't break existing tests.

riotfile.py
```diff
@@ -1731,9 +1731,29 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
                 ),
             ],
         ),
+        Venv(
+            name="unittest",
+            command="python -m pytest {cmdargs} tests/contrib/unittest_plugin/",
+            pkgs={"msgpack": latest},
+            env={"DD_CIVISIBILITY_AGENTLESS_ENABLED": "0"},
+            venvs=[
```
emmettbutler, line 1739: Unless there's a clear reason not to, I recommend running these tests under all versions of Python from 2.7 to 3.11, this library's [list of supported versions](https://github.com/DataDog/dd-trace-py/blob/bf1b3c2594b1d1beed86ba0393314ac5d1faca35/setup.py#L601).

ddtrace/contrib/pytest/plugin.py
```diff
@@ -250,7 +251,8 @@ def _start_test_suite_span(item, test_module_span, should_enable_coverage=False)
     test_suite_span.set_tag_str(test.FRAMEWORK_VERSION, pytest.__version__)
     test_suite_span.set_tag_str(test.COMMAND, _get_pytest_command(item.config))
     test_suite_span.set_tag_str(_EVENT_TYPE, _SUITE_TYPE)
-    test_suite_span.set_tag_str(_SESSION_ID, str(test_session_span.span_id))
+    if test_session_span:
```
emmettbutler, line 254: It seems possible that the changes in this file fix a bug that isn't directly related to adding unittest support. It might be beneficial to pull them out into a separate pull request.

ddtrace/internal/ci_visibility/encoder.py
```diff
@@ -32,7 +32,8 @@ class CIVisibilityEncoderV01(BufferedEncoder):
     ALLOWED_METADATA_KEYS = ("language", "library_version", "runtime-id", "env")
     PAYLOAD_FORMAT_VERSION = 1
     TEST_SUITE_EVENT_VERSION = 1
-    TEST_EVENT_VERSION = 2
+    # Change to 1 to allow unittest tests without module or session spans
+    TEST_EVENT_VERSION = 1
```
emmettbutler, line 36: How does this affect existing uses of `TEST_EVENT_VERSION`?

releasenotes/notes/add-unittest-support-024c4664f6546c21.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    Adds unittest support for test visibility only
```
emmettbutler, line 4: It's unclear to me what the "only" in this note is referring to. What are the other things that "unittest" might "support"?

tests/contrib/pytest/test_pytest.py
```diff
@@ -1228,11 +1234,18 @@ def test_ok(self):
         rec = self.inline_run("--ddtrace", file_name)
         rec.assertoutcome(failed=1, passed=1)
         spans = self.pop_spans()
-        test_span_skipped = spans[0]
-        test_span_ok = spans[1]
-        test_suite_span = spans[4]
-        test_session_span = spans[2]
-        test_module_span = spans[3]
+        if len(spans) >= 7:
```
emmettbutler, line 1237: The use of number literals in this code makes it hard to change. You might consider finding a way to abstract these different cases into a function or class that uses descriptive naming to clarify the separation of concerns between these two paths.

tests/contrib/unittest_plugin/test_unittest.py
```diff
@@ -0,0 +1,388 @@
...
+        assert spans[2].get_tag(test.TYPE) == SpanTypes.TEST
+        assert spans[2].get_tag(test.NAME) == 'test_will_fail_first'
+        assert spans[2].get_tag(test.MODULE) == 'tests.contrib.unittest_plugin.test_unittest'
+        assert spans[2].get_tag(test.SUITE) == 'UnittestExampleTestCase'
+        assert spans[2].get_tag(test.TEST_STATUS) == test.Status.FAIL.value
+        assert spans[2].get_tag(test.SKIP_REASON) is None
+        assert spans[2].get_tag(ERROR_MSG) == 'False is not true'
+        assert spans[2].get_tag(ERROR_TYPE) == 'builtins.AssertionError'
+
+        assert spans[3].get_tag(test.TEST_TYPE) == SpanTypes.TEST
```
emmettbutler, line 379: These repeated assertions are another instance where a bit of deduplication would noticeably improve the changeability of this code.

emmettbutler (changes requested):
Nice work, @ericlaz. You've got the foundation of this integration figured out, from the look of things. I left a bunch of comments that I hope will help as you continue to polish this change. Let me know how else I can help!

ddtrace/contrib/pytest/plugin.py
ericlaz: Good catch, will draft a quick PR for it

releasenotes/notes/add-unittest-support-024c4664f6546c21.yaml
ericlaz: Apologies, my approach is to split the unittest support into test level visibility initially and full visibility (including session, module, suite and test-level visibility). Do you think splitting them might make them easier for the reviewer? Otherwise, the full PR is available [here work in progress](https://github.com/DataDog/dd-trace-py/pull/6611)

riotfile.py
ericlaz: Good catch, should be addressed in this new commit

riotfile.py
ericlaz: Good point, will take a look at this. Using `--no-ddtrace` might not be needed after all

ddtrace/contrib/pytest/plugin.py
```diff
@@ -214,7 +214,8 @@ def _start_test_module_span(pytest_package_item=None, pytest_module_item=None):
     test_module_span.set_tag_str(test.FRAMEWORK_VERSION, pytest.__version__)
     test_module_span.set_tag_str(test.COMMAND, _get_pytest_command(item.config))
     test_module_span.set_tag_str(_EVENT_TYPE, _MODULE_TYPE)
-    test_module_span.set_tag_str(_SESSION_ID, str(test_session_span.span_id))
+    if test_session_span:
```
romainkomorndatadog, line 217: There should not be a case in which `test_session_span` is not available here. https://github.com/DataDog/dd-trace-py/commit/ae7bab3af39943bfdcdefec6f30bfe168633d2d7 was meant to help ensure we always have the full session -> module -> suite -> test hierarchy.

Is this breaking in a unit test, or are there issues with it in real-world use?

ddtrace/internal/ci_visibility/encoder.py
romainkomorndatadog: Our `pytest` plugin is meant to use `TEST_EVENT_VERSION` `2`, so I think we should either make versioned instances of the encoder and have `pytest` use v2 and `unittest` use v1, or just make sure `unittest` does support version `2`.

tests/contrib/pytest/test_pytest.py
```diff
@@ -1159,10 +1159,16 @@ def test_ok(self):
         rec.assertoutcome(skipped=1, passed=1)
         spans = self.pop_spans()
         test_span_skipped = spans[0]
-        test_span_ok = spans[1]
-        test_suite_span = spans[4]
-        test_session_span = spans[2]
-        test_module_span = spans[3]
+        if len(spans) >= 6:
```
romainkomorndatadog, line 1162: I don't think we should have a scenario where we expect the number of spans to be able to change.

romainkomorndatadog (commented):
Do we have to find a way to make the `pytest` plugin and the `unittest` patching mutually exclusive, so that we don't run into issues where a `pytest` suite that includes a `unittest` test case causes weird behavior?

ddtrace/contrib/pytest/plugin.py
ericlaz: This should've been addressed in this new commit. `unittest` is no longer enabled if `pytest` is running as we support `unittest` with `pytest` out of the box.

tests/contrib/pytest/test_pytest.py
ericlaz: Good point, I just reverted `test_pytest.py` as `unittest` won't run if `pytest` is enabled (we support `unittest` with our `pytest` integration anyway)

ddtrace/internal/ci_visibility/encoder.py
ericlaz: An update on this: `TEST_EVENT_VERSION` is now set to 2 (just like before) if the `DD_CIVISIBILITY_UNITTEST_BETA_ENABLED` is not True / or not set. This should not affect existing uses of `dd-trace-py` if the env var unittest is not set.

It is planned to rollback to `TEST_EVENT_VERSION = 2` with the full PR soon

tests/contrib/unittest_plugin/test_unittest.py
```diff
@@ -0,0 +1,361 @@
...
+            assert spans[0].get_tag(test.NAME) == "test_will_pass_first"
+            assert spans[0].get_tag(test.MODULE) == "tests.contrib.unittest_plugin.test_unittest"
+            assert spans[0].get_tag(test.SUITE) == "UnittestExampleTestCase"
+            assert spans[0].get_tag(test.TEST_STATUS) == test.Status.PASS.value
+
+        def test_unittest_pass_multiple(self):
+            """Tests with`unittest` tests which should pass."""
+            _set_tracer(self.tracer)
+
+            class UnittestExampleTestCase(unittest.TestCase):
```
romainkomorndatadog, line 61: For a follow-up task/PR:

Does `unittest.TestLoader()` let you load test cases from multiple test classes? If so, I think it'd be worth writing at least one test where you have multiple test case classes, so we can make sure we're not accidentally overwriting the suite.

It might also be worth having one test where we do what we do with `pytest` and actually create modules on disk and call `unittest` on them.

romainkomorndatadog (dismissed):
Code and tests look good to me. None of my comments are blocking, so you can resolve them as you see fit.

ddtrace/settings/config.py
```diff
@@ -333,6 +333,7 @@ def __init__(self):
         self._ci_visibility_intelligent_testrunner_disabled = asbool(
             os.getenv("DD_CIVISIBILITY_ITR_DISABLED", default=False)
         )
+        self._ci_visibility_unittest_enabled = asbool(os.getenv("DD_CIVISIBILITY_UNITTEST_BETA_ENABLED", default=False))
```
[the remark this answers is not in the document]
juan-fernandez: yeah that's a good catch, I'd avoid it too. 

This is just so we can merge a partially developed support (only test visibility) without affecting customers, but it's not like we're going through any beta phase. 

`DD_CIVISIBILITY_UNITTEST_ENABLED` sounds good to me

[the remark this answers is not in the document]
ericlaz: Should be changed now, thanks for addressing it!

tests/contrib/unittest_plugin/test_unittest.py
ericlaz: Good catch! I just added a test for this specific setup.

romainkomorndatadog (approved):

emmettbutler (approved):

riotfile.py
```diff
@@ -1726,6 +1726,13 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
                 ),
             ],
         ),
+        Venv(
+            name="unittest",
+            command="pytest --no-ddtrace {cmdargs} tests/contrib/unittest_plugin/",
+            pkgs={"msgpack": latest},
+            env={"DD_CIVISIBILITY_UNITTEST_ENABLED": "1"},
+            pys=select_pys(min_version="2.7", max_version="3.11"),
```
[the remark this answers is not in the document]
ericlaz: Thanks! I wasn't aware of this

emmettbutler (approved):

romainkomorndatadog (approved):

releasenotes/notes/add-unittest-support-024c4664f6546c21.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    Adds test level visibility support for unittest
```
mabdinur, line 4: nit:
```suggestion
    ci visibility: Adds test level visibility for `unittest <https://docs.python.org/3/library/unittest.html>`_

```

ddtrace/settings/config.py
```diff
@@ -334,6 +334,7 @@ def __init__(self):
         self._ci_visibility_intelligent_testrunner_enabled = asbool(
             os.getenv("DD_CIVISIBILITY_ITR_ENABLED", default=False)
         )
+        self._ci_visibility_unittest_enabled = asbool(os.getenv("DD_CIVISIBILITY_UNITTEST_ENABLED", default=False))
```
mabdinur, line 337: We need to document this configuration here: https://github.com/DataDog/dd-trace-py/blob/1.x/docs/configuration.rst?plain=1#L19

ddtrace/contrib/unittest/patch.py
```diff
@@ -0,0 +1,183 @@
...
+from ddtrace.internal.utils.wrappers import unwrap as _u
+from ddtrace.vendor import wrapt
+
+
+# unittest default settings
+config._add(
+    "unittest",
+    dict(
+        _default_service="unittest",
+        operation_name=os.getenv("DD_UNITTEST_OPERATION_NAME", default="unittest.test"),
```
mabdinur, line 25: We should document this environment variable too. Also Is this configuration set in other apm client libraries?

ddtrace/contrib/unittest/patch.py
```diff
@@ -0,0 +1,183 @@
...
+    if (
+        not config._ci_visibility_unittest_enabled
+        or getattr(unittest, "_datadog_patch", False)
+        or _CIVisibility.enabled
+    ):
+        return
+
+    _CIVisibility.enable(config=ddtrace.config.unittest)
+
+    unittest._datadog_patch = True
```
mabdinur, line 92: Should this attribute be set using `setattr`? From a quick glance this should raise an AttributeError

ddtrace/contrib/unittest/patch.py
```diff
@@ -0,0 +1,183 @@
...
+    if is_unittest_support_enabled() and _is_valid_result(instance, args):
+        _set_test_span_status(args, test.Status.SKIP.value)
+
+    return result
+
+
+def start_test_wrapper_unittest(func, instance, args, kwargs):
+    if is_unittest_support_enabled():
+        tracer = getattr(unittest, "_datadog_tracer", _CIVisibility._instance.tracer)
+        span = tracer._start_span(
```
mabdinur, line 156: nit: Why aren't we using `tracer.trace(...)` here? Are we just avoiding activating the span? If so, we should leave a comment to document this behavior

ddtrace/contrib/unittest/patch.py
```diff
@@ -0,0 +1,183 @@
...
+        span.set_tag_str(test.NAME, _extract_test_method_name(instance))
+        span.set_tag_str(test.SUITE, suite_name)
+        span.set_tag_str(test.MODULE, _extract_module_name_from_test_method(instance))
+
+        span.set_tag_str(test.STATUS, test.Status.FAIL.value)
+        span.set_tag_str(test.CLASS_HIERARCHY, suite_name)
+        _CIVisibility.set_codeowners_of(_extract_test_file_name(instance), span=span)
+
+        _store_span(instance, span)
+        result = func(*args, **kwargs)
```
mabdinur, line 179: Can we use a contextmanager here? This will allow use to set the error status and message on a span if an exception is raised in `result = func(*args, **kwargs)`. 

ex: 

```
with tracer.trace(ddtrace.config.unittest.operation_name,
         service=_CIVisibility._instance._service,
         resource="unittest.test",
         span_type=SpanTypes.TEST) as span:
        
         span.set_tag_str(_EVENT_TYPE, SpanTypes.TEST)

        span.set_tag_str(COMPONENT, COMPONENT_VALUE)
        span.set_tag_str(SPAN_KIND, KIND)

        span.set_tag_str(test.FRAMEWORK, FRAMEWORK)
        span.set_tag_str(test.TYPE, SpanTypes.TEST)
        suite_name = _extract_suite_name_from_test_method(instance)
        span.set_tag_str(test.NAME, _extract_test_method_name(instance))
        span.set_tag_str(test.SUITE, suite_name)
        span.set_tag_str(test.MODULE, _extract_module_name_from_test_method(instance))

        span.set_tag_str(test.STATUS, test.Status.FAIL.value)
        span.set_tag_str(test.CLASS_HIERARCHY, suite_name)
        _CIVisibility.set_codeowners_of(_extract_test_file_name(instance), span=span)
        _store_span(instance, span)
        return result
   ```

File: ddtrace/contrib/unittest/patch.py [hunk 19]
```diff
@@ -0,0 +1,183 @@
...
+    _u(unittest.TextTestResult, "addError")
+    _u(unittest.TextTestResult, "addSkip")
+    _u(unittest.TestCase, "run")
+
+    unittest._datadog_patch = False
+    _CIVisibility.disable()
+
+
+def _set_test_span_status(args, status):
+    test_item = args[0]
```
mabdinur, line 119: We are we only checking the args for the status of the test? Can the status of a test be passed in as a keyword argument? If so we should use this helper: https://github.com/DataDog/dd-trace-py/blob/00f9ed9a279b0786f38d3f5c51599cf6337ff053/ddtrace/internal/utils/__init__.py#L17

ddtrace/contrib/unittest/patch.py
```diff
@@ -0,0 +1,183 @@
...
+        return
+
+    _u(unittest.TextTestResult, "addSuccess")
+    _u(unittest.TextTestResult, "addFailure")
+    _u(unittest.TextTestResult, "addError")
+    _u(unittest.TextTestResult, "addSkip")
+    _u(unittest.TestCase, "run")
+
+    unittest._datadog_patch = False
+    _CIVisibility.disable()
```
mabdinur, line 115: Will disabling the civisibility writer here disable civisiblity for all integrations (ex: pytest)? Is there a valid use case where more than one civisibility integration is enabled and used? Are planning to test this interaction?

ddtrace/contrib/unittest/patch.py
```diff
@@ -0,0 +1,183 @@
...
+    span.set_tag_str(test.STATUS, status)
+    if status == test.Status.FAIL.value:
+        exc_info = args[1]
+        span.set_exc_info(exc_info[0], exc_info[1], exc_info[2])
+    elif status == test.Status.SKIP.value:
+        span.set_tag_str(test.SKIP_REASON, _extract_test_skip_reason(args))
+
+
+def add_success_test_wrapper(func, instance, args, kwargs):
+    if is_unittest_support_enabled() and _is_valid_result(instance, args):
```
mabdinur, line 132: Same concern here. We only parse the args and not the kwargs.

tests/contrib/unittest_plugin/test_unittest.py
```diff
@@ -0,0 +1,448 @@
...
+from ddtrace.contrib.unittest.patch import _set_tracer
+from ddtrace.contrib.unittest.patch import patch
+from ddtrace.ext import SpanTypes
+from ddtrace.ext import test
+from ddtrace.internal.constants import COMPONENT
+from tests.utils import TracerTestCase
+from tests.utils import override_env
+
+
+class UnittestTestCase(TracerTestCase):
```
mabdinur, line 21: These tests look good. Can we add one or two snapshot tests to validate the structure of the tracer generated by the unittest integration: https://ddtrace.readthedocs.io/en/stable/contributing-integrations.html#snapshot-tests. Let me know if you need help with this, the doc linked above is new

mabdinur (commented):
Good job. I left some nits and questions. Also heads up this PR could break your proposal but resolving the conflicts/test failures should be straight forward: https://github.com/DataDog/dd-trace-py/pull/6671

ddtrace/contrib/unittest/patch.py
ericlaz: It was set using `setattr` but the linter was raising an error about it. See: [CI workflow](https://app.circleci.com/pipelines/github/DataDog/dd-trace-py/44127/workflows/d4a371fc-0f36-4dab-8f89-86f91ae42cf2/jobs/2911016/parallel-runs/0/steps/0-104)

ddtrace/settings/config.py
juan-fernandez: hey @mabdinur. I don't think we need to instrument this feature flag because it's purely to avoid shipping a half finished feature to customers. This PR is only the first piece of the unittest visibility support (the last one is https://github.com/DataDog/dd-trace-py/pull/6611, which will remove this flag). Doing all the work in a single PR would've made it way harder to review and that's why we split it.

ddtrace/contrib/unittest/patch.py
romainkomorndatadog: When do setting an attribute cause an `AttributeError` exception that also doesn't happen with `setattr()` ?

Doesn't it behave the same?
```
>>> my_none = None
>>> my_none.blah = True
AttributeError: 'NoneType' object has no attribute 'blah'
>>> setattr(my_none, "blah", True)
AttributeError: 'NoneType' object has no attribute 'blah'
```
vs
```
>>> my_func = lambda: None
>>> my_func.blah = True
>>> setattr(my_func, "blah", False)
```

ddtrace/contrib/unittest/patch.py
romainkomorndatadog: Isn't one of the points of `unittest`'s runner that it'll capture those exceptions (like `pytest`), to then mark the test as failed?

Using a context here would at least capture exceptions in `start_test_wrapper_unittest()` itself, even if the ones in `func()` end up getting caught, so maybe that's worth it.

Then again, I'm actually not sure what would happen at our intake if we had an exception here that still ended up sending a span. For example, if we had an exception setting the module, suite, or name tags, then imagine the intake would discard it anyway.

ddtrace/contrib/unittest/patch.py
ericlaz: Yes, this is used in `pytest`: [docs link](https://ddtrace.readthedocs.io/en/stable/integrations.html#id183). This should've been addressed now.

ddtrace/contrib/unittest/patch.py
ericlaz: I'm not sure if I follow, the `add_success_test_wrapper` function gets called whenever `unittest` calls `TextTestResult.addSuccess` which happens whenever a test passes. We only need to set the span as `PASS`. Furthermore, the `kwargs` are empty in this function call.

ddtrace/contrib/unittest/patch.py
ericlaz: Yes, it should disable the other CIVisibility integrations. `pytest` will always override the `unittest` integration since `pytest` supports `unittest` out of the box, so in case `pytest` is enabled this will cause `unittest` to be automatically unpached.

ddtrace/contrib/unittest/patch.py
ericlaz: The `args` parameter refers to a tuple of `(test_item, reason)`. I have just changed the variable naming to make it clearer. I do not think it is necessary to use the helper in this case.

ddtrace/contrib/unittest/patch.py
ericlaz: @mabdinur, this should've been addressed now. Thanks!

ddtrace/contrib/unittest/patch.py
ericlaz: Just switched to using a contextmanager instead, thanks!

mabdinur (dismissed):

.circleci/config.templ.yml
```diff
@@ -784,6 +784,12 @@ jobs:
       - run_test:
           pattern: 'pytest'
 
+  unittest:
+    <<: *contrib_job
+    steps:
+      - run_test:
+          pattern: 'unittest'
```
mabdinur, line 791:
```suggestion
          pattern: 'unittest'
          snapshot: true
        
```

riotfile.py
```diff
@@ -1644,6 +1644,25 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
...
+            name="unittest-snapshot",
+            command="pytest --no-ddtrace {cmdargs} tests/contrib/unittest_plugin/snapshot/",
+            pkgs={"msgpack": latest},
+            env={
+                "DD_CIVISIBILITY_UNITTEST_ENABLED": "1",
+                "DD_TRACE_AGENT_URL": "http://localhost:9126",
+                "AGENT_VERSION": "testagent",
+            },
+            pys="3.7",
+        ),
```
mabdinur, line 1665: We shouldn't need this. Running the circleci job with `snapshot: True` configuration should set the ddtrace agent url to the expected value.

romainkomorndatadog (approved):

mabdinur (approved):

PR #6588, opened 2023-08-05
feat(datastreams): track Kafka lag

piochelepiotr:
Data Streams monitoring can generate the Kafka lag in seconds.
In order to do that correctly, we need to track the commits & produced offsets.

This PR adds that to the Python tracer.
A few limits of the approach are:
- if auto commit is enabled, we consider a message that is read as if it was acked. In reality, the message will be acked later (auto commits are done in the background on regular intervals)
- In order for the produce callback to be called, the produce.poll() method needs to be called
- At shutdown, it's important we don't loose any payloads, since it might lead to wrongly thinking some messages were not acked. Even with this PR changes, it's still possible if the app crashes abruptly to lose some payloads.

I tested the changes by running the tracer locally:

https://app.datadoghq.com/data-streams/map?env=test&selectedNode=python-forwarder_0&selectedTab=throughput&start=1691246632434&end=1691247532434&paused=false

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/tracer.py
```diff
@@ -300,6 +305,8 @@ def __init__(
 
         self._hooks = _hooks.Hooks()
         atexit.register(self._atexit)
+        signal.signal(signal.SIGTERM, signal_handler)
```
mabdinur, line 308: This signal handler can overwrite (or be overwritten) by user's application. Here's how we wrap existing signal handlers for the serverless product: https://github.com/DataDog/dd-trace-py/blob/5183394cf2cc0eedc57398ed60e1217928db4e8a/ddtrace/contrib/aws_lambda/patch.py#L68. 

Also is registering an atexit hook sufficient? We do not intercept the a sigterm signal to send any apm events. Doing so for DSM seems a bit overkill. I worry that this could lead to some undefined behavior since we can not guarantee the state of the ddtrace library when an application is killed suddenly.

ddtrace/tracer.py
```diff
@@ -1045,18 +1052,18 @@ def shutdown(self, timeout=None):
         """
         with self._shutdown_lock:
             # Thread safety: Ensures tracer is shutdown synchronously
+            try:
```
mabdinur, line 1055: Why do we need to move this up? Can we leave comment here

releasenotes/notes/kafka-lag-support-data-streams-c6b7aed85bf2cf14.yaml
```diff
@@ -0,0 +1,3 @@
+features:
+  - |
+    data_streams: Start tracking Kafka lag in seconds.
```
mabdinur, line 3: We should use third person present tense here ([docs](https://ddtrace.readthedocs.io/en/stable/releasenotes.html#style-guidelines)). 

Also can we be more explicit about the customer impact of this feature. Does this power a certain dashboard/view? Whats the minimum time resolution? If the kafka lag less than 1 second, is this value rounded to zero?
 
```suggestion
    data_streams: Starts tracking Kafka lag in seconds.
```

tests/datastreams/test_processor.py
```diff
@@ -1,6 +1,8 @@
 import time
```
mabdinur, line 1: Can we add test coverage for the dsm signal handlers? 

Here are some scenarios we could test:
 - a signal handler already exists
 - A process generating kafka events forks
 - A process generating kafka events is killed gracefully
 - A process generating kafka events is killed abruptly

ddtrace/contrib/kafka/patch.py
```diff
@@ -71,6 +79,7 @@ def patch():
 
     trace_utils.wrap(TracedProducer, "produce", traced_produce)
     trace_utils.wrap(TracedConsumer, "poll", traced_poll)
+    trace_utils.wrap(TracedConsumer, "commit", traced_commit)
```
mabdinur, line 82: Why are we wrapping the tracedconsumer? Can we move logic from `traced_commit` to `TracedConsumer.commit`

ddtrace/tracer.py
piochelepiotr: In Kubernetes, applications are closed using a signal.
If we don't send the data to the backend in this case, we drop some payloads. Which for this use case, causes us to incorrectly track lag.
I noticed it while testing the tracer, and trying to understand why lag was tracked incorrectly after shutdown.

piochelepiotr: I can move the signal handling to DSM only, and not the whole tracer.

ddtrace/tracer.py
piochelepiotr: some other parts of the code crash during shutdown. I could fix it, but I don't think it's in the scope of this PR

ddtrace/contrib/kafka/patch.py
piochelepiotr: How is that different from `poll` or `produce` ?
I'm not sure I understand your suggestion.

mabdinur: I guess this suggestion would also apply to poll or produce. There's no need to refactor this code right now. I think the current approach is fine.

mabdinur (dismissed):

mabdinur (dismissed):

mabdinur (approved):

hauntsaninja:
This addition is problematic, causes `ValueError: signal only works in main thread of the main interpreter`. Or things like:
```
  File "/root/.pyenv/versions/3.11.5/lib/python3.11/site-packages/ddtrace/__init__.py", line 42, in <module>
    tracer = Tracer()
             ^^^^^^^^
  File "/root/.pyenv/versions/3.11.5/lib/python3.11/site-packages/ddtrace/tracer.py", line 300, in __init__
    register_on_exit_signal(self._atexit)
  File "/root/.pyenv/versions/3.11.5/lib/python3.11/site-packages/ddtrace/internal/atexit.py", line 70, in register_on_exit_signal
    signals.handle_signal(signal.SIGINT, handle_exit)
  File "/root/.pyenv/versions/3.11.5/lib/python3.11/site-packages/ddtrace/internal/utils/signals.py", line 27, in handle_signal
    return signal.signal(sig, wrap_signals)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: Can't set signal handler for SIGINT while SIGINT is being deferred within a DeferSigint context.
```

mabdinur:
> This addition is problematic, causes `ValueError: signal only works in main thread of the main interpreter`. Or things like:
> 
> ```
>   File "/root/.pyenv/versions/3.11.5/lib/python3.11/site-packages/ddtrace/__init__.py", line 42, in <module>
>     tracer = Tracer()
>              ^^^^^^^^
>   File "/root/.pyenv/versions/3.11.5/lib/python3.11/site-packages/ddtrace/tracer.py", line 300, in __init__
>     register_on_exit_signal(self._atexit)
>   File "/root/.pyenv/versions/3.11.5/lib/python3.11/site-packages/ddtrace/internal/atexit.py", line 70, in register_on_exit_signal
>     signals.handle_signal(signal.SIGINT, handle_exit)
>   File "/root/.pyenv/versions/3.11.5/lib/python3.11/site-packages/ddtrace/internal/utils/signals.py", line 27, in handle_signal
>     return signal.signal(sig, wrap_signals)
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> ValueError: Can't set signal handler for SIGINT while SIGINT is being deferred within a DeferSigint context.
> ```

Hey @hauntsaninja,

Thanks for calling this out! Can you open a github issue for this bug and include a small reproduction. This will help us better track this error. 

cc: @piochelepiotr

hauntsaninja:
I opened https://github.com/DataDog/dd-trace-py/issues/7014, apologies, don't currently have time to make a minimal repro

hvtuananh:
I added a minimal repro to #7014

PR #6592, opened 2023-08-07
ci: fix github action trigger

gnufede:
## Checklist

## Reviewer Checklist

- [ ] Title is accurate.

### DIFF
--- a/.github/workflows/system-tests.yml
+++ b/.github/workflows/system-tests.yml
@@ -7,8 +7,9 @@ on:
       - 1.x
   pull_request:
     paths:
-      - ddtrace/*
-      - setup*
+      - ddtrace/**
+      - setup.py
+      - setup.cfg
       - pyproject.toml
   workflow_dispatch: {}
   schedule:


.github/workflows/system-tests.yml
```diff
@@ -7,8 +7,9 @@ on:
       - 1.x
   pull_request:
     paths:
-      - ddtrace/*
-      - setup*
```
P403n1x87, line 11: oh doesn't `setup*` work to match both `setup.py` and `setup.cfg`?

P403n1x87 (approved):

.github/workflows/system-tests.yml
romainkomorndatadog: Seems like it should according to the last example on https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#patterns-to-match-file-paths .

romainkomorndatadog: `ddtrace/*` would only match files in `ddtrace/` itself, though, so `ddtrace/**` is needed for anything nested more than one level deep.

romainkomorndatadog (approved):

juanjux (approved):

.github/workflows/system-tests.yml
gnufede: I changed `setup*` just to be more explicit in case any other file with a `setup` prefix is added in the future

PR #6611, opened 2023-08-08
feat(unittest): add full test suite level visibility for unittest

ericlaz:
## Motivation

Given that the [PR on test level visibility](https://github.com/DataDog/dd-trace-py/pull/6559) has been merged, it is time to propose this PR on TSLV for `unittest` suites.

## Changelog

This PR aims to add TSLV support for `unittest` supports. This feature is no longer behind a feature flag and it can be enabled using `ddtrace-run python -m unittest` or manually patching `unittest` using `patch(...)`.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/ext/test.py
```diff
@@ -32,8 +32,8 @@
 # Test Parameters
 PARAMETERS = "test.parameters"
 
-# Pytest Result (XFail, XPass)
-RESULT = TEST_RESULT = "pytest.result"
+# Test Result (XFail, XPass)
+RESULT = TEST_RESULT = "test.result"
```
romainkomorndatadog, line 36: I think this is pretty harmless but please double-check with backend/core/app folks to make sure we're not using `pytest.result` elsewhere.

https://github.com/search?q=org%3ADataDog%20pytest.result&type=code suggests we're not, but "famous last words".

romainkomorndatadog (approved):
Looks good to me. It's pretty straightforward. Got a few notes but nothing blocking.

ddtrace/contrib/unittest/patch.py
```diff
@@ -8,21 +9,35 @@
...
 
+log = get_logger(__name__)
+
 # unittest default settings
 config._add(
     "unittest",
     dict(
         _default_service="unittest",
         operation_name=os.getenv("DD_UNITTEST_OPERATION_NAME", default="unittest.test"),
+        strict_naming=asbool(os.getenv("DD_CIVISIBILITY_UNITTEST_STRICT_NAMING", default=True)),
```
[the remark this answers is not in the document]
romainkomorndatadog: Actually, I meant `config.rst`, but this isn't configuration for the tracer, so never mind. :)

ddtrace/contrib/unittest/patch.py
```diff
@@ -33,30 +48,98 @@ def get_version():
...
+
+def _is_test_suite(item):
+    return hasattr(item, "_datadog_object") and item._datadog_object == "suite"
+
+
+def _is_test(item):
+    if type(item) == unittest.TestSuite or not hasattr(item, "_testMethodName"):
+        return False
+    if ddtrace.config.unittest.strict_naming and not (
+        len(item._testMethodName) >= 4 and item._testMethodName[0:4] == "test"
```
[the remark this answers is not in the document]
ericlaz: Good catch! Should've been addressed now.

The `unittest` library uses `test` according to the [official docs](https://docs.python.org/3/library/unittest.html):

> The simplest [TestCase](https://docs.python.org/3/library/unittest.html#unittest.TestCase) subclass will simply implement a test method (i.e. a method whose name starts with test) in order to perform specific testing code

ddtrace/contrib/unittest/patch.py
```diff
@@ -33,30 +48,98 @@ def get_version():
...
+def _is_test_module(item):
+    return hasattr(item, "_datadog_object") and item._datadog_object == "module"
+
+
+def _is_test_suite(item):
+    return hasattr(item, "_datadog_object") and item._datadog_object == "suite"
+
+
+def _is_test(item):
+    if type(item) == unittest.TestSuite or not hasattr(item, "_testMethodName"):
```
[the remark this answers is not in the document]
ericlaz: Good catch! This has been addressed now

ddtrace/contrib/unittest/patch.py
```diff
@@ -65,31 +148,124 @@ def _extract_module_name_from_test_method(item):
...
-    return unittest and getattr(unittest, "_datadog_patch", False) and _CIVisibility.enabled
+def _extract_module_file_path(item):
+    if _is_test(item):
+        return os.path.relpath(inspect.getfile(item.__class__))
+
+    return ""
+
+
+def _generate_test_resource(suite_name, test_name):
+    return suite_name + "." + test_name
```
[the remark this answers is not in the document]
ericlaz: Good catch. Should be fixed now

ddtrace/contrib/unittest/patch.py
```diff
@@ -65,31 +148,124 @@ def _extract_module_name_from_test_method(item):
...
+
+def _is_invoked_by_cli(args):
+    return not args or (hasattr(_CIVisibility, "_datadog_entry") and _CIVisibility._datadog_entry == "cli")
+
+
+def _is_invoked_by_text_test_runner():
+    return hasattr(_CIVisibility, "_datadog_entry") and _CIVisibility._datadog_entry == "TextTestRunner"
+
+
+def _get_python_version():
```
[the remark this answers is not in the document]
ericlaz: Thank you! This is now addressed.

ddtrace/contrib/unittest/patch.py
```diff
@@ -65,31 +148,124 @@ def _extract_module_name_from_test_method(item):
...
+
+
+def _generate_module_suite_path(test_module_path, test_suite_name):
+    return test_module_path + "." + test_suite_name
+
+
+def _populate_suites_and_modules(test_objects, seen_suites, seen_modules):
+    for test_object in test_objects:
+        if not _is_test(test_object):
+            continue
```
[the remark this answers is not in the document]
ericlaz: This is expected. `unittest` populates lists of test objects with objects which happen to be empty, this is pretty common by the looks of it.

ddtrace/contrib/unittest/patch.py
```diff
@@ -102,85 +278,321 @@ def patch():
...
-    elif status == test.Status.SKIP.value:
-        span.set_tag_str(test.SKIP_REASON, reason)
+    if status == test.Status.SKIP.value:
+        span.set_tag_str(test.SKIP_REASON, skip_reason)
+
+
+def _set_test_span_result(test_item, result):
+    span = _extract_span(test_item)
+    if not span:
+        return
```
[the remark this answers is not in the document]
ericlaz: Good point! It now logs as expected.

ddtrace/contrib/unittest/patch.py
```diff
@@ -34,30 +50,92 @@ def get_version():
...
+        return _CIVisibility._unittest_data["suites"][suite_identifier].get("suite_span")
+    return None
 
 
-def _extract_class_hierarchy_name(item):
+def _update_status_item(item, status):
+    existing_status = item.get_tag(test.STATUS)
+    if existing_status and (status == test.Status.SKIP.value or existing_status == test.Status.FAIL.value):
+        return
+    item.set_tag_str(test.STATUS, status)
```
mabdinur, line 135: nit: In some functions we return None and other we don't. We should try to be consistent. 
```suggestion
        return None
    item.set_tag_str(test.STATUS, status)
    return None
```

ddtrace/contrib/unittest/patch.py
```diff
@@ -66,31 +144,128 @@ def _extract_module_name_from_test_method(item):
...
+
+def _extract_test_reason(item):
     return item[1]
 
 
 def _extract_test_file_name(item):
     return os.path.basename(inspect.getfile(item.__class__))
 
 
-def is_unittest_support_enabled():
```
mabdinur, line 77: This function is now part of our public api. We can't just remove it. We will need to deprecate it first. Here's an example on how to do that: https://github.com/DataDog/dd-trace-py/commit/c0b64cdc648845918c463155ad1a5fe8ad2dc309. We will also need to add a deprecation release note.

ddtrace/contrib/unittest/patch.py
```diff
@@ -66,31 +144,128 @@ def _extract_module_name_from_test_method(item):
...
+    del _CIVisibility._unittest_data
+
+
+def _update_remaining_suites_and_modules(test_module_suite_path, test_module_path, test_module_span, test_suite_span):
+    _CIVisibility._unittest_data["suites"][test_module_suite_path]["remaining_tests"] -= 1
+    if _CIVisibility._unittest_data["suites"][test_module_suite_path]["remaining_tests"] == 0:
+        _CIVisibility._unittest_data["modules"][test_module_path]["remaining_suites"] -= 1
+        _finish_span(test_suite_span)
+    if _CIVisibility._unittest_data["modules"][test_module_path]["remaining_suites"] == 0:
+        _finish_span(test_module_span)
```
mabdinur, line 261: This method has many nested dictionary look ups. Can this fail in unexpected ways? Should we wrap this in a try-except and log errors?

ddtrace/contrib/unittest/patch.py
```diff
@@ -103,85 +278,328 @@ def patch():
...
+        if test_suite_span is None and test_module_suite_path in _CIVisibility._unittest_data["suites"]:
+            test_suite_span = _start_test_suite_span(instance)
+            _CIVisibility._unittest_data["suites"][test_module_suite_path]["suite_span"] = test_suite_span
+        if not test_module_span or not test_suite_span:
+            log.debug("Suite and/or module span not found for test: %s", test_name)
+            return func(*args, **kwargs)
+        span = _start_test_span(instance, test_suite_span)
+        result = func(*args, **kwargs)
+        _update_status_item(test_suite_span, span.get_tag(test.STATUS))
+        span.finish()
```
mabdinur, line 392: We should use a context manager here. This will ensure span.finish() is always called.
```suggestion
        with _start_test_span(instance, test_suite_span) as span:
          result = func(*args, **kwargs)
          _update_status_item(test_suite_span, span.get_tag(test.STATUS))

File: releasenotes/notes/add-unittest-support-024c4664f6546c21.yaml [hunk 13]
```diff
@@ -1,4 +1,4 @@
 ---
 features:
   - |
-    CI Visibility: adds test level visibility for `unittest <https://docs.python.org/3/library/unittest.html>`_
+    CI Visibility: adds test level visibility for `unittest <https://docs.python.org/3/library/unittest.html>`_
```
mabdinur, line 4:
```suggestion
    CI Visibility: adds test level visibility for `unittest <https://docs.python.org/3/library/unittest.html>`_

```

ddtrace/contrib/unittest/patch.py
```diff
@@ -34,30 +50,92 @@ def get_version():
...
 
 
-def _store_span(item, span):
+def _store_test_span(item, span):
     """Store span at `unittest` instance."""
     item._datadog_span = span
 
 
+def _store_module_identifier(test_object):
+    """Store module identifier at `unittest` module instance."""
```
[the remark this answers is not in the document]
romainkomorndatadog: This is only going to go into `2.0`, can we do proper Python type hints instead?

tests/.suitespec.json
```diff
@@ -1101,4 +1101,4 @@
             "tests/snapshots/tests.contrib.{suite}.*"
         ]
     }
-}
+}
```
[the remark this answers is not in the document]
romainkomorndatadog: Not entirely relevant, but why isn't `flake8` catching this? This is the kind of formatting thing that our pre-commit hooks should catch, and that `hatch run lint:fmt` should fix, and that `pre_check` too should reject.

ddtrace/contrib/unittest/patch.py
```diff
@@ -66,31 +144,128 @@ def _extract_module_name_from_test_method(item):
...
 
-def _extract_test_skip_reason(item):
+def _extract_module_name_from_module(item):
+    if _is_test(item):
+        return type(item).__module__
+    return ""
+
+
+def _extract_test_reason(item):
     return item[1]
```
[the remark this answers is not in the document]
ericlaz: Yes, it will be using the test object which is a tuple like `(test_class, test_name)`.

ddtrace/contrib/unittest/patch.py
juan-fernandez: Do we need to go through the deprecation process for this one? unittest is in beta right now (before this PR is merged) and behind a feature flag. Do we commit to public APIs in that case?

ddtrace/contrib/unittest/patch.py
ericlaz: It should not, we're checking if `test_module_path` and `test_module_suite_path` exists in the `_CIVisibility._unittest_data` dictionary [here](https://github.com/DataDog/dd-trace-py/blob/a7b14cb8a7239134b39995d15e55ba476fe93df0/ddtrace/contrib/unittest/patch.py#L381)

ddtrace/contrib/unittest/patch.py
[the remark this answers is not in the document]
emmettbutler: I second the recommendation for Python-native type hinting.

tests/contrib/unittest_plugin/test_unittest.py
```diff
@@ -169,29 +300,59 @@ def test_will_be_skipped_second(self):
...
+
+            assert spans[3].name == TEST_OPERATION_NAME
+            assert spans[3].get_tag(test.NAME) == "test_will_be_skipped_first"
+            assert spans[3].get_tag(test.SKIP_REASON) == "demonstrating skipping with a reason"
+            assert spans[3].get_tag(test.MODULE) == "tests.contrib.unittest_plugin.test_unittest"
+            assert spans[3].get_tag(test.SUITE) == "UnittestExampleTestCase"
+            assert spans[3].get_tag(MODULE_ID) == str(test_module_span.span_id)
+            assert spans[3].get_tag(SUITE_ID) == str(test_suite_span.span_id)
+
+            assert spans[4].name == TEST_OPERATION_NAME
```
emmettbutler, line 329: These big blocks of assertions are duplicated a few times. Consider deduplicating them with a function.

ddtrace/contrib/unittest/patch.py
```diff
@@ -34,30 +50,93 @@ def get_version():
...
+
+
 def _extract_span(item):
     """Extract span from `unittest` instance."""
     return getattr(item, "_datadog_span", None)
 
 
-def _extract_test_method_name(item):
+def _extract_command_name_from_session(session):
+    """Extract command name from `unittest` instance."""
```
emmettbutler, line 100: This docstring doesn't add any information that isn't contained in the function name.

ddtrace/contrib/unittest/patch.py
emmettbutler: It's not documented anywhere, but anecdotally I believe the answer is no, we do not maintain public APIs for beta products.

ddtrace/contrib/unittest/patch.py
```diff
@@ -65,31 +148,124 @@ def _extract_module_name_from_test_method(item):
...
+
+def _set_identifier(item, name):
+    item._datadog_object = name
 
 
 def _is_valid_result(instance, args):
     return instance and type(instance) == unittest.runner.TextTestResult and args
 
 
+def _is_valid_test_call(kwargs):
```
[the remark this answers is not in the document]
emmettbutler: This is a good opportunity for an explanatory docstring.

ddtrace/contrib/unittest/patch.py
emmettbutler: Beyond the potential for failure, this code is kind of hard to read. Intermediate variables could help with that.

ddtrace/contrib/unittest/patch.py
```diff
@@ -103,85 +279,327 @@ def patch():
...
+        return func(*args, **kwargs)
+    _CIVisibility._datadog_entry = "TextTestRunner"
+    try:
+        if not hasattr(_CIVisibility, "_datadog_session_span"):
+            _CIVisibility._datadog_session_span = _start_test_session_span(instance)
+            _CIVisibility._datadog_expected_sessions = 0
+            _CIVisibility._datadog_finished_sessions = 0
+        _CIVisibility._datadog_expected_sessions += 1
+        result = func(*args, **kwargs)
+    except Exception as e:
```
emmettbutler, line 588: Can this exception handler expect a more specific exception type? If so, making that change would reduce the possibility of errors being unintentionally swallowed.

ddtrace/contrib/unittest/patch.py
```diff
@@ -103,85 +279,327 @@ def patch():
...
+
+
+def handle_text_test_runner_wrapper(func, instance, args, kwargs):
+    """
+    Creates session and module spans if the unittest is called through the `TextTestRunner` method
+    """
+    if _is_invoked_by_cli(instance):
+        return func(*args, **kwargs)
+    _CIVisibility._datadog_entry = "TextTestRunner"
+    try:
```
emmettbutler, line 581: Which part of this large `try` block is expected to raise an exception? Removing everything from it except the specific lines that generate exceptions would make this code more maintainable.

emmettbutler (changes requested):

tests/contrib/unittest_plugin/test_unittest.py
romainkomorndatadog: I agree that they're long and duplicated, but IMO, debugging failing tests is so much simpler when you have a specific line to go to (I don't even like the `for` loop all that much).

I've looked at the [test span class](https://github.com/DataDog/dd-trace-py/blob/1.x/tests/utils.py#L596), and thought about replacing these asserts with test spans and using `span.matches()`, but I think it would not end up being much shorter or less unwieldy.

ericlaz: I agree with @romainkomorndatadog. I was debugging my tests earlier and found that the `for` loop-type tests are much harder since you do not know which span failed.

ddtrace/contrib/unittest/patch.py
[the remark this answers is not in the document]
ericlaz: Good idea! This is now addressed, it supports type hinting on all functions.

ddtrace/contrib/unittest/patch.py
ericlaz: I added some new variables to clear up this function, hopefully it is more readable now. Thank you!

ddtrace/contrib/unittest/patch.py
ericlaz: Good catch! The `try` block is supposed to be on the `result = func(...)` line, this is now addressed.

emmettbutler (approved):

ddtrace/contrib/unittest/patch.py
```diff
@@ -33,64 +50,253 @@ def get_version():
...
-def _store_span(item, span):
-    """Store span at `unittest` instance."""
+def _store_test_span(item, span: ddtrace.Span):
+    """Store datadog span at `unittest` test instance."""
     item._datadog_span = span
 
 
-def _extract_span(item):
-    """Extract span from `unittest` instance."""
+def _store_module_identifier(test_object: unittest.TextTestRunner):
```
mabdinur, line 63: The 2.x branch dropped support for python2 so I guess this works. What I had in mind was something like [this](https://github.com/DataDog/dd-trace-py/blob/c6808b2d6d016aa600b7c52c50b172eeca44bd0d/ddtrace/internal/sma.py#L39). Using type hints in comments better aligns with how we currently use mypy in ddtrace. 

But this works. I think we should move towards using python3 syntax anyway.

releasenotes/notes/add-tslv-unittest-support-f2967051df9aa9aa.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    CI Visibility: adds full test suite level visibility for `unittest <https://docs.python.org/3/library/unittest.html>`_
```
mabdinur, line 4:
```suggestion
    CI Visibility: Adds full test suite level visibility for `unittest <https://docs.python.org/3/library/unittest.html>`_

```

ddtrace/contrib/unittest/patch.py
```diff
@@ -33,64 +50,253 @@ def get_version():
...
+def _extract_test_file_name(item) -> str:
     return os.path.basename(inspect.getfile(item.__class__))
 
 
-def is_unittest_support_enabled():
-    return unittest and getattr(unittest, "_datadog_patch", False) and _CIVisibility.enabled
+def _extract_module_file_path(item) -> str:
+    if _is_test(item):
+        return os.path.relpath(inspect.getfile(item.__class__))
+
```
mabdinur, line 172: nit
```suggestion
```

ddtrace/internal/ci_visibility/encoder.py
```diff
@@ -35,11 +34,7 @@ class CIVisibilityEncoderV01(BufferedEncoder):
     ALLOWED_METADATA_KEYS = ("language", "library_version", "runtime-id", "env")
     PAYLOAD_FORMAT_VERSION = 1
     TEST_SUITE_EVENT_VERSION = 1
-    # Change to 1 to allow unittest tests without module or session spans
-    if config._ci_visibility_unittest_enabled:
-        TEST_EVENT_VERSION = 1
-    else:
-        TEST_EVENT_VERSION = 2
+    TEST_EVENT_VERSION = 2
```
mabdinur, line 37: With this change ``TEST_EVENT_VERSION`` is always 2. Is this documented somewhere? Should we even call this out?

riotfile.py
```diff
@@ -103,6 +103,7 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
         "DD_CIVISIBILITY_AGENTLESS_ENABLED": "1",
         "DD_CIVISIBILITY_CODE_COVERAGE_ENABLED": "1",
         "DD_CIVISIBILITY_ITR_ENABLED": "1",
+        "DD_PATCH_MODULES": "unittest:false",
```
mabdinur, line 106: Why do we need this? Can we add comment to describe how enabling the unittest integration impacts ddtrace tests.

mabdinur (approved):
Left some nits. None of these suggestions are blocking

riotfile.py
ericlaz: Some integrations use `unittest` TestCases which, since it is patched, ends up generating more spans. This is a problem with our `pytest` integration too, since it supports `unittest` as shown [here](https://docs.pytest.org/en/7.1.x/how-to/unittest.html).

ddtrace/internal/ci_visibility/encoder.py
ericlaz: I do not think we should call this out, `TEST_EVENT_VERSION` was temporarily set to 1 due to the `unittest` beta. This is now reverted to pre-unittest branches.

PR #6612, opened 2023-08-08
chore(logging): remove DD_CALL_BASIC_CONFIG

Yun-Kim:
## Summary

This PR removes the `DD_CALL_BASIC_CONFIG` environment variable and `DD_LOG_FORMAT` constant from the public API, and features a subtle change in the library's logging functionality to avoid configuring the user application's root-level logger. Moving forward, we will configure the `ddtrace` level logger instead.

Additionally, this PR adds a new default behavior to ensure that ddtrace will always log to both stdout (stream) and/or file if configured to log to file. Previously ddtrace would only log to stream OR file, but this isn't intuitive for users.

## Background

`DD_CALL_BASIC_CONFIG` was deprecated as of #3168 (February 2022). This environment variable is currently used to enable the ddtrace library to call `logging.basicConfig()` to configure and format the root logger to print logs to stdout (when being run in `ddtrace-run` or through tracer debugging mode).

However, the ddtrace library is not a logging library and calling `logging.basicConfig()` will configure the root-level logger and invalidate any future calls to `logging.basicConfig()`, meaning this will potentially interfere with user logging configuration. User applications should be responsible for calling `logging.basicConfig()` instead of ddtrace.

## Proposed Solution

Instead of calling loggging.basicConfig(), we instead now configure the ddtrace level logger at module startup (no longer at sitecustomize.py or tracer startup) to add a stdout Streamhandler (or file logger if specified by `DD_TRACE_LOG_FILE`) to avoid interfering with root-level logger configurations. Note that we still configure logging injection at sitecustomize.py and tracer startup since that requires logging to be patched.

Note: this PR leaves one usage of `logging.basicConfig()` in `ddtrace/commands/ddtrace_run.py`. This will not impact user application logging because our `ddtrace-run` script replaces the currently running process with the bootstrap script and user application through `os.execl(...)`, meaning this will reset a new logging configuration to default and be unaffected once the bootstrap sitecustomize and user application starts running. This approach will leave ddtrace-run debugging logs to be printed to stdout but still avoids interfering with user application logging configurations.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


docs/upgrading.rst
```diff
@@ -41,7 +41,6 @@ Use the following patterns to identify the deprecated environment variables in a
 
     git grep -e "DATADOG_[A-Z_]*" \
       -e "DATADOG_SERVICE_NAME" \
-      -e "DD_CALL_BASIC_CONFIG" \
```
majorgreys, line 44: These upgrade instructions are specific to 0.x, so we might want to leave them unchanged. But for 2.x, we should make sure to have upgrade instructions for 1.x users.

Yun-Kim: Good catch - I'll revert this change and make a separate PR for upgrade docs.

ddtrace/_logger.py
```diff
@@ -32,14 +37,18 @@ def configure_ddtrace_logger():
...
 
 def _configure_ddtrace_debug_logger(logger):
     debug_enabled = asbool(os.environ.get("DD_TRACE_DEBUG", "false"))
-    if debug_enabled:
+    logs_injection_enabled = asbool(os.environ.get("DD_LOGS_INJECTION", "false"))
+
+    # Logs injection enabled requires a handler present on ddtrace
+    if debug_enabled or logs_injection_enabled:
+        if not os.environ.get("DD_TRACE_LOG_FILE"):
+            logger.addHandler(logging.StreamHandler())
```
brettlangdon, line 51: should we have a function similar to `_configure_ddtrace_file_logger`, but which configures the stream handler?

should we always configure the stream handler? (with a default log level of like `WARNING` or something?)

ddtrace/_logger.py
```diff
@@ -33,13 +38,14 @@ def configure_ddtrace_logger():
     """
     ddtrace_logger = logging.getLogger("ddtrace")
 
+    if not os.environ.get("DD_TRACE_LOG_FILE"):
+        ddtrace_logger.addHandler(logging.StreamHandler())
```
brettlangdon, line 42: should this be and and not or?

e.g. do we always configure a stream handler, and then optionally trigger a log handler if asked for?

Yun-Kim: We always need to configure a stream handler, otherwise logs injection will break if we remove `DD_CALL_BASIC_CONFIG` since it needs an active handler set on startup. 

My reasoning here is to either have a `StreamHandler` or a `FileHandler`, depending on if `DD_TRACE_LOG_FILE` is set, but avoid having both at the `ddtrace` level. This way we don't want to send logs to stdout if users want it sent to a file, and vice versa.

ddtrace/_logger.py
Yun-Kim: Log levels for unconfingured loggers are `WARNING` by default. To your point, I've changed this logic to always create a `StreamHandler` on startup (but unconfigured) so that it's available to be configured for trace debugging and log injections down the line.

ddtrace/tracer.py
```diff
@@ -81,22 +80,9 @@
...
-        message="`logging.basicConfig()` should be called in a user's application.",
-        removal_version="2.0.0",
-    )
-    if config.logs_injection:
-        # We need to ensure logging is patched in case the tracer logs during initialization
-        patch(logging=True)
-        logging.basicConfig(level=logging.DEBUG, format=DD_LOG_FORMAT)
-    else:
-        logging.basicConfig(level=logging.DEBUG)
+if config._debug_mode and config.logs_injection:
```
brettlangdon, line 83: same here, debug mode shouldn't matter right?

releasenotes/notes/remove-deprecated-items-2.0-89eac06472c59554.yaml
```diff
@@ -89,4 +89,22 @@ upgrade:
...
+    `log injection docs <https://ddtrace.readthedocs.io/en/stable/advanced_usage.html#update-log-format>`_.
+
+
+    "
+  - ".. _remove-basic-config:
+
+
+    ``DD_CALL_BASIC_CONFIG`` is removed.
+    There is no special configuration necessary to replace ``DD_CALL_BASIC_CONFIG``.
+    The ddtrace logger will log to standard out by default or to a file specified by ``DD_TRACE_LOG_FILE``.
```
brettlangdon, line 107:
```suggestion
    The ddtrace logger will log to standard out by default or additionally to a file specified by ``DD_TRACE_LOG_FILE``.
```

? this is a nit pick, not sure this really adds much clarification

brettlangdon (commented):
right now users will get us setting the log format by default for logs injection, do we need to still support that?

will something break for user's if we stop doing that? 🤔 

I suppose that is why this is a breaking change...

ddtrace/_logger.py
Yun-Kim: > My reasoning here is to either have a StreamHandler or a FileHandler, depending on if DD_TRACE_LOG_FILE is set, but avoid having both at the ddtrace level. This way we don't want to send logs to stdout if users want it sent to a file, and vice versa.

After offline discussion: we have decided to go with adding a `StreamHandler` by default regardless if `DD_TRACE_LOG_FILE` is specified, meaning that in cases where `DD_TRACE_LOG_FILE` is specified we will log to both stdout and log file. We want to avoid the above as it is potentially unclear to users where their stdout logs might have went if they added file logging as well in this scenario.

ddtrace/_logger.py
```diff
@@ -62,11 +66,21 @@ def _configure_ddtrace_file_logger(logger):
...
+def _configure_log_injection():
+    """
+    Ensures that logging is patched before we inject trace information into logs.
+    """
+    from ddtrace import patch
+
+    patch(logging=True)
+    ddtrace_logger = logging.getLogger("ddtrace")
+    if ddtrace_logger.hasHandlers():
+        ddtrace_logger.handlers[0].setFormatter(logging.Formatter(DD_LOG_FORMAT))
```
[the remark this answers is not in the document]
Yun-Kim: I'm pretty confident user code will always run after `_configure_log_injection()` is called since it's called on tracer module load and sitecustomize.py.

mabdinur (approved):

PR #6671, opened 2023-08-15
chore(telemetry): add integration version to telemetry integrations payload

wconti27:
This PR adds version to the APM Telemetry integrations payload. Version is added by requiring a `get_version` method for all integrations.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/contrib/asyncpg/patch.py
```diff
@@ -43,6 +43,13 @@
 
 log = get_logger(__name__)
 
+def get_version():
+    import asyncpg
```
mabdinur, line 47: Do we need to inline this import?

mabdinur (commented):
Left some nits but the overall approach looks good to me. We should also update our "creating integrations" doc to include defining `get_version` method. 

@emmettbutler I think your review would be helpful here

ddtrace/contrib/asyncpg/patch.py
emmettbutler: I'd recommend removing this and making the import on line 29 unconditional.

ddtrace/contrib/elasticsearch/patch.py
```diff
@@ -50,8 +50,16 @@ def _es_modules():
...
+versions = {}
+
+def get_version_tuple(elasticsearch):
+    if getattr(elasticsearch, "__name__", None):
+        versions[elasticsearch.__name__] = getattr(elasticsearch, "__versionstr__", "0.0.0")
+    return getattr(elasticsearch, "__version__", "0.0.0")
+
+
+def get_version():
+    return versions
```
emmettbutler, line 62: I think the fact that this is a dictionary will cause problems for `add_integration`, which expects a string argument.

emmettbutler (commented):
This approach makes sense to me. I second the recommendation to add a bit about `get_version` to [this document](https://github.com/DataDog/dd-trace-py/blob/d995ad39764e545382479f5edc22c32562d88ae6/docs/contributing-integrations.rst).

ddtrace/contrib/aiohttp/patch.py
```diff
@@ -42,6 +43,11 @@
     ),
 )
 
+def get_version():
+    import aiohttp
+
+    return getattr(aiohttp, "__version__", "0.0.0")
```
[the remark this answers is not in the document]
wconti27: I was just always using getattr for extra safety

ddtrace/contrib/elasticsearch/patch.py
wconti27: I updated the monkey patch code to account for this when calling `module.get_version()`. I think it's important that we include the exact package that was patched, in this case elasticsearch2, elasticsearch3 .... and it's corresponding version.

ddtrace/contrib/pytest_bdd/__init__.py
```diff
@@ -32,3 +32,12 @@
         _default_service="pytest_bdd",
     ),
 )
+
+
+def get_version():
```
romainkomorndatadog, line 37: I'd say this should use the same logic as https://github.com/DataDog/dd-trace-py/blame/2a1f4cab08b326a7b97ff4e0515da9db093f3dea/ddtrace/contrib/pytest_bdd/plugin.py#L81-L86 , but it should actually probably use the same code.

ddtrace/_monkey.py
```diff
@@ -163,8 +163,18 @@ def on_import(hook):
                 "tracers", "integration_errors", 1, (("integration_name", module), ("error_type", type(e).__name__))
             )
         else:
+            version = ""
             imported_module.patch()
-            telemetry_writer.add_integration(module, True, PATCH_MODULES.get(module) is True, "")
+            if hasattr(imported_module, "get_version"):
```
mabdinur, line 168: Can we avoid the ` hasattr` check? If we add `get_version` to every integration then this check should always pass. That way we have a consistent integrations api. Integrations that do not have a version (ex: logging) can return an empty string.

ddtrace/_monkey.py
```diff
@@ -163,8 +163,18 @@ def on_import(hook):
...
+                version = imported_module.get_version()
+                if type(version) == dict:
```
mabdinur, line 170: `get_version()` should return a consistent type (either string or dict whatever is the most common). Maybe should should a second optional method `get_versions()` that returns a dict. 

Or we can replace `get_version` with `get_versions` for one consistent api.

mabdinur (commented):
We spoke about this offline but we should update `dd-trace-py/contrib-patch-tests.py` to validate that `get_version()` method is defined in each integration and it returns an expected value.

.circleci/config.templ.yml
```diff
@@ -16,7 +16,7 @@ mongo_image: &mongo_image mongo:3.6@sha256:19c11a8f1064fd2bb713ef1270f79a742a184
 httpbin_image: &httpbin_image kennethreitz/httpbin@sha256:2c7abc4803080c22928265744410173b6fea3b898872c01c5fd0f0f9df4a59fb
 vertica_image: &vertica_image sumitchawla/vertica:latest
 rabbitmq_image: &rabbitmq_image rabbitmq:3.7-alpine
-testagent_image: &testagent_image ghcr.io/datadog/dd-apm-test-agent/ddapm-test-agent:v1.11.0
+testagent_image: &testagent_image williamconti549/dd-apm-test-agent:integration-versions
```
mabdinur, line 19: Is this just a placeholder? We should get a new release of the test agent before merging this PR. Can we call out this dependency in the PR description?

ddtrace/_monkey.py
```diff
@@ -163,8 +163,18 @@ def on_import(hook):
...
-            telemetry_writer.add_integration(module, True, PATCH_MODULES.get(module) is True, "")
+            if imported_module.get_version() != "":
+                version = imported_module.get_version()
+            elif hasattr(imported_module, "get_versions"):
+                version = imported_module.get_versions()
+                for k, v in version.items():
+                    telemetry_writer.add_integration(k, True, PATCH_MODULES.get(module) is True, "", version=v)
+
+            if type(version) == str:
+                telemetry_writer.add_integration(module, True, PATCH_MODULES.get(module) is True, "", version=version)
```
mabdinur, line 176:
```suggestion
```

ddtrace/_monkey.py
```diff
@@ -163,8 +163,18 @@ def on_import(hook):
                 "tracers", "integration_errors", 1, (("integration_name", module), ("error_type", type(e).__name__))
             )
         else:
+            version = ""
             imported_module.patch()
-            telemetry_writer.add_integration(module, True, PATCH_MODULES.get(module) is True, "")
+            if imported_module.get_version() != "":
+                version = imported_module.get_version()
```
mabdinur, line 169:
```suggestion
                version = imported_module.get_version()
                telemetry_writer.add_integration(module, True, PATCH_MODULES.get(module) is True, "", version=version)
```

ddtrace/_monkey.py
```diff
@@ -163,8 +163,18 @@ def on_import(hook):
...
+            elif hasattr(imported_module, "get_versions"):
+                version = imported_module.get_versions()
+                for k, v in version.items():
```
mabdinur, line 172:
```suggestion
                for name, v in versions.items():
```

mabdinur (commented):
Can we update the PR description to call out what is being changed and the value this change provides. 

Aside from my comments this change looks good to me.

.circleci/config.templ.yml
wconti27: this was just testing for a future project and is not necessary for this PR, will revert

ddtrace/_monkey.py
```diff
@@ -163,8 +163,18 @@ def on_import(hook):
                 "tracers", "integration_errors", 1, (("integration_name", module), ("error_type", type(e).__name__))
             )
         else:
+            version = ""
```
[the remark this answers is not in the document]
wconti27: removed

ddtrace/_monkey.py
```diff
@@ -163,8 +163,18 @@ def on_import(hook):
...
             imported_module.patch()
-            telemetry_writer.add_integration(module, True, PATCH_MODULES.get(module) is True, "")
+            if imported_module.get_version() != "":
+                version = imported_module.get_version()
+            elif hasattr(imported_module, "get_versions"):
```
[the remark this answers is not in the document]
wconti27: changed to do this

ddtrace/contrib/kombu/patch.py
```diff
@@ -30,6 +30,10 @@
 from .utils import get_routing_key_from_args
 
 
+def get_version():
+    return str(getattr(kombu, "__version__", ""))
```
mabdinur, line 34: Do we need the str? kombu.__version__ should always be a string: https://github.com/celery/kombu/blob/v1.0.1/kombu/__init__.py#L3

tests/contrib/futures/test_futures_patch_generated.py
```diff
@@ -27,3 +26,6 @@ def assert_not_module_patched(self, concurrent_futures):
 
     def assert_not_module_double_patched(self, concurrent_futures):
         pass
+
+    def assert_module_implements_get_version(self):
```
mabdinur, line 30: This method is unimplemented. Is this intentional?

ddtrace/contrib/aiohttp_jinja2/patch.py
```diff
@@ -15,6 +17,10 @@
 )
 
 
+def get_version():
+    return getattr(aiohttp_jinja2, "__version__", "")
```
mabdinur, line 21: nit: we can probably add typing to all the get version methods. This isn't blocking

```suggestion
    # type: () -> str
    return getattr(aiohttp_jinja2, "__version__", "")
```

mabdinur (approved):
Thanks for sticking with it. I left some non blocking comments. Overall LGTM

ddtrace/contrib/kombu/patch.py
wconti27: kombu returns unicode for Python2

ddtrace/contrib/aiohttp_jinja2/patch.py
wconti27: Done.

tests/contrib/futures/test_futures_patch_generated.py
wconti27: Yes, since futures has no version. Otherwise the tests default to the base class case which is testing for get_version and results in a fail

mabdinur (dismissed):

mabdinur (approved):

PR #6693, opened 2023-08-17
fix: make sampling tags conform to expectations

emmettbutler:
This change makes the tracer's sampling tagging behavior conform to the expectations described in https://github.com/DataDog/system-tests/pull/1487

## How to review

Use [this view](https://github.com/DataDog/dd-trace-py/pull/6693/files?file-filters%5B%5D=.py&show-viewed-files=true) to look at the changes to Python files, since it's a lot more responsive in the browser than the default.

1. Read these [system-tests](https://github.com/DataDog/system-tests/blob/8b0bdc3392792fa45fb97845f777688abb052a85/tests/parametric/test_sampling_span_tags.py) to understand the behavior they validate.
2. Read `tests/integration/test_sampling.py` to see the library tests that perform the same validations.
3. Look at `snapshots/test_sampling_with*` to see how test expectations have changed in light of these validations.
4. Examine `ddtrace/sampler.py` to see how it implements this behavior, especially the function call `from_datadog_sampler`.
4. Glance at some of the hundreds of other snapshots to confirm that their changes make sense in light of what you've learned.

## Change Summary

Filter the file list by `snapshots/test_sampling_with` to find the snapshot files that directly validate the behavior tested by [these system-tests](https://github.com/DataDog/system-tests/blob/8b0bdc3392792fa45fb97845f777688abb052a85/tests/parametric/test_sampling_span_tags.py#L74). Changes to these snapshot files include:

* Removing the `.dm` tag from spans not first in their chunks
* Adding the `.dm` tag to first-in-chunk spans that had their sampling decision set manually (`-4`) or by the tracer (`-3`)
* Removing the `_psr` tags from spans to which it doesn't apply
* Adding three test cases that involve the `DD_TRACE_RATE_LIMIT` configuration

My intent is that each of these snapshot file changes maps directly on to one of the "reasons" listed for the test skips in the [system-tests](https://github.com/DataDog/system-tests/blob/8b0bdc3392792fa45fb97845f777688abb052a85/tests/parametric/test_sampling_span_tags.py#L74). Find the corresponding test code for each snapshot in `tests/integration/test_sampling.py`.

`ddtrace/tracer.py` is changed to avoid propagating the `.dm` tag to child spans that exist in the same execution as their parents.

`ddtrace/span.py` is changed to remove `_psr` tags from spans to which they don't apply. For example, `_dd.p.dm: -1` indicates that the sampling decision was made on the basis of a sample rate from the agent, which is the only case in which `agent_psr` is meaningful. When `dm != -1`, `agent_psr` should not be set.

There are significant changes to `ddtrace/sampler.py`. The most fundamental of these is the [change](https://github.com/DataDog/dd-trace-py/pull/6693/files?file-filters%5B%5D=.py&show-viewed-files=true#diff-51bda2f09879ed6e54b15cc9da5ed1da04023d3465018dd7794d35384dd85409L357-R356) from the use of inheritance between sampler classes to the use of dependency injection via the new `from_datadog_sampler` method. This fixes a subtle bug in which executions of `sample()` were calling methods of subclasses rather than the intended superclass methods. Avoiding the use of `super()` sidesteps the situation that led to this issue and makes it straightforward to adjust the logic that chooses between the various sampler implementations. This also allows the removal of some hard-to-follow `isinstance` and `super` calls.

In `sampler.py`, rate limiting is applied unconditionally as a second stage after a sampling decision has been made.

In `sampler.py`, the `agent_psr` tag is only applied when a sample rate received from the agent was actually used to make the sampling decision.

All changes other than the ones listed here are changes to test expectations to make existing tests conform to the new behavior. A majority of these expectation changes were made by a script that

* Removed the `.dm` tag from any span with `parent_id != 0`
* Removed the `agent_psr` tag from any span with `_dd.p.dm != -1`

## Risk

Because these changes conform to the behavior in the system-tests, I think their risk is pretty low. The worst-case scenario is that this change causes inaccurate metrics in Datadog products, in which case it can be reverted.

The performance impact indicated by the bot comment is [probably](https://github.com/DataDog/dd-trace-py/pull/6693#pullrequestreview-1603313611) a result of the removal of an early exit from `sample()`. In my view it's worthwhile to trade CPU time for code clarity in this case.

## Other Stuff

Once this is included in a release, we can promote https://github.com/DataDog/system-tests/pull/1535 to "ready for review". I've run those tests locally and told them to install the tracer from this branch.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/internal/utils/__init__.py
```diff
@@ -75,6 +75,6 @@ def _get_metas_to_propagate(context):
     # type: (Any) -> List[Tuple[str, str]]
     metas_to_propagate = []
     for k, v in context._meta.items():
-        if isinstance(k, six.string_types) and k.startswith("_dd.p."):
+        if isinstance(k, six.string_types) and k.startswith("_dd.p"):
```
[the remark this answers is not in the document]
emmettbutler: Nothing, it was unintentional

ddtrace/sampler.py
```diff
@@ -348,25 +343,30 @@ def sample(self, span):
...
             if rule.matches(span):
                 sampler = rule
+                rule_matched = True
                 break
-        else:
-            # No rules match so use agent based sampling
-            return super(DatadogSampler, self).sample(span)
+
+        if not rule_matched:
+            sampler = RateByServiceSampler.from_datadog_sampler(self)
```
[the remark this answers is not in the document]
ZStriker19: Following up on this, just to clarify, isn't there overlap between `RateByServiceSampler` and DatadogSampler's `SamplingRule`'s? 

`RateByServiceSampler` has a dict of  service name + env keys and `RateSampler` values, that it uses to make a sampling decision for a span.

Whereas `rules` (which `DatadogSampler` tries to use first) is a list of `SamplingRule`s which care about service and name. 
 
This is just due to legacy that we essentially have two ways to configure sampling based off of a service right? Rules (aka `DD_TRACE_SAMPLING_RULES`) would be the recommended way to do this, however we have the legacy capability allowing users to create their own samplers and pass them in as they want, which we have to maintain, therefore we need to keep the `RateByServiceSampler` in here like this.

ddtrace/span.py
```diff
@@ -277,6 +280,13 @@ def _finish_ns(self, finish_time_ns):
         for cb in self._on_finish_callbacks:
             cb(self)
 
+    def _override_sampling_decision(self, decision):
+        self.context.sampling_priority = decision
+        set_sampling_decision_maker(self.context, SamplingMechanism.MANUAL)
+        for key in (SAMPLING_RULE_DECISION, SAMPLING_AGENT_DECISION, SAMPLING_LIMIT_DECISION):
```
ZStriker19, line 286: we make sure to delete all of the tags that shouldn't be on the span if we make a manual keep or drop decision.

riotfile.py
```diff
@@ -1837,7 +1837,7 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
         ),
         Venv(
             name="rq",
-            command="pytest tests/contrib/rq",
+            command="pytest {cmdargs} tests/contrib/rq",
```
ZStriker19, line 1840: Why'd we need to add this?

ZStriker19: I think I figured it out in https://github.com/DataDog/dd-trace-py/pull/6772#issuecomment-1696498213

ddtrace/sampler.py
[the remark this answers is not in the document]
emmettbutler: Yes, I believe you've got it right. If we didn't need to maintain backward compatibility we'd probably consolidate those two classes into one.

ZStriker19 (approved):

brettlangdon:
Is the increase in performance overhead expected?

ddtrace/sampler.py
```diff
@@ -348,25 +343,30 @@ def sample(self, span):
...
-            allowed = self.limiter.is_allowed(span.start_ns)
-            if not allowed:
-                self._set_sampler_decision(span, self.limiter, allowed)
-                return False
+        if rule_matched or has_configured_rate_limit:
+            self._set_sampler_decision(span, sampler, sampled)
 
-        return sampled
+        # Ensure all allowed traces adhere to the global rate limit
+        allowed = self.limiter.is_allowed(span.start_ns)
```
brettlangdon, line 364: I think we can keep this behind the `if sampled:` gate.

The limiter is meant to limit the number of sampled traces, non-sampled traces should not be taken into account.

e.g. I want a 50% sample rate, but please don't go over 20 traces/sec sampled.

iiuc these changes change that behavior?

emmettbutler:
@brettlangdon no, it's not. I just pushed a commit that removes a call to `dict.copy()` that seems slow and unnecessary.

ddtrace/sampler.py
brettlangdon: I might be wrong here... seems like this is the PR which made that change

https://github.com/DataDog/dd-trace-py/commit/8c73f10f8d831ec578e0eca36bf067df1ea12f7c

I cannot figure out the motivation from the PR description, cc @majorgreys do you remember?

ddtrace/span.py
```diff
@@ -277,6 +280,13 @@ def _finish_ns(self, finish_time_ns):
...
+            if key in self._local_root._metrics:
+                del self._local_root._metrics[key]
```
ZStriker19, line 288: Regarding performance, could it be that we're basically always deleting 3 tags from every span but the parent span? Is that expensive enough that it could cause the performance issue?

ddtrace/sampler.py
```diff
@@ -348,25 +343,32 @@ def sample(self, span):
...
+
+        rule_matched = False
         for rule in self.rules:
             if rule.matches(span):
                 sampler = rule
+                rule_matched = True
                 break
-        else:
-            # No rules match so use agent based sampling
-            return super(DatadogSampler, self).sample(span)
```
emmettbutler, line 357: The removal of this early return is the only part of this diff that seems remotely likely to degrade performance, as far as I can tell.

emmettbutler:
@brettlangdon - @ZStriker19 and I both added some ideas about performance in the comments.

https://github.com/DataDog/dd-trace-py/pull/6693#pullrequestreview-1603313611
https://github.com/DataDog/dd-trace-py/pull/6693#pullrequestreview-1603312296

ddtrace/sampler.py
brettlangdon: It would make sense why this change would have an impact on the span start/finish micro benchmarks because this code runs on root spans when sampling.

I wonder if it is worth having a start/finish test that doesn't trigger a sampling decision, and/or have a test dedicated to sampling?

It does make me a bit worried about the `scenario:flasksimple-appsec-telemetry` since that one should normalize the expensive cost of sampling / give a more "realistic" scenario... but also why does only that one show an increase?

🤔

emmettbutler: The latest changes reduce the performance impact to what I think are acceptable levels. What do you think? https://github.com/DataDog/dd-trace-py/pull/6693#issuecomment-1683029594

brettlangdon: nice, seems like higher memory usage, but that is fine, as long as we don't impact execution time, should be ok.

emmettbutler:
@ZStriker19 @brettlangdon let me know if there are any other changes you'd like to see to this pull request.

mabdinur (approved):

PR #6712, opened 2023-08-21
chore(telemetry): ensure metrics are aggegated every 10 secs but sent after 60 secs

mabdinur:
## Description
- Decouples the aggregation of metrics in the telemetry writer from when telemetry events are sent to the agent
   - The usage of `DD_TELEMETRY_HEARTBEAT_INTERVAL` remains the same. This configuration determines the interval that telemetry payloads are sent to the agent.
  - `TelemetryWriter.interval` is no longer equal to `DD_TELEMETRY_HEARTBEAT_INTERVAL` in all scenarios. The periodic interval of the telemetry writer is now set to a value between 0 and 10 seconds. This will give us a better resolution for telemetry metrics. 

- Example: heartbeat interval is 120seconds and TelemetryWriter.interval is 10seconds. 
        - In a given 120 seconds window `TelemetryWriter.periodic()` will be called 12 times. 
        - For each `TelemetryWriter.periodic()` call, metrics and logs will be aggregated into a telemetry payload. 
        - On the 12th (and last) call, all telemetry events will be batched and sent to the agent.

## Motivation

Align the behavior of the python telemetry client with the .NET tracer. This behavior will also be implemented in all other tracers.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/internal/telemetry/writer.py
```diff
@@ -196,10 +196,10 @@ def enable(self, start_worker_thread=True):
         if self.status == ServiceStatus.RUNNING:
             return True
 
-        if start_worker_thread:
+        if self._is_periodic:
             self.start()
             atexit.register(self.app_shutdown)
```
mabdinur, line 201:
```suggestion
            atexit.register(self.app_shutdown)
            return True
```

tests/telemetry/test_telemetry.py
```diff
@@ -135,21 +135,35 @@ def test_enable_fork_heartbeat(test_agent_session, run_python_code_in_subprocess
...
     assert stderr == b""
 
 
+def test_heartbeat_interval_invalid_configuration(run_python_code_in_subprocess):
+    """assert that DD_TELEMETRY_HEARTBEAT_INTERVAL config sets the telemetry writer interval"""
+    heartbeat_interval = "9"
+    env = os.environ.copy()
+    env["DD_TELEMETRY_HEARTBEAT_INTERVAL"] = heartbeat_interval
+
+    _, stderr, status, _ = run_python_code_in_subprocess("from ddtrace import config", env=env)
```
mabdinur, line 158:
```suggestion
    _, stderr, status, _ = run_python_code_in_subprocess("from ddtrace import config; assert config._ _telemetry_heartbeat_interval == 10", env=env)
```

tests/telemetry/test_telemetry.py
```diff
@@ -135,21 +135,35 @@ def test_enable_fork_heartbeat(test_agent_session, run_python_code_in_subprocess
...
     assert stderr == b""
 
 
+def test_heartbeat_interval_invalid_configuration(run_python_code_in_subprocess):
+    """assert that DD_TELEMETRY_HEARTBEAT_INTERVAL config sets the telemetry writer interval"""
+    heartbeat_interval = "9"
+    env = os.environ.copy()
+    env["DD_TELEMETRY_HEARTBEAT_INTERVAL"] = heartbeat_interval
+
+    _, stderr, status, _ = run_python_code_in_subprocess("from ddtrace import config; assert config._telemetry_heartbeat_interval == 10", env=env)
```
mabdinur, line 158:
```suggestion
    _, stderr, status, _ = run_python_code_in_subprocess(
        "from ddtrace import config;assert config._telemetry_heartbeat_interval == 10", 
        env=env
    )
```

emmettbutler (dismissed):

emmettbutler (approved):

tests/telemetry/app.py
```diff
@@ -20,3 +20,9 @@ def metrics_view():
         1.0,
     )
     return "OK", 200
+
+
+@app.route("/flush_metrics")
+def flush_metrics():
+    telemetry_writer.periodic(force_flush=True)
+    return "OK", 200
```
mabdinur, line 28:
```suggestion
```

tests/telemetry/test_telemetry_metrics_e2e.py
```diff
@@ -29,7 +28,6 @@ def gunicorn_server(telemetry_metrics_enabled="true", token=None):
     cmd = ["ddtrace-run", "gunicorn", "-w", "1", "-b", "0.0.0.0:8000", "tests.telemetry.app:app"]
     env = _build_env()
     env["DD_TELEMETRY_METRICS_ENABLED"] = telemetry_metrics_enabled
```
mabdinur, line 30:
```suggestion
    env["DD_TELEMETRY_METRICS_ENABLED"] = telemetry_metrics_enabled
    env["DD_TELEMETRY_HEARTBEAT_INTERVAL"] = "1.0"
```

tests/telemetry/test_telemetry_metrics_e2e.py
```diff
@@ -90,11 +88,12 @@ def test_telemetry_metrics_enabled_on_gunicorn_child_process(test_agent_session)
         gunicorn_client.get("/count_metric")
         response = gunicorn_client.get("/count_metric")
         assert response.status_code == 200
-        # DD_TELEMETRY_HEARTBEAT_INTERVAL is set to 1 second
-        time.sleep(1)
+        gunicorn_client.get("/flush_metrics")
+
```
mabdinur, line 92:
```suggestion
        # DD_TELEMETRY_HEARTBEAT_INTERVAL is set to 1 second
        time.sleep(1)
```

tests/telemetry/test_telemetry_metrics_e2e.py
```diff
@@ -90,11 +88,12 @@ def test_telemetry_metrics_enabled_on_gunicorn_child_process(test_agent_session)
...
         gunicorn_client.get("/count_metric")
         response = gunicorn_client.get("/count_metric")
         assert response.status_code == 200
+        gunicorn_client.get("/flush_metrics")
```
mabdinur, line 96:
```suggestion
```

ddtrace/internal/telemetry/writer.py
```diff
@@ -212,11 +214,10 @@ def disable(self):
         """
         self._disabled = True
         self.reset_queues()
-
-        if self.is_periodic:
+        if self._is_periodic and self.status is ServiceStatus.RUNNING:
             atexit.unregister(self.stop)
             self.stop()
-        else:
+       else:
```
mabdinur, line 220:
```suggestion
        else:
```

tests/telemetry/test_telemetry_metrics_e2e.py
```diff
@@ -3,7 +3,6 @@
 import os
 import subprocess
 import sys
```
mabdinur, line 5:
```suggestion
import sys
import time
```

tabgok (approved):

ddtrace/internal/telemetry/writer.py
```diff
@@ -552,6 +536,20 @@ def periodic(self):
         if logs_metrics:
             self._generate_logs_event(logs_metrics)
 
+        if self._is_periodic and force_flush is False:
```
brettlangdon, line 539: the above should happen/be flushed every 10 seconds, but the below only once every 60 seconds?

a comment here is going to be really important. I can easily see someone putting flushing of new data in the wrong location and something not working correctly.

also, what happens if the heartbeat interval is not divisible by 10, is that ok?

maybe the stuff that happens less frequently should be moved to it's own function fro clarity?

mabdinur: > the above should happen/be flushed every 10 seconds, but the below only once every 60 seconds?
Yup. We aligned on this spec across all tracer teams a few weeks ago 


> what happens if the heartbeat interval is not divisible by 10, is that ok?

We will round up. Ex: If the heartbeat is set to 12 seconds, telemetry events will be sent every 20. I think this is fine. The heartbeat interval does not have to exact. The key thing here is that metrics are aggregated into buckets with a max size of 10 seconds.

To be honest I don't see a valid case for including `DD_TELEMETRY_HEARTBEAT` in our public api. This is not a user facing feature and this configuration is only used in system-tests. We should honestly make it private but this would involve updating system tests and changing this environment variable across all tracers. It may not be worth the effort. 

`I will add better documentation`

mabdinur, line 539:
```suggestion
        # Telemetry metrics and logs should be aggregated in to payloads every time periodic is called.
        # This ensures metrics and logs are submitted in 0 to 10 second buckets.
        # All other events should be aggregated using `config._telemetry_heartbeat_interval`.
        if self._is_periodic and force_flush is False:
```

brettlangdon (approved):

PR #6731, opened 2023-08-23
fix(profiling): rationalize ddup string lifetimes and span_id conversion

sanchda:
* upgrade libdatadog to 3.0.0
* fix bug where span_ids were interpreted as signed integer
* fix bug where span attributes are less reliable than I thought
* fix bug with string lifetimes in ddup due to realloc

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/internal/datadog/profiling/include/exporter.hpp
```diff
@@ -130,7 +131,7 @@ class Profile {
   // strategy to dedup + store strings
   using StringTable = std::unordered_set<std::string_view>;
```
r1viollet, line 132: why not use directly the unordered_set as storage ? References to values are guaranteed to be stable.

ddtrace/internal/datadog/profiling/src/exporter.cpp
```diff
@@ -176,12 +176,13 @@ Uploader::set_runtime_id(std::string_view id)
 }
 
 bool
-Uploader::upload(const Profile* profile)
+Uploader::upload(Profile* profile)
 {
     ddog_prof_Profile_SerializeResult result = ddog_prof_Profile_serialize(profile->ddog_profile, nullptr, nullptr);
+    profile->reset();
```
r1viollet, line 182: I don't expect an upload function to reset the profile.

setup.py
```diff
@@ -52,7 +52,7 @@
     HERE, os.path.join("ddtrace", "internal", "datadog", "profiling", "libdatadog")
 )
 
-LIBDATADOG_PROF_VERSION = "v2.1.0"
+LIBDATADOG_PROF_VERSION = "v3.0.0"
```
r1viollet, line 55: Thanks for the libdatadog upgrades

ddtrace/internal/datadog/profiling/include/exporter.hpp
```diff
@@ -7,6 +7,7 @@
 
 #include <array>
 #include <chrono>
+#include <deque>
 #include <memory>
```
r1viollet, line 11: Minor: is the vector include still needed?

ddtrace/internal/datadog/profiling/src/exporter.cpp
r1viollet: I did not check if there was a better place to reset the profile.

ddtrace/internal/datadog/profiling/include/exporter.hpp
sanchda: Yes

ddtrace/internal/datadog/profiling/src/exporter.cpp
P403n1x87: Do we expect to be able to flush the data to multiple sinks? E.g. a file _and_ the network? If not I'd be OK with a reset here.

ddtrace/internal/datadog/profiling/src/interface.cpp
```diff
@@ -293,7 +293,8 @@ ddup_upload()
         // Rationalize return for interface
         std::cout << "WHOA NOT INITIALIZED" << std::endl;
     }
-    new std::thread(ddup_upload_impl, g_profile); // set it and forget it
+    std::thread(ddup_upload_impl, g_profile).detach(); // set it and forget it
```
P403n1x87, line 296: Is there a way for us to `join` the thread instead? Or would that not be any better than `detach`ing?

ddtrace/internal/datadog/profiling/src/exporter.cpp
```diff
@@ -620,19 +614,21 @@ Profile::push_task_name(std::string_view task_name)
...
     }
     return true;
 }
 
 bool
-Profile::push_local_root_span_id(int64_t local_root_span_id)
+Profile::push_local_root_span_id(uint64_t local_root_span_id)
 {
-    if (!push_label(ExportLabelKey::local_root_span_id, local_root_span_id)) {
+    int64_t recoded_id = reinterpret_cast<int64_t&>(local_root_span_id);
```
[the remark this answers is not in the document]
sanchda: Yeah, this is a type-punning procedure.  It'll change the type understood by the compiler without modifying any of the bits of the value.

ddtrace/internal/datadog/profiling/src/interface.cpp
sanchda: Ah, that's a good question.  I guess there are scenarios where the average thread lifetime might exceed the upload interval (e.g., if the user specifies a really short upload), but in those scenarios a `join()` could massively slow down the profiler.  At the same time, forceful termination (e.g., `kill`) could work, but I think it bypasses normal thread destruction.  I _think_ that's OK, but I need to see whether it introduces a leak.

Will check it out and try to tighten this up.

ddtrace/internal/datadog/profiling/src/exporter.cpp
sanchda: I need to reason about this a bit more.  There's something I don't understand, since I have a reset in the higher-level interface already (but it's in the wrapper for `upload()` called by the client).

@P403n1x87: to your question, I do plan on supporting serialize-to-file just like the current profiler does, and I do anticipate that setting being additive with pushing through the network.

The general comment came up in an earlier review.  I think the main concern is that the semantic meaning of the word "upload" encompasses only a step of what we're doing here.  Will address with a refactor.

ddtrace/internal/datadog/profiling/src/interface.cpp
P403n1x87: > I guess there are scenarios where the average thread lifetime might exceed the upload interval (e.g., if the user specifies a really short upload), but in those scenarios a `join()` could massively slow down the profiler.

True that. I think currently we do join the threads on shutdown, but skip it on fork (because the thread would be dead anyway). This of course might be problematic if the child process is short-lived and has produced data that needs to be flushed out. The `join` in this case would add a potentially sizeable delay. However, joining on shutdown in this case would not be any different from what we currently do.

ddtrace/internal/datadog/profiling/include/exporter.hpp
sanchda: I wrote a comment, but it looks like GitHub ate it up.  Writing succinctly.

The version of gcc we use in CI doesn't support heterogeneous access for unordered_set.  That means I'd need to convert the incoming `string_view` into a `string` in order to do any find-type comparison.  I was considering this to be a deal-breaker, since I expect the overall string duplication rate to be high.

Updating gcc is easy, but it isn't an option because some of the other extensions fail their builds on newer-GCC and I don't want to investigate all of that.

Arguably, I could use a different backing store.  :shrug:

ddtrace/internal/datadog/profiling/src/interface.cpp
sanchda: One of the interesting things here is that all of the storage is held outside of Python, so if all of the Python threads shut down, then as long as the process group doesn't get forcibly closed (does cpython kill itself after it's joined all of its threads?) these native threads will hang out until they're done uploading.  They won't have to deal with module cleanup.

So, I'd say we need to join on shutdown only if cpython forces us to (because it'll kill our threads otherwise).  I'd need to check that out--do you know?

ddtrace/internal/datadog/profiling/src/exporter.cpp
r1viollet: Some reasons behind my comment: 
- The reset is done just before we start writing to the profile (to ensure the duration of the profile is accurate). 
- This runs in an exporter thread where we would not expect the profile to be updated.

As mentioned by @sanchda there seems to be a reset in the right place, so we will look at this more thoroughly.

majorgreys (commented):
otherwise lgtm

majorgreys (approved):

PR #6735, opened 2023-08-23
fix(datastreams): support datastream propagation with raw message delivery enabled

tabgok:
When a subscription is set up between SNS and a consumer (such as SQS),
there is an option to enable "Raw Message Delivery" which changes the
format of delivered data, primarily the MessageAttributes and MessageBody
sections.

Prior to this change, datastream context was dropped/ignored on the
receive because the code assumed the format of delivered messages when
Raw Delivery was disabled.

After this change, the patch handles both raw message delivery and
non-raw delivery.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


emmettbutler (approved):

emmettbutler (changes requested):
Looks great! Since this is a "fix", please add a release note.

tabgok:
> Looks great! Since this is a "fix", please add a release note.

Done!

ddtrace/contrib/botocore/patch.py
```diff
@@ -94,6 +94,14 @@ class TraceInjectionDecodingError(Exception):
     pass
 
 
+def _encode_data(trace_data):
```
ZStriker19, line 97: this is smart. I never think about adding methods solely to make testing easier/possible.

ddtrace/contrib/botocore/patch.py
```diff
@@ -640,6 +648,18 @@ def patched_api_call(original_func, instance, args, kwargs):
...
+                                and "_datadog" in message["MessageAttributes"]
+                                and "BinaryValue" in message["MessageAttributes"]["_datadog"]
+                            ):
+                                # Raw message delivery
+                                message_body = message
+                                context_json = json.loads(
+                                    message_body["MessageAttributes"]["_datadog"]["BinaryValue"].decode()
+                                )
+                            else:
+                                log.debug("DataStreams did not handle message: %r", message)
```
ZStriker19, line 662: Should we make this a bit more specific, like explain that it wasn't handled because it's in a format we don't support? Or is general the best we can do since we don't know why the message wasn't handled?

ZStriker19 (commented):
Just a question about log message but otherwise looks great!

emmettbutler (approved):

ddtrace/contrib/botocore/patch.py
tabgok: A hard-learned lesson =p

ddtrace/contrib/botocore/patch.py
tabgok: The latter - so far its been an exploration into AWS behavior.  The other cases are the known issues, this catches the unknown unknowns for the purposes of gathering info from customers.

ZStriker19 (approved):

ZStriker19 (dismissed):

emmettbutler (approved):

PR #6798, opened 2023-08-30
fix(profiling): ensure correct thread link

P403n1x87:
We use the C API to retrieve the thread ID of the current thread to create a link between threads and objects. This way we protect ourselves from frameworks such as gevent that might turn thread IDs into task IDs.

## Testing strategy

Tested with a sample application that the endpoint information is attached to profiles with this change.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/profiling/_threading.pyx
```diff
@@ -10,9 +10,20 @@ from six.moves import _thread
...
-IF UNAME_SYSNAME == "Linux":
-    from cpython cimport PyLong_FromLong
+from cpython cimport PyLong_FromLong
+
+
+cdef extern from "<Python.h>":
+    # This one is provided as an opaque struct from Cython's
+    # cpython/pystate.pxd, but we need to access some of its fields so we
+    # redefine it here.
+    ctypedef struct PyThreadState:
```
sanchda, line 20: I'm confused about this.  `ctypedef` is defining the C layout for a struct named PyThreadState as a struct with a single `unsigned long` field.

On the other hand, cpython's internal definition for PyThreadState is a typedef to the `_ts` [struct in cpython](https://github.com/python/cpython/blob/main/Include/cpython/pystate.h#L64).

If one were to use this struct to bring a PyThreadState object into cpython, then on Linux this redefinition will pun the `PyThreadState *next` to a 64-bit integer (and on Windows, it'll take the bottom 32 bits of the pointer because a long is 32-bits there) called `thread_id`.  This will be a moderately unique value, but it isn't the same as PyThreadState's `thread_id`.

ddtrace/profiling/_threading.pyx
```diff
@@ -108,28 +119,22 @@ class _ThreadLink(_thread_link_base):
     ):
         # type: (...) -> None
         """Link an object to the current running thread."""
-        self._thread_id_to_object[_thread.get_ident()] = weakref.ref(obj)
+        # Because threads might become tasks with some frameworks (e.g. gevent),
+        # we retrieve the thread ID using the C API instead of the Python API.
+        self._thread_id_to_object[<object>PyLong_FromLong(PyThreadState_Get().thread_id)] = weakref.ref(obj)
```
sanchda, line 124: If my previous comment was correct, then I think the keys to `_thread_id_to_object` are actually pointers.  This would be easy to verify by printing the numerical value of `thread_id`.  If it's a pointer, it'll be a moderately large value.  If it's a true thread_id, then it should be small (especially if the test is run in a container running on a PID namespace).

sanchda: I don't know enough about the dynamical operation of Python interpreter to figure out whether my observation matters.  I think it does because the interpreter keeps these threads as a linked list, so as long as threads have differently-scaled lifetimes then `next` for a given thread will change much more rapidly than its own ID.  I can see how this might not matter for some tests though.

sanchda (approved):
I have one concern, but if you feel it isn't an issue, everything else looks good to me.

ddtrace/profiling/_threading.pyx
P403n1x87: 🤔 I'm not sure I follow here. I believe this relates to the previous comment. If so, here we are really getting the `thread_id`, not just some random bytes from a data structure (manual tests with a sample application confirm that indeed we are now resolving span objects when querying by thread ID). So this change should now implement the intended behaviour. Please let me know if I misinterpreted your question! 🙏

majorgreys (approved):

PR #6808, opened 2023-08-31
fix(ci-visibility): use unicode strings and values for spans payload so msgpack is correct

romainkomorndatadog:
msgpack in Python 2.7 requires strings to be unicode in order to produce the correct msgpack format expected by our backend.

Another potential fix for this would've been to `import msgpack` and use `msgpack.packb(..., use_bin_type=False)`, but that option was discarded since this repo provides its own `msgpack.packb` implementaton.

The tests only use single-key dictionaries because key ordering is not stable in Python 2.7 and may lead to flaky tests.

Since the `bytes` and `str` type are synonyms in Python 2.7, and are the only values that need to be coerced to unicode, there should not be increased risk from `_py2_payload_force_unicode_strings`.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


romainkomorndatadog:
I considered some other options:
- updating our `packb` implementation to take some flag that automatically coerces strings to unicode
- converting to JSON and back first (which might give us unicode for all strings, but I haven't checked), but that seems even more costly than what I'm doing here
- using the upstream `msgpack` module (as mentioned in the description), but I assume we introduced our own for performance reasons
- updating all the places we set keys in our structs to use `ensure_text()` which is probably what we should've done from the start, but at this point it seems like a lot of heavy lifting for a runtime we don't intend to support all that actively

ericlaz (approved):
Looks good to me overall.

emmettbutler (approved):

romainkomorndatadog:
@Kyle-Verhoog , since you authored a good bit of the `msgpack.packb` implementation (or so `git blame` says :) ), I feel like it'd be good to get your approval here.

nikita-tkachenko-datadog (approved):

PR #6813, opened 2023-08-31
chore(kafka): migrate kafka datastreams to Core API

tabgok:
## Summary
This change adjusts the data streams monitoring (DSM) code in the
Kafka library to be called via Core API events rather than
checkpoints.
    
This change allows the code to separate DSM logic into DSM specific folders
and files.
    
The key supporting changes are:
1. Creating a singleton in the DSM module
2. Creating events "kafka.produce.start" and "kafka.consume.start"
3. Moving the DSM logic out of the shared contrib and into a
   product-specific page

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/contrib/kafka/patch.py
```diff
@@ -131,10 +133,7 @@ def traced_produce(func, instance, args, kwargs):
     partition = kwargs.get("partition", -1)
     if config._data_streams_enabled:
         # inject data streams context
-        headers = kwargs.get("headers", {})
-        pathway = pin.tracer.data_streams_processor.set_checkpoint(["direction:out", "topic:" + topic, "type:kafka"])
-        headers[PROPAGATION_KEY] = pathway.encode()
-        kwargs["headers"] = headers
+        core.dispatch("kafka.produce.start", [instance, args, kwargs])
```
tabgok, line 136: do we have a pattern for what is placed in the dispatch vs what is attached to the context (via `core.set_item()`) ?

ddtrace/internal/datastreams/__init__.py
```diff
@@ -0,0 +1,11 @@
+from . import kafka
+
+_processor = None
```
tabgok, line 3: switching to a singleton while still supporting per-tracer processors.

ddtrace/internal/datastreams/kafka.py
```diff
@@ -0,0 +1,28 @@
+from ddtrace import config
+from ddtrace.internal import core
+
+from ddtrace.internal.datastreams.processor import PROPAGATION_KEY
+
+
+def dsm_kafka_message_produce(instance, args, kwargs):
+    from . import data_streams_processor as processor
```
tabgok, line 8: maybe there's a more elegant way to support singletons?

ddtrace/internal/datastreams/kafka.py
```diff
@@ -0,0 +1,28 @@
...
+    headers = {header[0]: header[1] for header in (message.headers() or [])}
+    topic = core.get_item("kafka_topic")
+    group = instance._group_id
+
+    ctx = processor().decode_pathway(headers.get(PROPAGATION_KEY, None))
+    ctx.set_checkpoint(["direction:in", "group:"+group, "topic:" + topic, "type:kafka"])
+
+
+if config._data_streams_enabled:
+    core.on("kafka.produce.start", dsm_kafka_message_produce)
```
tabgok, line 27: I feel like the dispatch names should be standardized. thoughts?

tests/contrib/kafka/test_kafka.py
```diff
@@ -69,6 +71,12 @@ def tracer():
         t.shutdown()
         unpatch()
 
+@pytest.fixture
+def dsm_processor(tracer):
+    processor = tracer.data_streams_processor
+    with mock.patch("ddtrace.internal.datastreams.data_streams_processor", return_value=processor):
```
tabgok, line 77: override the singletone

ddtrace/contrib/kafka/patch.py
tabgok: As an extension:

I'm not sure where the differentiation is between core and product for extracting information.  I'm currently on the side of putting the extraction code into the product, which means passing/attaching the already-available information and then deriving values in the product.  However, this could mean each product does its own derivation.  

In the case of "kafka_topic" I'd pass "message" to the product code and evaluate `message.product` there.

emmettbutler: > do we have a pattern for what is placed in the dispatch vs what is attached to the context

No. Hopefully one will arise as we transition more integrations to the core API.

> I'm not sure where the differentiation is between core and product for extracting information.

In my opinion, parsing the message once in the integration code (this file) and passing kafka_topic to `dispatch` is the simpler separation of concerns. Aside from not forcing the Tracing product to know how to parse a Kafka message, it also means that the possibly-expensive parsing operation only happens once rather than per-product.

ddtrace/internal/datastreams/kafka.py
emmettbutler: I agree. I don't think we can do it up front, but as we make more `on` calls I'm on the lookout for standardizations that could make sense.

emmettbutler (commented):
Makes sense to me so far. All of the uses of the Core API are aligned with what I had in mind when I set it up.

emmettbutler (dismissed):

jbertran (dismissed):

emmettbutler (approved):

PR #6853, opened 2023-09-07
chore(telemetry): report tested integration versions

wconti27:
## Motivation

With this PR we want to add a method to the generated patch for all contribs that will emit a telemetry payload describing the tested integration and its tested version. The goal of this PR is to get a list of all integrations tested during CI, along with the range of versions tested, which will be reported to metabase. The reasoning for adding another testcase instead of simply using information available in metabase from customer telemetry on what integrations are used is as follows:
 - we want the ability to know if an integration @ a specific version is being supported and actively tested, but not in use by any customers; using customer telemetry would therefore not cover this case and not allow us to deprecate support for unused integrations. 

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


mabdinur (commented):
Although this approach currently works it uses the instrumentation telemetry client in an unsupported way. To support python3.12 we recently changed when/how telemtry events are sent. This change will prevent integrations will not be sent in test runs: https://github.com/DataDog/dd-trace-py/pull/6859. We can hack around this issue but I don't think this is the best approach. 


Can we get the tested integration versions by parsing the venvs in https://github.com/DataDog/dd-trace-py/blob/1.x/riotfile.py? This way we can collect and submit the integration data to the test agent once. We can also make this nightly job on the 1.x branch so this data is not collected on every ci run.

docker-compose.yml
```diff
@@ -129,13 +129,13 @@ services:
...
     testagent:
-        image: ghcr.io/datadog/dd-apm-test-agent/ddapm-test-agent:v1.11.0
+        image: docker.io/williamconti549/dd-apm-test-agent:integration-versions
         ports:
             - "127.0.0.1:9126:8126"
         volumes:
             - ./tests/snapshots:/snapshots
         environment:
-            - LOG_LEVEL=WARNING
+            - LOG_LEVEL=ERROR
```
mabdinur, line 138: Why do we need to change the log level? What warnings are we seeing?

.circleci/config.templ.yml
```diff
@@ -770,6 +776,9 @@ jobs:
 
   pytest:
     <<: *contrib_job
+    docker:
+      - image: *ddtrace_dev_image
```
mabdinur, line 780: Why is this needed?

tests/contrib/patch.py
```diff
@@ -737,3 +769,25 @@ def patch_wrapper(wrapped, _, args, kwrags):
...
 
                 self.assertEqual(out, b"OK", "stderr:\n%s" % err.decode())
+
+        def test_and_emit_get_version(self):
+            """Each contrib module should implement a get_version() function. This function is used for
+            the APM Telemetry integrations payload event, and by APM analytics to inform of dd-trace-py
+            current supported integration versions.
+            """
+            # import pdb
+            # breakpoint()
```
mabdinur, line 779: Make sure to remove this

tests/contrib/psycopg/test_psycopg_patch_generated.py
```diff
@@ -36,4 +39,8 @@ def assert_module_implements_get_version(self):
         versions = get_versions()
         assert "psycopg" in versions
         assert versions["psycopg"] != ""
```
mabdinur, line 41: nit:
```suggestion
        assert versions.get("psycopg")
```

mabdinur (commented):
This is beyond the scope of this PR but some of the modules have the name `test_integrationname_patch_generated`. We need to [remove](https://github.com/DataDog/dd-trace-py/blob/v2.1.0rc1/scripts/contrib-patch-tests.py#L58) the generated part of the name and make sure the patch test is actually making the expected assertions.

docker-compose.yml
wconti27: Reduce noise during debugging but will be switched back

.circleci/config.templ.yml
wconti27: this adds the test agent to the job, it originally wasn't being used for pytest

tests/contrib/patch.py
wconti27: removed

ddtrace/contrib/elasticsearch/patch.py
```diff
@@ -50,7 +50,9 @@ def _es_modules():
     )
     for module_name in module_names:
         try:
-            yield import_module(module_name)
+            module = import_module(module_name)
+            versions[module.__name__] = getattr(module, "__versionstr__", "")
```
mabdinur, line 54:
```suggestion
            versions[module_name] = getattr(module, "__versionstr__", "")
```

docker-compose.yml
```diff
@@ -131,13 +131,13 @@ services:
...
     testagent:
-        image: ghcr.io/datadog/dd-apm-test-agent/ddapm-test-agent:v1.11.0
+        image: ghcr.io/datadog/dd-apm-test-agent/ddapm-test-agent:v1.13.1
         ports:
             - "127.0.0.1:9126:8126"
         volumes:
             - ./tests/snapshots:/snapshots
         environment:
-            - LOG_LEVEL=WARNING
+            - LOG_LEVEL=DEBUG
```
mabdinur, line 140:
```suggestion
            - LOG_LEVEL=WARNING
```

tests/contrib/aiohttp/test_aiohttp_patch_generated.py
```diff
@@ -19,6 +19,7 @@ class TestAiohttpPatch(PatchTestCase.Base):
     __module_name__ = "aiohttp"
```
mabdinur, line 19: We can do this in a follow up but the file name should be `tests/contrib/aiohttp/test_aiohttp_patch.py` (remove the _generated.py)

majorgreys (approved):

Issue #5325, opened 2023-03-14
WSGI Middleware fails to call `close()` on iterables


petersalas:
The WSGI middleware attempts to [call close on `iter(x)` rather than `x`](https://github.com/DataDog/dd-trace-py/blob/caca9d3b875d1a6d4d3708d948477d0df31572a9/ddtrace/contrib/wsgi/wsgi.py#L195-L217). This works correctly if the iterable happens to be its own iterator  but not generally. Per the [WSGI spec](https://peps.python.org/pep-0333/#specification-details) close should be attempted on the _iterable_ rather than its iterator.

In Django applications this can prevent the [`request_finished` signal](https://docs.djangoproject.com/en/4.1/ref/signals/#django.core.signals.request_finished) from being sent and cause database connection leaks.

I was able to work around this by adding middleware that introduces a wrapper iterable that ensures that `close()` on its iterator is forwarded to the original iterable.

emmettbutler:
Though #5858 is mentioned here, it looks like that change doesn't resolve this issue.

emmettbutler:
@petersalas does https://github.com/DataDog/dd-trace-py/pull/6692 fix the issue you're describing? Do you have any suggestions about how we might replicate the failure mode you mentioned?

petersalas:
Looking at that fix, I think it would only fix it in certain scenarios (namely: iterables that are also iterators, but not their _own_ iterator). Here's the workaround I used:

```python
class IterableCloseForwarder:
    """Wraps an iterable to ensure that `iter(self).close()` will forward the
    `close()` call to the wrapped iterable.

    Works around a bug in ddtrace-py where it assumes that iter(x).close() is
    equivalent to x.close().

    https://github.com/DataDog/dd-trace-py/blob/975266967de6ca3d1a753f843cd3ea5b6ce56f45/ddtrace/contrib/wsgi/wsgi.py#L195-L217
    """

    def __init__(self, iterable):
        self._iterable = iterable
        self._iterator = iter(iterable)

    def __next__(self):
        return next(self._iterator)

    def __iter__(self):
        return self

    def close(self):
        if getattr(self._iterable, "close", None):
            self._iterable.close()

    @classmethod
    def middleware(cls, app):
        return lambda *args, **kwargs: cls(app(*args, **kwargs))
```

and then I ensured that `IterableCloseForwarder.middleware(app)` was the WSGI app passed to `ddtrace.contrib.wsgi.DDWSGIMiddleware`.

[...]

PR #6858, opened 2023-09-07
fix(wsgi): handle django wsgi app being an iterator

emmettbutler:
This change fixes https://github.com/DataDog/dd-trace-py/issues/5325 by having the WSGI middleware handle cases when the wrapped iterable object is an iterator. This condition can arise when a Django app attempts to send its "request finished" signal, in which case it may cause connection leaks. Standard methods of distinguishing an iterable from its iterator, such as checking for the presence of __iter__ and __next__ methods, don't work in this case for unknown reasons. Instead of adding brittle special-case detection logic to the middleware, this new argument allows users to indicate when this is the case.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


emmettbutler:
@petersalas let me know if this adjustment to your suggested fix makes sense to you.

emmettbutler:
@petersalas let me know what you think of this addition to the change you proposed

petersalas:
I added a repro of the original issue here: https://gist.github.com/petersalas/aa1ce634e2cd83fdf2dc22431c3a1b9b -- let me know if you're not able to reproduce it.

Re: the changes, I'm not sure I totally follow the behavior given the `wrapt.ObjectProxy` -- but if it fixes the Django repro above that's good enough for me. :)

emmettbutler:
@petersalas thanks so much for the reproduction code. I adapted it into a test and changed this pull request such that it passes. The solution I've implemented here allows you to work around the issue by passing `app_is_iterator=True` to the `DDWSGIMiddleware` constructor.

petersalas:
Glad to hear it! For my own understanding, what's the scenario where the new bool should _not_ be set to `True`?

emmettbutler:
@petersalas [This test](https://github.com/DataDog/dd-trace-py/blob/6fd5712010c0ad614306c0dbfbe3691fccf052b8/tests/contrib/wsgi/test_wsgi.py#L194) was one of the many tests of the `DDWSGIMiddleware` that failed when the new logic was applied unconditionally. I suspect that the difference has to do with [this](https://docs.python.org/3/glossary.html#term-iterator) note from the Python documentation about subtle differences between different types of iterators:

> A container object (such as a [list](https://docs.python.org/3/library/stdtypes.html#list)) produces a fresh new iterator each time you pass it to the [iter()](https://docs.python.org/3/library/functions.html#iter) function or use it in a [for](https://docs.python.org/3/reference/compound_stmts.html#for) loop. Attempting this with an iterator will just return the same exhausted iterator object used in the previous iteration pass, making it appear like an empty container.

That test ends up giving the `TracedIterable` a generator object, whereas the test this pull request adds gives it a Django `HttpResponse` object. I suspect that these two types differ in the behavior described here, though I haven't been able to prove it yet.

I looked for a long time for a particular attribute or attributes that could be used to differentiate between objects that could handle this new logic and objects that still needed the old style. `__next__`, `__iter__`, and `__getitem__`, the ones that by specification differentiate iterables from iterators, followed the same set/unset pattern in both of these test cases, making them not a useful differentiator. The same was true of `__len__` in most cases. Because I couldn't find an intuitive way of having `TracedIterable` make this decision internally, I decided to push that responsibility to the integration code, where it makes sense to pass off this behavior as "Django does something weird here".

petersalas:
@emmettbutler, thanks for clarifying! It looks like the issue you're describing is a side-effect of using `wrapt.ObjectProxy` -- when I [removed that bit](https://github.com/petersalas/dd-trace-py/commit/629f7261a8f9428a723d73f9d45a415f3203da0a) all the `wsgi` tests passed. That said, I'm not sure whether there are other reasons to use `wrapt.ObjectProxy` beyond (the now explicit) `__next__` delegation, so I'll leave it to you to decide what to do with that information :)

emmettbutler:
@petersalas I think that because `TracedIterable` instances are returned to application code via a WSGI app's `__call__` function, they must inherit from `ObjectProxy` to avoid changing the contract between application code and the WSGI spec (one of the core objectives of dd-trace-py). Adding tests that validate this could be helpful in a future change.

Thanks again for the contributions!

mabdinur (approved):

PR #6867, opened 2023-09-08
ci: use single resource to decrease benchmark memory usage

emmettbutler:
This change reduces the number of different resource names tested during sampling benchmarking. The theory behind this change is that the benchmark scenario generates all combinations of resource, name, and operation, which prior to this change was more than a gigabyte of data. Reducing the number of resources to 1 keeps this multiplicative factor the same as it was before resource was added without changing the utility of the benchmark.

What's still unclear to me is why the benchmark comment isn't showing a large reduction in memory usage. This could be for the same reason that the memory increase didn't appear in the comment on #6821.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


mabdinur (approved):
Why does setting `num_resources` to 100 cause memory issues in the benchmarking platform? 

Can we update the PR description to include why this change is necessary.

emmettbutler:
@mabdinur I updated the description. I'm curious to know what you think about it.

PR #6890, opened 2023-09-12
chore(kafka): use core api for data streams monitoring

tabgok:
This commit swaps out all remaining DSM code in the kafka integration to
use the core API.
    
This commit also contains the addition of two tests and fixes to the
commit wrapper which were required to safely transition.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/contrib/kafka/patch.py
```diff
@@ -229,13 +191,6 @@ def traced_commit(func, instance, args, kwargs):
...
-    if config._data_streams_enabled:
-        message = get_argument_value(args, kwargs, 0, "message")
-        offsets = kwargs.get("offsets", [])
-        if message is not None:
-            offsets = [TopicPartition(message.topic(), message.partition(), offset=message.offset())]
-        for offset in offsets:
-            pin.tracer.data_streams_processor.track_kafka_commit(
-                instance._group_id, offset.topic, offset.partition, offset.offset or -1, time.time()
-            )
+    core.dispatch("kafka.commit.start", [instance, args, kwargs])
```
piochelepiotr, line 194: are we sure the consumer of this event will be executed before the function is called?

tabgok: The calls to core and events are (and will be for the foreseeable future) synchronous.  APM, DSM, and ASM all have the same requirements to be able to modify parameters and take action prior to execution, so its a necessary behavior

tests/contrib/kafka/test_kafka.py
```diff
@@ -101,6 +104,22 @@ def consumer(tracer, kafka_topic):
     _consumer.close()
 
 
+@pytest.fixture
+def non_auto_commit_consumer(tracer, kafka_topic):
```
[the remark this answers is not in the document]
tabgok: The short answer is "yes", the longer answer is this split should help split up tests so we aren't having to use subprocesses to try different Agent configurations anymore (i.e. for datastreams we may be able to get away with a single suite of tests with `DD_DATA_STREAMS_ENABLED` set instead of spawning a subprocess per test as we do now).

With that said, the immediate answer is "I'm refactoring in parts, so the tests will come later"

ddtrace/internal/datastreams/kafka.py
```diff
@@ -24,7 +65,34 @@ def dsm_kafka_message_consume(instance, message):
     ctx = processor().decode_pathway(headers.get(PROPAGATION_KEY, None))
     ctx.set_checkpoint(["direction:in", "group:" + group, "topic:" + topic, "type:kafka"])
 
+    if instance._auto_commit:
+        # it's not exactly true, but if auto commit is enabled, we consider that a message is acknowledged
+        # when it's read.
+        reported_offset = message.offset() + 1 if isinstance(message.offset(), INT_TYPES) else -1
```
tabgok, line 71: this (and similar) changes fix an off-by-one error; when passing in a message to `commit()` confluent_kafka uses `message + 1` for the offset as the convention is to point the offset at the NEXT message.

ddtrace/contrib/kafka/patch.py
```diff
@@ -76,7 +72,10 @@ def poll(self, timeout=None):
         return super(TracedConsumer, self).poll(timeout)
 
     def commit(self, message=None, *args, **kwargs):
-        return super(TracedConsumer, self).commit(message, args, kwargs)
+        if message:
+            return super(TracedConsumer, self).commit(message, *args, **kwargs)
```
tabgok, line 76: two changes here:

1. (message, args, kwargs) => (message, *args, **kwargs)
2. Support of calling commit with offsets only (message and offsets are mutually exclusive args to commit().

ddtrace/internal/datastreams/__init__.py
```diff
@@ -1,14 +1,23 @@
-from . import kafka  # noqa:F401
+from ddtrace import config
+from ddtrace import tracer
 
+from ...internal.utils.importlib import require_modules
 
+
+required_modules = ["confluent_kafka"]
```
tabgok, line 7: I copied this pattern from contrib, is it necessary/correct?

ddtrace/internal/datastreams/__init__.py
```diff
@@ -1,14 +1,23 @@
...
+
 
 def data_streams_processor():
     from . import processor
 
     global _processor
-    if not _processor:
-        _processor = processor.DataStreamsProcessor()
+    if config.data_streams_enabled and not _processor:
+        _processor = processor.DataStreamsProcessor(tracer._agent_url)
```
tabgok, line 21: is this a breaking change?  We're moving from "attach your own processor" to "use the global processor".   I *feel* like its not a breaking change, because the stats should end up the same, but I'd like another opinion.

ddtrace/internal/datastreams/__init__.py
emmettbutler: Cool! It makes sense here as far as I can tell.

ddtrace/internal/datastreams/__init__.py
emmettbutler: It doesn't look to me like a breaking change.

emmettbutler (approved):
very nice!

emmettbutler (approved):

tests/contrib/kafka/test_kafka.py
```diff
@@ -424,3 +444,52 @@ def test():
     out, err, status, _ = ddtrace_run_python_code_in_subprocess(code, env=env)
     assert status == 0, out.decode()
     assert err == b"", err.decode()
+
```
tabgok, line 447: TODO: add in a third test without message or offsets

pyproject.toml
```diff
@@ -154,6 +154,8 @@ exclude-modules = '''
   | ddtrace.appsec._iast._ast.aspects
   | ddtrace.appsec._iast._taint_utils
   | ddtrace.appsec._iast.taint_sinks.sql_injection
+  # DSM specific contribs
```
tabgok, line 157: add a release note for the fixes

ddtrace/internal/datastreams/kafka.py
```diff
@@ -13,6 +23,29 @@ def dsm_kafka_message_produce(instance, args, kwargs):
...
+    try:
+        on_delivery = get_argument_value(args, kwargs, on_delivery_arg, on_delivery_kwarg)
+    except ArgumentError:
+        on_delivery_kwarg = "callback"
+        on_delivery_arg = 4
+        on_delivery = get_argument_value(args, kwargs, on_delivery_arg, on_delivery_kwarg, optional=True)
+
+    def wrapped_callback(err, msg):
+        if err is None:
+            reported_offset = msg.offset() + 1 if isinstance(msg.offset(), INT_TYPES) else -1
```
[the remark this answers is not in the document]
hokitam: nvm just saw your comment below!

tests/contrib/kafka/test_kafka.py
tabgok: This will be taken care of at a later date, its not really supported right now

hokitam (approved):

emmettbutler (dismissed):

pyproject.toml
```diff
@@ -154,6 +154,8 @@ exclude-modules = '''
...
+  | ddtrace.internal.datastreams.kafka
```
mabdinur, line 158: Why do need to skip slots check for this module. Is it because of the `confluent_kafka` import?

releasenotes/notes/dsm-fix-offsets-and-metrics-7e34b2734667e986.yaml
```diff
@@ -0,0 +1,3 @@
+fixes:
+  - |
+    DSM: fix off-by-one metric issue and error where statistics weren't calculated when the core API was used.
```
mabdinur, line 3: nit
```suggestion
    DSM: fix off-by-one metric issue and error where statistics weren't calculated when the core API was used.

```

tests/contrib/kafka/test_kafka.py
```diff
@@ -16,6 +16,8 @@
 from ddtrace.contrib.kafka.patch import unpatch
```
mabdinur, line 16: We should add a [subprocess](https://github.com/DataDog/dd-trace-py/blob/c6808b2d6d016aa600b7c52c50b172eeca44bd0d/tests/contrib/redis/test_redis_cluster_asyncio.py#L342) test to ensure the dsm atexit hooks generate the expected traces/payloads.

ddtrace/internal/datastreams/kafka.py
```diff
@@ -13,6 +23,29 @@ def dsm_kafka_message_produce(instance, args, kwargs):
...
 
+    on_delivery_kwarg = "on_delivery"
+    on_delivery_arg = 5
+    on_delivery = None
+    try:
+        on_delivery = get_argument_value(args, kwargs, on_delivery_arg, on_delivery_kwarg)
+    except ArgumentError:
+        on_delivery_kwarg = "callback"
+        on_delivery_arg = 4
+        on_delivery = get_argument_value(args, kwargs, on_delivery_arg, on_delivery_kwarg, optional=True)
```
mabdinur, line 34: nit: Instead of catching an argument error can we parse the arguments based on the version of `confluent_kafka` library. 

ex:
```suggestion
   if confluent_kafka.VERSION >= (2,0): #random example not sure if this is the right version
        on_delivery = get_argument_value(args, kwargs, 5, "on_delivery", True)
    else:
        on_delivery = get_argument_value(args, kwargs, 4, "callback", optional=True)
```

pyproject.toml
tabgok: yes, exactly

ddtrace/internal/datastreams/kafka.py
tabgok: I'm not entirely sure on the reasoning here; I can dig in some

hokitam (approved):

tests/datastreams/test_processor.py
```diff
@@ -52,3 +54,19 @@ def test_kafka_offset_monitoring():
     assert processor._buckets[bucket_time_ns].latest_produce_offsets[PartitionKey("topic1", 1)] == 34
     assert processor._buckets[bucket_time_ns].latest_produce_offsets[PartitionKey("topic1", 2)] == 10
     assert processor._buckets[bucket_time_ns].latest_commit_offsets[ConsumerPartitionKey("group1", "topic1", 1)] == 14
+
+
+def test_processor_atexit(ddtrace_run_python_code_in_subprocess):
```
mabdinur, line 59: Was this test supposed to use the `ddtrace_run_python_code_in_subprocess` fixture?

mabdinur: I think we should still have a subprocess test to ensure _atexit it actually called when a process shutsdown

tabgok: changed to run an actual subprocess and assert its called

mabdinur (approved):

tabgok:
https://github.com/DataDog/dd-trace-py/pull/7130/checks?check_run_id=17327403810

PR #6891, opened 2023-09-12
fix(openai): ensure streamed spans with error are manually finished

### LINKED ISSUE #6769: fastapi + openai chat completion streaming: exceptions cause whole trace to disappear
### Summary of problem
When using `ddtrace-run` with [FastAPI](https://github.com/tiangolo/fastapi) and calling the `openai.ChatCompletion.create` with the `stream` parameter set to `True`, if an exception occurs in the function (e.g. invalid API key), whether caught or not, the entire trace of the request is gone. This doesn't happen if `openai` patching is disabled or `stream` is disabled, and happens for both `openai.ChatCompletion.create` and `openai.ChatCompletion.acreate`.

1.18.2

pip-23.2.1, Python 3.11.4

```
aiohttp==3.8.5
aiosignal==1.3.1
annotated-types==0.5.0
anyio==3.7.1
async-timeout==4.0.3
attrs==23.1.0
bytecode==0.14.2
cattrs==23.1.2
certifi==2023.7.22
charset-normalizer==3.2.0
click==8.1.7
ddsketch==2.0.4
ddtrace==1.18.2
Deprecated==1.2.14
envier==0.4.0
fastapi==0.103.0
frozenlist==1.4.0
gunicorn==21.2.0
h11==0.14.0
httptools==0.6.0
idna==3.4
importlib-metadata==6.8.0
multidict==6.0.4
openai==0.27.9
opentelemetry-api==1.19.0
packaging==23.1
protobuf==4.24.2
pydantic==2.3.0
pydantic_core==2.6.3
python-dotenv==1.0.0
PyYAML==6.0.1
requests==2.31.0
six==1.16.0
sniffio==1.3.0
starlette==0.27.0
tqdm==4.66.1
typing_extensions==4.7.1
urllib3==2.0.4
uvicorn==0.23.2
uvloop==0.17.0
watchfiles==0.20.0
websockets==11.0.3
wrapt==1.15.0
xmltodict==0.13.0
yarl==1.9.2
zipp==3.16.2
```

Have Datadog Agent running locally with default settings.
Install all the requirements provided above into a Python venv, then create a small fastapi app to reproduce it by creating a directory named e.g. `bug` with the following structure:
bug
├── app
.│ ......└── main.py

The content of `main.py` contains a single endpoint which creates a chat completion with streaming enabled:
```py
import openai
from fastapi import FastAPI
from fastapi.responses import JSONResponse

openai.api_key = "sk-abc" # Intentionally set wrong to trigger exception

app = FastAPI()

@app.get("/test") # get or post doesn't make a difference
def test(): # async def doesn't make a difference
  messages = [
    {"role": "user", "content": "hello"},
  ]
  try:
      for response_fragment in openai.ChatCompletion.create(
          model="gpt-3.5-turbo",
          messages=messages,
          stream=True,
      ):
        print(response_fragment)
  except Exception as e:
      print(e)
      return JSONResponse("not ok", 500)
  return JSONResponse("ok")
```

Run the fastapi server with `dd-trace` and `uvicorn`: 
```sh
DD_TRACE_LOG_FILE=ddtrace_logs.log DD_TRACE_DEBUG=true DD_SERVICE=test-service ddtrace-run uvicorn app.main:app --reload
``` 
And hit the endpoint at `http://127.0.0.1:8000/test`: 
```sh
curl http://127.0.0.1:8000/test
```
Observe that no trace is sent at all for the requests in the `ddtrace_logs.log` file.

No traces are sent for the request at all, making it impossible to track requests when `openai` throws exceptions.

Traces are sent for the requests with the `openai` spans showing errors. When streaming is disabled or openai patching is disabled by adding the environment variable `DD_PATCH_MODULES = 'openai:false'`, it works as expected.

Logs when there's no trace sent with exceptions (`stream=True`):
image
Logs when there are traces sent successfully with exceptions (`stream=False`):
image

Yun-Kim:
Resolves #6769.

This PR fixes an unhandled case in our OpenAI integration specifically regarding streamed `Chat/Completion` requests that have errors and result in an empty response. In this case, streamed spans were not finished (as we avoided finishing streamed response spans until the underlying generator was exhausted) because we did not handle empty response cases. The fix here is to add manual span finishing for streamed response spans with error, as we already do that for non-streamed spans.

The only risk is if there are requests that we trace that might result in a non-empty response even with an error. This is highly unlikely as we shouldn't get any response if the corresponding request was faulty.

Note: this PR also moves tagging prompt token usage information to the `EndpointHook.process_response()` instead of the `EndpointHook.process_request()` handler as keeping that in the latter results in prompt token information being recorded, even if there were no actual prompt/completion operation happening for OpenAI due to the erroneous request.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.

### DIFF
--- a/ddtrace/contrib/openai/_endpoint_hooks.py
+++ b/ddtrace/contrib/openai/_endpoint_hooks.py
@@ -91,7 +91,6 @@ class _BaseCompletionHook(_EndpointHook):
             try:
                 num_prompt_tokens = span.get_metric("openai.response.usage.prompt_tokens") or 0
                 num_completion_tokens = yield
-
                 span.set_metric("openai.response.usage.completion_tokens", num_completion_tokens)
                 total_tokens = num_prompt_tokens + num_completion_tokens
                 span.set_metric("openai.response.usage.total_tokens", total_tokens)
@@ -180,8 +179,15 @@ class _CompletionHook(_BaseCompletionHook):
             elif prompt:
                 for idx, p in enumerate(prompt):
                     span.set_tag_str("openai.request.prompt.%d" % idx, integration.trunc(str(p)))
+        return
+
+    def _record_response(self, pin, integration, span, args, kwargs, resp, error):
+        if not resp:
+            return self._handle_response(pin, span, integration, resp)
+        prompt = kwargs.get("prompt", "")
         if kwargs.get("stream"):
             num_prompt_tokens = 0
+            estimated = False
             if isinstance(prompt, str) or isinstance(prompt, list) and isinstance(prompt[0], int):
                 estimated, prompt_tokens = _compute_prompt_token_count(prompt, kwargs.get("model"))
                 num_prompt_tokens += prompt_tokens
@@ -191,10 +197,6 @@ class _CompletionHook(_BaseCompletionHook):
                     num_prompt_tokens += prompt_tokens
             span.set_metric("openai.request.prompt_tokens_estimated", int(estimated))
             span.set_metric("openai.response.usage.prompt_tokens", num_prompt_tokens)
-        return
-
-    def _record_response(self, pin, integration, span, args, kwargs, resp, error):
-        if not resp or kwargs.get("stream"):
             return self._handle_response(pin, span, integration, resp)
         if "choices" in resp:
             choices = resp["choices"]
@@ -212,7 +214,6 @@ class _CompletionHook(_BaseCompletionHook):
                     span.set_tag_str("openai.response.choices.%d.text" % idx, integration.trunc(choice.get("text")))
         integration.record_usage(span, resp.get("usage"))
         if integration.is_pc_sampled_log(span):
-            prompt = kwargs.get("prompt", "")
             integration.log(
                 span,
                 "info" if error is None else "error",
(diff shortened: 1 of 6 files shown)


ddtrace/contrib/openai/_endpoint_hooks.py
```diff
@@ -254,19 +255,20 @@ def _record_request(self, pin, integration, span, args, kwargs):
...
+
+    def _record_response(self, pin, integration, span, args, kwargs, resp, error):
+        if not resp:
+            return self._handle_response(pin, span, integration, resp)
+        messages = kwargs.get("messages")
         if kwargs.get("stream"):
-            # streamed responses do not have a usage field, so we have to
-            # estimate the number of tokens returned.
             est_num_message_tokens = 0
+            estimated = False
```
mabdinur, line 266: Estimated is a boolean here but it's then cast to an int. Dumb question. Is this intentional?

mabdinur (approved):
I don't think I fully understand the risk of this PR. Why is an empty response an issue here. Would we generate spans that have invalid/unexpected format? 

Also are Chat/Completion endpoints the only ones that be streamed? Would we need to handle this edge case in other endpoints?


Aside from nits and dumb questions LGTM

ddtrace/contrib/openai/_endpoint_hooks.py
Yun-Kim: `estimated` is just a placeholder span metric value that we use internally. For simplicity we cast it to an int when setting it as a span metric but in reality it's just 0(false) or 1 (true).

Yun-Kim:
@mabdinur Thanks for the great comments! I'll try to answer your concerns below in order:

> I don't think I fully understand the risk of this PR. Why is an empty response an issue here. Would we generate spans that have invalid/unexpected format?

This is a potential risk because then we would potentially be prematurely finishing spans before we have a chance to tag response parameters or metadata. This is most likely a non-issue, I'm just being paranoid of potential cases where it might be an erroneous request but still returns a response. I just can't prove that this will never happen which is the only reason I added this on the PR description.

> Also are Chat/Completion endpoints the only ones that be streamed? Would we need to handle this edge case in other endpoints?

Chat and Completion endpoints are the only ones AFAIK that are streamable. There is also the `ListFineTune` endpoint which can be streamed, but interestingly enough doesn't have the same issue (error spans finish).

PR #6925, opened 2023-09-18
fix(internal): do not replace sys.modules

### LINKED ISSUE #6923: importing ddtrace causes an exception atexit when working with flask and gevent
### Summary of problem

The act of importing `ddtrace` causes a `gevent` error on python shutdown when trying to use Flask CLI (but can also be reproduced without using the CLI explicitly, see below).

`1.18.5`

`pip 23.2`

The libraries listed are the result of running `pip install flask gevent ddtrace`.

<details>

> attrs==23.1.0
> blinker==1.6.2
> bytecode==0.15.0
> cattrs==23.1.2
> click==8.1.7
> ddsketch==2.0.4
> ddtrace==1.18.5
> Deprecated==1.2.14
> envier==0.4.0
> Flask==2.3.3
> gevent==23.9.1
> greenlet==3.0.0rc3
> importlib-metadata==6.8.0
> itsdangerous==2.1.2
> Jinja2==3.1.2
> MarkupSafe==2.1.3
> opentelemetry-api==1.20.0
> protobuf==4.24.3
> six==1.16.0
> typing_extensions==4.7.1
> Werkzeug==2.3.7
> wrapt==1.15.0
> xmltodict==0.13.0
> zipp==3.16.2
> zope.event==5.0
> zope.interface==6.0

</details>

This is the minimal `app.py` to reproduce the error:

```python
# app.py
from gevent import monkey

monkey.patch_all()

import ddtrace
from flask import Flask

app = Flask("testy")
```
Running `flask --help` will give the error below.

The reason seems to be the fact that flask is imported before `monkey.patch_all()` is called, but _only if `ddtrace` is imported as well_. Removing the `import ddtrace` line removes the exception.

The error is also visible when just importing flask manually before `monkey.patch_all()`:

```python
# oops.py
from gevent import monkey

import flask

monkey.patch_all()

import ddtrace
```

`python oops.py` gives the same error. Importing flask before running `monkey.patch_all()` is a mistake of course, but one that cannot be avoided when running `flask` as CLI directly.

As far as I can tell, this is caused by `ModuleWatchdog.install()` on import, as commenting that line seems to fix the error. Running `ModuleWatchdog.uninstall()` "manually" before shutdown doesn't seem to help.

This is shown _always_ at shutdown, and it seems it cannot be suppressed:

```
Exception ignored in: <function _removeHandlerRef at 0x7fc252f6fb00>
  File "/usr/lib/python3.11/logging/__init__.py", line 855, in _removeHandlerRef
  File "/usr/lib/python3.11/logging/__init__.py", line 235, in _acquireLock
  File "/usr/lib/python3.11/threading.py", line 164, in acquire
  File "/tmp/testimir/.venv/lib/python3.11/site-packages/gevent/thread.py", line 67, in get_ident
TypeError: 'NoneType' object is not callable
```

To not see the error :1st_place_medal:

P403n1x87:
We refactor the ModuleWatchdog class to avoid replacing the sys.modules dictionary with a proxy. Instead, we use a WeakValueDictionary to prevent adding strong references to modules that might be unloaded, and also preventing the origin mapping from growing unchecked over time.

Fixes #6923.

## Testing Strategy

The existing test suites should be good enough to catch any regressions from this substantial change. As for the mentioned issue, the change was tested manually on the reproducer provided in the description provided by the reporter.

## Risk

This is a substantial change in behaviour for `ModuleWatchdog`. However, we expect things to run smoother now that `sys.modules` is not _hacked_, at the cost of a potentially slightly delayed memory clean-up.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


mabdinur (dismissed):
Fix looks good to me but the conflicts might be a pain to resolve

ddtrace/internal/module.py
```diff
@@ -499,7 +457,7 @@ def register_module_hook(cls, module, hook):
         instance = cast(ModuleWatchdog, cls._instance)
         instance._hook_map[module].append(hook)
         try:
-            module_object = instance[module]
+            module_object = sys.modules[module]
```
mabdinur, line 460: Would this create a strong reference to the module object? Is this an issue?

mabdinur (approved):

ddtrace/internal/module.py
P403n1x87: We are not storing the result of this lookup anywhere. So unless the `hook` itself stores the module somewhere (which is out of our control here), we have a net-zero refcount.

majorgreys (approved):

PR #6970, opened 2023-09-21
chore(botocore): don't log an error when kinesis payload isn't json

meadsteve:
We are using boto3 to push protobuf encoded data to kinesis. We noticed the following error in our logs:

```
Unable to parse payload, unable to inject trace context.
```

It seems that any data that isn't json encoded will cause this error message. I don't believe there's any requirement for data to be json encoded so I believe this should not be an error level log.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.

### DIFF
--- a/ddtrace/contrib/botocore/patch.py
+++ b/ddtrace/contrib/botocore/patch.py
@@ -292,7 +292,7 @@ def get_kinesis_data_object(data):
     - json string
     - byte encoded json string
     - base64 encoded json string
-    If it's neither of these, then we leave the message as it is.
+    If it's none of these, then we leave the message as it is.
     """
 
     # check if data is a json string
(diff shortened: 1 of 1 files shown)


ddtrace/contrib/botocore/patch.py
```diff
@@ -313,7 +313,7 @@ def get_kinesis_data_object(data):
         data_str = base64.b64decode(data).decode("ascii")
         return get_json_from_str(data_str)
     except Exception:
-        log.error("Unable to parse payload, unable to inject trace context.")
+        log.debug("Unable to parse payload, unable to inject trace context.")
```
meadsteve, line 316: debug level here because all the earlier checks logged at debug level

meadsteve:
thanks for looking at this @emmettbutler. If I want this fix in production soon should I also create a PR against the 1.x branch? I'm not sure what the timeline for 2.x is.

emmettbutler:
@meadsteve the quickest route to that is probably to make a pull request with the same diff against the 1.19 branch. before too long we'll make a bugfix 1.19.x release.

Issue #6973, opened 2023-09-21
Kafka TracedConsumer causes the exception "message and offsets are mutually exclusive"


shshe:
### Summary of problem

Kafka TracedConsumer causes the error "message and offsets are mutually exclusive" when calling `consumer.commit()` and having `config['enable.auto.commit'] = 'false'`.

Setting `DD_TRACE_KAFKA_ENABLED=false` resolves this error.

```
Traceback (most recent call last):
...
  File "... app.py", line ..., in handle
    consumer.commit(message=msg)
  File "/usr/local/lib/python3.9/site-packages/ddtrace/contrib/kafka/patch.py", line 241, in traced_commit
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/ddtrace/contrib/kafka/patch.py", line 79, in commit
    return super(TracedConsumer, self).commit(message, args, kwargs)
ValueError: message and offsets are mutually exclusive
```

### Which version of dd-trace-py are you using?

1.20.0

### Which version of pip are you using?

pip 22.2.2

### Which libraries and their versions are you using?

<details>
  <summary>`pip freeze`</summary>
confluent-kafka = "==2.2.0"
ddtrace = "==1.20.0"
</details>

### How can we reproduce your problem?

```python
from confluent_kafka import Consumer

config = {'enable.auto.commit':  'false'}
consumer = Consumer(config)
while True:
    try:
        msg = consumer.poll(1.0)
        consumer.commit(message=msg)
    except:
        ...
```

### What is the result that you get?

```
Traceback (most recent call last):
...
  File "... app.py", line ..., in handle
    consumer.commit(message=msg)
  File "/usr/local/lib/python3.9/site-packages/ddtrace/contrib/kafka/patch.py", line 241, in traced_commit
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/ddtrace/contrib/kafka/patch.py", line 79, in commit
    return super(TracedConsumer, self).commit(message, args, kwargs)
ValueError: message and offsets are mutually exclusive
```

### What is the result that you expected?
No error.

PR #6976, opened 2023-09-21
fix(kafka): avoid overriding  methods in Producers and Consumers

mabdinur:
Resolves: https://github.com/DataDog/dd-trace-py/issues/6973

This PR achieves two goals:
 1) Fixes a bug where calling `confluent_kafka.Consumer(....).commit(offset=......)` raises the following error: `ValueError: message and offsets are mutually exclusive`. This is because `TracedConsumer`  intercepts this invocation and then calls: `Consumer(....).commit(message=None, offset=......)`.
 2) Cleans up the kafka patching code. There is no need to override a method just to call the method defined in the super class.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/contrib/kafka/patch.py
```diff
@@ -230,10 +221,12 @@ def traced_commit(func, instance, args, kwargs):
         return func(*args, **kwargs)
 
     if config._data_streams_enabled:
-        message = get_argument_value(args, kwargs, 0, "message")
-        offsets = kwargs.get("offsets", [])
+        message = get_argument_value(args, kwargs, 0, "message", True)
+        # messgae and offset are mutally exclusive. Only one parameter can be passed.
```
ZStriker19, line 225:
```suggestion
        # message and offset are mutually exclusive. Only one parameter can be passed.
```

releasenotes/notes/fix-kafka-commit-offsets-d86039c29e04259e.yaml
```diff
@@ -0,0 +1,4 @@
+---
+fixes:
+  - |
+    kafka: Fixes ``ValueError`` raised when ``Consumer.commit(offsets=...)`` is called
```
ZStriker19, line 4:
```suggestion
    kafka: Fixes ``ValueError`` raised when ``Consumer.commit(offsets=...)`` is called.
```

ddtrace/contrib/kafka/patch.py
```diff
@@ -230,10 +221,12 @@ def traced_commit(func, instance, args, kwargs):
...
         if message is not None:
             offsets = [TopicPartition(message.topic(), message.partition(), offset=message.offset())]
+        else:
+            offsets = get_argument_value(args, kwargs, 1, "offsets", True)
```
ZStriker19, line 229: If this returns `None`, then don't we try to iterate through `None` which will throw an error?

releasenotes/notes/fix-kafka-commit-offsets-d86039c29e04259e.yaml
ZStriker19, line 4:
```suggestion
    kafka: Fixes ``ValueError`` raised when ``Consumer.commit(offsets=...)`` is called.
```

ddtrace/contrib/kafka/patch.py
ZStriker19, line 225:
```suggestion
        # message and offset are mutually exclusive. Only one parameter can be passed.
```

ZStriker19 (commented):
nit and question

ddtrace/contrib/kafka/patch.py
mabdinur: Consumer.commit() can be called with the following arguments: `message`, `offsets` and `asynchronous`. 

The following function calls are valid:
`Consumer.commit(message=....)`
`Consumer.commit(message=..., asynchronous=....)`
`Consumer.commit(offsets=None, asynchronous=...)`
`Consumer.commit(asynchronous=...)`

The following function calls are INVALID. This raises a ValueError:
`Consumer.commit(message=...., offsets=.....)`


There is a scenario where both message and offsets are not passed as keyword arguments. In this case `get_argument_value(....)` should not raise an ArgumentError. It should just return `None`.



Docs for reference: https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/index.html#confluent_kafka.Consumer.commit

ZStriker19:
Kafka test is failing it seems

mabdinur:
> Kafka test is failing it seems

It's failing because I forgot to push up a new snapshot file

We should be good

ZStriker19 (approved):

PR #7002, opened 2023-09-22
ci: run system tests on 2.x

emmettbutler:
- [x] foobar


majorgreys:
Should we no longer run this scheduled job on 0.x and 1.x, given we are not making changes to them?

.github/workflows/system-tests.yml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - 0.x
       - 1.x
```
mabdinur, line 7: ?
```suggestion
```

emmettbutler:
Obsoleted by @ZStriker19's work in https://github.com/DataDog/dd-trace-py/pull/7035

PR #7074, opened 2023-09-27
chore(tracer): revert "limit memory usage for extra services reports"

emmettbutler:
Reverts DataDog/dd-trace-py#7060, which broke the `internal` test suite due to that suite not being run pre-merge.

- [x] foobar


gnufede (approved):

mabdinur (approved):

mabdinur:
Can we update the description to include why this revert is needed

PR #7089, opened 2023-09-28
fix(git): fix git version extraction

ericlaz:
## Motivation

This PR aims to fix Git version extraction which was relying on getting the last argument of `git --version` output. This is invalid since it could happen that the output: `git version 2.39.3 (Apple Git-145)`

[JIRA issue](https://datadoghq.atlassian.net/browse/CIAPP-7884?atlOrigin=eyJpIjoiNGJhODU0YzM3NDlkNGMxZWIzNWI4NWE3NzJmOWE0NmYiLCJwIjoiaiJ9)

## What has been changed?

Instead, we extract the Git version with the 3rd argument (`2.39.3`), which is in line with how `dd-trace-js` operates.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.

### DIFF
--- a/ddtrace/ext/git.py
+++ b/ddtrace/ext/git.py
@@ -143,7 +143,11 @@ def extract_user_info(cwd=None):
 
 def extract_git_version(cwd=None):
     output = _git_subprocess_cmd("--version")
-    version_info = tuple([int(part) for part in output.split()[2].split(".")])
+    try:
+        version_info = tuple([int(part) for part in output.split()[2].split(".")])
+    except ValueError:
+        log.error("Git version not found, it is not following the desired version format.")
+        return 0, 0, 0
     return version_info
 
 
--- a/ddtrace/ext/git.py
+++ b/ddtrace/ext/git.py
@@ -146,7 +146,7 @@ def extract_git_version(cwd=None):
     try:
         version_info = tuple([int(part) for part in output.split()[2].split(".")])
     except ValueError:
-        log.error("Git version not found, it is not following the desired version format.")
+        log.error("Git version not found, it is not following the desired version format: {}".format(output))
         return 0, 0, 0
     return version_info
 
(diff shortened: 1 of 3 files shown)


ddtrace/ext/git.py
```diff
@@ -143,7 +143,7 @@ def extract_user_info(cwd=None):
 
 def extract_git_version(cwd=None):
     output = _git_subprocess_cmd("--version")
-    version_info = tuple([int(part) for part in output.split()[-1].split(".")])
+    version_info = tuple([int(part) for part in output.split()[2].split(".")])
```
mabdinur, line 146: We should wrap this in a try-except and log an error if the output is in an invalid format. I am not sure if we should crash a test if we fail to extract a git version. 

Expecting `git --version` to have the following format: `git version 2.39.2` where the version only contains integers seems a bit brittle.  If the output was ``, `git version 2.39.2a1`, `git 2.49.21`, or  just `2.79.1` this function will raise a value error. I am not sure if any of these outputs are valid so feel free to ignore this suggestion.

mabdinur (approved):

ddtrace/ext/git.py
ericlaz: Good point. I will add a try except block.

ddtrace/ext/git.py
```diff
@@ -143,7 +143,11 @@ def extract_user_info(cwd=None):
 
 def extract_git_version(cwd=None):
     output = _git_subprocess_cmd("--version")
-    version_info = tuple([int(part) for part in output.split()[-1].split(".")])
+    try:
+        version_info = tuple([int(part) for part in output.split()[2].split(".")])
+    except ValueError:
+        log.error("Git version not found, it is not following the desired version format.")
```
mabdinur, line 149: nit: This will help us debug issues in the future
```suggestion
        log.error("Git version not found, it is not following the desired version format: {}".format(output))
```

mabdinur: Can we add a regression test where `_git_subprocess_cmd()` returns an invalid string. We don't want this to fail downstream.  We should still generate valid spans if the git version returns `0.0.0`

emmettbutler (approved):

mabdinur (approved):

PR #4360, opened 2022-10-25
refactor(iast): improve performance of get_info_frame

avara1986:
## Description
Improve the performance of `ddtrace.appsec.iast.stacktrace.get_info_frame`. Migrate to C++ the function to iterate over the stack. 

Python: 175.524807ms
Cython: 140.699406ms
C: 7.812132ms

## Reviewer Checklist
- [ ] Title is accurate.
- [ ] Description motivates each change.
- [ ] No unnecessary changes were introduced in this PR.
- [ ] PR cannot be broken up into smaller PRs.
- [ ] Avoid breaking [API](https://ddtrace.readthedocs.io/en/stable/versioning.html#interfaces) changes unless absolutely necessary.
- [ ] Tests provided or description of manual testing performed is included in the code or PR.
- [ ] Release note has been added for fixes and features, or else `changelog/no-changelog` label added.
- [ ] All relevant GitHub issues are correctly linked.
- [ ] Backports are identified and tagged with Mergifyio.
- [ ] Add to milestone.

PR #7112, opened 2023-09-30
fix: reference leaks and incorrect C-API usage in _stacktrace.c

pablogsal:
I was just reading this file and found many problems and leaks so I thought that a PR may be welcomed :wink: 

There are several problems and possible improvements with "_stacktrace.c":

* The macros for 3.11+ have some nested calls such as:

  #define GET_FILENAME(frame) PyObject_GetAttrString(PyFrame_GetCode(frame), "co_filename")

  The problem with these nested calls is that the nested functions
  return strong references that are never cleaned and also the call can
  fail, which will crash the interpreter or raise SystemError and doesn't
  allow the code to properly do error management.

* The macros for 3.11+ that return PyObject* are strong references that
  must be properly managed or otherwise the code will leak.

* There are many calls that are not checked for errors such as
  PyBytes_AsString, PyFrame_GetBack, PyThreadState_GetFrame, ...

* When PyThreadState_Get or PyThreadState_GetFrame return NULL, that's
  considered a hard error (for instance this can happen at interpreter
  teardown) and returning None is incorrect. The error must be
  propagated.

* There invalid paths are innecesarily obscured and the only situation
  where None should be returned (although I would advise to return an
  error instead) is if the argument passed is not a unicode/string object.

* The function just receives one argument, so it should use METH_O
  instead of adding the extra complication of using METH_VARARGS.

The leaks can be confirmed using `memray` or a debug version of python. Just use this example `setup.py` to compile the extension:

```
from setuptools import setup, Extension

extension_module = Extension(
    '_stacktrace',
    sources=['_stacktrace.c'],
)

setup(
    name='_stacktrace',
    version='1.0',
    ext_modules=[extension_module],
)
```

To check for leaks one possibility is using a debug version of the interpreter and print the total reference count bere and after:

```
import _stacktrace as s
import sys, gc

for _ in range(100):
    gc.collect()
    a = sys.gettotalrefcount()
    s.get_info_frame(b"")
    gc.collect()
    print(sys.gettotalrefcount() -a)
```

This prints:

```
7
7
7
7
```

because there are 7 references being leaked (at least).

The other is using `memray` or another memory profieler. Use this drive script:

```
# Test script
import _stacktrace as s
import gc

for _ in range(1_000_000):
    s.get_info_frame(b"")

gc.collect()
```

Then use memray:

```
$ memray run --trace-python-allocators --native -o lel.bin -f test.py
$ python -m memray flamegraph lel.bin --leaks -f
```

leaks

Note that frame objects do not appear in `memray` because we only leak ~7 frames and they use a freelist:

https://github.com/python/cpython/blob/97ce15c5f8743fb8b8967c6a05d6aec9fef9cbc7/Objects/frameobject.c#L810

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


pablogsal:
CC: @avara1986 @gnufede as this was introduced in https://github.com/DataDog/dd-trace-py/pull/4360

ddtrace/appsec/_iast/_stacktrace.c
```diff
@@ -39,68 +51,82 @@
...
     int line;
+    PyObject* filename_o = NULL;
+    PyObject* result = NULL;
+    PyObject *cwd_bytes = NULL;
+    char* cwd = NULL;
 
-    PyObject *cwd_obj = Py_None, *cwd_bytes;
-    char* cwd;
-    if (!PyArg_ParseTuple(args, "O", &cwd_obj))
+    if (!PyUnicode_FSConverter(cwd_obj, &cwd_bytes)) {
```
pablogsal, line 67: Change this if you want to return `None` for some reason in casee you pass something invalid. Right now it shows:

```
>>> import _stacktrace as s
>>> s.get_info_frame(34.4)
TypeError: expected str, bytes or os.PathLike object, not float
```

ddtrace/appsec/_iast/_stacktrace.c
P403n1x87, line 1: Just a note, can be addressed in a follow-up PR.

Can we base the preprocessor version checks on the `PY_VERSION_HEX` variable instead, e.g.
`#if PY_VERSION_HEX >= 0x030b0000` instead of `#if PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION >= 11`

ddtrace/appsec/_iast/_stacktrace.c
```diff
@@ -39,68 +51,82 @@
...
+    { "get_info_frame", (PyCFunction)get_file_and_line, METH_O, "stacktrace functions" },
     { NULL, NULL, 0, NULL }
 };
 
 static struct PyModuleDef stacktrace = { PyModuleDef_HEAD_INIT,
                                          "ddtrace.appsec._iast._stacktrace",
                                          "stacktrace module",
                                          -1,
-                                         StacktraceMethods };
+                                         StacktraceMethods};
```
P403n1x87, line 124:
```suggestion
                                         StacktraceMethods };
```
unrelated change?

ddtrace/appsec/_iast/_stacktrace.c
```diff
@@ -39,68 +51,82 @@
...
 
 PyMODINIT_FUNC
 PyInit__stacktrace(void)
 {
-    PyObject* m;
-    m = PyModule_Create(&stacktrace);
+    PyObject* m = PyModule_Create(&stacktrace);
```
P403n1x87, line 129: unrelated change?

P403n1x87 (commented):
Thanks, @pablogsal! I've left a few comments with nits and a note with a suggestion for further coding improvements.

ddtrace/appsec/_iast/_stacktrace.c
juanjux: I find the later more readable to be honest but I guess the former is faster.

juanjux:
LGTM once @P403n1x87 nits are addressed and thanks!

ddtrace/appsec/_iast/_stacktrace.c
P403n1x87: I agree on readability, but _technically_ the condition is not correct (in fact it is, if we consider that we dropped support for Python 2 on 2.x, and if we assume that there will never be a Python 4 🙂 ).

ddtrace/appsec/_iast/_stacktrace.c
pablogsal: Technically is unrelated but I added it as cleanup s I was already changing most of the code. I can revert if you want.

avara1986 (dismissed):
LGTM, thank you so much, @pablogsal ! I learned many things from your PR! I will add `memray` to our tests ASAP :heart: :+1:

ddtrace/appsec/_iast/_stacktrace.c
avara1986: You need to run this before committing your changes:

```bash
clang-format --style="{BasedOnStyle: Mozilla, IndentWidth: 4, ColumnLimit: 120}" -i ddtrace/appsec/_iast/_stacktrace.c
```

Alternatively, you can use pre-commit 🙃

pablogsal: Fixed!

pablogsal:
Code updated to address feedback. Please check this small change I made for readability:

https://github.com/DataDog/dd-trace-py/pull/7112#discussion_r1342770811

Feel free to revert the commit if you prefer the other version

avara1986 (approved):

pablogsal:
Wow, the CI seems to be quite flaky 😟 Is there something I can do in this PR to help?

P403n1x87:
> Wow, the CI seems to be quite flaky 😟 Is there something I can do in this PR to help?

Thanks for offering to put up with our CI 🙂.  I think the Windows build issues are being addressed in #7126. As for some of the other failures, I think we expect some framework tests to fail at the moment, and we might also have some quirks about PRs from external contributors. So as long as all the required jobs pass I think we're good 🤞.

pablogsal:
> Thanks for offering to put up with our CI 🙂. I think the Windows build issues are being addressed in https://github.com/DataDog/dd-trace-py/pull/7126. As for some of the other failures, I think we expect some framework tests to fail at the moment, and we might also have some quirks about PRs from external contributors. So as long as all the required jobs pass I think we're good 🤞.

Cool! I was wondering if something we did here was affecting the other tests but it was difficult for me to properly asses it :)

P403n1x87:
Ah I see we have the no-changelog label on, but we should probably add a release note since we're fixing a problem here. We use `reno` for maintaining the changelog, so e.g.
```
pipx install reno
reno new fix-iast-mem-leak
```
and a brief description of what the fix does under the `fixes` section would be grand.

avara1986 (approved):

gnufede (approved):

PR #7113, opened 2023-10-01
feat(tracing): elasticsearch 8 and async support

chadawagner:
## Description

Adds support for async elasticsearch clients and version 8

## Checklist

## Reviewer Checklist

- [ ] Title is accurate.


riotfile.py
```diff
@@ -711,20 +711,19 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
...
+                            latest,
                         ]
                     },
                 ),
-                Venv(pys=select_pys(), pkgs={"elasticsearch1": ["~=1.10.0"]}),
-                Venv(pys=select_pys(), pkgs={"elasticsearch2": ["~=2.5.0"]}),
-                Venv(pys=select_pys(), pkgs={"elasticsearch5": ["~=5.5.0"]}),
-                Venv(pys=select_pys(), pkgs={"elasticsearch6": ["~=6.8.0", latest]}),
-                Venv(pys=select_pys(), pkgs={"elasticsearch7": ["~=7.11.0"]}),
+                Venv(pys=select_pys(), pkgs={"elasticsearch1": latest}),
```
emmettbutler, line 721: These versions should probably stay pinned to the earliest supported version in accordance with the [compatibility policy](https://github.com/DataDog/dd-trace-py/blob/46466889078eea0b7cf454b06f7456ee6be37c7a/docs/contributing-integrations.rst#what-versions-of-a-given-library-should-an-integration-support).

releasenotes/notes/feat-elasticsearch-async-1907c207f5e435ca.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    elasticsearch: This introduces tracing for elasticsearch 8 and async client.
```
emmettbutler, line 4:
```suggestion
    elasticsearch: This introduces tracing for Elasticsearch 8 and the async Elasticsearch client.
```

docs/index.rst
```diff
@@ -85,7 +85,7 @@ contacting support.
 +--------------------------------------------------+---------------+----------------+
 | :ref:`dogpile.cache`                             | >= 0.6        | Yes            |
 +--------------------------------------------------+---------------+----------------+
-| :ref:`elasticsearch`                             | >= 1.6, < 8.0 | Yes            |
```
emmettbutler, line 88: This change should be reflected in the [public corporate documentation](https://github.com/DataDog/documentation/blob/9b9442134a92f1be867987ff4fe8724b9fdc29b4/content/en/tracing/trace_collection/compatibility/python.md).

docker-compose.yml
```diff
@@ -3,8 +3,10 @@ version: "3"
 
 services:
     elasticsearch:
-        image: elasticsearch:2.4-alpine
-        platform: linux/amd64
+        image: elasticsearch:7.17.13
```
emmettbutler, line 6: Similarly to my comment on `riotfile.py`, I think this pin needs to remain as old as possible to align with the compatibility policy. To test against ES8, you might consider adding an `elasticsearch8` service to the docker-compose file.

ddtrace/_monkey.py
```diff
@@ -110,11 +110,13 @@
 _MODULES_FOR_CONTRIB = {
     "elasticsearch": (
         "elasticsearch",
+        "elasticsearch1",
         "elasticsearch2",
         "elasticsearch5",
         "elasticsearch6",
         "elasticsearch7",
         "opensearchpy",
+        "elastic_transport",
```
emmettbutler, line 119: Why isn't elasticsearch8 included here? An explanatory comment could help with the reader's understanding.

emmettbutler (changes requested):
Thank you for the in-depth contribution. These changes are looking great! The main changes I'm requesting are related to the [version compatibility policy](https://github.com/DataDog/dd-trace-py/blob/46466889078eea0b7cf454b06f7456ee6be37c7a/docs/contributing-integrations.rst#what-versions-of-a-given-library-should-an-integration-support).

Let me know how else I can help!

riotfile.py
chadawagner: That's exactly what I based the changes here on, did I misread it? 😅

```
For each third-party library ddtrace provides integrations for:
- Test the latest major version. Specifically, test the min-max of the latest major version (the oldest and newest minor releases from the major version) going back 2 years.
- Test the latest minor release of non-latest major versions going back 2 years.
- If no new releases within the past 2 years, test the latest version.
- If support for legacy Python versions (2.7, 3.5, 3.6) was dropped in a major release, then test the latest minor release to support the legacy Python version only.
```

which I interpreted as:

- min-max of 8 (`8.0.0` has a bug that fails on a sort like `name:desc`, so I skipped that with a note)
- for 7, `7.15` was released in Sep 2021
- for others, test the latest version

I wasn't sure whether I should consider those versioned modules as separate integrations when applying these criteria? i.e. test min-max of all of those? Sounds like that's what you're saying, if so I should test min _and_ max of each of those, not just min.

Elasticsearch also has the fun quirk that `7.14+` will refuse to connect to unofficial server versions like AWS OpenSearch, so likely a lot of folks are pinned to `7.13.4`. If we're adding more versions to test we should probably include that one.

docker-compose.yml
chadawagner: Elasticsearch 2 was EOL in 2018, and `7.17` is currently the oldest non-EOL version. I wouldn't have touched this except clients beyond `7.13` will reject connections to older servers as not being recognized as "official".

As you said, we could add new/multiple services to docker-compose, if you feel it's necessary, but my thought was that moving to the oldest compatible version would be sufficient for coverage and preferable for CI workload

ddtrace/_monkey.py
chadawagner: Comment added

Starting with version 8 they moved the transport to a separate package, so we patch only that and nothing actually in the elasticsearch module itself.

riotfile.py
chadawagner: For most of these it turns out `latest` was already a synonym for the previous requirement, but I restored to what they were. Added `7.13` for the pinned folks out there

docker-compose.yml
chadawagner: @emmettbutler I pushed a new commit which runs tests against both es2 and es7 servers. I still feel like this is not needed, it doesn't really increase coverage of anything we're testing with the integration, and all we're really testing is whether the tests themselves are sending what the server expects (and for a 5-years-EOL'd server). In fact I don't even think we need the `opensearch` image some tests run against, those can just be run against es. But not my repo, so your call

docs/index.rst
chadawagner: https://github.com/DataDog/documentation/pull/20075
That's a beast of a fork to change one line (so I made it two) 😅

emmettbutler (changes requested):
Looks good to me. With the changes you suggested to the docker compose file, I think this is mergeable.

docker-compose.yml
chadawagner: @emmettbutler Ok no worries, I'll revert that part then. Thanks!

chadawagner:
@emmettbutler good to go when CI passes, thanks!

emmettbutler (approved):

chadawagner:
@emmettbutler ah shoot, hold on. It looks like the async spans are closing immediately, let me see if there's an easy fix or if this approach is not going to work 😞

chadawagner:
@emmettbutler ok spans are correctly spanning now, should be good to go. I'm not sure why some unrelated tests seem to be failing, is there a way to re-run those? I can also rebase and force push to see if that clears anything up

emmettbutler (approved):

chadawagner:
@emmettbutler thanks for approving! Do these required "Expected — Waiting for status to be reported" checks happen automatically eventually or does something need to kick them off?

chadawagner:
> @emmettbutler thanks for approving! Do these required "Expected — Waiting for status to be reported" checks happen automatically eventually or does something need to kick them off?

Ah ok, I see "This workflow requires approval from a maintainer," which I think you already did. Sorry, I saw some test failures earlier and a warning that the branch was out of date, so I rebased, which I guess must have reset your workflow approval. When you get a chance could you please kick that off again? Thanks!

antyser:
Hi, I upgrade to the latest ddtrace version 2.1.5 but I still cannot get the elasticserach 8 working. Is this matched to the latest ddtrace version in https://pypi.org/project/ddtrace/ ? Thank you.

chadawagner:
@antyser This PR was merged a couple days after the 2.1.0 rc was cut, so I assume it will go out with 2.2. @emmettbutler might know when that is expected to be

chadawagner:
(that was about a month ago, and past cadence seems like minor versions every couple weeks, so I hope soon!)

emmettbutler:
There will probably be another minor version release in the next two weeks, according to current plans.

PR #7125, opened 2023-10-02
feat(unittest): add ITR support

ericlaz:
## Changes

This PR aims to add ITR support for the `unittest` integration, specifically:
* Adds per test coverage data to test spans
* Fetches test to skip and skips tests thanks to ITR
* Allows users to add unskippable tests with the `skipIf` decorator

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


emmettbutler (changes requested):
I'll wait to do a proper review until there's a pull request description

ddtrace/contrib/unittest/patch.py
```diff
@@ -56,11 +63,67 @@ def _enable_unittest_if_not_started():
...
+def _is_skipped_test(test_object) -> bool:
+    testMethod = getattr(test_object, test_object._testMethodName, "")
+    return (
+        (hasattr(test_object.__class__, "__unittest_skip__") and test_object.__class__.__unittest_skip__)
+        or (hasattr(testMethod, "__unittest_skip__") and testMethod.__unittest_skip__)
+        or (hasattr(test_object, "_dd_itr_skip") and test_object._dd_itr_skip)
+    )
+
+
+def is_skipped_by_itr(test_object) -> bool:
```
romainkomorndatadog, line 112: Probably keep this as `_is_skipped_by_itr` to keep it an internal API.

ddtrace/contrib/unittest/patch.py
```diff
@@ -481,7 +636,19 @@ def _start_test_session_span(instance) -> ddtrace.Span:
...
+        test.ITR_TEST_CODE_COVERAGE_ENABLED,
+        "true" if _CIVisibility._instance._collect_coverage_enabled else "false",
+    )
+    if _CIVisibility.test_skipping_enabled():
+        test_session_span.set_tag_str(test.ITR_TEST_SKIPPING_ENABLED, "true")
+        test_session_span.set_tag(test.ITR_TEST_SKIPPING_TYPE, TEST)
+        test_session_span.set_tag(test.ITR_TEST_SKIPPING_TESTS_SKIPPED, "false")
+        test_session_span.set_tag(test.ITR_DD_CI_ITR_TESTS_SKIPPED, "false")
+        test_session_span.set_tag_str(test.ITR_FORCED_RUN, "false")
+        test_session_span.set_tag_str(test.ITR_UNSKIPPABLE, "false")
```
romainkomorndatadog, line 649: Why the differences between `set_tag` and `set_tag_str` ?

ddtrace/contrib/unittest/patch.py
```diff
@@ -516,6 +683,20 @@ def _start_test_module_span(instance) -> ddtrace.Span:
...
+        "true" if _CIVisibility._instance._collect_coverage_enabled else "false",
+    )
+    if _CIVisibility.test_skipping_enabled():
+        test_module_span.set_tag_str(test.ITR_TEST_SKIPPING_ENABLED, "true")
+        test_module_span.set_tag(test.ITR_TEST_SKIPPING_TYPE, TEST)
+        test_module_span.set_tag_str(test.ITR_TEST_SKIPPING_TESTS_SKIPPED, "false")
+        test_module_span.set_tag_str(test.ITR_DD_CI_ITR_TESTS_SKIPPED, "false")
+        test_module_span.set_tag_str(test.ITR_FORCED_RUN, "false")
+        test_module_span.set_tag_str(test.ITR_UNSKIPPABLE, "false")
+        test_module_span.set_metric(test.ITR_TEST_SKIPPING_COUNT, 0)
```
romainkomorndatadog, line 697: Feels like these `set_tag`/`set_tag_str` should match the ones above (eg: https://github.com/DataDog/dd-trace-py/pull/7125/files#diff-34d77150382cf4bb09fce56b5ae8364e0f3293de8791431903a983dc80f5e63fR646 uses `set_tag()` for `test.ITR_TEST_SKIPPING_TESTS_SKIPPED` while https://github.com/DataDog/dd-trace-py/pull/7125/files#diff-34d77150382cf4bb09fce56b5ae8364e0f3293de8791431903a983dc80f5e63fR693 uses `sset_tag_str()`).

tests/contrib/unittest_plugin/test_unittest_snapshot.py
romainkomorndatadog, line 1: I know this is a little bit out of scope but could you please add a (or modify an existing) test that uses custom configurations so we can see it in the spans?

ddtrace/contrib/unittest/patch.py
```diff
@@ -298,6 +373,19 @@ def _update_remaining_suites_and_modules(
         _finish_span(test_module_span)
 
 
+def _update_test_skipping_count_span(span: ddtrace):
```
romainkomorndatadog, line 376: Is this typehint right? 🧐

ddtrace/contrib/unittest/patch.py
```diff
@@ -423,13 +529,63 @@ def handle_test_wrapper(func, instance, args: tuple, kwargs: dict):
...
+                        span.set_tag_str(test.ITR_SKIPPED, "true")
+                        span.set_tag_str(test.SKIP_REASON, SKIPPED_BY_ITR_REASON)
+
+            if _is_test_coverage_enabled(instance):
+                coverage = _start_coverage(root_directory)
+                instance._coverage = coverage
+            if is_skipped_by_itr(instance):
+                result = args[0]
+                result.startTest(test=instance)
+                result.addSkip(test=instance, reason=SKIPPED_BY_ITR_REASON)
```
romainkomorndatadog, line 579: Do you know when `result` is ever `None` ? [TestCase.run()](https://github.com/python/cpython/blob/main/Lib/unittest/case.py#L599) and [its docs](https://docs.python.org/3/library/unittest.html#unittest.TestCase.run) don't say much.

I'm wondering if you also need to call [TestResult.stopTest()](https://docs.python.org/3/library/unittest.html#unittest.TestResult.stopTest) if you're going to manually run the test.

I think I like the way we do it in the `pytest` plugin where we just add the skip marker (with our reason) and then still let `pytest` "run" the test as if it were a regular one. It makes me feel a little less paranoid that we might forget to do something the underlying runner does on its own.

romainkomorndatadog (commented):
Looks pretty good to me. Nice job. :) 

I have a couple of questions in the comments, but overall this looks pretty mergeable to me.

ddtrace/contrib/unittest/patch.py
```diff
@@ -415,6 +504,22 @@ def add_xpass_test_wrapper(func, instance, args: tuple, kwargs: dict):
     return func(*args, **kwargs)
 
 
+def _mark_test_as_unskippable(obj):
+    test_name = obj.__name__
+    test_suite_name = str(obj).split(".")[0].split()[1]
+    test_module_path = os.path.relpath(obj.__code__.co_filename)
+    test_module_suite_name = _generate_module_suite_test_path(test_module_path, test_suite_name, test_name)
+    _initialize_unittest_data()
```
romainkomorndatadog, line 512: Is this necessary because the `skipIf` decorator's actually called when `unittest` is discovering tests?

Also, minor nit: if you're going to call `_initialize_unittest_data()` it no matter what, I'd move it to the top of the function.

ddtrace/contrib/unittest/patch.py
```diff
@@ -431,13 +536,64 @@ def handle_test_wrapper(func, instance, args: tuple, kwargs: dict):
...
+                ):
+                    global _global_skipped_elements
+                    _global_skipped_elements += 1
+
+                    test_module_span._metrics[test.ITR_TEST_SKIPPING_COUNT] += 1
+                    test_module_span.set_tag_str(test.ITR_TEST_SKIPPING_TESTS_SKIPPED, "true")
+                    test_module_span.set_tag_str(test.ITR_DD_CI_ITR_TESTS_SKIPPED, "true")
+
+                    test_session_span.set_tag_str(test.ITR_TEST_SKIPPING_TESTS_SKIPPED, "true")
+                    test_session_span.set_tag_str(test.ITR_DD_CI_ITR_TESTS_SKIPPED, "true")
```
romainkomorndatadog, line 569: These should only happen if the test is not unskippable.

ddtrace/contrib/unittest/patch.py
```diff
@@ -431,13 +536,64 @@ def handle_test_wrapper(func, instance, args: tuple, kwargs: dict):
...
+                        test_module_span.set_tag_str(test.ITR_FORCED_RUN, "true")
+                        test_session_span.set_tag_str(test.ITR_FORCED_RUN, "true")
+                    else:
+                        instance._dd_itr_skip = True
+                        span.set_tag_str(test.ITR_SKIPPED, "true")
+                        span.set_tag_str(test.SKIP_REASON, SKIPPED_BY_ITR_REASON)
+
+            if _is_test_coverage_enabled(instance):
+                coverage = _start_coverage(root_directory)
+                instance._coverage = coverage
```
romainkomorndatadog, line 582: We shouldn't run coverage for tests that ITR skips, so I'd move this to the `else` below.

romainkomorndatadog (approved):
LGTM. Thanks for doing all this. :)

ddtrace/contrib/unittest/patch.py
```diff
@@ -51,18 +58,73 @@ def get_version():
...
+def _is_test_coverage_enabled(test_object) -> bool:
+    return _CIVisibility._instance._collect_coverage_enabled and not _is_skipped_test(test_object)
+
+
+def _is_skipped_test(test_object) -> bool:
+    testMethod = getattr(test_object, test_object._testMethodName, "")
+    return (
+        (hasattr(test_object.__class__, "__unittest_skip__") and test_object.__class__.__unittest_skip__)
+        or (hasattr(testMethod, "__unittest_skip__") and testMethod.__unittest_skip__)
+        or (hasattr(test_object, "_dd_itr_skip") and test_object._dd_itr_skip)
```
emmettbutler, line 109:
```suggestion
        or _is_skipped_by_itr(test_object)
```

ddtrace/contrib/unittest/patch.py
```diff
@@ -431,13 +536,63 @@ def handle_test_wrapper(func, instance, args: tuple, kwargs: dict):
...
+                    len(args)
+                    and _CIVisibility._instance._should_skip_path(test_module_suite_path_without_extension, test_name)
+                    and not _is_skipped_test(instance)
+                ):
+                    if _is_marked_as_unskippable(instance):
+                        span.set_tag_str(test.ITR_FORCED_RUN, "true")
+                        test_module_span.set_tag_str(test.ITR_FORCED_RUN, "true")
+                        test_session_span.set_tag_str(test.ITR_FORCED_RUN, "true")
+                    else:
+                        global _global_skipped_elements
```
emmettbutler, line 566: The number of nested conditionals in this code makes it fairly difficult for me to think about what this deeply-nested block means. Consider using encapsulation to make this code less nested and more focused on a single level of abstraction.

ddtrace/contrib/unittest/patch.py
```diff
@@ -522,6 +690,20 @@ def _start_test_module_span(instance) -> ddtrace.Span:
...
+        test.ITR_TEST_CODE_COVERAGE_ENABLED,
+        "true" if _CIVisibility._instance._collect_coverage_enabled else "false",
+    )
+    if _CIVisibility.test_skipping_enabled():
+        test_module_span.set_tag_str(test.ITR_TEST_SKIPPING_ENABLED, "true")
+        test_module_span.set_tag_str(test.ITR_TEST_SKIPPING_TYPE, TEST)
+        test_module_span.set_tag_str(test.ITR_TEST_SKIPPING_TESTS_SKIPPED, "false")
+        test_module_span.set_tag_str(test.ITR_DD_CI_ITR_TESTS_SKIPPED, "false")
+        test_module_span.set_tag_str(test.ITR_FORCED_RUN, "false")
+        test_module_span.set_tag_str(test.ITR_UNSKIPPABLE, "false")
```
emmettbutler, line 703: The duplication of logic between this block and line 651 increases the likelihood of subtle bugs being introduced where one block is changed independently of the other. I recommend deduplicating this code.

emmettbutler (changes requested):
Great work! I made a few readability recommendations.

ddtrace/contrib/unittest/patch.py
ericlaz: Good suggestion, should've been addressed now

emmettbutler (approved):

PR #7179, opened 2023-10-05
fix(rcm): start error when remote config subscriber thread is stopped

avara1986:
When the remote config poller is trying to start a stopped subscriber, `Subscriber.is_running` is False (OK), but `Subscriber._th_worker` is a stopped thread (ERROR: the thread `_th_worker` must be equal to None).

This PR introduces:
- Validation that the thread is None before trying to start it.
- Setting it to None when calling `Subscriber.stop`. If not, an exception is raised when calling `Subscriber._th_worker.start()` on a stopped thread.
- Joining threads when stopping them in tests. We had a potential flaky test error because subscribers were not stopped correctly.

Extra ball: Removed an unnecessary loop on `_th_worker` (PeriodicThread).

## Checklist

## Reviewer Checklist

- [ ] Title is accurate.

### DIFF
--- a/ddtrace/internal/remoteconfig/worker.py
+++ b/ddtrace/internal/remoteconfig/worker.py
@@ -92,7 +92,7 @@ class RemoteConfigPoller(periodic.PeriodicService):
         log.debug("[%s][P: %s] Remote Config Poller fork. Starting Pubsub services", os.getpid(), os.getppid())
         self._client.renew_id()
         for pubsub in self._client.get_pubsubs():
-            pubsub.restart_subscriber()
+            pubsub.restart_subscriber(True)
 
     def _poll_data(self, test_tracer=None):
         """Force subscribers to poll new data. This function is only used in tests"""
@@ -124,7 +124,7 @@ class RemoteConfigPoller(periodic.PeriodicService):
         forksafe.unregister(self.start_subscribers)
         atexit.unregister(self.disable)
 
-        self.stop()
+        self.stop(join=join)
 
     def _stop_service(self, *args, **kwargs):
         # type: (...) -> None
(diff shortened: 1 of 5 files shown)


ddtrace/internal/remoteconfig/worker.py
```diff
@@ -92,7 +92,7 @@ def start_subscribers(self):
         log.debug("[%s][P: %s] Remote Config Poller fork. Starting Pubsub services", os.getpid(), os.getppid())
         self._client.renew_id()
         for pubsub in self._client.get_pubsubs():
-            pubsub.restart_subscriber()
+            pubsub.restart_subscriber(True)
```
P403n1x87, line 95: Since the old thread is not running there shouldn't be anything to join. Since we dropped support for Python 2 I don't think it is a problem here, but IIRC this used to be an issue with old interpreters. Just a note in case something weird surfaces.

ddtrace/internal/remoteconfig/worker.py
```diff
@@ -174,6 +174,7 @@ def __enter__(self):
 
     def __exit__(self, *args):
         # type: (...) -> None
+        self.stop_subscribers(True)
```
P403n1x87, line 177: Joining on process shutdown is likely to cause a delay, so I'd join here only if we really have to, to ensure a proper state throughout. The impact is most noticeable for applications that spawn short-lived processes, in which we might end up adding some latency.

gnufede (dismissed):

ddtrace/internal/remoteconfig/worker.py
avara1986: `self.stop_subscribers(True)` is needed, if not, the test fails. `__exit__` is a function that we use in tests only. :+1:

P403n1x87 (dismissed):

gnufede (approved):

emmettbutler (approved):

majorgreys (commented):
Can you explain the test changes? I am not sure where the old behavior has been fixed.

emmettbutler (approved):

PR #7197, opened 2023-10-06
chore(debugging): support more decorators

P403n1x87:
We add support for more decorators. In particular, we add support for common decorators, such as `@wraps`. We also add support for class decorators, and more general object decorators by inspecting the object fields. To avoid running into an unbound problem, we don't make the inference recursive and only stop at the first level.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/debugging/_function/discovery.py
```diff
@@ -135,23 +136,80 @@ def _undecorate(f, name, path):
...
+    q = deque([f])  # FIFO: use popleft and append
+
+    while q:
+        g = q.popleft()
+
+        # Look for a wrapped function
+        for attr in ("__wrapped__", "func"):
+            try:
+                wrapped = object.__getattribute__(g, attr)
+                if _isinstance(wrapped, FunctionType) and wrapped not in seen_functions:
```
shatzi, line 156: nit: this if logic is repeated multiple times. not sure if it would be cleaner to have in an helper function but there is also return in there... maybe there is a nice python way to do this. maybe not.

what I came up with is:
```
if isUnseenFunction(wrapped): 
    return wrapped if match(wrapped) else addToQueue(wrapped)
```
not sure if this any better.

ddtrace/debugging/_function/discovery.py
```diff
@@ -135,23 +136,80 @@ def _undecorate(f, name, path):
...
+                        return c
+                    q.append(c)
+                    seen_functions.add(c)
+
+        # Look for a function attribute (method decoration)
+        # DEV: We don't recurse over arbitrary objects. We stop at the first
+        # depth level.
+        try:
+            for v in object.__getattribute__(g, "__dict__").values():
+                if _isinstance(v, FunctionType) and v not in seen_functions and match(v):
```
shatzi, line 195: nit: return v if isUnseenAndMatch(v) or just return v if match(v) - not sure what checking seen_function is important here. it is because it is faster?

ddtrace/debugging/_function/discovery.py
```diff
@@ -135,23 +136,80 @@ def _undecorate(f, name, path):
...
+                wrapped = object.__getattribute__(g, attr)
+                if _isinstance(wrapped, FunctionType) and wrapped not in seen_functions:
+                    if match(wrapped):
+                        return wrapped
+                    q.append(wrapped)
+                    seen_functions.add(wrapped)
+            except AttributeError:
+                pass
+
+        if _isinstance(g, partial):
```
shatzi, line 164: add comment and maybe link to partial docs to match any other if branch here.

ddtrace/debugging/_function/discovery.py
```diff
@@ -135,23 +136,80 @@ def _undecorate(f, name, path):
...
+    if _isinstance(f, FunctionType) and match(f):
+        return f
+
+    seen_functions = {f}
+    q = deque([f])  # FIFO: use popleft and append
+
+    while q:
+        g = q.popleft()
+
+        # Look for a wrapped function
```
shatzi, line 152: maybe add a link to github or docs where we cam up with __wrapped__ or func.

tests/debugging/function/test_discovery.py
```diff
@@ -157,3 +158,27 @@ def f():
...
+    @Decorator
+    def f():
+        pass
+
+    code = _undecorate(f, name="f", path=abspath(__file__)).__code__
+    assert code.co_name == "f"
+    assert code.co_filename == abspath(__file__)
+
+
+def test_discovery_wrapped_decoration():
```
shatzi, line 177: should we also test for partial?

shatzi (dismissed):
Looks good

ddtrace/debugging/_function/discovery.py
P403n1x87: Since Python doesn't have inlining, I think it's best to be explicit and save a function call for simple expressions like this

tests/debugging/function/test_discovery.py
P403n1x87: `wraps` uses `partial` for its implementation
```
def wraps(wrapped,
          assigned = WRAPPER_ASSIGNMENTS,
          updated = WRAPPER_UPDATES):
    """Decorator factory to apply update_wrapper() to a wrapper function

       Returns a decorator that invokes update_wrapper() with the decorated
       function as the wrapper argument and the arguments to wraps() as the
       remaining arguments. Default arguments are as for update_wrapper().
       This is a convenience function to simplify applying partial() to
       update_wrapper().
    """
    return partial(update_wrapper, wrapped=wrapped,
                   assigned=assigned, updated=updated)
```

ddtrace/debugging/_function/discovery.py
P403n1x87: I'll expand the comment here as these things are generally implementation details and are not documented.

tests/debugging/function/test_discovery.py
shatzi: I see. this is cool. while it two birds in one stone situation here, I think we shouldn't depend on wraps implementation and have it explicit if wraps impl changed in the future. WDYT?

shatzi (approved):

tests/debugging/function/test_discovery.py
P403n1x87: I think it's hard for this implementation to change in a way that won't use `partial` in the future, but we can add a dedicated test case anyway.

PR #7221, opened 2023-10-11
feat(ci_visibility): add pytest hooks for module, suite, and test naming

romainkomorndatadog:
Introducing `pytest` hooks that users can override in `conftest.py` if they wish to alter the way modules, suites, and tests are named in the CI test visibility product.

The hooks are accompanied by a somewhat unsightly refactoring of some of the pytest plugin so that the `item` that is passed around in functions is a pytest Item, instead of sometimes being a module. Much of this code is expected to be refactored as part of an upcoming project to introduce a new API around the CI Visibility product, so the refactor in this change is minimal, only to support the introduction of this feature.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/contrib/pytest/newhooks.py
```diff
@@ -0,0 +1,29 @@
+"""pytest-ddtrace hooks.
```
emmettbutler, line 1: Including the word "new" in the module name raises the question of whether there are also "old" hooks somewhere that have maybe been deprecated. If there is a clear reason to use "new" here, I recommend stating it in the docstring. If not, I'd suggest removing the "new" naming.

ddtrace/contrib/pytest/newhooks.py
```diff
@@ -0,0 +1,29 @@
...
+Note that these names will affect th display and reporting of tests in the Datadog UI, as well as information stored
+the Intelligent Test Runner. Differing hook implementations may impact the behavior of Datadog CI Visibility products.
+"""
+
+import pytest
+
+
+@pytest.hookspec(firstresult=True)
+def pytest_ddtrace_get_item_module_name(item):
+    # type (Pytest.item) -> str
```
emmettbutler, line 16: Making these type comments into Py3 annotations would increase the library's interoperability with type-checking tools and reduce the risk of type-related confusion when working with this code.

ddtrace/contrib/pytest/plugin.py
```diff
@@ -756,3 +753,42 @@ def pytest_runtest_makereport(item, call):
...
+    return pytest_module_item.nodeid.rpartition("/")[0].replace("/", ".")
+
+
+@pytest.hookimpl(trylast=True)
+def pytest_ddtrace_get_item_suite_name(item):
+    """
+    Extract suite name from a `pytest.Item` instance.
+    If the module path doesn't exist, the suite path will be reported in full.
+    """
+    # breakpoint()
```
emmettbutler, line 781: This should be removed because it lacks context and might be confusing to readers.

emmettbutler (changes requested):

ddtrace/contrib/pytest/newhooks.py
romainkomorndatadog: I followed `pytest-xdist`'s naming scheme (and that's the plugin Pytest's own docs point to for "writing new hooks" in [their docs](https://docs.pytest.org/en/7.1.x/how-to/writing_hook_functions.html#declaring-new-hooks) . It's somewhat consistent with [other plugins](https://github.com/search?q=path%3A**%2Fnewhooks.py+NOT+xdist&type=code).

I think the point is that they're new hooks being introduced to `pytest` by the plugin. FWIW, I also found it a bit confusing, and followed more because it looked like convention than anything.

ddtrace/contrib/pytest/plugin.py
romainkomorndatadog: Hah. Foiled.

ddtrace/contrib/pytest/plugin.py
```diff
@@ -756,3 +753,42 @@ def pytest_runtest_makereport(item, call):
...
+    pytest_module_item = _find_pytest_item(item, pytest.Module)
```
romainkomorndatadog:
```suggestion
    """
    pytest_module_item = _find_pytest_item(item, pytest.Module)
```

romainkomorndatadog:
Rewrote this to stack on top of #7725 (instead of the other way around) so this can ship as a feature in `2.x`. The diff will look a little smaller once that merges.

emmettbutler (approved):

juan-fernandez (approved):

emmettbutler (approved):

PR #7259, opened 2023-10-16
feat(datastreams): add kafka payload size metric

tabgok:
Prior to this change we did not send information about kafka payload size.  After this change, the kafka payload size is included as a metric alongside with other datastream pathway metrics.

This change allows Datadog to calculate and display kafka payload size metrics.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


tests/contrib/kafka/test_kafka.py
```diff
@@ -287,12 +288,43 @@ def retry_until_not_none(factory):
...
+@pytest.mark.parametrize("key_and_length", [("test-key", 8), ("你".encode("utf-8"), 3), (b"t2", 2)])
+def test_data_streams_payload_size(
+    dsm_processor, deserializing_consumer, serializing_producer, kafka_topic, payload_and_length, key_and_length
+):
+    payload, payload_length = payload_and_length
+    key, key_length = key_and_length
+    expected_payload_size = float(payload_length + key_length)
+    expected_payload_size += 8  # to account for headers we add here
+    expected_payload_size += len(PROPAGATION_KEY)  # Add in header key length
+    expected_payload_size += 20  # to account for path header we add
```
emmettbutler, line 301: I think the decision to iteratively build up the expected payload size makes this code nicely self-documenting. One could go even further and turn the comments into constant names, though that's a matter of taste and not necessary.

ddtrace/internal/datastreams/kafka.py
```diff
@@ -11,15 +11,42 @@
...
 INT_TYPES = (int,)
+MESSAGE_ARG_POSITION = 1
+KEY_ARG_POSITION = 2
+KEY_KWARG_NAME = "key"
+
+
+def _calculate_byte_size(data):
+    if isinstance(data, str):
+        # We encode here to handle non-ascii characters
+        return len(data.encode("utf-8"))
```
emmettbutler, line 22: This will [raise an exception](https://docs.python.org/3/library/stdtypes.html#str.encode) on non-ascii characters. For the purpose of calculating payload size, maybe `errors='replace'` is appropriate.

ddtrace/internal/datastreams/processor.py
```diff
@@ -342,15 +344,20 @@ def new_pathway(self, now_sec=None):
...
+    def set_checkpoint(self, tags, now_sec=None, payload_size=0):
         if not now_sec:
             now_sec = time.time()
         if hasattr(self._current_context, "value"):
             ctx = self._current_context.value
         else:
             ctx = self.new_pathway()
             self._current_context.value = ctx
-        ctx.set_checkpoint(tags, now_sec=now_sec)
+        if "direction:out" in tags:
```
emmettbutler, line 355: Assuming that `tags` here means "span tags", this appears to be a coupling between the datastreams and tracing products. Storing the directionality of the message could be a good use case for `ddtrace.internal.core`.

emmettbutler (changes requested):
The only blocking comment is the one about `str.encode`. Nice work!

ddtrace/internal/datastreams/kafka.py
```diff
@@ -54,8 +81,18 @@ def dsm_kafka_message_consume(instance, message):
     topic = core.get_item("kafka_topic")
     group = instance._group_id
 
+    payload_size = 0
+    if hasattr(message, "len"):
+        # message.len() is only support for some versions of confluent_kafka
```
vandonr, line 86:
```suggestion
        # message.len() is only supported for some versions of confluent_kafka
```

ddtrace/internal/datastreams/processor.py
tabgok: in this case "tags" is separate - its a dependency on the tags for custom metrics being generated.  I will change the name to clarify!

ddtrace/internal/datastreams/kafka.py
tabgok: thank you!

tabgok: thanks! fixed

kr-igor (approved):
LGTM. Thank you for working on this!

emmettbutler (approved):

tests/contrib/kafka/test_kafka.py
```diff
@@ -16,6 +16,7 @@
 from ddtrace.contrib.kafka.patch import unpatch
 from ddtrace.filters import TraceFilter
 import ddtrace.internal.datastreams  # noqa: F401 - used as part of mock patching
+from ddtrace.internal.datastreams.kafka import PROPAGATION_KEY
```
rarguelloF, line 19: I'm personally not a big fan of using non-test constants in tests (it can lead to bugs - if for example  change the value for the constant, the tests won't fail).

ddtrace/internal/datastreams/processor.py
```diff
@@ -142,9 +143,9 @@ def __init__(self, agent_url, interval=None, timeout=1.0, retry_attempts=3):
         self.start()
 
     def on_checkpoint_creation(
-        self, hash_value, parent_hash, edge_tags, now_sec, edge_latency_sec, full_pathway_latency_sec
+        self, hash_value, parent_hash, edge_tags, now_sec, edge_latency_sec, full_pathway_latency_sec, payload_size=0
     ):
-        # type: (int, int, List[str], float, float, float) -> None
+        # type: (int, int, List[str], float, float, float, Optional[int]) -> None
```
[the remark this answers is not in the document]
vandonr: I think that path is taken for GRPC and HTTP calls for instance, for which we don't compute payload size for now

ddtrace/internal/datastreams/kafka.py
```diff
@@ -11,15 +11,42 @@
...
+
+    if isinstance(data, bytes):
+        return len(data)
+
+    if isinstance(data, dict):
+        total = 0
+        for k, v in data.items():
+            total += _calculate_byte_size(k)
+            total += _calculate_byte_size(v)
+        return total
```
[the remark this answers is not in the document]
vandonr: The way it's done in .net (which is kinda the reference) is like that, if it's not bytes or string (or a combination) we just give up and say the size is zero.
We could handle other types like integers for example, but we have to keep it in sync between tracers.

ddtrace/internal/datastreams/processor.py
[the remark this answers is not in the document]
tabgok: We're not computing this value for all contribs (rabbit mq, celery, etc) or for all protocols, so by making it optional it isn't a breaking change

ddtrace/internal/datastreams/kafka.py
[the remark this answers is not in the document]
tabgok: For kafka, specifically, the documentation and code enforce types; in these cases its only bytes and strings.  We'd need to worry for other contribs that don't strictly type for us

ddtrace/internal/datastreams/processor.py
```diff
@@ -342,15 +344,20 @@ def new_pathway(self, now_sec=None):
         ctx = DataStreamsCtx(self, 0, now_sec, now_sec)
         return ctx
 
-    def set_checkpoint(self, tags, now_sec=None):
+    def set_checkpoint(self, tags, now_sec=None, payload_size=0):
```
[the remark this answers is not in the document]
tabgok: keeps it from being a breaking change

tests/contrib/kafka/test_kafka.py
tabgok: In this case, if the value changes, I don't want the test to fail.  Perhaps this is a call for another test to ensure that it doesn't change for cross-version compatability?

PR #7263, opened 2023-10-17
fix(tracing): fix msgpack import error

mabdinur:
## Description

Reverts: DataDog/dd-trace-py#7099

- This change adds a performance regression to v0.5 and is no longer required.
- This change fixes an import error raised when `DD_TRACE_API` is set to `v0.5`. This occurs because msgpack is not installed by ddtrace.  

## Motivation
[Import msgpack](https://github.com/DataDog/dd-trace-py/blob/v2.1.1/ddtrace/internal/_encoding.pyx#L425) is problematic and breaks v0.5 encoding. To fix this issue we can ensure the ddtrace library installs msgpack (or we can vendor it) but this comes with it's own drawbacks. Instead of fixing the importerror this PR simply reverts the commit that introduced this import.

- [x] checkbox


ddtrace/internal/_encoding.pyx
```diff
@@ -866,75 +845,6 @@ cdef class MsgpackEncoderV05(MsgpackEncoderBase):
 
         return 0
 
-    cpdef _verify_encoding(self, encoded_bytes):
-        import msgpack
```
mabdinur, line 870: This is the problematic import. msgpack is not installed by ddtrace.

tests/utils.py
```diff
@@ -501,7 +501,7 @@ def __init__(self, *args, **kwargs):
         # original call
         if len(args) == 0 and "agent_url" not in kwargs:
             kwargs["agent_url"] = agent.get_trace_url()
-        kwargs["api_version"] = kwargs.get("api_version", "v0.5")
+        kwargs["api_version"] = kwargs.get("api_version", "v0.4")
```
mabdinur, line 504:
```suggestion
        kwargs["api_version"] = kwargs.get("api_version", "v0.5")
```

majorgreys (approved):

PR #7285, opened 2023-10-18
chore(debugging): snapshot pruning

P403n1x87:
We implement an algorithm to prune snapshot JSON messages after they have been serialised to ensure that they fit within the maximum log message size accepted by the event platform.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/debugging/_encoding.py
```diff
@@ -121,15 +123,110 @@ def _build_log_track_payload(
...
         # type: (str, Optional[str]) -> None
         self._service = service
         self._host = host
 
     def encode(self, log_signal):
         # type: (LogSignal) -> bytes
-        return json.dumps(_build_log_track_payload(self._service, log_signal, self._host)).encode("utf-8")
+        return self.pruned(json.dumps(_build_log_track_payload(self._service, log_signal, self._host))).encode("utf-8")
+
+    def pruned(self, log_signal_json: str) -> str:
```
shatzi, line 207: I think the json tree need to have more information on the tree structure to know which nodes can be removed and also at what cost.

for example:
```
"field": {
     type: "complexObject",
     fields: {  <-- can't prune here
         "a": { type:"int",value:42} <-- can be replaced with truncated node
         "b": { type:"object",isNull:true} <-- can be replaced, but what the point?
         "c": { type:"string", noCatpureReason:"depth"} <-- already not captured, so what the point replacing this one
     }
}
```
the main problem is that the fields: {...} can't be replaced, if we want we need to replace the entire object.
So, our algorithm need to check the "json" fields for each node and decided if that is a prunable node or not.

P403n1x87: The algorithm prunes just the leaf nodes, so the `fields` one won't be pruned. As for the other comments, the point is that we want to reduce the size 🙂 . `notCapturedReason`s are also contributing to the size of the final payload.

shatzi: Yup. I think at some point truncate only the leafs would not be enough, and you would need to remove a parent object. an example is a list of 100 items that all the items have noCaptureReason:"depth" is a good candidate to remove.

shatzi: While strings are caped to 256 bytes, I see your point they would be the preferred things to trim. However, we are spending too many bytes on {type:"something.something","notCapturedReason":"depth"} that add no value to the user. can we update our algorithm to prioritize prune those items before strings?

P403n1x87: > can we update our algorithm to prioritize prune those items before strings?

To do that we would need either make almost a feature-complete JSON parser, which is what we are trying to avoid, or at the very least check for a substring in each leaf node. In either cases we would be doing more work. My take is that we should evaluate whether the extra processing cost is worth the more ad-hoc pruning. My feeling is that, if we are above the 1 MB threshold, chances are we are likely to remove the large strings anyway.

shatzi (approved):

PR #7355, opened 2023-10-25
fix(profiler): remove ddup threading

sanchda:
This patch removes threading from the ddup extension (the wrapper over libdatadog).  Instead, it offers only a blocking `upload()` call.  This is aligned with the current strategy using the Python exporter, which temporarily (and harmlessly) blocks.

The GIL is toggled in _ddup.pyx, before/after the call into native code.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/internal/datadog/profiling/src/exporter.cpp
```diff
@@ -176,7 +194,38 @@ Uploader::set_runtime_id(std::string_view id)
...
+    bool ret = upload_impl(profile);
+    thread_working.store(false);
+    return ret;
+}
+
+void
 Uploader::upload(const Profile* profile)
+{
+    // If the upload thread is busy, give it some time.  We give it up to 3 seconds, which is arbitrary
+    wait_for_thread(3);
```
r1viollet, line 213: waiting here means we are at a new cycle and still have not finished uploading?
Would we stop profiling for 3 seconds here ?

r1viollet:
This is better! Thanks for looking into this.
Do we need to consider a special case for the last upload ?

ddtrace/internal/datadog/profiling/src/exporter.cpp
sanchda: Exports are scheduled separately from sampling, so this is actually quite complicated.  We'll block the caller, but sampling will proceed.  This means that samples will continue to accrue in the current buffer while the uploader is blocked.

r1viollet (approved):
LGTM

ddtrace/internal/datadog/profiling/include/exporter.hpp
```diff
@@ -6,16 +6,20 @@
...
+#include <optional>
 #include <string>
 #include <string_view>
+#include <thread>
 #include <unordered_map>
 #include <unordered_set>
 #include <vector>
 
-extern "C" {
+extern "C"
```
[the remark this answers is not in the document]
sanchda: @P403n1x87 Fixed!  Will push the style fixups as a separate PR--thanks

ddtrace/internal/datadog/profiling/src/exporter.cpp
```diff
@@ -176,7 +194,38 @@ Uploader::set_runtime_id(std::string_view id)
...
+        upload_thread->join();
+        upload_thread.reset();
+    }
+
+    // Launch the thread.  We mark the working state here, potentially
+    // before the thread even launches, but the fence at the top of this
+    // function should ensure the thread has been created by the time
+    // joinability is checked
+    thread_working.store(true);
+    upload_thread.emplace(&Uploader::thread_upload_impl, this, profile);
```
P403n1x87, line 224: Apologies if I take a step back here, but I believe the upload is already happening in a periodic service (hence a periodic thread). Why do we have the need to start another thread to do the actual upload? Can't we just release the GIL while the upload here takes place? Pardon me if this has already been discussed, but I don't have any recollection of this at the moment 🙁

sanchda: The short answer:  it's not a need at all.  I didn't see a strong reason to go one way or the other, so I chose this one because I thought it would lead to a cleaner design.

The long answer:
There were a few minor reasons, but I confess that they're somewhat flimsy.

* I thought it would be easier to understand the behavior of the library ("ddup") as a unit if its thread dispatch and scheduling were internal.  I'm personally willing to accept a little bit more local complexity in order to be able to draw more rigid lines between components.
* Since the periodic scheduler is responsible for reacting to on_shutdown signals, I had a (very minor) concern about maintaining reactivity.
* I didn't see a particularly strong proscription against the dispatch of async operations within the periodic scheduler (this is such a one).  I also figured that from the kernel work-scheduling perspective, it was a wash either way (the periodic scheduler is hopefully sleeping when we're serializing+uploading).
* I'd like to cook up a more competent solution to transient HTTP errors (double-buffering is... not that solution), but I wanted to be able to think of the problem in terms that would fall under a single semantic umbrella (language/runtime).

WDYT?  What are the important factors at play here?

P403n1x87: My feeling is that if we went with the standard "GIL release on I/O" approach we could save the extra logic required to keep note of the thread activity. This is already done on the Python side by the PeriodicService class anyway. So I feel it would make the code slightly easier to reason about.

sanchda: All right.  It's an easy enough thing to move around that imho there's no harm in doing it that way.  I think I've gotten most of that fix landed, testing now.

sanchda: Oops, forgot to release the GIL.  Fixing.

sanchda: OK, should be better in 88828b4

ddtrace/internal/datadog/profiling/include/exporter.hpp
```diff
@@ -9,6 +9,7 @@
 #include <chrono>
 #include <deque>
 #include <memory>
+#include <optional>
```
P403n1x87, line 12: I think we no longer have a use for this?

ddtrace/internal/datadog/profiling/include/exporter.hpp
```diff
@@ -76,6 +77,7 @@ class Uploader {
 
 public:
   Uploader(std::string_view _url, ddog_prof_Exporter *ddog_exporter);
+  ~Uploader();
```
P403n1x87, line 80: do we need this?

P403n1x87 (commented):
Neato! ✨

P403n1x87:
Ah I forgot to ask: we probably don't have an easy reproducer of the issue we are fixing that we can include to validate this change, and catch future regressions, correct?

ZStriker19:
Closing this for now, but feel free to reopen if we want to continue work on this!

ddtrace/internal/datadog/profiling/include/exporter.hpp
sanchda: Nope, the default is fine

majorgreys (approved):

Issue #7347, opened 2023-10-25
duplicate ddtrace logger events when not using ddtrace-run in 2.1.1


kravnaut:
### Summary of problem

It looks like in this PR https://github.com/DataDog/dd-trace-py/pull/6580 the was a line added that forces ddtrace logger to have  a stream handler: `ddtrace_logger.addHandler(logging.StreamHandler())` It seems that with that change, when not using `ddtrace-run` this leads to conflicts with the configured logger handlers. So if we don't explicitly configure ddtrace in our own logging config then we see ddtrace log twice - once with it's streamhandler and once through our configured handlers.

### Which version of dd-trace-py are you using?

2.1.1

### How can we reproduce your problem?

Run app using ddtrace without ddtrace-run

### What is the result that you get?

Duplicate logging out of ddtrace and child loggers into stderr and our standard logging handlers

### What is the result that you expected?

Expecting it to honor our logging config especially since it's configured prior to loading ddtrace modules

### Note

Did not really test much with ddtrace-run because we avoid using it

PR #7359, opened 2023-10-25
fix(logging): only add StreamHandler to logger if not already configured

ZStriker19:
https://github.com/DataDog/dd-trace-py/issues/7347

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/_logger.py
```diff
@@ -37,7 +37,9 @@ def configure_ddtrace_logger():
 
     """
     ddtrace_logger = logging.getLogger("ddtrace")
-    ddtrace_logger.addHandler(logging.StreamHandler())
+    # don't add a handler if already configured by user
+    if not ddtrace_logger.handlers:
+        ddtrace_logger.addHandler(logging.StreamHandler())
     _configure_ddtrace_debug_logger(ddtrace_logger)
     _configure_ddtrace_file_logger(ddtrace_logger)
```
brettlangdon, line 44: will we have the same issue with the file logger?

ddtrace/_logger.py
```diff
@@ -37,7 +37,9 @@ def configure_ddtrace_logger():
 
     """
     ddtrace_logger = logging.getLogger("ddtrace")
-    ddtrace_logger.addHandler(logging.StreamHandler())
+    # don't add a handler if already configured by user
+    if not ddtrace_logger.handlers:
```
brettlangdon, line 41: I wonder if we should do something more like:

```python
stream_handler = logging.StreamHandler()

def configure_ddtrace_logger():
    if stream_handler not in ddtrace_logger:
        ddtrace_logger.addHandler(stream_handler)
```


not 100% sure what is the "best" approach, but this was a consideration that came up when reviewing.

ZStriker19: but what if the user has already added a FileHandler or some other kind of handler?

brettlangdon: the different is... "is our specific StreamHandler already set" vs "is any StreamHandler set". we'd want the former.

e.g.

```python
>>> import logging
>>> s1 = logging.StreamHandler()
>>> s2 = logging.StreamHandler()
>>> s3 = logging.StreamHandler()
>>> handlers = [s1, s2]
>>> s1 in handlers
True
>>> s2 in handlers
True
>>> s3 in handlers
False
```

brettlangdon: but again... is that what we want? I am not sure what normal/best practices are, but seems like a possible option.

ZStriker19: Shouldn't we only care about just adding a handler if the customer hasn't added one so we make sure we log either way, but also make sure that if the customer configures ddtrace logging, we don't mess with it?

brettlangdon: > Shouldn't we only care about just adding a handler if the customer hasn't added one

I think I was thinking about it more like "do not configure/add our handlers more than once", but not sure I know if it should be "do not configure if any handlers have been added".

🤔

We easily can end up in a case where we configure both the StreamHandler and a RotatingFileHandler, since our configuration can do both.

ok, not sure I have a good sense for the right answer here.

ZStriker19: Will sync with team on thoughts.

ddtrace/_logger.py
ZStriker19: Yeah but if a customer doesn't explicitly set `DD_TRACE_LOG_FILE` ddtrace won't add the handler. 

So it's like: 
"I configured ddtrace logging, but then also wanted it to send to this file, so I added this envar. " - That makes sense to me. If you add an envar for logging, you expect it do something. It'd be confusing if configuring logging kept the envar from having an effect.

The other scenario I view as: "I configured ddtrace logging, and now it's also just adding its own StreamHandler() by default" - which is probably not what users would want if they already configured it.

releasenotes/notes/only-add-log-handler-when-not-already-configured-60cb797f5381a6ef.yaml
```diff
@@ -0,0 +1,8 @@
+---
+fixes:
```
ZStriker19, line 2: If we put the stream handler behind a flag that you can use to toggle it off, is it a feat or a fix?

ddtrace/_logger.py
ZStriker19: Added an envar to toggle it

emmettbutler (approved):

releasenotes/notes/only-add-log-handler-when-not-already-configured-60cb797f5381a6ef.yaml
emmettbutler: I think it's open to interpretation. Personally I think a "fix" is appropriate here because this change is in response to a bug report.

ddtrace/_logger.py
```diff
@@ -37,7 +37,9 @@ def configure_ddtrace_logger():
 
     """
     ddtrace_logger = logging.getLogger("ddtrace")
-    ddtrace_logger.addHandler(logging.StreamHandler())
+    if asbool(os.environ.get("DD_TRACE_STREAM_HANDLER", "true")):
```
brettlangdon, line 40: 🚲 🏠 ... should this name have `LOG` in it?

ZStriker19: sure!

PR #7413, opened 2023-10-30
fix(tracing): resolves concurrent dictionary modification error raised by the trace writer

### LINKED ISSUE #7411: AgentWriter thread gets killed due to concurrent dictionary modification
### Summary of problem
The `AgentWriter` thread that's responsible for flushing traces to the agent gets killed due to something modifying a dictionary while iterating over it [here](https://github.com/DataDog/dd-trace-py/blob/v1.18.6/ddtrace/internal/writer/writer.py#L456). Stack trace:

```python
Exception in thread ddtrace.internal.writer.writer:AgentWriter:
  File "/python-install/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "/mnt/kubernetes/sandbox/responsiveness-model-evaluator/lib/python3.10/site-packages/ddtrace/internal/periodic.py", line 56, in run
    self._target()
  File "/mnt/kubernetes/sandbox/responsiveness-model-evaluator/lib/python3.10/site-packages/ddtrace/internal/writer/writer.py", line 461, in periodic
    self.flush_queue(raise_exc=False)
  File "/mnt/kubernetes/sandbox/responsiveness-model-evaluator/lib/python3.10/site-packages/ddtrace/internal/writer/writer.py", line 415, in flush_queue
    self._flush_queue_with_client(client, raise_exc=raise_exc)
  File "/mnt/kubernetes/sandbox/responsiveness-model-evaluator/lib/python3.10/site-packages/ddtrace/internal/writer/writer.py", line 456, in _flush_queue_with_client
    for name, metric_tags in self._metrics.items():
RuntimeError: dictionary changed size during iteration
```

This causes traces to stop being sent to the agent, and the following warning message is logged over and over:

```python
2023-10-25 09:06:52,043 WARNING ddtrace.internal.writer.writer: trace buffer (53414 traces 8388538b/8388608b) cannot fit trace of size 157b, dropping (writer status: running) [in /mnt/kubernetes/sandbox/responsiveness-model-evaluator/lib/python3.10/site-packages/ddtrace/internal/writer/writer.py:396, 139702107621120 (ThreadPoolExecutor-3_2), 586 (MainProcess)]
```

1.18.6, but looking at the code the problem still seems to be there in the latest 2.x version.

I'm not sure how to reproduce it, but this keeps happening for several of our Python applications. The only way to fix it is to restart the application.

mabdinur:
## Description

Resolves #7411. Instead of batching health metrics in the trace writer this change submits health metrics directly to the dogstatsd client. With this refactor we no longer need to iterate over the `_metrics` dictionary.

Note: Although health metrics are now sent directly to statd client, this PR does NOT remove `self._metrics`. Instead it repurposes it so that it is only used to calculate the tracer keep rate (`tracer.kr`). The tracer keep rate is used by the agent to make sampling decisions.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/internal/writer/writer.py
```diff
@@ -420,14 +420,17 @@ def _flush_queue_with_client(self, client, raise_exc=False):
                 )
         finally:
             if config.health_metrics_enabled and self.dogstatsd:
+                # _metrics is updated when traces are queued.
+                # Here we get a copy of the metrics dictionary as early as possible
+                metrics = self._metrics.copy()
```
P403n1x87, line 425: Is copying better than guarding this data structure with a lock, in terms of performance? (I'm assuming the issue occurs because of concurrent access from different threads here).

mabdinur: We add values to the metrics dictionary multiple times each times per trace chunk (a few times when `tracer.flush()` is called and a few times when `tracer.write()` is called). Acquiring a lock every time `_metrics_dist` is invoked could come with a significant performance hit.

The current approach hides copying the dictionary behind a the `health_metrics_enabled` flag. This flag is disabled by default so most users will not get this performance. We also need a deep copy here since self._metrics has nested dictionaries. I made this fix in my latest commit.

ddtrace/internal/writer/writer.py
```diff
@@ -420,14 +421,17 @@ def _flush_queue_with_client(self, client, raise_exc=False):
                 )
         finally:
             if config.health_metrics_enabled and self.dogstatsd:
+                # _metrics is updated when traces are queued.
+                # Here we get a deep copy of the nested metrics dictionary as early as possible
+                metrics = copy.deepcopy(self._metrics)
```
Rfank2021, line 426: If after `deepcopy` there is change to `self._metircs`, those change will lose due to `self._metrics_reset()` in `self.flush_queue`?

romainkomorndatadog: Yeah, maybe it'd be better to `deepcopy` in `flush_queue`, immediately call `_metrics_reset()`, then pass the copied object to `_flush_queue_with_client()`.

That said, metrics are somewhat best-effort (eg: `self._metrics_reset()` is called in the `finally:` above, so we clear metrics no matter what).

Rfank2021: LGTM.

ddtrace/internal/writer/writer.py
P403n1x87: But `flush` would be called very rarely compared to `write`, so most of the time the lock would be acquired immediately. I have no feel for how big the metrics dict could be, so I wonder if the amortised cost of deepcopy would really be cheaper than locking 🤔

ddtrace/internal/writer/writer.py
```diff
@@ -420,14 +421,17 @@ def _flush_queue_with_client(self, client, raise_exc=False):
...
                 namespace = self.STATSD_NAMESPACE
                 # Note that we cannot use the batching functionality of dogstatsd because
                 # it's not thread-safe.
                 # https://github.com/DataDog/datadogpy/issues/439
                 # This really isn't ideal as now we're going to do a ton of socket calls.
                 self.dogstatsd.distribution("datadog.%s.http.sent.bytes" % namespace, len(encoded))
                 self.dogstatsd.distribution("datadog.%s.http.sent.traces" % namespace, n_traces)
```
mabdinur, line 433:
```suggestion
        finally:
              with self._conn_lck:
                    if config.health_metrics_enabled and self.dogstatsd:
                        # _metrics is updated when traces are queued.
                        namespace = self.STATSD_NAMESPACE
                        # Note that we cannot use the batching functionality of dogstatsd because
                        # it's not thread-safe.
                        # https://github.com/DataDog/datadogpy/issues/439
                        # This really isn't ideal as now we're going to do a ton of socket calls.
                        self.dogstatsd.distribution("datadog.%s.http.sent.bytes" % namespace, len(encoded))
                        self.dogstatsd.distribution("datadog.%s.http.sent.traces" % namespace, n_traces)
```

ddtrace/internal/writer/writer.py
```diff
@@ -420,14 +421,17 @@ def _flush_queue_with_client(self, client, raise_exc=False):
...
-                for name, metric_tags in self._metrics.items():
+                for name, metric_tags in metrics.items():
                     for tags, count in metric_tags.items():
                         self.dogstatsd.distribution("datadog.%s.%s" % (namespace, name), count, tags=list(tags))
```
mabdinur, line 436:
```suggestion
                for name, metric_tags in list(metrics.items()):
                    for tags, count in list(metric_tags.items()):
                        self.dogstatsd.distribution("datadog.%s.%s" % (namespace, name), count, tags=list(tags))
```

ddtrace/internal/writer/writer.py
majorgreys: Are we, in effect, implementing a thread-safe batching for dogstatsd (which at the time when this code was added had an open issue)? 

This issue looks to be resolved: https://github.com/DataDog/datadogpy/issues/439

We should remove the metrics batching currently implemented in the `AgentWriter`.

mabdinur: So instead of batching metrics we should call `self.dogstatsd.distribution(....)` directly?

majorgreys: @mabdinur right

majorgreys: cc @Kyle-Verhoog and @brettlangdon who can also speak to the choice to batch metrics within the `AgentWriter`.

brettlangdon: > We add values to the metrics dictionary multiple times each times per trace chunk (a few times when tracer.flush() is called and a few times when tracer.write() is called). Acquiring a lock every time _metrics_dist is invoked could come with a significant performance hit.

I think this is correct, iirc. Performance improvement, take the big hit once at time we flush metrics, rather than a lock every time we create them.

mabdinur: I updated the PR description. `TracerWriter._metrics` is no longer used to batch health metrics. It is now only used to calculate `tracer_kr` (removing `TracerWriter._metrics` will require a significant refactor to how we calculate `tracer_kr`).

releasenotes/notes/fix-writer-concurrent-dictionary-modification-d37e2f918c51578c.yaml
```diff
@@ -0,0 +1,4 @@
+---
+fixes:
+  - |
+    tracing: Fixes an issue where the thread responsible to sending traces is killed due to concurrent dictionary modification.
```
emmettbutler, line 4:
```suggestion
    tracing: Fixes an issue where the thread responsible for sending traces is killed due to concurrent dictionary modification.
```

emmettbutler (approved):

PR #7434, opened 2023-10-31
feat(kafka): trace propagation for kafka consumer/producer spans

wconti27:
## Description
This PR adds trace context propagation for Kafka. The behavior is hidden behind a feature flag and is disabled by default to prevent any unexpected changes to users' kafka tracing

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


tests/contrib/kafka/test_kafka.py
```diff
@@ -666,38 +666,111 @@ def test_data_streams_default_context_propagation(dummy_tracer, consumer, produc
...
 
     # None of these spans are part of the same trace
-    assert produce_span.trace_id != consume_span1.trace_id
-    assert consume_span1.trace_id != consume_span2.trace_id
+    assert produce_span.trace_id != consume_span.trace_id
+
+
+# Propagation should work when enabled
+def test_tracing_context_is_propagated_when_enabled(ddtrace_run_python_code_in_subprocess):
+    code = """
```
emmettbutler, line 710: Using [this decorator](https://github.com/DataDog/dd-trace-py/blob/5cb0b45c7179c21f203bff2ebe4008a7b112d0a3/tests/contrib/loguru/test_loguru_logging.py#L69) means you won't have to put such a large amount of Python code into a string.

emmettbutler (changes requested):
The code looks good. I recommend adding a release note and using the `subprocess` decorator on the test.

tests/contrib/kafka/test_kafka.py
wconti27: I recommitted with test refactor with the decorator. However the tests are now failing due to connection errors from kafka being printed to stdout. @mabdinur recommended to just use the original approach and keep the code within a string so that I am able to control the assertions.

ddtrace/contrib/kafka/patch.py
```diff
@@ -159,6 +164,17 @@ def traced_produce(func, instance, args, kwargs):
...
             span.set_tag(ANALYTICS_SAMPLE_RATE_KEY, rate)
+
+        if config.kafka.distributed_tracing_enabled:
+            # inject headers with Datadog tags:
+            try:
+                Propagator.inject(span.context, kwargs["headers"])
+            except KeyError:
+                try:
+                    Propagator.inject(span.context, args[6])
+                except IndexError:
```
mabdinur, line 175: In what scenarios are both key and index errors raised? Can we use `get_argument_value` and `set_argument_value here: https://github.com/DataDog/dd-trace-py/blob/fd1c9e12478089b91db24a17ea2efa994feba5ac/ddtrace/internal/utils/__init__.py#L17?

ddtrace/contrib/kafka/patch.py
```diff
@@ -167,12 +183,16 @@ def traced_poll(func, instance, args, kwargs):
     if not pin or not pin.enabled():
         return func(*args, **kwargs)
 
-    with pin.tracer.trace(
-        schematize_messaging_operation(kafkax.CONSUME, provider="kafka", direction=SpanDirection.PROCESSING),
+    message = func(*args, **kwargs)
```
mabdinur, line 186: Do we want to move the execution of the poll method (`func(*args, **kwargs)`) outside of the span? 

If we do then the duration of the span will not be accurate

tests/contrib/kafka/test_kafka.py
mabdinur: This test is reliant on pytest fixtures and pytest fixtures can not be used with the `@pytest.mark.subprocess(..)`. This is something we can improve in our testing framework.  

This test is less than ideal but works for now. Maybe another approach could be to simplify this test to mock the producer and consumer. With this approach we no longer need the `consumer, producer, kafka_topic` fixtures.

wconti27: IMO mocking the producer / consumer is not ideal since we want to test the propagation is working correctly.

ddtrace/contrib/kafka/patch.py
```diff
@@ -159,6 +166,15 @@ def traced_produce(func, instance, args, kwargs):
         rate = config.kafka.get_analytics_sample_rate()
         if rate is not None:
             span.set_tag(ANALYTICS_SAMPLE_RATE_KEY, rate)
+
+        # inject headers with Datadog tags if trace propagation is enabled
+        if config.kafka.distributed_tracing_enabled:
+            # inject headers with Datadog tags:
+            headers = get_argument_value(args, kwargs, 6, "headers", True)
+            if headers is None:
+                headers = {}
```
mabdinur, line 175: nit:
```suggestion
            headers = get_argument_value(args, kwargs, 6, "headers", True) or {}
```

ddtrace/contrib/kafka/patch.py
```diff
@@ -167,12 +183,28 @@ def traced_poll(func, instance, args, kwargs):
...
+    # we must get start time now since execute before starting a span in order to get distributed context
+    # if it exists
+    start_ns = time_ns()
+    # wrap in a try catch and raise exception after span is started
+    err = None
+    try:
+        message = func(*args, **kwargs)
+    except Exception as e:
+        err = e
+        pass
```
mabdinur, line 195:
```suggestion
```

mabdinur (approved):

ddtrace/contrib/kafka/patch.py
wconti27: We are recalculating the start time and manually setting this attribute on the span to make it accurate

ddtrace/contrib/kafka/patch.py
```diff
@@ -166,12 +180,28 @@ def traced_poll(func, instance, args, kwargs):
...
+    # if it exists
+    start_ns = time_ns()
+    # wrap in a try catch and raise exception after span is started
+    err = None
+    try:
+        message = func(*args, **kwargs)
+    except Exception as e:
+        err = e
+    ctx = None
+    if message is not None and config.kafka.distributed_tracing_enabled and message.headers() is not None:
```
tabgok, line 193: This causes us to ignore empty polls, correct?

tabgok: nit (style): you can say`and not message.headers()` instead

ddtrace/contrib/kafka/patch.py
```diff
@@ -166,12 +180,28 @@ def traced_poll(func, instance, args, kwargs):
...
+        ctx = Propagator.extract(dict(message.headers()))
```
tabgok, line 194: do we need to put this inside a try/except clause?

releasenotes/notes/add-kafka-distributed-tracing-150a31a858c2f7c5.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
```
tabgok, line 2: We're making some more hardware here - before we release this as a "feature", lets talk with Tracers, DSM, and APM eng/product.

tests/contrib/kafka/test_kafka.py
```diff
@@ -675,42 +675,114 @@ def test_data_streams_default_context_propagation(dummy_tracer, consumer, produc
 
 
 # It is not currently expected for kafka produce and consume spans to connect in a trace
-def test_tracing_context_is_not_propagated(dummy_tracer, consumer, producer, kafka_topic):
+def test_tracing_context_is_not_propagated_by_default(dummy_tracer, consumer, producer, kafka_topic):
```
tabgok, line 678: Should we be using versions?

tabgok (changes requested):
Minor nits and a few questions, from a code perspective.

Two major requests:
1. Lets review with APM, DSM, and Tracer Client libraries before pushing (I can take on pushing this through)
2. Lets define some language/integration-agnostic tests that specify the behavior we want (in another repo) prior to submitting these changes

emmettbutler (commented):
I'm going to wait on @tabgok's requested changes before doing another review. Please let me know if you'd like one sooner.

tabgok (approved):

ddtrace/contrib/kafka/patch.py
wconti27: We don't wrap these `extracts` anywhere else within a try except so it should be ok

ddtrace/contrib/kafka/patch.py
wconti27: Atleast for creating a distributed context, this will ignore empty polls. The span will still be created tho

ddtrace/contrib/kafka/patch.py
```diff
@@ -37,6 +42,7 @@
     "kafka",
     dict(
         _default_service=schematize_service_name("kafka"),
+        distributed_tracing_enabled=asbool(os.getenv("DD_KAFKA_DISTRIBUTED_TRACING_ENABLED", default=False)),
```
tabgok, line 45: DD_KAFKA_TRACING_ENABLED

ddtrace/contrib/kafka/patch.py
```diff
@@ -185,12 +198,28 @@ def traced_poll(func, instance, args, kwargs):
...
+    # if it exists
+    start_ns = time_ns()
+    # wrap in a try catch and raise exception after span is started
+    err = None
+    try:
+        message = func(*args, **kwargs)
+    except Exception as e:
+        err = e
+    ctx = None
+    if message and config.kafka.distributed_tracing_enabled and message.headers():
```
tabgok, line 211: Let's add in the "empty poll" toggle here as well

ddtrace/contrib/kafka/patch.py
wconti27: DD_KAFKA_PROPAGATION_ENABLED but 👍

ddtrace/contrib/kafka/patch.py
wconti27: Done

PR #7506, opened 2023-11-07
docs: change patch_all recommendations to ddtrace.auto where appropriate

emmettbutler:
This change replaces most documented recommendations to use `patch_all` to instead reference `import ddtrace.auto`. The latter approach is the preferred way to install instrumentation without `ddtrace-run`, because it uses the same code path as `ddtrace-run`, thus minimizing the possibility of different behavior.

This change also adds a small docstring to `ddtrace.auto` that can be used as a link target in generated documentation.

Other changes, related to `hatch run docs:build` in the testrunner container:
* Removes `Tracer` type comments that cause builds to fail
* Adds `..` to `sys.path`, enabling sphinx to import ddtrace
* Adds a fallback installation of `libenchant-dev` to the docs installation step
* Enables verbose sphinx logging
* Disables automatic instrumentation
* Installs all library dependencies

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/auto.py
```diff
@@ -1 +1,21 @@
+"""
+.. _ddtraceauto:
+
+Importing ``ddtrace.auto`` installs Datadog instrumentation in the runtime. It should be used
+when :ref:`ddtrace-run<ddtracerun>` is not an option. Using it with :ref:`ddtrace-run<ddtracerun>`
+is unsupported and may lead to undefined behavior::
```
brettlangdon, line 6: 🤔 interesting..... I don't think I've given this much thought before, and I think I would have assumed this "just worked"... but....

`ddtrace-run` loads `ddtrace.bootstrap.sitecustomize` as the `sitecustomize` module, but then `ddtrace.auto` loads it as `ddtrace.bootstrap.sitecustomize` so it gets reloaded since the `sys.module` cache has `sitecustomize` not `ddtrace.bootstrap.sitecustomize`.

I mean, this makes complete sense, but I hadn't stopped to think about it before.

```python
❯ ddtrace-run python
Python 3.11.0 (main, Feb 22 2023, 15:34:50) [Clang 14.0.0 (clang-1400.0.29.202)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> sys.modules["sitecustomize"]
<module 'sitecustomize' from '/Users/brett.langdon/datadog/dd-trace-py/ddtrace/bootstrap/sitecustomize.py'>
>>> import ddtrace.auto
>>> sys.modules["ddtrace.bootstrap.sitecustomize"]
<module 'ddtrace.bootstrap.sitecustomize' from '/Users/brett.langdon/datadog/dd-trace-py/ddtrace/bootstrap/sitecustomize.py'>
>>> sys.modules["sitecustomize"] is sys.modules["ddtrace.bootstrap.sitecustomize"]
False
>>>
```

ddtrace/contrib/httplib/__init__.py
```diff
@@ -7,8 +7,8 @@
 ~~~~~~~~
 
 The httplib integration is disabled by default. It can be enabled when using
-:ref:`ddtrace-run<ddtracerun>` using the ``DD_TRACE_HTTPLIB_ENABLED``
-environment variable::
+:ref:`ddtrace-run<ddtracerun>` or :mod:`import ddtrace.auto<ddtrace.auto`
```
brettlangdon, line 10:
```suggestion
:ref:`ddtrace-run<ddtracerun>` or :mod:`import ddtrace.auto<ddtrace.auto>`
```

?

scripts/docs/build.sh
```diff
@@ -1,7 +1,9 @@
 #!/usr/bin/env bash
 
+export DD_TRACE_ENABLED=0
+
 set -eux
 
 reno lint
-sphinx-build -W -b spelling docs docs/_build/html
-sphinx-build -W -b html docs docs/_build/html
+sphinx-build -vvv -W -b spelling docs docs/_build/html
```
brettlangdon, line 8: do we want to keep this verbosity? (I don't care either way, just checking that it was intentional)

templates/integration/__init__.py
```diff
@@ -7,7 +7,7 @@
 ~~~~~~~~
 
 The foo integration is enabled automatically when using
-:ref:`ddtrace-run <ddtracerun>` or :func:`patch_all() <ddtrace.patch_all>`.
+:ref:`ddtrace-run <ddtracerun>` or :mod:`import ddtrace.auto <ddtrace.auto>`.
```
brettlangdon, line 10: a few places use `:ref:` and a only a few use `:mod:`. I didn't notice a pattern for why one is used vs the other, is it intentional?

brettlangdon (approved):
very minor nits or thoughts, but otherwise lgtm, thanks!

scripts/docs/build.sh
emmettbutler: yeah that's intentional. it makes debugging easier.

ddtrace/auto.py
emmettbutler: "undefined behavior" here really just means that I haven't given it much thought and it's not how the interface is intended to be used.

templates/integration/__init__.py
emmettbutler: thanks for catching that. those oversights are the result of malformed `sed` commands.

ZStriker19 (approved):

PR #7548, opened 2023-11-10
fix(rcm): update the forking behavior in remote config

avara1986:
# Context
The Remote Configuration Publisher/Subscriber System https://github.com/DataDog/dd-trace-py/pull/5464 restarts all pubsub instances when the application forks (for example, [gunicorn workers](https://docs.gunicorn.org/en/stable/design.html#server-model), uwsgi workers, etc.).

image

Dynamic Instrumentation needs to update the pubsub instance at this point because the probe mechanism should run in the child process. For that, DI needs the callback as the method of an instance of Debugger, which lives in the child process.

https://github.com/DataDog/dd-trace-py/blob/2.x/ddtrace/debugging/_debugger.py#L276

# Problem description

When the application forks and restarts the subscribers, this happens before the debugger updates its callback instance. 

```
10348 starting subscribers
10348 restarting the debugger
10348 register callback 4429382528 <bound method Debugger._on_configuration of Debugger(status=<ServiceStatus.STOPPED: 'stopped'>)> 4406068560 shared_data 4398747792
```

This results in the registration of the callback in the child processes but the execution of callbacks using the instance of the parent process.

* Parent process: register a callback [on_configuration](https://github.com/DataDog/dd-trace-py/blob/2.x/ddtrace/debugging/_debugger.py#L654) for DI

```
94621 register callback <bound method Debugger._on_configuration of Debugger(status=<ServiceStatus.STOPPED: 'stopped'>)> PubSub Isntance id: 4363974784 Debugger._on_configuration id: 4360362576
```

* Child processes: register callbacks [on_configuration](https://github.com/DataDog/dd-trace-py/blob/2.x/ddtrace/debugging/_debugger.py#L654) for DI

```
94638 register callback <bound method Debugger._on_configuration of Debugger(status=<ServiceStatus.STOPPED: 'stopped'>)> PubSub Isntance id: 4392569856 Debugger._on_configuration id: 4364006016
94639 register callback <bound method Debugger._on_configuration of Debugger(status=<ServiceStatus.STOPPED: 'stopped'>)> PubSub Isntance id: 4392569856 Debugger._on_configuration id: 4364006016
94640 register callback <bound method Debugger._on_configuration of Debugger(status=<ServiceStatus.STOPPED: 'stopped'>)> PubSub Isntance id: 4392569856 Debugger._on_configuration id: 4364006016
```

* Child processes: exec callbacks.
```
94621 _exec_callback <bound method Debugger._on_configuration of Debugger(status=<ServiceStatus.RUNNING: 'running'>)> PubSub Isntance id: 4392569856 Debugger._on_configuration id:  4360362576
94638 _exec_callback <bound method Debugger._on_configuration of Debugger(status=<ServiceStatus.RUNNING: 'running'>)> PubSub Isntance id: 4392569856 Debugger._on_configuration id:  4360362576
94639 _exec_callback <bound method Debugger._on_configuration of Debugger(status=<ServiceStatus.RUNNING: 'running'>)> PubSub Isntance id: 4392569856 Debugger._on_configuration id:  4360362576
94640 _exec_callback <bound method Debugger._on_configuration of Debugger(status=<ServiceStatus.RUNNING: 'running'>)> PubSub Isntance id: 4392569856 Debugger._on_configuration id:  4360362576
```

As a result, we're registering callback id **4364006016** but calling **4360362576** (which is the instance of the parent process).

# PR Description
This PR removes the restart of publisher-subscriber instances in Remote Configuration at fork and delegates it to the products that have integration with Remote Config, namely AppSec and Dynamic Instrumentation.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/internal/remoteconfig/worker.py
```diff
@@ -80,17 +80,20 @@ def enable(self):
...
+    def reset_at_fork(self):
         # type: () -> None
-        """Subscribers need to be restarted when application forks"""
+        """Client Id needs to be restarted when application forks"""
         self._enable = False
         log.debug("[%s][P: %s] Remote Config Poller fork. Starting Pubsub services", os.getpid(), os.getppid())
         self._client.renew_id()
+
+    def restart_subscribers(self):
+        # type: () -> None
```
P403n1x87, line 96: I don't think anything should ever call this. Because it should be responsibility of all the products making use of RC to determine whether they need to restart after a fork, there should not be an attempt to restart all the subscribers at once. So probably the best thing to do here is remove this method entirely

ddtrace/internal/remoteconfig/worker.py
```diff
@@ -80,17 +80,20 @@ def enable(self):
...
+            forksafe.register(self.reset_at_fork)
             atexit.register(self.disable)
             return True
         return False
 
-    def start_subscribers(self):
+    def reset_at_fork(self):
         # type: () -> None
-        """Subscribers need to be restarted when application forks"""
+        """Client Id needs to be restarted when application forks"""
```
P403n1x87, line 90:
```suggestion
        """Client Id needs to be refreshed when application forks"""
```

ddtrace/internal/remoteconfig/worker.py
```diff
@@ -80,17 +80,20 @@ def enable(self):
...
         self._enable = False
         log.debug("[%s][P: %s] Remote Config Poller fork. Starting Pubsub services", os.getpid(), os.getppid())
```
P403n1x87, line 92:
```suggestion
        log.debug("[%s][P: %s] Remote Config Poller fork. Refreshing state", os.getpid(), os.getppid())
```

ddtrace/appsec/_remoteconfiguration.py
```diff
@@ -36,6 +39,10 @@ def __init__(self, _preprocess_results, callback):
         self._subscriber = self.__subscriber_class__(self.__shared_data__, callback, "ASM")
 
 
+def _forksafe_appsec_rc():
+    remoteconfig_poller.start_subscribers_by_product(APPSEC_PRODUCTS)
```
P403n1x87, line 43: Looks good. I'm just wondering if we can do something similar to what the debugger does, that is, first disable appsec, and then re-enable by simply re-registering the product. That seems to do the trick for DI, without the need for this extra method.

avara1986: `start_subscribers_by_product` disables and re-enables it. `start_subscribers_by_product` starts all products that we registered previously. If the product is not registered, `start_subscribers_by_product` does nothing. We can't do what you suggest because ASM enables and disables products dynamically. `ASM_DATA`, `ASM`, and `ASM_DD` depend on the callbacks that we receive from the `ASM_FEATURE` product callbacks.

P403n1x87:
I think this PR needs to be of type `fix` since we are fixing an issue, and we also need to backport it on other release branches.

P403n1x87 (changes requested):
We also need this patch
```
diff --git a/ddtrace/internal/remoteconfig/worker.py b/ddtrace/internal/remoteconfig/worker.py
index e507dd5be..ec2ae453f 100644
--- a/ddtrace/internal/remoteconfig/worker.py
+++ b/ddtrace/internal/remoteconfig/worker.py
@@ -150,14 +150,12 @@ class RemoteConfigPoller(periodic.PeriodicService):
         try:
             # By enabling on registration we ensure we start the RCM client only
             # if there is at least one registered product.
-            enabled = True
             if not skip_enabled:
-                enabled = self.enable()
+                self.enable()
 
-            if enabled:
-                self._client.register_product(product, pubsub_instance)
-                if not self._client.is_subscriber_running(pubsub_instance):
-                    pubsub_instance.start_subscriber()
+            self._client.register_product(product, pubsub_instance)
+            if not self._client.is_subscriber_running(pubsub_instance):
+                pubsub_instance.start_subscriber()
         except Exception:
             log.debug("error starting the RCM client", exc_info=True)
```
to fix DI

ddtrace/internal/remoteconfig/worker.py
```diff
@@ -144,17 +147,19 @@ def update_product_callback(self, product, callback):
 
     def register(self, product, pubsub_instance, skip_enabled=False):
         # type: (str, PubSub, bool) -> None
+
+        if not isinstance(pubsub_instance, PubSub):
+            raise TypeError("%s is not a PubSub instance" % str(pubsub_instance))
```
P403n1x87, line 152: 🤔  what is this for?

avara1986: When I updated this function with your suggestion, many Remote Config (RC) tests started to fail with "AttributeError: 'function' object has no attribute 'stop.'"

https://github.com/DataDog/dd-trace-py/pull/7548/files#diff-344d4f5be7adec7a15bb259a6da69b1a8a4bb49a813484f6eaafbaff56601657L171

It was hard to find the root of the problem, and I found that we are registering functions at some points instead of PubSub instances, but the error is raised at:

```python
    def stop_subscribers(self, join=False):
        # type: (bool) -> None
        """
        Disable the remote config service and drop, remote config can be re-enabled
        by calling ``enable`` again.
        """
        log.debug(
            "[%s][P: %s] Remote Config Poller fork. Stopping  Pubsub services",
            os.getpid(),
            self._parent_id,
        )
        for pubsub in self._client.get_pubsubs():
>           pubsub.stop(join=join)
```

P403n1x87: Ah I see, but we already have type annotations here, so I don't think we need this check at runtime? We don't do these checks in other places

avara1986:
```
We don't do these checks in other places
```
That's true, I'll update the PR, thanks for your suggestion!

P403n1x87 (approved):

P403n1x87 (commented):
We need a release note, otherwise LGTM!

juanjux (approved):

ZStriker19 (approved):

PR #7551, opened 2023-11-10
chore(aiohttp): add support for StreamResponse

### LINKED ISSUE #1619: aiohttp: support StreamResponse
Hi,

I have an aiohttp API with some async code inside that I'm trying to trace with Datadog. I'm using [documentation for aiohttp](http://pypi.datadoghq.com/trace/docs/web_integrations.html#aiohttp) for tracing: `trace_app(app, tracer, service="my-api")`

The problem that I encounter is that DD trace shows incorrect execution time as well as status 200 all the time. Moreover, all logs after the first `await` are not tagged with `dd.trace_id` and `dd.span_id`. I assume that the execution time is also measured until the first `await`.

For example, a curl to the endpoint that returns a file can take around 30 seconds and sometimes ends up with a timeout exception, while Datadog always reports a few ms exec time and code 200. 

Below is the problematic endpoint:

```python
class DownloadLogsView(BaseView):
    """Download logs."""

    async def get(self):

        log.info("Starting download")  # This log has correct dd.trace_id and dd.span_id

        app_id = self.request.match_info.get("app_id")
        client = self.request.app.get("elastic")

        filename = f"logs_{app_id}.txt"
        response = web.StreamResponse(
            headers={
                hdrs.CONTENT_TYPE: "text/plain; charset=utf-8",
                hdrs.CACHE_CONTROL: "no-cache",
                hdrs.CONTENT_DISPOSITION: f"attachment; filename={filename}",
                hdrs.TRANSFER_ENCODING: "chunked",
            }
        )
        await response.prepare(self.request)
       
       ### That is where trace and span ids are "lost" ###

        log.info(f"Download logs request for app: {app_id}")  # This log has dd.trace_id=0 and dd.span_id=0

        async for logline in es_client.scan_app_logs(client, app_id):  # Same for all logs from es_client
            await response.write(logline.encode("utf-8"))

        return response
```

```
ddtrace==0.40.0
```
```
aiohttp==3.6.2
elasticsearch[async]==7.8.0
python-json-logger==v0.1.10
gunicorn==20.0.4
```

The problematic endpoint is provided above.

Incomplete traces.

Traces with correct execution time and status codes, and corresponding logs to have correct trace ids.

emmettbutler:
This pull request makes the `aiohttp` integration wait to close spans for `aiohttp.web.StreamResponse`s until they are finished streaming data. It does this by skipping the `on_prepare` signal handler and instead closing the span in an `add_done_callback` on the asyncio task handling the request. The middleware only uses this behavior when `type(response) is aiohttp.web.StreamResponse`.

Fixes #1619

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/contrib/aiohttp/middlewares.py
```diff
@@ -60,7 +62,10 @@ async def attach_context(request):
         # DEV: aiohttp is special case maintains separate configuration from config api
         analytics_enabled = app[CONFIG_KEY]["analytics_enabled"]
         if (config.analytics_enabled and analytics_enabled is not False) or analytics_enabled is True:
-            request_span.set_tag(ANALYTICS_SAMPLE_RATE_KEY, app[CONFIG_KEY].get("analytics_sample_rate", True))
+            request_span.set_tag(
```
emmettbutler, line 65: This is a `black` change

ddtrace/contrib/aiohttp/middlewares.py
```diff
@@ -126,6 +126,18 @@ async def on_prepare(request, response):
     request_span.finish()
 
 
+async def on_prepare(request, response):
```
majorgreys, line 129: Do we know when this signal is sent? Is the server done responding? Is there anything after this signal is sent that we would want to capture with the request span? Can we do with just the callback?

emmettbutler: Maybe. My approach with this change is to avoid breaking any existing behavior. Do you think it's important to remove this signal handler if possible?

majorgreys: Not necessary. I had missed the fact that the signal was already being handled and that this change was ensuring streaming responses were closed. 

What happens right now with streaming responses?

emmettbutler: They have incorrect durations as described here https://github.com/DataDog/dd-trace-py/issues/1619

majorgreys (approved):

PR #7590, opened 2023-11-14
feat(propagation): assess and link conflicting contexts upon extraction

ZStriker19:
**What?**
Libraries must add the ability to read multiple propagator formats so that if the subsequent trace contexts with conflicting trace-ids are detected then an appropriate span link will be added to the span indicating the error.

**Design**
We need to add the span_links to the Context object, and can't go directly to adding them to the span. Therefore the links are added to the span on `start_span`, we add the span_links to the span if it's the local root span.

**Testing**
For the tests I wanted to make sure the Context object looked full correct so I created a test to look at the full object. 
I did include a test in which a span link is added, and also the tracecontext trace_id matches the primary context's trace_id so the the tracestate is added to the context, which is the behavior implemented [here](https://github.com/DataDog/dd-trace-py/pull/7538). However, since this PR essentially adds to that functionality I thought it good to include.

**Peformance**
Performance overhead of just adding SpanLink upon non-matching trace_ids for contexts is pretty minimal, each of these cases would generate one span link. If a user enabled all styles and had non-matching trace_ids for 3 styles it would be worse, but that seems pretty unusual. This case seems by far the most typical.
config yaml — dd-trace-py 2023-11-16 at 12 20 27 AM

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


releasenotes/notes/resolve-trace-contexts-43450ad9a5a0d59e.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    propagation: If a valid context is extracted from headers, and the following extracted
+    trace context's trace_id's do not match the valid context's, then add a span link to the
```
emmettbutler, line 5:
```suggestion
    trace context's ``trace_id``s do not match the valid context's, then add a span link to the
```

emmettbutler (approved):

PR #7597, opened 2023-11-14
ci(logging): update tests to reflect new 128 bit trace_id standard

zarirhamza:
Updates tests in new logging libraries to reflect new 128 bit trace_id standard. 
The logic remains untouched that 128bit trace_ids continue to only be logged iff `_128_bit_trace_id_logging_enabled` is `True`

## Checklist

## Reviewer Checklist

- [x] Title is accurate.

### DIFF
--- a/tests/contrib/logbook/test_logbook_logging.py
+++ b/tests/contrib/logbook/test_logbook_logging.py
@@ -18,6 +18,9 @@ handler = TestHandler()
 def _test_logging(span, env, service, version):
     dd_trace_id, dd_span_id = (span.trace_id, span.span_id) if span else (0, 0)
 
+    if dd_trace_id != 0 and config._128_bit_trace_id_enabled and not config._128_bit_trace_id_logging_enabled:
+        dd_trace_id = span._trace_id_64bits
+
     assert handler.records[0].message == "Hello!"
     assert handler.records[0].extra["dd.trace_id"] == str(dd_trace_id)
     assert handler.records[0].extra["dd.span_id"] == str(dd_span_id)
@@ -49,6 +52,8 @@ def test_log_trace_global_values():
     Check trace info includes global values over local span values
     """
 
+    handler.records.clear()
+
     span = tracer.trace("test.logging")
     span.set_tag(ENV_KEY, "local-env")
     span.set_tag(SERVICE_KEY, "local-service")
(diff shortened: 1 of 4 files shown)


tests/contrib/logbook/test_logbook_logging.py
```diff
@@ -49,6 +52,8 @@ def test_log_trace_global_values():
     Check trace info includes global values over local span values
     """
 
+    handler.records.clear()
```
mabdinur, line 55: nit: this can be a fixture that's auto applied to all logbook tests

tests/contrib/logbook/test_logbook_logging.py
```diff
@@ -18,6 +18,9 @@
 def _test_logging(span, env, service, version):
     dd_trace_id, dd_span_id = (span.trace_id, span.span_id) if span else (0, 0)
 
+    if dd_trace_id != 0 and config._128_bit_trace_id_enabled and not config._128_bit_trace_id_logging_enabled:
+        dd_trace_id = span._trace_id_64bits
```
mabdinur, line 22: nit: I dont think we need to check if `128_bit_trace_id` is enabled. 

`span._trace_id_64bits == trace_id` if `config._128_bit_trace_id_enabled == False`.

mabdinur (approved):
It would be nice to have test case for when 128 bit trace id logging is enabled. I think we are only test the default behavior. 

Changes so far look good to me. I don't think we need to block this PR on my current suggestions. Let's unblock ci

ZStriker19 (approved):

tests/contrib/logbook/test_logbook_logging.py
zarirhamza: Done

tests/contrib/logbook/test_logbook_logging.py
zarirhamza: Done

zarirhamza:
> It would be nice to have test case for when 128 bit trace id logging is enabled. I think we are only test the default behavior.
> 
> Changes so far look good to me. I don't think we need to block this PR on my current suggestions. Let's unblock ci

We do have tests that test 128 bit trace Id within the subprocesses. Those tests are explicitly defined with an env var making 128bit trace_ids enabled.

PR #7627, opened 2023-11-16
fix(pytest+unittest): add safety check for Windows paths

ericlaz:
[CIVIS-8034
](https://datadoghq.atlassian.net/browse/CIVIS-8034?atlOrigin=eyJpIjoiNTVhNTFmZjkxYWRkNGUwOThiYzExYjhkZTkxZjU5NzQiLCJwIjoiaiJ9)

## Motivation

This PR aims to fix a GitHub issue regarding ValueErrors on Windows machines when using different path drives.

## What does this PR change?

Adds a few type safety checks around `os.path.relpath` calls to ensure that the plugin will continue to work smoothly when encountering these edge cases.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/contrib/pytest/plugin.py
```diff
@@ -806,7 +806,14 @@ def pytest_ddtrace_get_item_suite_name(item):
     if test_module_path:
         if not pytest_module_item.nodeid.startswith(test_module_path):
             log.warning("Suite path is not under module path: '%s' '%s'", pytest_module_item.nodeid, test_module_path)
-        suite_path = os.path.relpath(pytest_module_item.nodeid, start=test_module_path)
+        try:
+            suite_path = os.path.relpath(pytest_module_item.nodeid, start=test_module_path)
+        except TypeError:
```
romainkomorndatadog, line 811: Shouldn't this be
```suggestion
        except ValueError:
```
?

romainkomorndatadog: Tempted to commit this based on https://docs.python.org/3/library/os.path.html .

ddtrace/contrib/unittest/patch.py
```diff
@@ -258,7 +258,22 @@ def _extract_test_file_name(item) -> str:
...
+            return ""
+        try:
+            module_file_path = os.path.relpath(test_module_object, start=os.getcwd())
+            return module_file_path
+        except ValueError:
+            log.debug(
+                "Tried to collect module file path but it is using different drive paths on Windows, "
+                "using absolute path instead"
+            )
+            return os.path.abspath(test_module_object)
```
romainkomorndatadog, line 276: I think we're reusing this pattern (`try: replath, except ValueError`) enough times that we should probably roll it into its own utility function.

It'd be nice to get this fix out sooner rather than later, though, so I'm open to making a follow-up PR to clean up.

romainkomorndatadog:
@ZStriker19 , if you're comfortable merging this so we can get in to `2.3.0`, and let us have a follow-up PR to clean up into a helper function and add some tests, go ahead. Otherwise we'll just wait update the PR on Monday and backport into `2.3`.

mabdinur (approved):
This fix looks good from a core prespective.

romainkomorndatadog (approved):

ddtrace/internal/ci_visibility/utils.py
```diff
@@ -10,12 +10,25 @@
 log = get_logger(__name__)
 
 
-def get_source_file_path_for_test_method(test_method_object) -> typing.Union[str, None]:
+def get_relative_or_absolute_path_for_path(path: str, start_directory: str):
     try:
-        source_file_path = os.path.relpath(inspect.getfile(test_method_object))
+        relative_path = os.path.relpath(path, start=start_directory)
```
P403n1x87, line 15: feel free to ignore: would it make sense to use `pathlib` for these path operations at this point, or would it make more sense to do a refactor later?

romainkomorndatadog: Seems like, with `pathlib`, we'd still end up with a `ValueError` for relative paths that can't be computed, so the logic would be the same.

For me, that adds up to "refactor later".

PR #7635, opened 2023-11-16
chore(rcm): make the RC enablement variable private

P403n1x87:
The DD_REMOTE_CONFIGURATION_ENABLED variable has been introduced to deal with threading issue and for testing purposes, and was not supposed to be part of the public API. We make the variable private to allow it to still be used in testing and debugging.

## Checklist

## Reviewer Checklist

- [ ] Title is accurate.


tests/commands/test_runner.py
```diff
@@ -346,48 +346,23 @@ def test_info_no_configs():
...
-        b""" https://docs.datadoghq.com/help/\x1b[0m"""
-        b"""\n\n\x1b[93mWARNING SERVICE NOT SET: It is recommended that a service tag be set for all traced """
-        b"""applications. For more information please see"""
-        b""" https://ddtrace.readthedocs.io/en/stable/troubleshooting.html\x1b[0m"""
-        b"""\n\n\x1b[93mWARNING ENV NOT SET: It is recommended that an env tag be set for all traced applications. """
-        b"""For more information please see https://ddtrace.readthedocs.io/en/stable/troubleshooting.html\x1b[0m"""
-        b"""\n\n\x1b[93mWARNING VERSION NOT SET: """
-        b"""It is recommended that a version tag be set for all traced applications. """
-        b"""For more information please see https://ddtrace.readthedocs.io/en/stable/troubleshooting.html\x1b[0m\n"""
+    assert (
```
P403n1x87, line 349: @ZStriker19 🆘

ddtrace/settings/config.py
```diff
@@ -328,10 +328,11 @@ def __init__(self):
 
         self.http = HttpConfig(header_tags=self.trace_http_header_tags)
         self._tracing_enabled = asbool(os.getenv("DD_TRACE_ENABLED", default=True))
-        self._remote_config_enabled = asbool(os.getenv("DD_REMOTE_CONFIGURATION_ENABLED", default=True))
+        self._remote_config_enabled = asbool(os.getenv("_DD_REMOTE_CONFIGURATION_ENABLED", default=True))
```
ZStriker19, line 331: Are we allowed to just make this internal now? We should probably at least have a release note for customers as I know some rely on RCM being disabled with their current configuration.

P403n1x87: Good point. With the regression fixed we don't ever expect users to need to disable RCM. Besides this changes also makes the RCM worker start-up lazy: if there are no products registered with RC then the workers won't start. If the env var is set, it would effectively become a no-op.

ZStriker19: Spoke with team, can we deprecate `DD_REMOTE_CONFIGURATION_ENABLED` and have that be a fall back, that way for users that are disabling it, they won't all of the sudden crash again.

P403n1x87: This PR follows the fix for the gevent regression #7649, and besides makes RC lazy, meaning that if there are no products requiring it, the RC workers won't start. Users not relying on RC won't have it started anyway so I don't expect any impact in this regard.

P403n1x87: RC is not something that can be used directly, it's more an internal dependencies for other products to use. In this sense, I think it makes sense for RC to be enabled "on-demand" via other products (e.g. a user requesting DI, ASM etc...)

PR #7649, opened 2023-11-17
fix(bootstrap): do not load new modules after module cleanup

P403n1x87:
We refactor the logic of the sitecustomize bootstrap script to make it harder to introduce new initialisation logic that can break some of the boilerplate bootstrapping logic, such as the module cleanup mechanism for the support of frameworks like gevent. We move all the logic into a dedicated module, preload, and expose post_preload hooks in case that some code needs to be run after the module cleanup. However, the problem is now slightly shifted towards these hooks. Care must be taken not to register any hooks that import modules that are not already imported during the execution of the preload script.

## Testing Strategy

The test that was supposed to catch the presence of the `threading` module after initialisation when cleaning up the imported modules failed because RC was disabled globally in tests with `DD_REMOTE_CONFIGURATION_ENABLED=1` in the `riotfile.py` script. We have modified the test to ensure that RC is enabled.

## Further Details

Because the changes in this PR include a refactor, we summarise the part of the old coding that actually caused the issue in the first place. The `sitecustomize.py` source included the following coding

```python
    # Only the import of the original sitecustomize.py is allowed after this
    # point.

    if sys.version_info >= (3, 7) and config._otel_enabled:
        from opentelemetry.trace import set_tracer_provider

        from ddtrace.opentelemetry import TracerProvider

        set_tracer_provider(TracerProvider())

    ...

    if config._remote_config_enabled:
        from ddtrace.internal.remoteconfig.worker import remoteconfig_poller

        remoteconfig_poller.enable()

    if asm_config._asm_enabled or config._remote_config_enabled:
        from ddtrace.appsec._remoteconfiguration import enable_appsec_rc

        enable_appsec_rc()
```

The first `if` block introduces two problems:
- it force-loads the `opentelemetry` module
- it imports more modules from `ddtrace` _after_ the comment that instructs not to add imports beyond that point

Note that the current implementation of the `opentelemetry` intergration also force-loads the `opentelemetry` module when `TracerProvider` is imported. Therefore, the whole logic should be registered as a module import hook.

The other `if` blocks further down in the `sitecustomize.py` source also import from `ddtrace`. In fact, they import features that require threading support, which ultimately forces the gevent-crucial `threading` module to be re-loaded too early after the loaded module clean-up.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


mabdinur:
From the diff it's not immediately clear which changes introduced a regression in our gevent support. Can we call this out in the the PR description?

majorgreys (commented):
@P403n1x87 Is there a reproduction that can be used for testing this fix?

P403n1x87:
> @P403n1x87 Is there a reproduction that can be used for testing this fix?

The modified test will ensure that the RC poller thread is started. Having disabled RC globally was what make the regression go undetected.

majorgreys (approved):

tests/telemetry/test_telemetry_metrics_e2e.py
```diff
@@ -172,7 +172,7 @@ def test_span_creation_and_finished_metrics_opentracing(test_agent_session, ddtr
 def test_span_creation_no_finish(test_agent_session, ddtrace_run_python_code_in_subprocess):
     code = """
 import ddtrace
-import opentelemetry
+import opentelemetry.trace
```
mabdinur, line 175: I still don't follow why this import is problematic. If we update `@ModuleWatchdog.after_module_imported("opentelemetry.trace")` to `@ModuleWatchdog.after_module_imported("opentelemetry")` would we still need this refactor?

mabdinur (approved):

tests/telemetry/test_telemetry_metrics_e2e.py
P403n1x87: `opentelemetry` is a namespace package. When imported, it essentially contains nothing
```
>>> import opentelemetry
>>> dir(opentelemetry)
['__doc__', '__file__', '__loader__', '__name__', '__package__', '__path__', '__spec__']
```
The `trace` sub-module needs to be imported directly for it to be added to `opentelemetry`
```
>>> import opentelemetry.trace
>>> dir(opentelemetry)
['__doc__', '__file__', '__loader__', '__name__', '__package__', '__path__', '__spec__', 'attributes', 'context', 'environment_variables', 'trace', 'util']
```
We could in theory make the suggested change `@ModuleWatchdog.after_module_imported("opentelemetry.trace") -> @ModuleWatchdog.after_module_imported("opentelemetry")` (which would require #7727), but because `opentelemetry` is just a namespace package, it is unlikely that somebody would import just that, as it is pretty much useless. The real module of interest is `opentelemetry.trace` here.

PR #7663, opened 2023-11-17
feat(openai): enable sending llmobs payloads

Yun-Kim:
This PR builds on top of #7621 to add the llmobs writer to the openai integration, enabling the openai integration to send prompt-response record payloads to the LLMobs ingest API. Note this is only sending records for the completion and chat completion endpoints. This PR also adds tests for the openai integration to send LLMObs payloads when configured.

Three configuration options are introduced to allow sending LLMObs payloads (from the openai integration):
- `DD_OPENAI_LLMOBS_ENABLED` which must be set to true, default false
- `DD_OPENAI_LLMOBS_PROMPT_COMPLETION_SAMPLE_RATE` to control the sample rate for generating/sending LLMObs payloads, default 100%
- `DD_APP_KEY`, which is required (along with `DD_API_KEY` which already exists) to send to the LLMObs ingest API

Note: this feature is in private preview and gated with a feature enable env var `DD_OPENAI_LLMOBS_ENABLED` so no release note is needed. This means the above configuration options also don't need to be documented until LLMObs goes GA.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/contrib/openai/patch.py
```diff
@@ -34,11 +34,14 @@
     "openai",
     {
         "logs_enabled": asbool(os.getenv("DD_OPENAI_LOGS_ENABLED", False)),
+        "llmobs_enabled": asbool(os.getenv("DD_OPENAI_LLMOBS_ENABLED", False)),
```
mabdinur, line 37: Can we document these new configurations

ddtrace/contrib/_trace_utils_llm.py
```diff
@@ -24,7 +24,7 @@
 class BaseLLMIntegration:
     _integration_name = "baseLLM"
 
-    def __init__(self, config, stats_url, site, api_key):
+    def __init__(self, config, stats_url, site, api_key, app_key=None):
```
mabdinur, line 27: Should app_key be an optional argument?  Do we need to raise an errorr/log warning if it is None

ddtrace/contrib/_trace_utils_llm.py
```diff
@@ -37,8 +37,16 @@ def __init__(self, config, stats_url, site, api_key):
             interval=float(os.getenv("_DD_%s_LOG_WRITER_INTERVAL" % self._integration_name.upper(), "1.0")),
             timeout=float(os.getenv("_DD_%s_LOG_WRITER_TIMEOUT" % self._integration_name.upper(), "2.0")),
         )
+        self._llmobs_writer = LLMObsWriter(
+            site=site,
+            api_key=api_key,
+            app_key=app_key,
+            interval=float(os.getenv("_DD_%s_LLM_WRITER_INTERVAL" % self._integration_name.upper(), "1.0")),
+            timeout=float(os.getenv("_DD_%s_LLM_WRITER_TIMEOUT" % self._integration_name.upper(), "2.0")),
```
mabdinur, line 45: Do we need to make the interval and timeout configurable via internal configurations? When will we need to toggle these values?

ddtrace/contrib/langchain/patch.py
```diff
@@ -52,6 +52,9 @@ def get_version():
         "logs_enabled": asbool(os.getenv("DD_LANGCHAIN_LOGS_ENABLED", False)),
         "metrics_enabled": asbool(os.getenv("DD_LANGCHAIN_METRICS_ENABLED", True)),
         "span_prompt_completion_sample_rate": float(os.getenv("DD_LANGCHAIN_SPAN_PROMPT_COMPLETION_SAMPLE_RATE", 1.0)),
+        "llmobs_prompt_completion_sample_rate": float(
```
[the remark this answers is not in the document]
mabdinur: Will the future PR be added soon? If not we should add a comment to document this gap in our support

ddtrace/contrib/openai/patch.py
```diff
@@ -34,11 +34,14 @@
...
         "metrics_enabled": asbool(os.getenv("DD_OPENAI_METRICS_ENABLED", True)),
         "span_prompt_completion_sample_rate": float(os.getenv("DD_OPENAI_SPAN_PROMPT_COMPLETION_SAMPLE_RATE", 1.0)),
+        "llmobs_prompt_completion_sample_rate": float(os.getenv("DD_OPENAI_LLMOBS_PROMPT_COMPLETION_SAMPLE_RATE", 1.0)),
         "log_prompt_completion_sample_rate": float(os.getenv("DD_OPENAI_LOG_PROMPT_COMPLETION_SAMPLE_RATE", 0.1)),
         "span_char_limit": int(os.getenv("DD_OPENAI_SPAN_CHAR_LIMIT", 128)),
         "_api_key": os.getenv("DD_API_KEY"),
+        "_app_key": os.getenv("DD_APP_KEY"),
```
mabdinur, line 44: These configurations are getting a bit messy. Can/should we use envier to organize them?

ddtrace/contrib/openai/patch.py
```diff
@@ -244,6 +247,72 @@ def record_usage(self, span, usage):
             span.set_metric("openai.response.usage.%s_tokens" % token_type, num_tokens)
             self._statsd.distribution("tokens.%s" % token_type, num_tokens, tags=tags)
 
+    def generate_completion_llm_records(self, resp, span, args, kwargs):
+        # type: (Any, Span, List[Any], Dict[str, Any]) -> None
```
mabdinur, line 251: The Type of response is currently Any, can we make this more defined?

ddtrace/contrib/openai/patch.py
```diff
@@ -244,6 +247,72 @@ def record_usage(self, span, usage):
...
+        """Generate payloads for the LLM Obs API from a completion."""
+        if not self._config.llmobs_enabled:
+            return
+        choices = resp.choices
+        n = kwargs.get("n", 1)
+        prompt = kwargs.get("prompt", "")
```
mabdinur, line 257: nit: We should initialize prompt to an array if we can

```suggestion
        prompt = kwargs.get("prompt", [""])
```

ddtrace/contrib/openai/patch.py
```diff
@@ -244,6 +247,72 @@ def record_usage(self, span, usage):
...
+        if isinstance(prompt, str):
+            prompt = [prompt]
+        # Note: LLMObs ingest endpoint only accepts a 1:1 prompt-response mapping per record,
+        #  so we need to deduplicate and send unique prompt-response records if n > 1.
+        for i in range(n):
+            unique_choices = choices[i::n]
```
mabdinur, line 263: So we're getting a slice that starts at index i and skips n values? So if n=3 choices=[A, B, C, D, E],
 iteration 1 -> unique_choices=[A, D], iteration 2 -> [B, E], iteration 3 -> [C]. Is this expected? This algorithm just seems funky.

ddtrace/contrib/openai/patch.py
```diff
@@ -244,6 +247,72 @@ def record_usage(self, span, usage):
...
+            attrs_dict = {
+                "type": "completion",
+                "id": resp.id,
+                "timestamp": resp.created * 1000,
```
mabdinur, line 267: Can this value be None?

ddtrace/contrib/openai/patch.py
```diff
@@ -266,12 +336,19 @@ def patch():
...
+        app_key=ddappkey,
     )
 
     if config.openai.logs_enabled:
         if not ddapikey:
             raise ValueError("DD_API_KEY is required for sending logs from the OpenAI integration")
         integration.start_log_writer()
+    if config.openai.llmobs_enabled:
+        if not ddapikey:
+            raise ValueError("DD_API_KEY is required for sending LLMObs payloads from the OpenAI integration")
```
mabdinur, line 348: We should give instructions on how to disable this feature.

ddtrace/contrib/_trace_utils_llm.py
Yun-Kim: `app_key` doesn't always need to be provided if `llmobs_enabled` is not set to true, in that case we raise an error on startup. I'll make the error more specific to show how to disable this feature.

ddtrace/contrib/_trace_utils_llm.py
Yun-Kim: These config values are only relevant to specific user cases for sending llmobs payloads at different intervals/timeouts. Most of the time these won't need to be toggled.

What do you mean by internal configurations?

ddtrace/contrib/openai/patch.py
Yun-Kim: We're planning on removing logs payloads entirely in the near future, so things like `logs_enabled`, `log_prompt_completion_sample_rate` will be removed and make this less messy - I don't think we need to reorganized at this time. 

If at that point (once we remove logs) is still too messy, we can look at cleaning up.

ddtrace/contrib/openai/patch.py
Yun-Kim: This is tough since openai can either return a json dictionary (`openai<1.0`) or a pydantic model (`openai>=1.0`). The former is a large complex dictionary which will make type hinting more gross, and the latter would be difficult to add a type hint for since this would now need to be version gated. 

I think we can ignore this for now unless you think this is a notable issue.

ddtrace/contrib/openai/patch.py
Yun-Kim: Nope, these are always guaranteed to be not-None values by the openai API.

ddtrace/contrib/openai/patch.py
Yun-Kim: The way openai completions work is that you send P number of prompts, specify N number of choices to return per prompt, meaning you will receive P * N choices in the response.

Let's say you send openai 3 prompts: [A, B, C], and specify N=2. Then you'll receive a completions in the form
```
[A1, A2, B1, B2, C1, C2]
```
Where A1 corresponds to the first response created for prompt A, A2 corresponds to the second response created for prompt A, and etc.

This algorithm is just making sure that we create a record with deduplicated, unique prompt-response pairs in the following form:
```
[A1, B1, C1], [A2, B2, C2]
```
i.e. this loop is just starting at index `i` and iterating over every `n`th choice which will correspond to a different prompt.

ddtrace/contrib/openai/patch.py
Yun-Kim: LLM Monitoring is in a private preview for now so this isn't meant to be public. Once it goes GA I'll add it to the docs.

majorgreys (approved):

mabdinur (approved):

alexbarksdale (approved):

Issue #7701, opened 2023-11-21
RuntimeError failed to encode SpanLink with 2.3.0


lecrepont01:
### Summary of problem

We have experienced `RuntimeError` issues with the extension starting from this morning and the dependabot update to 2.3.0.

Stack trace :
```
ValueError: Failed to encode SpanLink. span_id=None contains an unsupported type, <class 'NoneType'>
  File "ddtrace/internal/_encoding.pyx", line 488, in ddtrace.internal._encoding.MsgpackEncoderBase._pack_trace
  File "ddtrace/internal/_encoding.pyx", line 747, in ddtrace.internal._encoding.MsgpackEncoderV03.pack_span
  File "ddtrace/internal/_encoding.pyx", line 592, in ddtrace.internal._encoding.MsgpackEncoderV03._pack_links

RuntimeError: failed to pack span: <Span(id=6671936502756880443,trace_id=14943038461850520012489475877480943891,parent_id=4075607385283888244,name=fastapi.request)>. Exception: Failed to encode SpanLink. span_id=None contains an unsupported type, <class 'NoneType'>
  File "starlette/applications.py", line 122, in __call__
    await self.middleware_stack(scope, receive, send)
  File "ddtrace/contrib/asgi/middleware.py", line 275, in __call__
    reraise(exc_type, exc_val, exc_tb)
  File "six.py", line 719, in reraise
    raise value
  File "ddtrace/contrib/asgi/middleware.py", line 270, in __call__
    return await self.app(scope, receive, wrapped_send)
  File "starlette/middleware/errors.py", line 184, in __call__
    raise exc
  File "starlette/middleware/errors.py", line 162, in __call__
    await self.app(scope, receive, _send)
  File "starlette/middleware/cors.py", line 83, in __call__
    await self.app(scope, receive, send)
  File "starlette/middleware/exceptions.py", line 79, in __call__
    raise exc
  File "starlette/middleware/exceptions.py", line 68, in __call__
    await self.app(scope, receive, sender)
  File "fastapi/middleware/asyncexitstack.py", line 20, in __call__
    raise e
  File "fastapi/middleware/asyncexitstack.py", line 17, in __call__
    await self.app(scope, receive, send)
  File "starlette/routing.py", line 718, in __call__
    await route.handle(scope, receive, send)
  File "starlette/routing.py", line 276, in handle
    await self.app(scope, receive, send)
  File "starlette/routing.py", line 69, in app
    await response(scope, receive, send)
[...]

PR #7708, opened 2023-11-21
fix(propagation): check for span_id and trace_id before creating span link

ZStriker19:
Fixes: https://github.com/DataDog/dd-trace-py/issues/7701

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/propagation/http.py
```diff
@@ -839,7 +839,8 @@ def _resolve_contexts(contexts, styles_w_ctx, normalized_headers):
         links = []
         for context in contexts[1:]:
             style_w_ctx = styles_w_ctx[contexts.index(context)]
-            if context.trace_id != primary_context.trace_id:
+            # encoding expects at least trace_id and span_id
+            if context.span_id and context.trace_id and context.trace_id != primary_context.trace_id:
```
mabdinur, line 843: Should we also check if trace_id is None? Is this possible?

ddtrace/propagation/http.py
```diff
@@ -839,7 +839,8 @@ def _resolve_contexts(contexts, styles_w_ctx, normalized_headers):
...
                 links.append(
                     SpanLink(
```
mabdinur, line 845: Can we update the typing on the fields in the SpanLink class? We should make sure trace_id and span_id are always integers. It would be nice to also add a validator that prevents the ids  from being zero too. 

This can be done in a follow up. It's not a blocker

ddtrace/propagation/http.py
mabdinur, line 843: Some propagators get span ids to zero while other default to None. Can we make this behavior consistent? 

This isn't a blocker for this change.

ZStriker19: Actually, I think datadog, b3, and b3multi all return None, while tracecontext won't generate a context if the span_id or trace_id is 0, so I think we're good.

ddtrace/propagation/http.py
ZStriker19: I'll do it as follow up.

releasenotes/notes/propagation_check_id_valid_before_span_link-08be876337762a15.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    propagation: This fix resolves an issue where a ``Context`` generated from extracted headers could
+    lack a span_id or trace_id, leading to an encoding error for ``SpanLink``s.
```
mabdinur, line 5:
```suggestion
    lack a span_id or trace_id, leading ``SpanLink`` encoding errors.
```

mabdinur (approved):

PR #7725, opened 2023-11-21
fix(kafka): fix bug when setting message.key span tag for kafka serialization classes

### LINKED ISSUE #7674: Kafka failing writing span MESSAGE_KEY with a key that is not a string
### Summary of problem
Upgraded from 1.8.8 to 2.2.0 version to take advantage of the [fix](https://github.com/DataDog/dd-trace-py/issues/7174) of kafka DeserializingConsumer and SerializingProducer but it fails trying to write the key span: 

```
  File "/app/.venv/lib/python3.9/site-packages/xxxx/xxxx.py", line 35, in poll
    message = self._consumer.poll(timeout=self._timeout)
  File "/app/.venv/lib/python3.9/site-packages/ddtrace/contrib/kafka/patch.py", line 188, in traced_poll
    span.set_tag_str(kafkax.MESSAGE_KEY, ensure_text(message_key, errors="replace"))
  File "/app/.venv/lib/python3.9/site-packages/six.py", line 955, in ensure_text
    raise TypeError("not expecting type '%s'" % type(s))
TypeError: not expecting type '<class 'dict'>'
```

I think that parsing to string as you do for the [value](https://github.com/DataDog/dd-trace-py/blob/v2.2.0/ddtrace/contrib/kafka/patch.py#L155) could be enough if that's what you want. If not, directly not use a type at all.

2.2.0

pip 23.3.1

confluent-kafka 2.3.0

Try to consume a message with a key that is not `str` as you do on the tests. Exactly the same happens with the producer. Think that keys and values are going to be bytes that will be serialized from/deserialized into objects of arbitrary types.

`TypeError: not expecting type '<class 'dict'>'`

The span tags are parsed correctly.

wconti27:
## Description

Fixes #7674 . Sets `SerializedProducer` span tag for `message.key` by serializing key to bytes. Does not set the tag for a `DeserializedConsumer` unless the key is a `str` or `bytes` since we do not have the serialization function 

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/contrib/kafka/patch.py
```diff
@@ -27,6 +28,17 @@
 _DeserializingConsumer = (
     confluent_kafka.DeserializingConsumer if hasattr(confluent_kafka, "DeserializingConsumer") else None
 )
+_SerializationContext = (
+    confluent_kafka.serialization.SerializationContext
+    if hasattr(confluent_kafka.serialization, "SerializationContext")
```
mabdinur, line 33: What's the minimum version where `confluent_kafka.serialization` has the class `SerializationContext`? 

nit: Using the version of the integration here could be a little easier to maintain than a hasattr check.

ddtrace/contrib/kafka/patch.py
```diff
@@ -149,7 +162,15 @@ def traced_produce(func, instance, args, kwargs):
         span.set_tag_str(COMPONENT, config.kafka.integration_name)
         span.set_tag_str(SPAN_KIND, SpanKind.PRODUCER)
         span.set_tag_str(kafkax.TOPIC, topic)
-        span.set_tag_str(kafkax.MESSAGE_KEY, ensure_text(message_key, errors="replace"))
+
+        if isinstance(instance, _SerializingProducer):
```
mabdinur, line 166: `isinstance(instance, None)` will raise a TypeError

```suggestion
        if _SerializingProducer is not None and isinstance(instance, _SerializingProducer):
```

ddtrace/contrib/kafka/patch.py
```diff
@@ -149,7 +162,15 @@ def traced_produce(func, instance, args, kwargs):
...
+            if hasattr(instance, "_key_serializer"):
```
mabdinur, line 167: Using an internal attribute here is a bit brittle. This check can start failing at anytime and it could be painful to debug why the message was not set. 

We should try to use a public attribute if we can, if not we should atleast log a warning when this check fails.

ddtrace/contrib/kafka/patch.py
```diff
@@ -149,7 +162,15 @@ def traced_produce(func, instance, args, kwargs):
...
+                serialized_key = serialize_key(instance, topic, message_key, headers)
+                if serialized_key is not None:
+                    span.set_tag_str(kafkax.MESSAGE_KEY, serialized_key)
+        else:
+            span.set_tag_str(kafkax.MESSAGE_KEY, message_key)
```
mabdinur, line 172: Do we need the `ensure_text(....)` function here?

ddtrace/contrib/kafka/patch.py
```diff
@@ -184,7 +205,15 @@ def traced_poll(func, instance, args, kwargs):
             message_key = message.key() or ""
             message_offset = message.offset() or -1
             span.set_tag_str(kafkax.TOPIC, message.topic())
-            span.set_tag_str(kafkax.MESSAGE_KEY, ensure_text(message_key, errors="replace"))
+
+            # If this is a deserializing consumer, do not set the key as a tag since we
+            # do not have the serialization function
+            if (
+                not isinstance(instance, _DeserializingConsumer)
```
mabdinur, line 212: This class can be None too
?
```suggestion
                _DeserializingConsumer is not None and not isinstance(instance, _DeserializingConsumer)
```

ddtrace/contrib/kafka/patch.py
```diff
@@ -184,7 +205,15 @@ def traced_poll(func, instance, args, kwargs):
...
+                or isinstance(message_key, str)
+                or isinstance(message_key, bytes)
+            ):
+                span.set_tag_str(kafkax.MESSAGE_KEY, message_key)
```
mabdinur, line 216: What are the drawbacks of not setting this tag for `_DeserializingConsumer`? Can this break DSM? Do we need to call out this gap in our support in our public docs?

ddtrace/contrib/kafka/patch.py
```diff
@@ -203,3 +232,14 @@ def traced_commit(func, instance, args, kwargs):
     core.dispatch("kafka.commit.start", [instance, args, kwargs])
 
     return func(*args, **kwargs)
+
+
+def serialize_key(instance, topic, key, headers):
+    ctx = _SerializationContext(topic, _MessageField.KEY, headers)
```
mabdinur, line 238: We should check if `_SerializationContext` is None here. This code might be refactored/reused and this could introduce bugs

releasenotes/notes/fix-bug-for-setting-message-key-span-tag-from-kafka-serialization-classes-630b09004a97063a.yaml
```diff
@@ -0,0 +1,6 @@
+---
+fixes:
+  - |
+    This fix resolves an issue when setting the ``message.key`` span tag for kafka serializing producers and deserializing consumers. Serialized keys
```
mabdinur, line 4:
```suggestion
    kafka: This fix resolves an issue when setting the ``message.key`` span tag for kafka serializing producers and deserializing consumers. Serialized keys 
```

emmettbutler (commented):
I'll wait to review until existing change requests are resolved. Let me know if you want a review before then.

ddtrace/contrib/kafka/patch.py
wconti27: v1.4.0 is the version, I'll make this change!

ddtrace/contrib/kafka/patch.py
wconti27: `set_tag_str ` already uses `ensure_text` internally within the function

ddtrace/contrib/kafka/patch.py
wconti27: AFAIK we do not use the `message.key` tag for anything other than more information tagged on the span.

ddtrace/contrib/kafka/patch.py
wconti27: I will add the warning log

ddtrace/contrib/kafka/patch.py
wconti27: And at this moment if the key is not a string then we are already not setting it

ddtrace/contrib/kafka/patch.py
```diff
@@ -203,3 +231,18 @@ def traced_commit(func, instance, args, kwargs):
...
+        ctx = _SerializationContext(topic, _MessageField.KEY, headers)
+        if hasattr(instance, "_key_serializer") and instance._key_serializer is not None:
+            try:
+                key = instance._key_serializer(key, ctx)
+                return key
+            except Exception:
+                log.warning("Failed to set Consumer key tag: ", key, exc_info=True)
+                return None
+        else:
+            log.warning("Failed to set Consumer key tag, no method available to serialize key: ", key, exc_info=True)
```
mabdinur, line 247:
```suggestion
            log.warning("Failed to set Consumer key tag, no method available to serialize key: %s", str(key))
```

ddtrace/contrib/kafka/patch.py
```diff
@@ -203,3 +231,18 @@ def traced_commit(func, instance, args, kwargs):
...
+
+def serialize_key(instance, topic, key, headers):
+    if _SerializationContext is not None and _MessageField is not None:
+        ctx = _SerializationContext(topic, _MessageField.KEY, headers)
+        if hasattr(instance, "_key_serializer") and instance._key_serializer is not None:
+            try:
+                key = instance._key_serializer(key, ctx)
+                return key
+            except Exception:
+                log.warning("Failed to set Consumer key tag: ", key, exc_info=True)
```
mabdinur, line 244: We don't need to include a traceback in this warning. Also would could warning be noisy? If so, we should probably make it debug level

```suggestion
                log.warning("Failed to set Consumer key tag: %s", str(key))
```

releasenotes/notes/fix-bug-for-setting-message-key-span-tag-from-kafka-serialization-classes-630b09004a97063a.yaml
```diff
@@ -0,0 +1,6 @@
+---
+fixes:
+  - |
+    kafka: This fix resolves an issue when setting the ``message.key`` span tag for kafka serializing producers and deserializing consumers. Serialized keys 
+    are now used for the tag value. If a key is already deserialized by the consumer, no ``message.key`` span tag is set as we do not have access to the
+    serialization function.
```
mabdinur, line 6: Can we simplify this a bit. We should also call out the error this fix resolves. 

```suggestion
    kafka: Resolves ``TypeError`` raised by serializing producers and deserializing consumers when the ``message.key`` tag is set on spans.
```

mabdinur (approved):
left some nits overall LGTM

Also do we need to backport this fix to 1.20?

wconti27:
> left some nits overall LGTM
> 
> Also do we need to backport this fix to 1.20?

Yes we should backport to 1.20

ddtrace/contrib/kafka/patch.py
```diff
@@ -149,7 +162,15 @@ def traced_produce(func, instance, args, kwargs):
         span.set_tag_str(COMPONENT, config.kafka.integration_name)
         span.set_tag_str(SPAN_KIND, SpanKind.PRODUCER)
         span.set_tag_str(kafkax.TOPIC, topic)
-        span.set_tag_str(kafkax.MESSAGE_KEY, ensure_text(message_key, errors="replace"))
+
+        if isinstance(instance, _SerializingProducer):
+            if hasattr(instance, "_key_serializer"):
+                serialized_key = serialize_key(instance, topic, message_key, headers)
```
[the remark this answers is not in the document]
wconti27: Hi, manual benchmarking did not show any significant performance differences versus doing the serialization only once. Thanks

PR #7734, opened 2023-11-22
chore(openai): clean up OpenAI integration

Yun-Kim:
This PR cleans up the OpenAI integration by moving/removing duplicate code and hopefully making the patched endpoint hooks to be more straightforward for future readers.

## Refactor

To be more specific, this PR includes the following refactors:
- Adding a `_response_attrs` class variable for each endpoint hook class. This can be filled with the names of expected response attributes that will be automatically tagged (`resp.model`, `resp.id`, `resp.created`, etc). For more complicated or nested response attributes that cannot be tagged purely by their string value, `_record_response()` should be the place to add that tagging logic.
- The `_BaseCompletionHook._handle_response()` method has been renamed to `_handle_streamed_response()` and modified to only handle streamed responses, as non-stream responses don't need any further logic contained within `_handle_response()`. This means that all logic pertaining to prompt token computation from the `_Completion/_ChatCompletionHook` classes can be replaced by the logic in `_BaseCompletionHook._handle_streamed_response()`. 
- The Completion and Embedding hooks previously given single prompt/input arguments would tag them as `openai.request.prompt/input`, and if given multiple prompt/inputs would tag them as `openai.request.prompt/input.<index>`. They have now been changed to always tag the prompt/input arguments into a list even if only single arguments are provided, meaning the prompt/input tags will always be enumerated, i.e. `openai.request.prompt.0`.
- Tags which were mistakenly tagged based on prompt-completion sampling such as `openai.response.choices.X.finish_reason`, `openai.request.messages.X.role`, `openai.request.messages.X.name`, `openai.response.created (images)`, `openai.response.images_count`, are now always tagged.
- Extracted tagging logic for function/tool calls in the `_ChatCompletionHook` to a utility method in `contrib.openai.utils`.

## Outdated tag removal

In addition, this PR removes certain tags, including:
- permission information from the `_ModelRetrieve` hook. This is because the OpenAI API has now been modified to no longer output permission information, meaning we wouldn't tag this information to begin with.
- For the `_ModerationHook`, I've replaced the generic tagging of everything in the outputted moderation dictionary response in favor of a static list of categories we will always look for and tag. This is because the OpenAI API has changed which categories are outputted for this endpoint and this is a way for us to avoid this potentially tagging a lot of information we don't need.
- Completions/ChatCompletions/Edits: remove tagging `openai.response.choices_count`, `openai.response.choices.X.logprobs`, as neither are very useful pieces of information (if `n>1`, then this is redundant with the `n` keyword argument which is tagged). 
- Embeddings: `openai.response.embeddings.X.embedding-length` is always the same for all embedding outputs, so this was removed in favor of a `openai.response.embedding-length`.

## Cassette/Snapshot cleanup

Lastly, this PR also cleans up OpenAI test cassette data files and tests. Previously we had close to one test cassette (recorded request/response pair) and snapshot file per test, meaning this would quickly become a mess to manage. This PR attempts to solve this by:
- When applicable, making tests reuse the same test cassette data file (i.e. sync/async versions of the same test, although for v0 tests this is largely not possible due to the different openai-internal implementations of async responses).
- When applicable, make tests use the same snapshot, for example if multiple tests use the same method/endpoint we try to reuse the same cassette and test the same snapshot information.
- Rename cassette files to be more clear on what is being tested. Previously cassette files for async requests would follow this pattern: `completion.yaml --> acompletion.yaml`, but `completion_async.yaml` is probably more clear to reviewers.

## Testing Strategy

Aside from slight adjustments to tests to reuse the same cassette/snapshot files as mentioned above, there are no functional changes to the tests and thus no functional changes expected from this PR other than the dropping of outdated tags.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


emmettbutler (approved):

alexbarksdale (approved):

mabdinur:
Blocked by: https://github.com/DataDog/dd-trace-py/pull/7778

PR #7762, opened 2023-11-27
feat(aws): add trace context propagation for aws sns and sqs

wconti27:
## Description

Adds:
- Trace Context Extraction for AWS SQS / SNS / Kinesis

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/contrib/botocore/patch.py
```diff
@@ -481,212 +482,260 @@ def patched_api_call(original_func, instance, args, kwargs):
...
+                func_run = True
+                result = original_func(*args, **kwargs)
+            except Exception as e:
+                func_run_err = e
+            if result is not None and "Records" in result and len(result["Records"]) >= 1:
+                message_received = True
+                ctx = extract_DD_context(result["Records"])
+
+    # we don't want to create a span for empty poll consumer operations
+    if (func_run and message_received) or config.empty_poll_enabled or not func_run:
```
tabgok, line 532: The way this is written is cognitively complex (its not clear what func_run means, for instance, nor how empty_poll_enabled relates to endpoints other than sqs).

I recommend refactoring this by having a "default" trace function that is called for endpoints, then overriding based on the endpoint.

```
if endpoint == sqs:
    special_sqs_trace(*args, **kwargs)
elif endpoint == "events":
    default_trace(*args, **kwargs)
```

ddtrace/contrib/botocore/patch.py
```diff
@@ -481,212 +482,260 @@ def patched_api_call(original_func, instance, args, kwargs):
...
+            trace_operation,
+            service=schematize_service_name("{}.{}".format(pin.service, endpoint_name)),
+            span_type=SpanTypes.HTTP,
+            child_of=ctx if ctx is not None else pin.tracer.context_provider.active(),
+            activate=True,
+        ) as span:
+            span.set_tag_str(COMPONENT, config.botocore.integration_name)
+
+            if start_ns is not None:
+                span.start_ns = start_ns
```
tabgok, line 543: nit: can be `span.start_ns = start_ns or span.start_ns`  (up to you)

ddtrace/contrib/botocore/patch.py
tabgok: The idea is we can have special trace logic per endpoint and make it clear what logic is being applied to which endpoint

ddtrace/contrib/botocore/patch.py
```diff
@@ -481,225 +123,104 @@ def patched_api_call(original_func, instance, args, kwargs):
...
+    elif endpoint_name == "sqs":
+        return patched_sqs_api_call(
+            original_func=original_func,
+            instance=instance,
+            args=args,
+            kwargs=kwargs,
+            function_vars=function_vars,
+        )
+    else:
+        with pin.tracer.trace(
```
tabgok, line 158: Take the code below this and move it into a separate method, maybe "patch_fallthrough" (or something more obvious/clever) so that the `patch()` method is only handling the switching between patching mechanisms.

wconti27: Done!

ddtrace/contrib/botocore/patch.py
wconti27: Done!

tabgok (approved):

ddtrace/contrib/botocore/patch.py
```diff
@@ -72,6 +61,8 @@
         "operations": collections.defaultdict(Config._HTTPServerConfig),
         "tag_no_params": asbool(os.getenv("DD_AWS_TAG_NO_PARAMS", default=False)),
         "instrument_internals": asbool(os.getenv("DD_BOTOCORE_INSTRUMENT_INTERNALS", default=False)),
+        "propagation_enabled": asbool(os.getenv("DD_BOTOCORE_PROPAGATION_ENABLED", default=False)),
```
joeyzhao2018, line 64: Can I ask why the default is False?

joeyzhao2018: Also I wonder if we want to keep using keyword `BOTOCORE`. Does `AWS` or `AWSSDK` makes more sense when we want to keep it consistent for tracers in other languages?

ddtrace/internal/constants.py
```diff
@@ -8,14 +8,16 @@
...
 _PROPAGATION_STYLE_NONE = "none"
-_PROPAGATION_STYLE_DEFAULT = "tracecontext,datadog"
+_PROPAGATION_STYLE_DEFAULT = "tracecontext,datadog,xray"
 PROPAGATION_STYLE_ALL = (
     _PROPAGATION_STYLE_W3C_TRACECONTEXT,
     PROPAGATION_STYLE_DATADOG,
     PROPAGATION_STYLE_B3_MULTI,
     PROPAGATION_STYLE_B3_SINGLE,
     _PROPAGATION_STYLE_NONE,
+    _PROPAGATION_STYLE_AWS_XRAY,
```
joeyzhao2018, line 20: I'm struggling with the idea of adding XRAY as a propagation style here... Can you please add some explanation of how the propagation is used in the descriptions? Thank you 🙇

ddtrace/contrib/botocore/services/sqs.py
```diff
@@ -0,0 +1,279 @@
...
+            if "AttributeNames" not in params:
+                params.update({"AttributeNames": ["AWSTraceHeader"]})
+            elif "AWSTraceHeader" not in params["AttributeNames"]:
+                params.update({"AttributeNames": list(params["AttributeNames"]) + ["AWSTraceHeader"]})
+
+            try:
+                start_ns = time_ns()
+                func_run = True
+                # run the function before in order to extract possible parent context before starting span
+                result = original_func(*args, **kwargs)
```
joeyzhao2018, line 170: The original func call is not inside the tracer.trace(). This seems to be a variation from previous implementation? Or did I misunderstand this?

ddtrace/contrib/botocore/patch.py
wconti27: The default is set to `False` since we cannot change the default user experience for the integration. Adding propagation by default would be considered a breaking change. As for the naming, I don't really have a preference but wanted to stay consistent. We have other configurations using `botocore` and we generally use the contrib name within the configuration name.

ddtrace/internal/constants.py
wconti27: Will add to the description. But basically  `dd-trace-java` propagates all context for AWS integrations via XRay headers. This work was done in [this PR](https://github.com/DataDog/dd-trace-java/pull/3104). In order for a python consumer to extract context from a message created by a Java producer, we need to add this new propagator.  The other way around ( a python producer to a java consumer) already works since the java tracer tries to extract via both the XRay header, and if that isn't present, then the message's attributes, which is how we currently propagate in the Python tracer.

ddtrace/contrib/botocore/services/sqs.py
wconti27: Yes, this was done as a workaround to limitations with how we start spans in Python and was previously discussed with the Python guild. Basically, we need the parent context before starting a span, since there's no functionality to re-parent a span once its created. Only way to do this is by running the original function, then starting the span (with a parent context if one is extracted)

joeyzhao2018: Oh I missed the "ReceiveMessage" part.  This makes sense.

ddtrace/contrib/botocore/patch.py
joeyzhao2018: I think our intentions are the same, that is, not to make it a breaking change. I'll need some time to test this for serverless (AWS Lambda) cases.

ddtrace/propagation/http.py
```diff
@@ -817,12 +819,59 @@ def _inject(span_context, headers):
...
+                # value = int(value) * 1000000
+                # key = "e2e-start-time"
+                pass
+        if trace_id is not None and span_id is not None:
+            return Context(trace_id=trace_id, span_id=span_id, sampling_priority=sampling_priority, dd_origin=dd_origin)
+        else:
+            return None
+
+    @staticmethod
+    def _inject(span_context, headers):
```
mabdinur, line 863: Do we need to support inject for `AWSXRayPropagator`?

mabdinur (commented):
Can we simplify the commits to isolate the new feature from changes to tests and the refactor. This should make this PR easier to review. 

Overall I don't see anything that's a deal breaker

releasenotes/notes/refactor-botocore-contrib-and-add-trace-context-propagation-2876781c977c1377.yaml
```diff
@@ -0,0 +1,7 @@
+---
+features:
+  - |
+    Adds optional feature to propagate context between producers and consumers for AWS SQS, AWS SNS, and AWS Kinesis via `DD_BOTOCORE_PROPAGATION_ENABLED`
+    environment variable. Adds functionality for extracting Datadog trace context from `AWSTraceHeader` to be compatible with `dd-trace-java` propagation. 
+    Adds optional feature to disable tracing of AWS SQS `poll()` operation and AWS Kinesis 'get_records()' operation when no data is consumed via 
+    `DD_BOTOCORE_EMPTY_POLL_ENABLED` environment variable.
```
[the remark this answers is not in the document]
wconti27: Yup, simplified it a bunch

ddtrace/propagation/http.py
```diff
@@ -817,12 +819,59 @@ def _inject(span_context, headers):
...
+                        dd_trace_id = aws_trace_id_segments[i][8:]
+                        trace_id = int(dd_trace_id, 16)
+
+            elif key == "parent":
+                span_id = int(value, 16)
+            elif key == "sampled":
+                sampling_priority = int(value)
+            elif key == "_dd.origin":
+                dd_origin = str(value)
+            elif key == "t0":
```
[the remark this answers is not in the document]
wconti27: removed propagator

ddtrace/propagation/http.py
wconti27: removed propagator

ddtrace/contrib/botocore/services/kinesis.py
```diff
@@ -134,48 +136,79 @@ def patched_kinesis_api_call(original_func, instance, args, kwargs, function_var
...
-            raise
+    message_received = False
+    func_run = False
+    func_run_err = None
+    ctx = None
+    start_ns = None
+    result = None
+
+    if config.botocore.propagation_enabled:
+        if endpoint_name == "kinesis" and operation == "GetRecords":
```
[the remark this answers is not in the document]
wconti27: No, its only for kinesis endpoints. Though I always think erring on the side of caution its okay to have the extra check for safety.

ddtrace/contrib/botocore/services/kinesis.py
```diff
@@ -134,48 +136,79 @@ def patched_kinesis_api_call(original_func, instance, args, kwargs, function_var
...
+            try:
+                start_ns = time_ns()
+                func_run = True
+                result = original_func(*args, **kwargs)
+            except Exception as e:
+                func_run_err = e
+            if result is not None and "Records" in result and len(result["Records"]) >= 1:
+                message_received = True
```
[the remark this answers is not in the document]
wconti27: We are not guaranteed to have a `context` after running the function. Even though we received a message, maybe there was no context within the message, in which case we'd be skipping the span creation.

ddtrace/contrib/botocore/services/kinesis.py
```diff
@@ -134,48 +136,79 @@ def patched_kinesis_api_call(original_func, instance, args, kwargs, function_var
...
+                ctx = extract_DD_context(result["Records"])
+
+    # we don't want to create a span for empty poll consumer operations
+    if (func_run and message_received) or config.empty_poll_enabled or not func_run:
```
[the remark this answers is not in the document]
wconti27: The `not func_run` clause handles the case where we are tracing an operation other than `getRecords`, and `config.empty_poll_enabled` is `false`, in which case the span creation would be skipped if we didn't add this clause.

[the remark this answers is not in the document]
wconti27: For the second part of your comment, I have changed the logic for running the `getRecords` operation. Please see new code, which will always run it before starting the span, which means `config.botocore. propagation_enabled` does not prevent `config.empty_poll_enabled` from working properly.

ddtrace/contrib/botocore/services/kinesis.py
```diff
@@ -134,48 +136,79 @@ def patched_kinesis_api_call(original_func, instance, args, kwargs, function_var
...
+            trace_operation,
+            service=schematize_service_name("{}.{}".format(pin.service, endpoint_name)),
+            span_type=SpanTypes.HTTP,
+            child_of=ctx if ctx is not None else pin.tracer.context_provider.active(),
+            activate=True,
+        ) as span:
+            set_patched_api_call_span_tags(span, instance, args, params, endpoint_name, operation)
+
+            if start_ns is not None:
+                span.start_ns = start_ns
```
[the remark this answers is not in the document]
wconti27: Comment added. I don't think we need to update duration since it is `None` when a span is created and only set once a span is finished. Updating `start_ns` should be the only thing we need AFAIK.

ddtrace/contrib/botocore/services/kinesis.py
```diff
@@ -134,48 +136,79 @@ def patched_kinesis_api_call(original_func, instance, args, kwargs, function_var
...
+        with pin.tracer.start_span(
+            trace_operation,
+            service=schematize_service_name("{}.{}".format(pin.service, endpoint_name)),
+            span_type=SpanTypes.HTTP,
+            child_of=ctx if ctx is not None else pin.tracer.context_provider.active(),
+            activate=True,
+        ) as span:
+            set_patched_api_call_span_tags(span, instance, args, params, endpoint_name, operation)
+
+            if start_ns is not None:
```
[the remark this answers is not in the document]
wconti27: Sure thats fine to add to the condition

ddtrace/contrib/botocore/services/sqs.py
```diff
@@ -141,93 +143,132 @@ def patched_sqs_api_call(original_func, instance, args, kwargs, function_vars):
     endpoint_name = function_vars.get("endpoint_name")
     operation = function_vars.get("operation")
 
-    with pin.tracer.trace(
-        trace_operation,
-        service=schematize_service_name("{}.{}".format(pin.service, endpoint_name)),
-        span_type=SpanTypes.HTTP,
-    ) as span:
-        set_patched_api_call_span_tags(span, instance, args, params, endpoint_name, operation)
+    message_received = False
```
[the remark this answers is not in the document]
wconti27: Done

docs/configuration.rst
```diff
@@ -602,6 +602,23 @@ The following environment variables for the tracer are supported:
...
+         v2.5.0:
+
+   DD_BOTOCORE_EMPTY_POLL_ENABLED:
+      type: Boolean
+      default: True
+      description: |
+         Enables creation of consumer span when AWS SQS and AWS Kinesis ``poll()`` operations return no records. When disabled, no consumer span is created
+         if no records are returned.
+      version_added:
+         v2.5.0:
```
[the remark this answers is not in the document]
wconti27: Done

ddtrace/contrib/botocore/services/kinesis.py
```diff
@@ -134,48 +136,79 @@ def patched_kinesis_api_call(original_func, instance, args, kwargs, function_var
...
-            # `ClientError.response` contains the result, so we can still grab response metadata
-            set_response_metadata_tags(span, e.response)
-
-            # If we have a status code, and the status code is not an error,
-            #   then ignore the exception being raised
-            status_code = span.get_tag(http.STATUS_CODE)
-            if status_code and not config.botocore.operations[span.resource].is_error_code(int(status_code)):
-                span._ignore_exception(botocore.exceptions.ClientError)
-            raise
+    message_received = False
```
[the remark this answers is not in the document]
mabdinur: Discussed this offline. This would introduce a lot of duplication and negatively impact maintainability.  No changes required.

ddtrace/contrib/botocore/services/kinesis.py
```diff
@@ -134,48 +136,86 @@ def patched_kinesis_api_call(original_func, instance, args, kwargs, function_var
...
-            if status_code and not config.botocore.operations[span.resource].is_error_code(int(status_code)):
-                span._ignore_exception(botocore.exceptions.ClientError)
-            raise
+    """
+    We only want to create a span for the following cases:
+        - not func_run: The function is not `getRecords` and we need to run it
+        - func_run and message_received: Received a message when polling
+        - config.empty_poll_enabled: We want to trace empty poll operations
+    """
+    if (func_run and message_received) or config.empty_poll_enabled or not func_run:
```
mabdinur, line 164: Should we do this?

```suggestion
    if (func_run and message_received) or (config.empty_poll_enabled and not func_run):
```

ddtrace/contrib/botocore/services/kinesis.py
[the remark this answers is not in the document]
mabdinur: My main concern here is that the duration will then include all of the tracing/dsm logic. I guess we already do this in other integrations so it is not a big concern

mabdinur (approved):

ddtrace/contrib/botocore/services/kinesis.py
wconti27: No, this doesn't work since:

- If we have `config.empty_poll_enabled=false`, this will never run, regardless of `func_run`

ddtrace/contrib/botocore/services/kinesis.py
```diff
@@ -134,48 +136,79 @@ def patched_kinesis_api_call(original_func, instance, args, kwargs, function_var
...
+                        log.debug("Failed to report data streams monitoring info for kinesis", exc_info=True)
+
+                # raise error if it was encountered before the span was started
+                if func_run_err:
+                    raise func_run_err
+
+                set_response_metadata_tags(span, result)
+                return result
+
+            except botocore.exceptions.ClientError as e:
```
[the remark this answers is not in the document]
wconti27: I don't think so

Yun-Kim (approved):

keigohtr:
Hi,
I randomly found this PR and this was really what we needed.
When will it be released? We are looking forward to this.

PR #7786, opened 2023-11-30
refactor(debugging): support @exception in expressions

P403n1x87:
We add support for the expressions in conditions and log templates to be based on the value of the captured exception, if any.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/debugging/_signal/snapshot.py
```diff
@@ -182,6 +182,8 @@ def exit(self, retval, exc_info, duration):
         _locals = []
         if exc_info[1] is None:
             _locals.append(("@return", retval))
+        else:
+            _locals.append(("@exception", exc_info[1]))
```
shatzi, line 186: why do add @exception to locals here? do we want to capture exc as "@exception" ?

tests/debugging/test_debugger.py
```diff
@@ -1138,3 +1138,38 @@ def test_debugger_redacted_identifiers():
...
+    from tests.submod import stuff
+
+    snapshots = simple_debugger_test(
+        create_snapshot_function_probe(
+            probe_id="probe-instance-method",
+            module="tests.submod.stuff",
+            func_qname="throwexcstuff",
+            evaluate_at=ProbeEvaluateTimingForMethod.EXIT,
+            condition=DDExpression(
+                dsl="expr.__class__.__name__ == 'Expression'",
```
shatzi, line 1157: can we also test this with isUndefined(@excpetion) and instranceof(@exception,'Exception) (not sure the syntax and if RFC is completed around this one)

shatzi (approved):
LGTM

tests/debugging/test_debugger.py
P403n1x87: 🤔 I don't think we have `instanceof` in the DSL?

ddtrace/debugging/_signal/snapshot.py
P403n1x87: I've added this because we were already capturing the return value as `@return` in locals. I don't think it hurts to have it. Don't the other runtimes do this?

tests/debugging/test_debugger.py
P403n1x87: We can add the `isUndefined` test when we make the change to `isDefined`. WDYT?

ddtrace/debugging/_signal/snapshot.py
P403n1x87: From offline discussion: we decided to leave this change out for now, as there is no equivalent in other runtimes at the moment.

shatzi (approved):

Issue #7662, opened 2023-11-17
Trace Encoding Errors when using DD_TRACING_API_VERSION=v0.5


mabdinur:
### Summary of problem

When `DD_TRACING_API_VERSION` is set to `v0.5` a small fraction of spans are misencoded. This results in trace fields being swapped in the decoded payload (ex: resource name is swapped with the operation name). 

### Which version of dd-trace-py are you using?

ddtrace~=1.20.0

### Which version of pip are you using?

n/a

### Which libraries and their versions are you using?

<details>
  <summary>`pip freeze`</summary>
  
</details>

### How can we reproduce your problem?

https://gist.github.com/mabdinur/bf88de6184e359db072766282db786cd

### What is the result that you get?

Decoding errors in the trace agent and traces that fail to send

### What is the result that you expected?

No decoding errors

PR #7874, opened 2023-12-07
fix(v05): reset string table index map after rollbacks

mabdinur:
Resolves: https://github.com/DataDog/dd-trace-py/issues/7662

## Checklist

## Reviewer Checklist

- [ ] Title is accurate.


ddtrace/internal/_encoding.pyx
```diff
@@ -278,13 +278,28 @@ cdef class MsgpackStringTable(StringTable):
...
+        # This will resolve two classes of encoding errors:
+        #  - multiple strings referencing the same string table index. In this scenario
+        #    two different strings in the encoded trace could be serialized with the same _id. In this scenario 
+        #    two different strings could reference one string in the encoded trace (string swapping).
+        # - when the string table references an index of the string table that is not serialized. The encoded 
+        #    trace can not be decoded without accessing an invalid index. In this scenario the agent will 
+        #    return a 400 status code.
+        for string in list(self._table.keys()):
+            if self._table[string] >= self._next_id:
+                del self._table[string]
```
brettlangdon, line 291: we could probably also do a dict comprehension here too:

```python
self._table = {s: idx for s, idx in self._table.items() if idx < self._next_id}
```

just an observation, not sure it is any cleaning. could save on the extra data structures (the list) and operations (all the `del`s), but probably splitting hairs here.

ddtrace/internal/_encoding.pyx
```diff
@@ -278,13 +278,28 @@ cdef class MsgpackStringTable(StringTable):
...
     cdef get_bytes(self):
         cdef int ret
-        cdef stdint.uint32_t table_size = self._next_id
-        cdef int offset = MSGPACK_STRING_TABLE_LENGTH_PREFIX_SIZE - array_prefix_size(table_size)
-        cdef int old_pos = self.pk.length
-
+        cdef stdint.uint32_t table_size
+        cdef int offset
+        cdef int old_pos
         with self._lock:
```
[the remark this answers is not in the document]
mabdinur: I thought this change additional safety would be nice. 

We actually acquire a lock when we flush encoded traces and `get_bytes` is called which this lock is acquired. So a lock in this method is not even necessary.

ddtrace/internal/_encoding.pyx
```diff
@@ -278,13 +278,26 @@ cdef class MsgpackStringTable(StringTable):
...
 
+        # After rolling back the string table next_id we must remove all stale string -> _id pairs
+        # This will resolve two classes of encoding errors:
+        #  - multiple strings referencing the same string table index. In this scenario
+        #    two different strings in the encoded trace could be serialized with the same _id. In this scenario
+        #    two different strings could reference one string in the encoded trace (string swapping).
+        # - when the string table references an index of the string table that is not serialized. The encoded
+        #    trace can not be decoded without accessing an invalid index. In this scenario the agent will
+        #    return a 400 status code.
+        self._table = {s: idx for s, idx in self._table.items() if idx < self._next_id}
```
P403n1x87, line 289: 🤦 good catch! I wonder if a for loop from `_sp_id` to `_next_id` to delete entries from the existing `_table` dict would be more efficient than building a new dict here.

P403n1x87: Ah nvm, we don't have a reverse index here to do that 🙁

mabdinur: We could do that [here](https://github.com/DataDog/dd-trace-py/blob/20a57ab3f3c425ff1f64a59ccb2d5c2c4a61968e/ddtrace/internal/_encoding.pyx#L181-#L185) but it could be harder to maintain. We will also need to update this [contains](https://github.com/DataDog/dd-trace-py/blob/20a57ab3f3c425ff1f64a59ccb2d5c2c4a61968e/ddtrace/internal/_encoding.pyx#L208) method too.

I think it safer to just clear the table here. Users should not be hitting the BufferLimit regularly. If they do they will get noisy error logs asking them to increase the ddtrace buffer size limit.

tests/tracer/test_encoders.py
```diff
@@ -299,7 +299,43 @@ def allencodings(f):
...
+
+    # Encode a trace without triggering a rollback/BufferFull exception
+    ref_encoder = MsgpackEncoderV05(1 << 20, 1 << 20)
+    for tt in encoded_traces:
+        ref_encoder.put(tt)
+
+    # Ensure the two encoders have the same state
+    assert ref_encoder.encode() == rolledback_encoder.encode()
+
+
```
mabdinur, line 338:
```suggestion

@allencodings
```

brettlangdon:
> ## Benchmarks
> Benchmark execution time: 2023-12-07 21:42:32
> 
> Comparing candidate commit [2e65d4d](https://github.com/DataDog/dd-trace-py/commit/2e65d4daa48dc132c93c6dea467c37f0541d3ef5) in PR branch `munir/fix-incident-22445` with baseline commit [4287a9a](https://github.com/DataDog/dd-trace-py/commit/4287a9a8d9c5b4f9fddad4f770acdadf20751a46) in branch `main`.
> 
> Found 8 performance improvements and 0 performance regressions! Performance is the same for 94 metrics, 2 unstable metrics.
> 
> #### scenario:flasksimple-appsec-get
> * 🟩 `max_rss_usage` [-870.926KB; -696.203KB] or [-2.510%; -2.006%]
> 
> #### scenario:flasksimple-appsec-post
> * 🟩 `max_rss_usage` [-907.041KB; -726.034KB] or [-2.609%; -2.088%]
> 
> #### scenario:flasksimple-baseline
> * 🟩 `max_rss_usage` [-1.133MB; -1.012MB] or [-3.267%; -2.919%]
> 
> #### scenario:flasksimple-debugger
> * 🟩 `max_rss_usage` [-1.227MB; -1.035MB] or [-3.541%; -2.986%]
> 
> #### scenario:flasksimple-iast-get
> * 🟩 `max_rss_usage` [-1.142MB; -0.955MB] or [-3.300%; -2.760%]
> 
> #### scenario:flasksimple-profiler
> * 🟩 `max_rss_usage` [-1.131MB; -0.954MB] or [-3.268%; -2.757%]
> 
> #### scenario:flasksimple-tracer
> * 🟩 `max_rss_usage` [-1081.239KB; -902.044KB] or [-3.128%; -2.610%]
> 
> #### scenario:span-add-metrics
> * 🟩 `max_rss_usage` [-24.156MB; -24.006MB] or [-37.826%; -37.590%]

I am curious why this had a positive impact on max rss usage... ?

tests/tracer/test_encoders.py
```diff
@@ -299,6 +299,34 @@ def allencodings(f):
     return pytest.mark.parametrize("encoding", MSGPACK_ENCODERS.keys())(f)
 
 
+def test_msgpack_encoding_after_an_exception_was_raised():
+    """Ensure that the encoder's state is consistent after an Exception is raised during encoding"""
+    # Encode a trace after a rollback/BufferFull occurs exception
+    rolledback_encoder = MsgpackEncoderV05(1 << 12, 1 << 12)
+    trace = gen_trace(nspans=1, ntags=100, nmetrics=100, key_size=10, value_size=10)
+    rand_string = rands(size=20, chars=string.ascii_letters)
+    trace[-1].set_tag_str("some_tag", rand_string)
```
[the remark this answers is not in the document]
mabdinur: > don't we want to write this random string into the first span? if the trace is much larger than the buffer we'll bail before we get to the last span, right?

The trace only has one span now. So we are writing the random string in the first span. I will make this more clear.

> Are we sure rand_string is even being written into the string table at all?

Yes this tag is adding the the random string to the StringTable. If we fail to add strings to a string table a ValueError is raised ([link](https://github.com/DataDog/dd-trace-py/blob/4bff470887108a52aeacbc6d79d2c1ec2a501319/ddtrace/internal/_encoding.pyx#L262)). This will crash the test since we only except `BufferItemTooLarge` to be raised.  This test does not overflow the string. It would overflows the trace buffer.

does that matter with this bug?

Yes. If `rand_string` is not added this test will fail.

mabdinur, line 308:
```suggestion
    # trace only has one span
    trace[0].set_tag_str("some_tag", rand_string)
```

tests/tracer/test_encoders.py
```diff
@@ -299,6 +299,34 @@ def allencodings(f):
...
+    try:
+        # Encode a trace that will trigger a rollback/BufferFull exception
+        rolledback_encoder.put(trace)
+    except (BufferFull, BufferItemTooLarge):
```
mabdinur, line 312:
```suggestion
        # Encode a trace that will trigger a rollback/BufferItemTooLarge exception
        # BufferFull is not raised since only one span is being encoded
        rolledback_encoder.put(trace)
    except BufferItemTooLarge: 
```

tests/tracer/test_encoders.py
```diff
@@ -299,6 +299,34 @@ def allencodings(f):
...
+        pass
+    else:
+        pytest.fail("Expected BufferFull or BufferItemTooLarge exception")
```
mabdinur, line 315:
```suggestion
        pytest.fail("Encoding the trace did not overflow the trace buffer. We should increase the size of the span.")
```

tests/tracer/test_encoders.py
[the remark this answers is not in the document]
brettlangdon: oh, I think I missed the fact that this is one single span....

and... ah... I think the one piece I didn't understand is "when do we bail with buffer too full", it looks like we always encode the full span, and then check _after_ it is encoded if the buffer is now overflowed and then roll back... is that correct?

P403n1x87 (approved):

tests/tracer/test_encoders.py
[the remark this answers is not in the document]
mabdinur: The string table maps strings to indices and encodes this mapping as bytes via msgpack. The "trace buffer" encodes spans as a list of integers and also encodes this data via msgpack. In my reproduction we essentially want the trace buffer to overflow without overflowing the string table.

> I think the one piece I didn't understand is "when do we bail with buffer too full"

We bail on two scenarios:
  1) The string table is full. This [raises](https://github.com/DataDog/dd-trace-py/blob/4bff470887108a52aeacbc6d79d2c1ec2a501319/ddtrace/internal/_encoding.pyx#L262) a ValueError and we immediately stop encoding the trace. We are not testing this scenario.
  2) The trace buffer is full. We only detect this after the trace is already fully encoded ([link](https://github.com/DataDog/dd-trace-py/blob/4bff470887108a52aeacbc6d79d2c1ec2a501319/ddtrace/internal/_encoding.pyx#L506-#L519)). However since this test only encodes one very large span, we actually catch `BufferItemTooLarge`.  But it's essentially the same thing.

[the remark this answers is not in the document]
mabdinur: Another fix we could've implemented is to never rollback and allow payloads to surpass the max size. This was actually my first fix and how I isolated this issue to the rollback functionality. However since the agent can only accept payloads less than 50MB this could be dangerous/inefficient.

PR #7912, opened 2023-12-13
refactor(debugging): use new probe status endpoint

P403n1x87:
We refactor the upload logic to redirect the probe status messages to the new dedicated endpoint on the agent. The upload is done via a multipart POST request with the event.json file carrying the same JSON payload that was being uploaded along with log signals with in the previous implementation.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


emmettbutler (approved):

ddtrace/debugging/_encoding.py
```diff
@@ -74,7 +71,7 @@ def put(self, item: Any) -> int:
         """Enqueue the given item and returns its encoded size."""
 
     @abc.abstractmethod
-    def encode(self) -> Optional[bytes]:
+    def flush(self) -> Optional[bytes]:
         """Encode the given item."""
```
shatzi, line 75: change comment

ddtrace/debugging/_probe/status.py
```diff
@@ -56,39 +72,43 @@ def _payload(
...
+                else:
+                    meter.increment("upload.success")
+                    meter.distribution("upload.size", len(payload))
+        except Exception:
+            log.error("Failed to write payload", exc_info=True)
+            meter.increment("error")
 
+    def _write(self, probe: Probe, status: str, message: str, error: Optional[ErrorInfo] = None) -> None:
+        try:
+            self._write_payload_with_backoff(
```
shatzi, line 97: we are sending single message per probe status?

shatzi (approved):

ddtrace/debugging/_uploader.py
```diff
@@ -84,9 +84,9 @@ def upload(self) -> None:
...
     def periodic(self) -> None:
         """Upload the buffer content to the logs intake."""
-        count = self._encoder.count
+        count = self._queue.count
         if count:
-            payload = self._encoder.encode()
+            payload = self._queue.flush()
             if payload is not None:
                 try:
                     self._write_with_backoff(payload)
```
shatzi, line 92: I find that the implementation of _uploader and _probeStatusLogger while using similar terms and design have different details that make it hard to see. for example _probeStatusLogger use _write_payload_with_backoff that is called inside the flush while this class use flush to only generate payload and then call _write_with_backoff (different name).

I don't think this need to be on this PR but maybe we can look into simplify those two classes?

ddtrace/debugging/_probe/status.py
```diff
@@ -56,53 +73,68 @@ def _payload(
...
+            msgs.append(self._queue.get_nowait())
+
+        try:
+            self._write_payload_with_backoff(
+                multipart(
+                    parts=[
+                        FormData(
+                            name="event",
+                            filename="event.json",
+                            data=f"[{','.join(msgs)}]",
```
shatzi, line 116: are there limit on number of messages we send? as this is not log backend anymore... how big the messages can be?

shatzi (approved):

ddtrace/debugging/_probe/status.py
P403n1x87: On the assumption that we have a hard limit on probes, I don't think we ever risk of getting near the size limit (which I'd expect to be ~ 1 MB).

PR #7919, opened 2023-12-13
feat(botocore): expand support for bedrock

Yun-Kim:
This PR adds tracing support for the `bedrock-runtime` service in botocore. Specifically, this adds tracing support for the `invoke_model` and `invoke_model_with_response_stream` operations in bedrock, and attaches prompt/response tags onto the associated bedrock spans.

### Notes for Reviewers
**The majority of the lines of code changed in this PR are from snapshot test files, riot requirement lockfiles, and test cassette data files. The real number of lines of code changed is closer to ~300 (excluding tests).**

The major files that require reviewing are:
- `ddtrace/contrib/botocore/__init__.py`
- `ddtrace/contrib/botocore/patch.py`
- `ddtrace/contrib/botocore/services/bedrock.py`
- `releasenotes/notes/feat-bedrock-8cf3f9460a529ac6.yaml`
- and `tests/contrib/botocore/test_bedrock.py`.

### Span tagging
The following request tags will be set on Bedrock spans:
Screenshot 2024-01-09 at 3 11 40 PM

The following response tags will be set on Bedrock spans:
Screenshot 2024-01-09 at 3 12 07 PM

### Limitations

Note that AI21 does not have any models that support streamed response streaming, so any requests using that model provider will not have prompt/response tagging.
Bedrock also offers a `stability` provider list of models which are image-modality based. This PR does not include support for those models, but can be added in the future.
Note: the responses returned are in streamed responses, which means that if the user does not fully consume the response, then the span will not be finished. There is the possibility of a memory leak if the number of unfinished spans gets very large, but this is following the wsgi integration example and acceptable for now.

### Future Steps

This PR only introduces tracing support for Bedrock operations. Future PRs will be completed to send prompt/response payloads to LLMObs.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/contrib/botocore/services/bedrock.py
```diff
@@ -0,0 +1,292 @@
...
+    elif provider == "meta":
+        text = body.get("generation")
+        finish_reason = body.get("stop_reason")
+    elif provider == "stability":
+        # TODO: request/response formats are different for image-based models. Defer for now
+        pass
+
+    if not isinstance(text, list):
+        text = [text]
+        finish_reason = [finish_reason]
```
sabrenner, line 173: Should this also be turned into a list with the condition that it already isn't one, as is done in `_extract_streamed_response`? Looks like in the case of `cohere` it might already be a list.

sabrenner (approved):
The contents of `ddtrace/contrib/botocore/services/bedrock.py` LGTM! Just left a small question, but not blocking on it.

ddtrace/contrib/botocore/services/bedrock.py
Yun-Kim: Good catch! Will update.

ddtrace/contrib/botocore/__init__.py
```diff
@@ -27,8 +27,17 @@
 
    Can also be enabled with the ``DD_BOTOCORE_DISTRIBUTED_TRACING`` environment variable.
 
+    Example::
```
Yun-Kim, line 30: TODO: add mlobs as codeowner for bedrock/services.py

ddtrace/contrib/botocore/services/bedrock.py
```diff
@@ -0,0 +1,293 @@
...
+class _BedrockIntegration(BaseLLMIntegration):
+    _integration_name = "bedrock"
+
+
+class TracedBotocoreStreamingBody(wrapt.ObjectProxy):
+    #  Currently the corresponding span finishes only if
+    #  1) the user fully consumes the stream body
+    #  2) error during reading
+
+    def __init__(self, wrapped, span, integration):
```
Yun-Kim, line 23: Note: need to look at https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/tracing/trace_handlers.py#L78 as an example

Yun-Kim: Tested locally, this is a non-issue as checking for `len()` or `streaming_body.__len__` results in an attribute error without having to override `__getattribute__`.

ddtrace/contrib/botocore/services/bedrock.py
```diff
@@ -0,0 +1,297 @@
...
+_META = "meta"
+_STABILITY = "stability"
+
+
+class TracedBotocoreStreamingBody(wrapt.ObjectProxy):
+    #  Currently the corresponding span finishes only if
+    #  1) the user fully consumes the stream body
+    #  2) error during reading
+
+    def __init__(self, wrapped, span, integration):
```
mabdinur, line 27: nit: It would be nice if TracedBotocoreStreamingBody methods had doc strings and typing

ddtrace/contrib/botocore/services/bedrock.py
```diff
@@ -0,0 +1,297 @@
...
+            raise
+
+    def readlines(self):
+        try:
+            lines = self.__wrapped__.readlines()
+            for line in lines:
+                self._body.append(json.loads(line))
+            formatted_response = _extract_response(self._datadog_span, self._body[0])
+            self._process_response(formatted_response)
+            self._datadog_span.finish()
```
mabdinur, line 54: We should move the span finish to a "finally" block

ddtrace/contrib/botocore/services/bedrock.py
```diff
@@ -0,0 +1,297 @@
...
+
+    def __iter__(self):
+        try:
+            for line in self.__wrapped__:
+                self._body.append(json.loads(line["chunk"]["bytes"]))
+                yield line
+            metadata = _extract_streamed_response_metadata(self._datadog_span, self._body)
+            formatted_response = _extract_streamed_response(self._datadog_span, self._body)
+            self._process_response(formatted_response, metadata=metadata)
+            self._datadog_span.finish()
```
mabdinur, line 69: Move this to a finally block. This way if the iterable does not finish and is garbage collected then it will close the span

ddtrace/contrib/botocore/services/bedrock.py
```diff
@@ -0,0 +1,297 @@
...
+            raise
+
+    def _process_response(self, formatted_response: Dict[str, Any], metadata: Dict[str, Any] = None) -> None:
+        """
+        Sets the response tags on the span.
+        """
+        if metadata is not None:
+            for k, v in metadata.items():
+                self._datadog_span.set_tag_str("bedrock.%s" % k, str(v))
```
mabdinur, line 81: nit: we should use f strings in many places in the method. They are more efficient and easier to read

ddtrace/contrib/botocore/services/bedrock.py
```diff
@@ -0,0 +1,297 @@
...
+
+
+def _extract_response(span: Span, body: Dict[str, Any]) -> Dict[str, List[str]]:
+    """
+    Extracts text and finish_reason from the response body, which has different formats for different providers.
+    """
+    text, finish_reason = None, None
+    provider = span.get_tag("bedrock.request.model_provider")
+    if provider == _AI21:
+        text = body.get("completions")[0].get("data").get("text")
```
mabdinur, line 155: Are completions and results guaranteed to have at least one element? Should we check the size of the element first?

ddtrace/contrib/botocore/services/bedrock.py
```diff
@@ -0,0 +1,297 @@
...
+        for i in range(len(text)):
+            span.set_tag_str("bedrock.response.choices.%d.id" % i, str(body.get("generations")[i]["id"]))
+    elif provider == _META:
+        text = body.get("generation")
+        finish_reason = body.get("stop_reason")
+    elif provider == _STABILITY:
+        # TODO: request/response formats are different for image-based models. Defer for now
+        pass
+
+    if not isinstance(text, list):
```
mabdinur, line 175: Could the text be a tuple?

ddtrace/contrib/botocore/services/bedrock.py
```diff
@@ -0,0 +1,297 @@
...
+    if provider == _AI21:
+        pass  # note: ai21 does not support streamed responses
+    elif provider == _AMAZON:
+        text = "".join([chunk["outputText"] for chunk in streamed_body])
+        finish_reason = streamed_body[-1]["completionReason"]
+    elif provider == _ANTHROPIC:
+        text = "".join([chunk["completion"] for chunk in streamed_body])
+        finish_reason = streamed_body[-1]["stop_reason"]
+    elif provider == _COHERE:
+        if "is_finished" in streamed_body[0]:  # streamed response
```
mabdinur, line 198: Should we check the size of streamed_body first?

ddtrace/contrib/botocore/services/bedrock.py
```diff
@@ -0,0 +1,297 @@
...
+    )
+    try:
+        handle_bedrock_request(bedrock_span, integration, params)
+        result = original_func(*args, **kwargs)
+        result = handle_bedrock_response(bedrock_span, integration, result)
+        return result
+    except Exception:
+        bedrock_span.set_exc_info(*sys.exc_info())
+        bedrock_span.finish()
+        raise
```
mabdinur, line 297:
```suggestion
        raise
    finally:
        bedrock_span.finish()
```

riotfile.py
```diff
@@ -1341,12 +1341,18 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
...
-            pys=select_pys(min_version="3.7"),
             pkgs={
                 "moto[all]": latest,
                 "botocore": latest,
                 "pytest-randomly": latest,
             },
+            venvs=[
+                Venv(
+                    pys=select_pys(min_version="3.8"),
+                    pkgs={"vcrpy": latest},
```
mabdinur, line 1352: We should leave a comment to explain why we need a min version of 3.8 here

mabdinur (approved):
left some nits but overall LGTM

ddtrace/contrib/botocore/patch.py
```diff
@@ -152,6 +161,14 @@ def patched_api_call(original_func, instance, args, kwargs):
             kwargs=kwargs,
             function_vars=function_vars,
         )
+    elif endpoint_name == "bedrock-runtime" and operation.startswith("InvokeModel") and config.llmobs_enabled:
```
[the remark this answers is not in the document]
Yun-Kim: Good catch! Yeah, we'll still enable tracing if llmobs is disabled.

ddtrace/contrib/botocore/services/bedrock.py
Yun-Kim: This won't work, since `read(n_bytes)` can read only up to `n_bytes` and not fully consume the stream. We don't want to finish the span always when this method is called, only when the stream is fully read or on error.

ddtrace/contrib/botocore/services/bedrock.py
Yun-Kim: yup, the results are guaranteed to be this format. What I can do instead is to add error handling in case of index/attribute error, and return None on error.

ddtrace/contrib/botocore/services/bedrock.py
Yun-Kim: Nope, it'll always be a str or list

ddtrace/contrib/botocore/services/bedrock.py
Yun-Kim: We can't finish the span on all cases here, since the user has to consume the body before we can finish it.

PR #7921, opened 2023-12-14
chore: dont import telemetry writer from module.py

juanjux:
## Checklist

## Reviewer Checklist

- [x] Title is accurate.


avara1986 (approved):

ddtrace/internal/module.py
```diff
@@ -428,7 +428,7 @@ def after_import(self, module):
             for hook in hooks:
                 hook(module)
         if config._telemetry_enabled and config._telemetry_dependency_collection:
```
P403n1x87, line 430: This is still a potential source of circular import. We should not have any form of "product"-specific details in this general utility module

P403n1x87 (changes requested):

ddtrace/internal/module.py
juanjux: Is the "config"module product-specific? Anyway I can move the checks to the writer.

juanjux: @P403n1x87 change done, PTAL.

juanjux: @P403n1x87 change done, PTAL.

ddtrace/internal/module.py
```diff
@@ -427,8 +426,8 @@ def after_import(self, module):
             log.debug("Calling %d registered hooks on import of module '%s'", len(hooks), module.__name__)
             for hook in hooks:
                 hook(module)
-        if config._telemetry_enabled and config._telemetry_dependency_collection:
-            telemetry_writer._new_dependencies.add(str(module_path))
+
+        _new_imported_modules.add(str(module_path))
```
P403n1x87, line 430: With the new `BaseModuleWatchdog` we can support dedicated logic to be implemented via subclasses of it. Can't we have this logic implemented in a subclass of `BaseModuleWatchdog`, instead of adding to this implementation? For instance, we implement a subclass for some extra behaviour in DI https://github.com/P403n1x87/dd-trace-py/blob/adbc2ce42154c06cb2b45efe19d5ad5b1f2c93eb/ddtrace/debugging/_debugger.py#L87-L88

emmettbutler (approved):

ddtrace/internal/module.py
juanjux: Hi @P403n1x87 can you un-request the change since it's done?

ddtrace/internal/module.py
juanjux: I will take a look at this.

juanjux: After checking it again, I have an idea for a small refactor including a subclass of `BaseModuleWatchdog`. Initially didn't use it because I wanted to get all the modules since the tracer started, thus it made sense to reuse the `ModuleWatchdog` that is already installed instead of installing two pretty similar ones in our `__init__`. But I could subclass and start a `TelemetryDependenciesWatchdog` and initialize it with `sys.modules`. However if you agree I will do it in a next PR because it's a little outside of the scope of this one, that is fixing a potential circular import.

P403n1x87 (approved):

PR #7926, opened 2023-12-14
chore(tracing): report peer hostname or IP in dedicated tags

jbertran:
For peer service computation on the backend, we want to provide tags from which to derivate this information in GRPC spans. We can use the value from which we expose the existing `grpc.host` tag:

* If this value is an IP, then we report it in `network.destination.ip`
* If this value is not an IP, then we report it in `peer.hostname`

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


tests/snapshots/tests.contrib.grpc.test_grpc.test_method_service.json
```diff
@@ -20,6 +20,7 @@
       "grpc.method.service": "Servicer",
       "grpc.status.code": "StatusCode.OK",
       "language": "python",
+      "network.destination.ip": "::",
```
emmettbutler, line 23: This seems like an unexpected value for an IP address field to have. Is it intentional?

jbertran: `::` is IPv6 loopback range, equivalent to `0.0.0.0` for IPv4, and the [test code that generates the trace](https://github.com/DataDog/dd-trace-py/blob/ba7470328a8f7eccff18bc5e2f71e0c485a98642/tests/contrib/grpc/test_grpc.py#L802-L808) does indeed configure it as IPv6.

ddtrace/contrib/grpc/utils.py
```diff
@@ -43,6 +44,12 @@ def set_grpc_method_meta(span, method, method_kind):
 def set_grpc_client_meta(span, host, port):
     if host:
         span.set_tag_str(constants.GRPC_HOST_KEY, host)
+        try:
+            _ = ipaddress.ip_address(host)
```
ZStriker19, line 48: Do we need to assign a variable at all here? 
```suggestion
            ipaddress.ip_address(host)
```

ZStriker19 (approved):
Just a nit/question

PR #7935, opened 2023-12-15
chore: update to envier 0.5.0

P403n1x87:
Update the dependency on envier to the latest 0.5.0 release.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


emmettbutler (approved):
Did you manually edit requirements lockfiles? I'm a bit concerned that this change could make the locks fall out of sync with the riotfile used to generate them. I'm approving anyway because tests are passing and the fix for that issue is to simply run `compile-and-prune-test-requirements`, so the risk of doing something weird here is small.

P403n1x87:
> Did you manually edit requirements lockfiles? I'm a bit concerned that this change could make the locks fall out of sync with the riotfile used to generate them. I'm approving anyway because tests are passing and the fix for that issue is to simply run `compile-and-prune-test-requirements`, so the risk of doing something weird here is small.

Yes I did change the files manually and _then_ run the `compile-and-prune-test-requirements` script. That didn't seem to make any new changes, so I assumed we are fine here. I believe this is because we don't actually have a pin on `envier`, and so no hash is actually affected. I think perhaps we should have an equivalent of the `update` command of various package managers if we want to avoid this manual work in the future.

PR #7936, opened 2023-12-15
chore: use a ModuleWatchdog for the Telemetry imports

juanjux:
## Description

- Use a subclass of `BaseModuleWatchdog` installed on the `TelemetryWriter` instead of modifying the main one.
- Disable telemetry's dependency collection on unrelated telemetry tests using the config var when possible.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/internal/telemetry/writer.py
```diff
@@ -162,6 +164,34 @@ def get_headers(self, request):
...
+        module_path = origin(module)
+        self._new_imported.add(str(module_path))
+
+    def get_new_imports(self):
+        if self._initial:
+            try:
+                # On the first call, use sys.modules to cover all imports before we started. This is not
+                # done on __init__ because we want to do this slow operation on the writer's periodic call
+                # and not on instantiation.
+                new_imports = [str(origin(i)) for i in sys.modules.values()]
```
P403n1x87, line 183: My feeling is that we might need to handle at least `RuntimeError`, like done in https://github.com/P403n1x87/dd-trace-py/blob/acfee55cdec740ba4af79cedddda845580d2afa6/ddtrace/internal/module.py#L405, because `sys.modules` might mutate while iterated upon

ddtrace/internal/telemetry/writer.py
```diff
@@ -205,6 +235,7 @@ def __init__(self, is_periodic=True):
         self._debug = asbool(os.environ.get("DD_TELEMETRY_DEBUG", "false"))
 
         self._client = _TelemetryClient(self.ENDPOINT_V2)
+        self._module_watchdog = TelemetryWriterModuleWatchdog()
```
P403n1x87, line 238: `ModuleWatchdog`s are singletons so we can avoid storing a reference here. We can simply call the `install` class method when `enable` is called, and `uninstall` when `disabled` is called.

P403n1x87 (commented):
🙌  great refactor!

ddtrace/internal/telemetry/writer.py
```diff
@@ -436,8 +472,7 @@ def _flush_integrations_queue(self):
 
     def _flush_new_imported_dependencies(self) -> List[str]:
         with self._lock:
-            new_deps = list(_new_imported_modules)
-            _new_imported_modules.clear()
+            new_deps = self._module_watchdog.get_new_imports()
```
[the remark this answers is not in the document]
juanjux: `_new_imported_dependencies` is removed from this class, the new list is inside the new module watchdog (which does its `clear()` on `get_new_imports`).

ddtrace/internal/telemetry/writer.py
juanjux: Done!

ddtrace/internal/telemetry/writer.py
juanjux: Good advice, didn't notice it was a singleton.

P403n1x87 (approved):

juanjux:
Merging since the previous PR seems to do the CI flaky on telemetry tests and this one should fix that.

PR #7942, opened 2023-12-18
feat(tracer): use qualified name in wrap decorator span name

Toilal:
It seems like a simple way to remove a Todo comment in the source code.

## Checklist

## Reviewer Checklist

- [ ] Title is accurate.

### DIFF
--- a/ddtrace/tracer.py
+++ b/ddtrace/tracer.py
@@ -933,8 +933,7 @@ class Tracer(object):
         """
 
         def wrap_decorator(f: AnyCallable) -> AnyCallable:
-            # FIXME[matt] include the class name for methods.
-            span_name = name if name else "%s.%s" % (f.__module__, f.__name__)
+            span_name = name if name else "%s.%s" % (f.__module__, f.__qualname__)
 
             # detect if the the given function is a coroutine to use the
             # right decorator; this initial check ensures that the


ddtrace/tracer.py
```diff
@@ -933,8 +933,7 @@ def execute():
         """
 
         def wrap_decorator(f: AnyCallable) -> AnyCallable:
-            # FIXME[matt] include the class name for methods.
-            span_name = name if name else "%s.%s" % (f.__module__, f.__name__)
+            span_name = name if name else "%s.%s" % (f.__module__, f.__qualname__)
```
mabdinur, line 936: Thanks for the contribution. 

Is the motivation for this change just to address the FIXME comment or would you like to have this default name on spans generated by `tracer.wrap(...)`. 

Updating the default names of spans could break monitors, dashboards, and metrics for many datadog customers. Instead of  
```suggestion
            span_name = name if name else "%s.%s" % (f.__module__, f.__name__)
```

mabdinur, line 936: Thanks for the contribution 😄 

Is the motivation for this change just to address the comment or would you like to have this default behavior? 

This comment was added years ago when ddtrace was still in a beta state. Updating the default names of spans could break monitors, dashboards, and metrics for many existing datadog customers. Instead of addressing the comment I would prefer to just remove it.

If this feature has the potential to significantly improve your experience with ddtrace then we can hide this change behind an environment variable and make it opt in.

```suggestion
            span_name = name if name else "%s.%s" % (f.__module__, f.__name__)
```

emmettbutler (commented):
I second @mabdinur's comments. This isn't mergeable as-is. Even so, thank you for pointing out this opportunity to clean up some code.

PR #7964, opened 2023-12-20
feat: lazy load instrumentation telemetry everywhere

astuyve:
## Description
In a previous PR, we lazily-loaded `telemetry_writer` during init, unless it was disabled. This is important because it has a very high cold start cost, as detailed in [the pr](https://github.com/DataDog/dd-trace-py/pull/7683).

Unfortunately another [PR](https://github.com/DataDog/dd-trace-py/commit/c328aea335b8a0f319d96220c592ac442c66bb79#diff-b79457a3006017376c6ffaaeb34f3ffafa4f7f12e8f28dc96307faa196fd3e58R659) which landed in the same release also imported this upon init, so the effect of #7683 was basically moot.

I then discovered another path where we import the telemetry writer without checking if it is enabled.

This PR fixes this, and thus we see no telemetry writer loaded in the cold start trace:
image
https://ddserverless.datadoghq.com/apm/trace/14863221237695016030?colorBy=service&env=dev&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=11572349096161165479&spanViewType=metadata&timeHint=1703088776346&view=spans

## Checklist

## Reviewer Checklist

- [x] Title is accurate.

### DIFF
--- a/ddtrace/_monkey.py
+++ b/ddtrace/_monkey.py
@@ -11,9 +11,6 @@ from .settings import _config as config
 from .settings.asm import config as asm_config
 
 
-if config._telemetry_enabled:
-    from .internal import telemetry
-
 if TYPE_CHECKING:  # pragma: no cover
     from typing import Any  # noqa:F401
     from typing import Callable  # noqa:F401
@@ -165,6 +162,9 @@ def _on_import_factory(module, prefix="ddtrace.contrib", raise_errors=True, patc
         try:
             imported_module = importlib.import_module(path)
         except Exception as e:
+            if config._telemetry_enabled:
+                from .internal import telemetry
+
             if raise_errors:
                 raise
             error_msg = "failed to import ddtrace module %r when patching on import" % (path,)
--- a/ddtrace/_monkey.py
+++ b/ddtrace/_monkey.py
@@ -157,14 +157,13 @@ def _on_import_factory(module, prefix="ddtrace.contrib", raise_errors=True, patc
     """Factory to create an import hook for the provided module name"""
 
     def on_import(hook):
+        if config._telemetry_enabled:
+            from .internal import telemetry
         # Import and patch module
         path = "%s.%s" % (prefix, module)
         try:
             imported_module = importlib.import_module(path)
         except Exception as e:
-            if config._telemetry_enabled:
-                from .internal import telemetry
-
             if raise_errors:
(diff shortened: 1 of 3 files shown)


ddtrace/_monkey.py
```diff
@@ -167,23 +169,25 @@ def on_import(hook):
...
+                    versions = imported_module.get_versions()
+                    for name, v in versions.items():
+                        telemetry.telemetry_writer.add_integration(
+                            name, True, PATCH_MODULES.get(module) is True, "", version=v
+                        )
+                else:
+                    version = imported_module.get_version()
                     telemetry.telemetry_writer.add_integration(
-                        name, True, PATCH_MODULES.get(module) is True, "", version=v
+                        module, True, PATCH_MODULES.get(module) is True, "", version=version
```
astuyve, line 189: this wasn't changed, I think github is mis-highlighting here, we simply wrapped the whole block in a conditional

duncanista (approved):
LGTM!

gnufede (approved):

ddtrace/_monkey.py
```diff
@@ -5,13 +5,15 @@
...
 
-from .internal import telemetry
 from .internal.logger import get_logger
 from .internal.utils import formats
 from .settings import _config as config
 from .settings.asm import config as asm_config
 
 
+if config._telemetry_enabled:
+    from .internal import telemetry
```
mabdinur, line 15: nit: We can move this import to where it's used (`on_import` function)

mabdinur (approved):

ddtrace/_monkey.py
astuyve: yeah that's way better

astuyve: well hmmm, I guess technically it's never needed unless the requested module can't be imported. I could load it at the top of `on_import` if you'd rather, but in some cases we may import it and never use it.

Up to you.

mabdinur: No issues here. I just wanted to move the import to where it was used. This way we check if telemetry is enabled in two places instead of 3. Also I can see a world where telemetry is enabled via remote config. If that were to happen with this proposal then we would get an import error.

astuyve: Okay done!

PR #7971, opened 2023-12-21
chore(asm): upgrade test to avoid flakyness

christophe-papazian:
Some value in a remote config test can rarely be None instead of True and both are ok.

This PR adds a new mechanism to allow method `assert_called_with` from `mock` to accept at some point several alternate values as ok. It could be used later in other appsec tests if required.

It also adds some error message to get more info about values that failed to match.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


tests/appsec/utils.py
```diff
@@ -0,0 +1,12 @@
+import sys
+
+
+class Mock_Alternative:
```
gnufede, line 4: nit: I don't love the "Mock_Alternative" name , maybe we call this "Either" or "In" so it would be used as `Either(True, None)` or `In(True, None)`? WDYT?

gnufede (approved):

tests/appsec/utils.py
christophe-papazian: I'm open to a new name. Either is restricted to only two options in english, and In is too short and you're not sure if it starts with an i and an l. Any other suggestions?

christophe-papazian: Well, I went with Either, it looks like the best alternative ;)

PR #8028, opened 2024-01-06
fix(asyncpg): working patch for asyncpg.Pool instances

ekzhang:
Hello! I noticed that using `ddtrace.patch(asyncpg=True)` doesn't work as expected with connection pools, the most common way of using asyncpg database connections.

I'm trying to use `Pin.override()` on the connection object when it is created, which works for direct connections created with `asyncpg.connect()`. However, it does not work with pooled connections.

```python
import asyncpg
import ddtrace
from ddtrace import Pin

ddtrace.patch(asyncpg=True)

async def init_db_connection(conn):
    # print("DEBUG:", getattr(conn, "__getddpin__", None))
    Pin.override(conn, service="my-database-123")  # Let Datadog know what service this is.

pool = await asyncpg.create_pool(dsn="postgres://...", init=init_db_connection)
async with pool.acquire() as conn:
    await conn.fetch(...)
```

This Pin `service` override doesn't work with pools, even though it works with ordinary connections. If you uncomment the `DEBUG:` line above, you can see that the `__getddpin__` method is not present on the connection objects from a pool, even though it is on other asyncpg connections.

I noticed it's because of this [line of code](https://github.com/MagicStack/asyncpg/blob/d7faaff57a7a9c0029a31f09564d30ab35007907/asyncpg/pool.py#L501-L508) from asyncpg.Pool, which calls `asyncpg.connection.connect()`. But ddtrace only patches `asyncpg.connect()`, which is an alias for that.

We need to patch both methods for it to work!

Happy to add a release note, but I can't get `riot run` to work either locally or on a Linux machine. It freezes without making progress. Is there a way to manually add the release note file?

## Checklist

## Reviewer Checklist

- [ ] Title is accurate.

### DIFF
--- a/ddtrace/contrib/asyncpg/patch.py
+++ b/ddtrace/contrib/asyncpg/patch.py
@@ -128,6 +128,7 @@ async def _traced_protocol_execute(asyncpg, pin, func, instance, args, kwargs):
 
 def _patch(asyncpg: ModuleType) -> None:
     wrap(asyncpg, "connect", _traced_connect(asyncpg))
+    wrap(asyncpg.connect_utils, "_connect_addr", _traced_connect(asyncpg))
     for method in ("execute", "bind_execute", "query", "bind_execute_many"):
         wrap(asyncpg.protocol, "Protocol.%s" % method, _traced_protocol_execute(asyncpg))
 
@@ -147,6 +148,7 @@ def patch():
 
 def _unpatch(asyncpg: ModuleType) -> None:
     unwrap(asyncpg, "connect")
+    unwrap(asyncpg.connect_utils, "_connect_addr")
     for method in ("execute", "bind_execute", "query", "bind_execute_many"):
         unwrap(asyncpg.protocol.Protocol, method)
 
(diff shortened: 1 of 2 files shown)


ekzhang:
(cc @aksh-at this is why APM traces are still mixed up)

emmettbutler (changes requested):
Thank you for the contribution, @ekzhang! With a few additions, we could certainly merge this.

* With Docker running, try prefixing your riot command with `scripts/ddtest pip install riot && `. This will run the command inside of a container that is provisioned to be able to make release notes.
* Consider adding a regression test that verifies that the new patching is having the intended effect.

Let me know if you'd like any assistance with either of these. Cheers!

ekzhang:
I just made a best effort to add tests as you advised, though I can't run them at the moment — running tests and generating the release note file name don't seem possible right now as an external contributor since I can't get feedback from your development scripts.

ekzhang:
Hi, we didn't get a response. We have been vendoring this patch in our own codebase now. Given the bugs and inscrutability of datadog's client libraries, we're going to look to other vendors.

PR #8050, opened 2024-01-09
ci(kafka): make test suite more resilient to randomization

emmettbutler:
This pull request increases the reliability of the `kafka` test suite by undoing Pin overrides that change the tracer instance stored on Kafka producer and consumer objects once the test that uses them is finished. I don't have an exact sequence of events figured out, but it seems like the dummy tracer was continuing to issue write retries even after the test using it had completed, polluting other tests' snapshot results. Stopping this from happening appears to have significantly increased the suite's reliability in my testing.

These are other changes that I attempted during debugging. None of them had much of an effect on test reliability, but I've decided to keep them in as solid defensive programming improvements that may make our lives a little easier in the future. These changes are:

* Disabling retries during the sending of traces to the agent, making the number of HTTP requests more predictable
* Setting `config.kafka.trace_empty_poll_enabled=False` in every test where it's applicable. This ensures that extra traces are not generated when multiple `poll()` calls happen, which appears to have been a somewhat common occurrence.
* Removes `poll()` calls from a few tests in which they were redundant with other tests' assertions
* Change all applicable `poll()` calls to avoid setting a timeout and running in a loop; instead calling `poll()` with no arguments to block until a message is received. This reduces the chance of multiple `poll()` invocations (which result in multiple traces, breaking snapshot assertions).
* Slightly relaxes some offset assertions to account for the topic having old data in it from previous tests.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


tests/contrib/kafka/test_kafka.py
```diff
@@ -80,7 +79,10 @@ def empty_kafka_topic(request):
 @pytest.fixture
 def dummy_tracer():
     patch()
-    yield DummyTracer()
+    t = DummyTracer()
+    # disable backoff because it makes these tests less reliable
+    t._writer._send_payload_with_backoff = t._writer._send_payload
```
mabdinur, line 84: Nice catch! Should we also make this the default behavior for all tests that use the global tracer?

This issue is likely impacting all snapshot tests (or maybe update all/most tests to use the dummy tracer).

mabdinur (approved):

tests/contrib/kafka/test_kafka.py
majorgreys: Trying this out:

https://github.com/DataDog/dd-trace-py/commit/77e924afe0a554c2d652953d5af9245a0988f0a6

https://app.circleci.com/pipelines/github/DataDog/dd-trace-py/54570/workflows/57e8fb2b-8f34-4e35-a780-372342834b48

emmettbutler: It's definitely worth a shot. Based on my testing I'm not highly confident that it will help, but it certainly won't hurt most tests.

Issue #8119, opened 2024-01-16
KeyError in ddtrace/internal/utils/cache.py


chase-peach:
### Summary of problem

We are _very rarely_ receiving a KeyError internally to the ddtrace library when attempting to make a grpc call. To my knowledge, it has only happened once since we upgraded to `2.4.0` ~two weeks ago.

### Which version of dd-trace-py are you using?

The latest as of now: `2.4.0`

### Which version of pip are you using?

pip 23.3.1

### Which libraries and their versions are you using?

<details>
  <summary>`pip freeze`</summary>
  
</details>

```
aiohttp==3.9.1
aiohttp-retry==2.8.3
aiosignal==1.3.1
alembic==1.9.1
alembic-autogenerate-enums==0.0.2
appnope==0.1.3
argon2-cffi==21.3.0
argon2-cffi-bindings==21.2.0
asn1crypto==1.5.1
astroid==2.15.8
asttokens==2.4.1
async-timeout==4.0.3
attrs==23.2.0
autopep8==1.7.0
bai2==0.9.2
bcrypt==4.1.2
black==22.12.0
blinker==1.7.0
Brotli==1.1.0
bytecode==0.15.1
cachetools==5.3.2
-e git+https://github.com/peachfinance/python-ach.git@d4113363ea0acb939ce0270d537c651af10d51e6#egg=carta_ach
cattrs==23.2.3
certifi==2023.7.22
certvalidator==0.11.1
cffi==1.16.0
cfgv==3.4.0
charset-normalizer==3.3.2
click==8.1.7
clickclick==20.10.2
cloudpickle==3.0.0
colorama==0.4.6
configparser==6.0.0
connexion==2.14.2
convertdate==2.4.0
coverage==7.3.3
crontab==1.0.0
cryptography==41.0.6
-e git+https://github.com/peachfinance/cybersource-rest-client-python.git@305a50b757ed5eb9dc2b585baef4e15bb1131b95#egg=cybersource_rest_client_python
DateTime==5.4
ddsketch==2.0.4
ddtrace==2.4.0
decorator==5.1.1
defusedxml==0.7.1
Deprecated==1.2.14
dill==0.3.7
distlib==0.3.8
dnspython==2.4.2
docopt==0.6.2
elasticsearch==7.17.9
elementpath==4.1.5
email-validator==1.3.1
endesive==2.0.15
envier==0.5.0
execnet==2.0.2
executing==2.0.1
expandvars==0.9.0
[...]

PR #8120, opened 2024-01-16
fix(internal): make caching thread-safe

P403n1x87:
The current implementation of the caching utilities was not entirely thread-safe. We refactor the logic to make sure that concurrent accesses are fully supported.

Addresses #8119.

## Checklist

## Reviewer Checklist

- [x] Title is accurate.


ddtrace/internal/utils/cache.py
```diff
@@ -38,23 +39,27 @@ def get(self, key, f):  # type: ignore[override]
...
         """
-        if len(self) >= self.maxsize:
-            for _, h in zip(range(self.maxsize >> 1), sorted(self, key=lambda h: self[h][1])):
-                del self[h]
 
         _ = super(LFUCache, self).get(key, miss)
         if _ is not miss:
-            value, count = _
-            self[key] = (value, count + 1)
+            with self.count_lock:
```
mabdinur, line 45: Should we also acquire a lock for the `get(key, miss)`  call?

ddtrace/internal/utils/cache.py
```diff
@@ -38,23 +39,27 @@ def get(self, key, f):  # type: ignore[override]
...
+                value, count = _
+                self[key] = (value, count + 1)
             return value
 
         with self.lock:
             _ = super(LFUCache, self).get(key, miss)
```
mabdinur, line 51: Just curious why do we need to get the cached value again. Can we just use the result from line 43?

releasenotes/notes/fix-thread-safe-cache-b3ccf646feb364c6.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    Fixed an issue that could have caused an exception as a result of a
```
mabdinur, line 4: Can we call out the impact to library and different products. What symptoms does this fix address

mabdinur (approved):
Left some nits but the PR looks good as is

ddtrace/internal/utils/cache.py
P403n1x87: If the function is slow and we have a concurrent access to the same key, the first call to acquire the lock computes the value, so the second can check again when it gets the lock, instead of recomputing the value itself.

ddtrace/internal/utils/cache.py
P403n1x87: the operation is atomic, so we should be good here

majorgreys (approved):

releasenotes/notes/fix-thread-safe-cache-b3ccf646feb364c6.yaml
P403n1x87: that's hard to say really, as the `@cached` decorator is used in many places, so potentially it affected all products

ddtrace/internal/utils/cache.py
```diff
@@ -38,23 +39,27 @@ def get(self, key, f):  # type: ignore[override]
...
-                value, count = _
-                self[key] = (value, count + 1)
+                with self.count_lock:
+                    value, count = _
+                    self[key] = (value, count + 1)
                 return value
 
+            # Cache miss: ensure that we have enough space in the cache
+            while len(self) >= self.maxsize:
+                for h in sorted(self, key=lambda h: self[h][1])[: self.maxsize >> 1]:
```
gnufede, line 60: Some comment regarding how we're making room would be great, as this one-liner is not trivial to follow

gnufede (approved):

PR #8156, opened 2024-01-19
fix(tracing): do not mark StopIteration spans with error

ZStriker19:
Following this: https://github.com/miguelgrinberg/flask-sock/issues/64#issuecomment-1664709789
StopIteration is just a normal expected exception upon socket close. I am wondering if we should re-raise though? If we do, then we need to add additional code so this isn't marked as an error.

## Checklist

## Reviewer Checklist

- [x] Title is accurate

### DIFF
--- a/ddtrace/contrib/redis_utils.py
+++ b/ddtrace/contrib/redis_utils.py
@@ -71,7 +71,7 @@ async def _run_redis_command_async(ctx: core.ExecutionContext, func, args, kwarg
     try:
         result = await func(*args, **kwargs)
         return result
-    except Exception:
+    except BaseException:
         rowcount = 0
         raise
     finally:
--- a/releasenotes/notes/fix-redis-catch-cancelled-error-c39431f98276a0b0.yaml
+++ b/releasenotes/notes/fix-redis-catch-cancelled-error-c39431f98276a0b0.yaml
@@ -1,0 +1,5 @@
+---
+fixes:
+  - |
+    redis: This fix resolves an issue in the redis exception handling where an UnboundLocalError 
+    was raised instead of the expected BaseException.
--- a/tests/contrib/redis/test_redis_asyncio.py
+++ b/tests/contrib/redis/test_redis_asyncio.py
@@ -1,3 +1,4 @@
+import asyncio
 import typing
 from unittest import mock
 
@@ -233,3 +234,12 @@ async def test_client_name(snapshot_context):
         with tracer.trace("web-request", service="test"):
             redis_client = get_redis_instance(10, client_name="testing-client-name")
             await redis_client.get("blah")
+
+
+@pytest.mark.asyncio
+async def test_asyncio_task_cancelled(redis_client):
+    with mock.patch.object(
+        redis.asyncio.connection.ConnectionPool, "get_connection", side_effect=asyncio.CancelledError
+    ):
+        with pytest.raises(asyncio.CancelledError):
+            await redis_client.get("foo")


ddtrace/contrib/wsgi/wsgi.py
```diff
@@ -132,6 +132,9 @@ def blocked_view():
                 core.dispatch("wsgi.request.prepare", (ctx, start_response))
                 try:
                     closing_iterable = self.app(environ, ctx.get_item("intercept_start_response"))
+                # StopIteration should not mark span with error: https://github.com/miguelgrinberg/flask-sock/issues/64
+                except StopIteration:
+                    core.dispatch("wsgi.app.success", (ctx, closing_iterable))
```
mabdinur, line 137: We should reraise this exception. It used to halt the execution of loops and generators

mabdinur: I would expect some tests to fail if we don't reraise here. If everything passed then this is worth looking into further

emmettbutler (changes requested):
I agree with @mabdinur that the least invasive approach to this problem is to allow the exception to continue propagating while also avoiding marking associated spans as errors. I think I did something similar with `SystemExit(0)` in https://github.com/DataDog/dd-trace-py/pull/7557.

ZStriker19:
Agreed, I saw your change but wasn't sure it belonged, after more thought, this seems like the best way.

ddtrace/span.py
```diff
@@ -482,6 +482,9 @@ def set_exc_info(self, exc_type, exc_val, exc_tb):
         # SystemExit(0) is not an error
         if issubclass(exc_type, SystemExit) and exc_val.code == 0:
             return
+        # StopIteration should not mark span as an error: https://github.com/miguelgrinberg/flask-sock/issues/64
+        if exc_type(StopIteration):
```
P403n1x87, line 486:
```suggestion
        if exc_type is StopIteration:
```
?

tests/contrib/wsgi/test_wsgi.py
```diff
@@ -30,11 +30,20 @@ def chunked_response_generator_error(start_response):
             raise GeneratorExit()
 
 
+def raise_stop_iteration_exception(start_response):
+    status = "200 OK"
+    headers = [("Content-type", "text/plain")]
+    start_response(status, headers)
+    raise StopIteration("StopIteration raised in wsgi app")
```
P403n1x87, line 37: A more natural way of generating a `StopIteration` could be
```
>>> def g():yield 42
...
>>> _ = g()
>>> next(_), next(_)
StopIteration
```

tests/contrib/wsgi/test_wsgi.py
```diff
@@ -30,11 +30,25 @@ def chunked_response_generator_error(start_response):
...
+def raise_stop_iteration_exception(start_response):
+    status = "200 OK"
+    headers = [("Content-type", "text/plain")]
+    start_response(status, headers)
+
+    def g():
+        yield 13
+
+    it = g()
+    next(it), next(it)
```
mabdinur, line 42: nit: This endpoint feels a bit off. Can we just raise StopIteration here instead of creating a generator and calling next twice. Some thing like this could also work `next(list().__iter__())`

mabdinur (approved):
left one nit aside from that LGTM

tests/contrib/wsgi/test_wsgi.py
ZStriker19: https://github.com/DataDog/dd-trace-py/pull/8156#discussion_r1463510049 Sure, I think they're basically the same but happy to make it simpler.

Jhoong-Harvey:
Hi @ZStriker19, is this good to merge in?

brettlangdon:
@ZStriker19 the span module was moved, would you be able to update this changeset from `main` so we can work on getting it merged in?

emmettbutler (commented):
I'm going to wait to review until the branch is up to date with the base. @ZStriker19 let me know if you want help with this or if you want me to take ownership of getting this merged.

releasenotes/notes/ignore_stopiteration_exception-43806201b7fc8005.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    wsgi: Previously ``StopIteration`` exceptions which are expected to be raised after closing a websocket connection,
```
mabdinur, line 4: This isn't a wsgi specific fix. It should impact all spans.

```suggestion
    tracing: Ensures the status of a Span is not set to error when a``StopIteration`` exception is raised
```

tabgok:
Replacing with https://github.com/DataDog/dd-trace-py/pull/9788

PR #8185, opened 2024-01-24
chore(di): trigger context-creating probes first

P403n1x87:
The span function probe can create new tracing context, affecting the correlation with the data emitted by other probes (e.g. log, span decoration) wrapped on the same function. We make sure that context-creating probes are triggered before those that only consume the currently active context.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


ddtrace/debugging/_probe/model.py
```diff
@@ -241,7 +243,7 @@ class SpanProbeMixin(object):
 
 @attr.s
 class SpanFunctionProbe(Probe, FunctionLocationMixin, SpanProbeMixin, ProbeConditionMixin):
-    pass
+    __context_mutator__ = True
```
jpbempel, line 246: SpanDecorationProbe are also context-mutating...
though, I need to execute SpanProbe first VS SpanDecorationProbe

P403n1x87: Good point. Here I'm regarding the context as the object that bears the correlation IDs essentially. Mutating the tags on a span is mutating the span itself, and I'm not regarding the span as a context, even though I think the tracer actually does). Perhaps the more correct terminology would be "creation" rather than "mutation", as essentially a span probe _creates_ a new context, rather than mutate (it mutates the context _hierarchy_).

jpbempel (approved):
LGTM but still not convinced by the categorization :)

shatzi (approved):

P403n1x87:
> LGTM but still not convinced by the categorization :)

🙌  Thanks. For now this is just an implementation detail, so shouldn't be an issue 🙂

PR #8203, opened 2024-01-26
chore(profiling): convert libdatadog build apparatus to cmake

sanchda:
This modifies the installation fixture for the "ddup" (libdatadog wrapper) module from normal build_ext to a cmake build_ext.

There's a few reasons for this.
* When I first added libdatadog, our cmake support wasn't quite so rich
* I have an upcoming PR which shares a common native dependency with ddup, so I need a way to specify a dependency between the extensions
* Makes it easier to fine-tune compilation details
* Reviewers of native code components will be more familiar with cmake than with build_ext and friends

## Checklist

## Reviewer Checklist

- [x] Title is accurate


ddtrace/internal/datadog/profiling/CMakeLists.txt
```diff
@@ -0,0 +1,63 @@
+cmake_minimum_required(VERSION 3.19)
+
+# We're building a Python extension and expect to be called from setup.py
+set(EXTENSION_NAME "" CACHE STRING "Name of the extension")
```
danielsn, line 4: Is this a placeholder, or should this be something like "datadog profiling"

ddtrace/internal/datadog/profiling/CMakeLists.txt
```diff
@@ -0,0 +1,63 @@
...
+include(FindPython3)
+add_subdirectory(dd_wrapper "${LIB_INSTALL_DIR}")
+
+# Find the Python interpreter
+find_package(Python3 COMPONENTS Interpreter REQUIRED)
+if (NOT Python3_INCLUDE_DIRS)
+  message(FATAL_ERROR "Python3_INCLUDE_DIRS not found")
+endif()
+
+# Cythonize the .pyx file
```
danielsn, line 23: This comment should probably go on line 28

ddtrace/internal/datadog/profiling/CMakeLists.txt
```diff
@@ -0,0 +1,63 @@
...
+set (ENV{PY_MICRO_VERSION} ${PY_MICRO_VERSION})
+
+add_custom_command(
+    OUTPUT ${CMAKE_CURRENT_LIST_DIR}/_ddup.cpp
+    COMMAND ${PYTHON_EXECUTABLE} -m cython ${CMAKE_CURRENT_LIST_DIR}/_ddup.pyx -o ${CMAKE_CURRENT_LIST_DIR}/_ddup.cpp
+    DEPENDS ${CMAKE_CURRENT_LIST_DIR}/_ddup.pyx
+)
+
+# Specify the target C-extension that we want to build
+# prefix/suffix stuff is to ensure the extension name is literal
```
danielsn, line 35: What does this mean

ddtrace/internal/datadog/profiling/CMakeLists.txt
```diff
@@ -0,0 +1,63 @@
...
+    dd_wrapper/include
+    ${Datadog_INCLUDE_DIRS}
+    ${Python3_INCLUDE_DIRS}
+)
+
+target_link_libraries(${EXTENSION_NAME} PRIVATE
+    dd_wrapper
+)
+
+# We want the extension to be built with position-independent code
```
danielsn, line 54: link as to why?

ddtrace/internal/datadog/profiling/dd_wrapper/include/exporter.hpp
```diff
@@ -208,7 +208,7 @@ class Profile {
   void push_frame(std::string_view name,     // for ddog_prof_Function
                   std::string_view filename, // for ddog_prof_Function
                   uint64_t address,          // for ddog_prof_Location
-                  int64_t line               // for ddog_prof_Line
+                  int64_t line               // for ddog_prof_Location
```
danielsn, line 211: Does a `ddog_prof_Location` have both an `address` and `line`?

ddtrace/internal/datadog/profiling/dd_wrapper/src/exporter.cpp
```diff
@@ -196,11 +196,7 @@ Uploader::upload(const Profile* profile)
     ddog_prof_Exporter_File file[] = {
       {
           .name = to_slice("auto.pprof"),
-          .file =
-              {
-                  .ptr = encoded->buffer.ptr,
-                  .len = encoded->buffer.len,
-              },
+          .file = ddog_Vec_U8_as_slice(&encoded->buffer),
```
danielsn, line 199: nice

ddtrace/internal/datadog/profiling/dd_wrapper/src/exporter.cpp
```diff
@@ -210,8 +206,15 @@ Uploader::upload(const Profile* profile)
...
-      ddog_exporter.get(), start, end, { .ptr = file, .len = 1 }, &tags, nullptr, nullptr, 5000);
+    auto build_res = ddog_prof_Exporter_Request_build(ddog_exporter.get(),
+                                                      start,
+                                                      end,
+                                                      ddog_prof_Exporter_Slice_File_empty(),
+                                                      { .ptr = file, .len = 1 },
+                                                      &tags,
+                                                      nullptr,
+                                                      nullptr,
+                                                      5000);
```
danielsn, line 217: magic constant

ddtrace/internal/datadog/profiling/dd_wrapper/src/exporter.cpp
```diff
@@ -316,8 +319,15 @@ Profile::Profile(ProfileType type, unsigned int _max_nframes)
...
 
-    ddog_prof_Period default_sampler = { samplers[0], 1 }; // Mandated by pprof, but probably unused
-    ddog_profile = ddog_prof_Profile_new({ &samplers[0], samplers.size() }, &default_sampler, nullptr);
+    ddog_prof_Period default_period = { samplers[0], 1 }; // Mandated by pprof, but probably unused
+    ddog_prof_Profile_NewResult res = ddog_prof_Profile_new({ &samplers[0], samplers.size() }, &default_period, nullptr);
+
+    // Check that the profile was created properly
+    if (res.tag != DDOG_PROF_PROFILE_NEW_RESULT_OK) {
+      ddog_Error_drop(&res.err);
+      throw std::runtime_error("Could not create profile");
```
danielsn, line 328: Maybe also take the actual message from the error, to allow better debugging?

ddtrace/internal/datadog/profiling/dd_wrapper/src/exporter.cpp
```diff
@@ -370,30 +383,23 @@ Profile::start_sample(unsigned int nframes)
 void
 Profile::push_frame_impl(std::string_view name, std::string_view filename, uint64_t address, int64_t line)
 {
-    if (cur_frame >= lines.size() || cur_frame >= locations.size())
+    if (cur_frame >= locations.size())
```
danielsn, line 386: How can this happen?

ddtrace/internal/datadog/profiling/dd_wrapper/src/exporter.cpp
```diff
@@ -370,30 +383,23 @@ Profile::start_sample(unsigned int nframes)
...
+        ddog_prof_Function{
+          to_slice(name),
+          {},
+          to_slice(filename),
+          line,
+        },
         address,
-        { &lines[cur_frame], 1 },
-        false,
+        line,
```
danielsn, line 401: Same `line` twice

ddtrace/internal/datadog/profiling/dd_wrapper/src/exporter.cpp
```diff
@@ -472,12 +478,11 @@ Profile::flush_sample()
         .labels = { &labels[0], cur_label },
     };
 
-    ddog_prof_Profile_AddResult address = ddog_prof_Profile_add(ddog_profile, sample);
-    if (address.tag == DDOG_PROF_PROFILE_ADD_RESULT_ERR) {
-        std::string ddog_errmsg(ddog_Error_message(&address.err).ptr);
+    ddog_prof_Profile_Result res = ddog_prof_Profile_add(ddog_profile, sample, 0);
+    if (res.tag == DDOG_PROF_PROFILE_RESULT_ERR) {
+        std::string ddog_errmsg(ddog_Error_message(&res.err).ptr);
```
danielsn, line 483: is this string then used?

ddtrace/internal/datadog/profiling/cmake/tools/fetch_libdatadog.sh
```diff
@@ -0,0 +1,74 @@
+#!/bin/bash
+
+# Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+# This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2021-Present Datadog, Inc.
+
+# http://redsymbol.net/articles/unofficial-bash-strict-mode/
+set -euo pipefail
```
danielsn, line 7: Is setting `-x` useful for debugging?

ddtrace/internal/datadog/profiling/CMakeLists.txt
sanchda: There's actually an elaborate reason for this.  I updated the comment to better reflect the convention imposed on us by our setup fixture.

ddtrace/internal/datadog/profiling/CMakeLists.txt
sanchda: Indeed--thank you.  Fixed

ddtrace/internal/datadog/profiling/CMakeLists.txt
sanchda: Added a clarifying comment, but it has to do with retaining a strong conformance between the name of the extension from the perspective of setup.py versus the name of the project and its assets in cmake

ddtrace/internal/datadog/profiling/CMakeLists.txt
sanchda: This is required for dynamic libraries and python extensions are built as dynamic libraries.

ddtrace/internal/datadog/profiling/dd_wrapper/include/exporter.hpp
sanchda: Yeah, this change was made around version 5 when the structs for the API were flattened.

ddtrace/internal/datadog/profiling/dd_wrapper/src/exporter.cpp
sanchda: Good call--fixed

ddtrace/internal/datadog/profiling/dd_wrapper/src/exporter.cpp
sanchda: This was a rather defensive pattern used as a check because `locations` was a std::array and keeping an array-with-cursor around is somewhat brittle.  This made more sense when we had both lines and locations.

Given the flattening we saw in libdatadog5, I think it makes more sense to refactor the locations buffer as a reserved vector.  This eliminated `cur_frame` as well as `nframes` and simplifies the caller API.  I believe the overall number of destructions for any fixed period of time is equal in the limit either way, so I slightly prefer the simplified code this brought us.

ddtrace/internal/datadog/profiling/dd_wrapper/src/exporter.cpp
sanchda: Indeed, thank you.  The `line` quantity in function is actually the start _line_ in the source of the Function object.  At one point I thought, "we don't have any value for this, so let's just send the current line."  In reality, we don't even  use the start_line value in the BE or UI anywhere, so I just zero out the Function line now.

ddtrace/internal/datadog/profiling/dd_wrapper/src/exporter.cpp
sanchda: Missed that.  No, that was pointless on my part.  I have it printing the string as per the convention used elsewhere in the file, but as we discussed this will be changed to a push-to-logger pattern in a future PR.

ddtrace/internal/datadog/profiling/cmake/tools/fetch_libdatadog.sh
sanchda: Ah, indeed.  Fixed

hatch.toml
```diff
@@ -30,7 +30,7 @@ dependencies = [
 
 [envs.lint.scripts]
 black_check = [
-    "black --check {args:.}",
+    "black {args:.}",
```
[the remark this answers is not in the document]
sanchda: Absolutely unwanted, sorry

ddtrace/internal/datadog/profiling/dd_wrapper/src/exporter.cpp
```diff
@@ -347,61 +363,54 @@ Profile::insert_or_get(std::string_view sv)
...
-
-    locations[cur_frame] = {
-        {},
+    ddog_prof_Location loc = {
+        null_mapping, // No support for mappings in Python
+        {
+          to_slice(name),
+          {}, // No support for system name in Python
+          to_slice(filename),
+          0 // We don't know the start_line in the typical case
```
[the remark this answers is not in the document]
sanchda: AFAIK, named (or designated) initializers (forms 3 and 4 [here](https://en.cppreference.com/w/cpp/language/aggregate_initialization)) are a compiler extension until C++20 (this is somewhat annoying because it's been in C since C11, so it can be relied on in some other contexts).  I can only generally assume C++17 for now in our CI builds, and there's also the tradeoff between using "better" standards versus what end-users can reliably get in their environments.

Personally, I like named initializers and I think it improves readability, which is why I wrote it that way the first time.

ddtrace/internal/datadog/profiling/dd_wrapper/src/exporter.cpp
sanchda: Fixed

ddtrace/internal/datadog/profiling/dd_wrapper/src/exporter.cpp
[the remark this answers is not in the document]
sanchda: Added back in [b27f89b](https://github.com/DataDog/dd-trace-py/pull/8203/commits/b27f89bfdf9c5c5594345438e34b6cb0af317337)

danielsn (approved):

emmettbutler (approved):

setup.py
```diff
@@ -325,14 +288,20 @@ def build_extension_cmake(self, ext):
 
         # We derive the cmake build directory from the output directory, but put it in
         # a sibling directory to avoid polluting the final package
-        cmake_build_dir = os.path.abspath(self.build_lib.replace("lib.", "cmake."))
+        cmake_build_dir = os.path.abspath(os.path.join(self.build_lib.replace("lib.", "cmake."), ext.name))
```
P403n1x87, line 291: It would be great if we could use the equivalent utilities from the `pathlib` library

setup.py
```diff
@@ -369,18 +340,10 @@ def build_extension_cmake(self, ext):
...
-            subprocess.run([cmake_command, "--install", ".", *install_args], cwd=cmake_build_dir, check=True)
-        except subprocess.CalledProcessError as e:
-            print("WARNING: Command '{}' returned non-zero exit status {}.".format(e.cmd, e.returncode))
-            if not ext.permissive_build:
-                raise
-        except Exception:
-            print("WARNING: An error occurred while building the CMake extension.")
-            raise
+        cmake_command = os.environ.get("CMAKE_COMMAND", "cmake")
+        subprocess.run([cmake_command, *cmake_args], cwd=cmake_build_dir, check=True)
```
P403n1x87, line 344: Wondering if we could benefit from a `cmake` utility function to DRY this up a bit?

ddtrace/internal/datadog/profiling/ddup.py
P403n1x87, line 1: Nit: these signature changes seem unrelated to the PR at first sight. Can they be taken out into their own PR by any chance? If they are too tied to this change I think it's fine to leave them here though.

P403n1x87 (commented):
Left a few questions, but overall looking great! Thanks for this!

setup.py
sanchda: I took a stab at doing this in places where the precise sequence of operations was better captured in pathlib:
```
-HERE = os.path.dirname(os.path.abspath(__file__))
+HERE = Path(__file__).resolve().parent

 DEBUG_COMPILE = "DD_COMPILE_DEBUG" in os.environ

@@ -283,12 +284,12 @@ class CMakeBuild(build_ext):

     def build_extension_cmake(self, ext):
         # Define the build and output directories
-        output_dir = os.path.abspath(os.path.dirname(self.get_ext_fullpath(ext.name)))
-        extension_basename = os.path.basename(self.get_ext_fullpath(ext.name))
+        output_dir = Path(self.get_ext_fullpath(ext.name)).parent.resolve()
+        extension_basename = Path(self.get_ext_fullpath(ext.name)).name

         # We derive the cmake build directory from the output directory, but put it in
         # a sibling directory to avoid polluting the final package
-        cmake_build_dir = os.path.abspath(os.path.join(self.build_lib.replace("lib.", "cmake."), ext.name))
+        cmake_build_dir = Path(self.build_lib.replace("lib.", "cmake."), ext.name).resolve()
```

Is this what you were thinking, or do you think we should replace the `os.path.join()` constructs with pathlib as well, now that we're using it anyway (in other words, standardize on a single filesystem path management abstraction)?

ddtrace/internal/datadog/profiling/ddup.py
sanchda: Fair question.  My hand was forced wrt the underlying implementation, but I could just ignore the superfluous argument and submit a different PR for the signature changes.  It would keep this PR "pure" in the sense that we'd isolate the changes to the inner extension rather than the ddtrace .py code.

It doesn't take any time or really change anything, so I'll split the interface changes out from here.

setup.py
sanchda: That's a really good suggestion, but it would take me a while to clean this up properly.  Any complaints about doing this in a separate PR?

ddtrace/internal/datadog/profiling/ddup.py
sanchda: OK, I did what I said I'd do.  The Python-facing API is now undisturbed--we'll fix this in a followup PR

ddtrace/internal/datadog/profiling/dd_wrapper/src/exporter.cpp
```diff
@@ -316,19 +326,30 @@ Profile::Profile(ProfileType type, unsigned int _max_nframes)
...
     reset();
-
-    // Initialize the size for buffers
-    cur_frame = 0;
 }
 
 Profile::~Profile()
 {
-    ddog_prof_Profile_drop(ddog_profile);
+    ddog_prof_Profile_drop(&ddog_profile);
```
danielsn, line 352: How did this typecheck before!

danielsn (approved):

ddtrace/internal/datadog/profiling/dd_wrapper/src/exporter.cpp
sanchda: I changed the definition of `ddog_profile` in the class, so the overall type is the same.  The argument is still a pointer, but the underlying member no longer takes the value of a stack-allocated pointer.

Originally this wasn't a problem because the libdatadog3 interface returned the pointer directly (it was heap-allocated).  It also wasn't a problem in the add_echion PR (which is where I split this one from) since I read the docs and did it properly.

However, when I broke apart that PR, I took a half-measure in this PR which gave `ddog_profile` a pointer to a stack-allocated value.

The really, really curious thing is that I didn't get any segfaults from _THAT_, which is a complete mystery.  My best hypothesis is that there just so happens to be some kind of void in the stack and I got rid of it when I twiddled the interface.  Or I was testing the wrong thing.

Either way, fixed now.  🙃

setup.py
```diff
@@ -32,25 +34,20 @@
...
 
 DEBUG_COMPILE = "DD_COMPILE_DEBUG" in os.environ
 
 IS_PYSTON = hasattr(sys, "pyston_version_info")
 
-LIBDDWAF_DOWNLOAD_DIR = os.path.join(HERE, os.path.join("ddtrace", "appsec", "_ddwaf", "libddwaf"))
-IAST_DIR = os.path.join(HERE, os.path.join("ddtrace", "appsec", "_iast", "_taint_tracking"))
+LIBDDWAF_DOWNLOAD_DIR = os.path.join(HERE, "ddtrace", "appsec", "_ddwaf", "libddwaf")
+IAST_DIR = os.path.join(HERE, "ddtrace", "appsec", "_iast", "_taint_tracking")
+DDUP_DIR = os.path.join(HERE, "ddtrace", "internal", "datadog", "profiling")
```
P403n1x87, line 45:
```suggestion
LIBDDWAF_DOWNLOAD_DIR = HERE / "ddtrace" / "appsec" / "_ddwaf" / "libddwaf"
IAST_DIR = HERE / "ddtrace" / "appsec" / "_iast" / "_taint_tracking"
DDUP_DIR = HERE / "ddtrace" / "internal" / "datadog" / "profiling"
```

P403n1x87 (commented):
Just a nit otherwise LGTM

setup.py
P403n1x87: Ah those might have to be converted to `str` when used perhaps 🤔

sanchda: Converted setup.py to use pathlib when possible in [bc4fc7b](https://github.com/DataDog/dd-trace-py/pull/8203/commits/bc4fc7bb0ab855c6f65529a942528fc0524d7ffc)

P403n1x87 (approved):
👍

PR #8207, opened 2024-01-29
fix(ci_visibility): don't enable CIVisibility on API 403

romainkomorndatadog:
This changes the `CIVisibility` service to no longer succesfully `.enable()` if we encounter a `403` status code during API requests.

This change can be made due to the fact that we now call our settings API early on in the enabling process, and can make the decision not to continue enabling. Prior to this, sending data was best effort, but getting a 403 from the API means that subsequent queries (and span submissions) likely won't succeed, so there's no point in enabling the service.

Instead, we just supply an error message:
```
[Datadog CI Visibility] WARNING  ddtrace.internal.ci_visibility.recorder:recorder.py:510 Authentication error, disabling CI Visibility, please check Datadog API key
```

This isn't considered a breaking change because the end result (data not being submitted successfully) is the same.

This has been tested both using `DD_CIVISIBILITY_AGENTLESS_ENABLED=1` and an invalid API key, as well as with it set to `0`, relying on EVP proxy mode to a Datadog agent with an invalid API key.

This PR also updates a bunch of tests to properly mock the function that talks to our production settings API, since these would generate 403s, which in the past was invisible, but now would cause CI Visibility to be disabled.

This change also introduces a set of CI Visibility-specific integration environments which gives a little bit better visibility at the cost some delicious, though unfortunate, copy/pasta which.

Other note: this is labeled as a "fix" in the PR title because our PR title check script doesn't allow for it to start with `upgrade`.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


romainkomorndatadog:
I convinced myself this was gonna be a small change and here we are...

I don't think this should be backported, but I also don't think it's a "feature"... but since it has some impact, it's also not a "chore", so release-note-wise, I'm not sure.

juan-fernandez (approved):

ZStriker19:
At first, I was worried this relied on hitting the agent, in which case the tracer might start up and query before the agent was ready, but since it relies on the site, which should be ever present as long as networking is setup beforehand (which should be always), this change seems good to me.

ZStriker19 (approved):

romainkomorndatadog:
> At first, I was worried this relied on hitting the agent

Actually, that's a pretty good point, @ZStriker19 . We do rely on the agent in EVP proxy mode, but since this is specifically about a `403` status code specifically, we wouldn't get that unless the agent was up and able to proxy requests to our API, so it's essentially "the same".

romainkomorndatadog:
Checked off the checklist myself since it's been a few days and I've got two approvals.

PR #8249, opened 2024-01-31
fix(logging): disallow log propagation [APMS-11072]

emmettbutler:
This change stops `DDLogger` from propagating to its parents, which had been causing duplicate entries to be logged.

Here's a simple way to reproduce the duplicate logging behavior that this fixes:
```
import logging

from fastapi import FastAPI

app = FastAPI()

@app.get("/")
async def read_root():
    logging.warning("requested")
    return {"Hello": "World"}
```
Install requirements files from APMS-11072, and run with `ddtrace-run uvicorn app.main:app` and observe the following log output:
```
WARNING:root:requested
INFO:     127.0.0.1:64577 - "GET / HTTP/1.1" 200 OK
failed to send, dropping 1 traces to intake at http://localhost:8126/v0.5/traces after 3 retries
ERROR:ddtrace.internal.writer.writer:failed to send, dropping 1 traces to intake at http://localhost:8126/v0.5/traces after 3 retries
```
The last two lines are duplicates.

## Checklist

## Reviewer Checklist

- [ ] Title is accurate

### DIFF
--- a/ddtrace/internal/logger.py
+++ b/ddtrace/internal/logger.py
@@ -180,9 +180,9 @@ class DDLogger(logging.Logger):
                     if record.levelno >= hdlr.level:
                         hdlr.handle(record)
                 if not c.propagate or found >= 1:
-                    c = None  # break out
+                    c = None  # type: ignore[assignment]
                 else:
-                    c = c.parent
+                    c = c.parent  # type: ignore[assignment]
         else:
             # Increment the count of records we have skipped
             # DEV: `self.buckets[key]` is a tuple which is immutable so recreate instead
--- a/ddtrace/internal/logger.py
+++ b/ddtrace/internal/logger.py
@@ -172,7 +172,17 @@ class DDLogger(logging.Logger):
             self.buckets[key] = DDLogger.LoggingBucket(current_bucket, 0)
 
             # Call the base handle to actually log this record
-            super(DDLogger, self).handle(record)
+            c = self
+            found = 0
+            while c:
+                for hdlr in c.handlers:
+                    found = found + 1
+                    if record.levelno >= hdlr.level:
+                        hdlr.handle(record)
+                if not c.propagate or found >= 1:
+                    c = None  # break out
+                else:
+                    c = c.parent
         else:
             # Increment the count of records we have skipped
             # DEV: `self.buckets[key]` is a tuple which is immutable so recreate instead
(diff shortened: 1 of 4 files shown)


emmettbutler:
@majorgreys @ZStriker19 @erikayasuda what do you think of this? I'm surprised at how few tests failed when I made this change, but this can't be the ideal solution to duplicate log messages....right?

ddtrace/internal/logger.py
```diff
@@ -172,7 +172,17 @@ def handle(self, record):
...
-            super(DDLogger, self).handle(record)
+            c = self
+            found = 0
+            while c:
+                for hdlr in c.handlers:
+                    found = found + 1
+                    if record.levelno >= hdlr.level:
+                        hdlr.handle(record)
+                if not c.propagate or found >= 1:
+                    c = None  # type: ignore[assignment]
```
romainkomorndatadog, line 183: nit: I would `break` here to make it explicit that we're bailing out to the loop.

romainkomorndatadog (commented):
Are we still double-logging in the case of where the rate limit is not set, since we do
```
        if not self.rate_limit or self.getEffectiveLevel() == logging.DEBUG:
            super(DDLogger, self).handle(record)
            return
```
in `handle()`?

Overall I have mixed feelings about this.

I thought our philosophy here was to let end users deal with logging (including double logging). In #7955 , I introduced logic to ["take over the stream handler"](https://github.com/DataDog/dd-trace-py/pull/7955/files#diff-e530cfd776475242f4bd13c2087721cd66ede8ef7208ac3aadab49dea5444c32R96). I have some mixed feelings about it.

I'm tempted to say that, if users don't want double logging, we should give them an option to just disable our handler, rather than do our own propagation handling.

emmettbutler:
@romainkomorndatadog I've updated the description with a reproduction case that I'm pretty sure illustrates how this is more complicated than disabling the Datadog logging handler. Let me know what you think.

romainkomorndatadog:
@emmettbutler , that's what the handler-removal code I added in the [CIVIsibility utils](https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/internal/ci_visibility/utils.py#L116-L120) deals with.

In that logic, there's still potential double-logging since I'm just replacing our logger's, but if I simply remove our handler without replacing it, the double logging issue goes away.

emmettbutler:
Oh cool, I didn't know about that. I guess the fix for this issue is actually just a documentation change https://github.com/DataDog/dd-trace-py/pull/8266

PR #8265, opened 2024-02-02
fix: only use first message context for trace propagation in aws messaging

wconti27:
Small change to only use first message's extracted context during propagation for AWS messaging services

Also fixes bug where if context is not found, the span is created with a new context. After the fix, if a context is not found, the active context is used.

## Checklist

## Reviewer Checklist

- [x] Title is accurate

### DIFF
--- a/ddtrace/contrib/botocore/utils.py
+++ b/ddtrace/contrib/botocore/utils.py
@@ -197,7 +197,9 @@ def extract_DD_context(messages):
         message = messages[0]
         context_json = extract_trace_context_json(message)
         if context_json is not None:
-            ctx = HTTPPropagator.extract(context_json)
+            child_of = HTTPPropagator.extract(context_json)
+            if child_of.trace_id is not None:
+                ctx = child_of
     return ctx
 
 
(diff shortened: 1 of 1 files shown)


ddtrace/contrib/botocore/utils.py
```diff
@@ -193,23 +193,11 @@ def set_response_metadata_tags(span, result):
 
 def extract_DD_context(messages):
     ctx = None
-    if len(messages) == 1:
+    if len(messages) >= 1:
         message = messages[0]
         context_json = extract_trace_context_json(message)
```
mabdinur, line 198: Could a payload messages with different trace contexts? If this is a valid scenario should we try to reconcile differences (maybe through span links)?

Is there a valid scenario where the trace context is only sent in the 2nd, 3rd, 4th, .... message?

mabdinur (commented):
This change will also require a release note

ddtrace/contrib/botocore/utils.py
wconti27: This is a valid scenario, but not one that we're targeting due to the limitations of current trace context propagation. For now, we want to be consistent with other languages, which will only use the first message's context. We will address these problems once we get to the implementation of span links

mabdinur (approved):

PR #8306, opened 2024-02-06
chore(profiling): make ddup safe against state-tearing during fork in a native thread

sanchda:
This PR makes some sweeping changes to the organization of the "ddup" extension in order to make it more robust against state-tearing issues when native threads are interrupted during fork() operations.

The test apparatus for this PR is included within the native code itself. I'll incorporate these tests into a later PR when this gets closer to user testing (for now: no need to add latency + flakiness to everyone else...)

This PR prepares things for a later PR which will add a pure-native stack sampling mechanism (stack sampling v2).

A changelog entry will go into a later PR in which those stack-sampling features are made available.

## Checklist

- [X] Change(s) are motivated and described in the PR description
- [X] Testing strategy is described if automated tests are not included in the PR
- [X] Risks are described (performance impact, potential for breakage, maintainability)
- [X] Documentation is included (in-code, generated user docs, [public corp docs](https://github.com/DataDog/documentation/))
- [X] If this PR changes the public interface, I've notified `@DataDog/apm-tees`.
- [X] If change touches code that signs or publishes builds or packages, or handles credentials of any kind, I've requested a review from `@DataDog/security-design-and-guidance`.

## Reviewer Checklist

- [ ] Title is accurate


ddtrace/internal/datadog/profiling/dd_wrapper/test/threading.cpp
```diff
@@ -0,0 +1,149 @@
...
+    ddup_config_profiler_version("3.100");
+    ddup_config_max_nframes(256);
+    ddup_init();
+
+    // Creates two threads, waits for 1 second, then uploads
+    done.store(false);
+    std::thread t1(emulate_sampler);
+    std::thread t2(emulate_sampler);
+    std::thread t3(emulate_sampler);
+    std::thread t4(emulate_sampler);
```
danielsn, line 123: Why not do this in a loop?

ddtrace/internal/datadog/profiling/dd_wrapper/test/api.cpp
```diff
@@ -0,0 +1,201 @@
...
+lotsa_frames_lotsa_samples()
+{
+    ddup_config_service("my_test_service");
+    ddup_config_env("my_test_env");
+    ddup_config_version("0.0.1");
+    ddup_config_url("https://localhost:8126");
+    ddup_config_runtime("cpython");
+    ddup_config_runtime_version("3.10.6");
+    ddup_config_profiler_version("3.100");
+    ddup_config_max_nframes(512);
```
danielsn, line 155: This is repeated.  Helper function?

ddtrace/internal/datadog/profiling/dd_wrapper/src/uploader_builder.cpp
```diff
@@ -0,0 +1,124 @@
...
+{
+    std::lock_guard<std::mutex> lock(mtx);
+    if (!_runtime_version.empty())
+        runtime_version = _runtime_version;
+    return *this;
+}
+UploaderBuilder&
+UploaderBuilder::set_profiler_version(std::string_view _profiler_version)
+{
+    std::lock_guard<std::mutex> lock(mtx);
```
danielsn, line 54: are you going to repeatedly acquire and release the mutex if you have `.set_profiler_version().set_runtime_version()` etc?

ddtrace/internal/datadog/profiling/dd_wrapper/src/uploader_builder.cpp
```diff
@@ -0,0 +1,124 @@
...
+    return *this;
+}
+
+std::unique_ptr<Uploader>
+UploaderBuilder::build_ptr()
+{
+    // Setup the ddog_Exporter
+    ddog_Vec_Tag tags = ddog_Vec_Tag_new();
+
+    // Add the tags, emitting the first failure
```
danielsn, line 91: where do we emit the failure?

ddtrace/internal/datadog/profiling/dd_wrapper/src/uploader.cpp
```diff
@@ -0,0 +1,78 @@
...
+        errmsg = err_to_msg(&result.err, "Error serializing pprof");
+        ddog_Error_drop(&result.err);
+        std::cerr << errmsg << std::endl;
+        return false;
+    }
+    ddog_prof_EncodedProfile* encoded = &result.ok;
+
+    // If we have any custom tags, set them now
+    ddog_Vec_Tag tags = ddog_Vec_Tag_new();
+    add_tag(tags, ExportTagKey::profile_seq, std::to_string(profile_seq), errmsg);
```
danielsn, line 33: What if these fail?

ddtrace/internal/datadog/profiling/dd_wrapper/src/uploader.cpp
```diff
@@ -0,0 +1,78 @@
+#include "uploader.hpp"
+#include "libdatadog_helpers.hpp"
+
+using namespace Datadog;
+
+void
+DdogProfExporterDeleter::operator()(ddog_prof_Exporter* ptr) const
```
danielsn, line 7: where is this used?

ddtrace/internal/datadog/profiling/dd_wrapper/test/forking.cpp
```diff
@@ -0,0 +1,106 @@
...
+    ddup_config_runtime("cpython");
+    ddup_config_runtime_version("3.10.6");
+    ddup_config_profiler_version("3.100");
+    ddup_config_max_nframes(256);
+    ddup_init();
+
+    // Creates two threads
+    done.store(false);
+    std::thread t1(emulate_sampler);
+    std::thread t2(emulate_sampler);
```
danielsn, line 67: Would it be interesting to create `n` threads?

ddtrace/internal/datadog/profiling/dd_wrapper/src/sample_builder.cpp
```diff
@@ -0,0 +1,40 @@
...
+}
+
+SampleBuilder&
+SampleBuilder::set_max_nframes(unsigned int _max_nframes)
+{
+    const unsigned int backend_max_nframes = 512;
+
+    if (_max_nframes > 0)
+        max_nframes = _max_nframes;
+    if (max_nframes > backend_max_nframes)
```
danielsn, line 27: Should this signal an error?

ddtrace/internal/datadog/profiling/dd_wrapper/src/sample.cpp
```diff
@@ -0,0 +1,336 @@
+#include "sample.hpp"
+#include "global_cache.hpp"
+
+#include <thread>
+
+using namespace Datadog;
+
+Sample::Sample(SampleType type, unsigned int _max_nframes)
```
danielsn, line 8: does this happen under a lock?

ddtrace/internal/datadog/profiling/dd_wrapper/src/sample.cpp
```diff
@@ -0,0 +1,336 @@
...
+    };
+
+    locations.emplace_back(loc);
+}
+
+void
+Sample::push_frame(std::string_view name, std::string_view filename, uint64_t address, int64_t line)
+{
+
+    if (locations.size() <= max_nframes)
```
danielsn, line 75: otherwise we just drop it?

ddtrace/internal/datadog/profiling/dd_wrapper/src/global_cache.cpp
```diff
@@ -0,0 +1,52 @@
...
+GlobalCache::get(std::thread::id id)
+{
+    GlobalCache& global_store = get_singleton();
+
+    std::unique_lock<std::mutex> lock(global_store.sample_storage_mtx);
+    auto& samples = global_store.sample_cache;
+    auto it = samples.find(id); // Attempt to find the sample once
+
+    if (it == samples.end()) {
+        // Sample not found, unlock the mutex while creating the sample
```
danielsn, line 25: is the issue that the `sample_builder` gets a lock and is not reentrant?

ddtrace/internal/datadog/profiling/dd_wrapper/src/global_cache.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+        auto new_sample = global_store.sample_builder.build();
+        if (!new_sample) {
+            throw std::runtime_error("Error creating sample");
+        }
+
+        // Lock again to insert the new sample
+        lock.lock();
+        auto [new_it, inserted] = samples.emplace(
+          std::piecewise_construct, std::forward_as_tuple(id), std::forward_as_tuple(std::move(new_sample.value())));
```
danielsn, line 36: modern C++ makes my eyes hurt

ddtrace/internal/datadog/profiling/dd_wrapper/src/interface.cpp
```diff
@@ -1,33 +1,39 @@
...
+// Store a pointer to the current profile.  This can be assumed set in inner
+// functions, but any function that may be called by a new thread should check
+thread_local Datadog::Sample* g_profile;
+
+// When a fork is detected, we need to reinitialize this state.
+// This handler will be called in the single thread of the child process after the fork
+void
+ddup_fork_handler()
+{
+    Datadog::Profile::mark_dirty();
```
danielsn, line 25: don't we also need a barrier for the uploader?

ddtrace/internal/datadog/profiling/dd_wrapper/src/interface.cpp
```diff
@@ -139,18 +161,26 @@ ddup_init()
...
-        g_profile_real[1] = profile_builder.build_ptr();
-        g_profile = g_profile_real[g_prof_flag];
-        g_uploader = uploader_builder.build_ptr();
-        is_initialized = true;
+        // install the ddup_fork_handler for pthread_atfork
+        // Right now, only do things in the child _after_ fork
+        pthread_atfork(nullptr, nullptr, ddup_fork_handler);
+
+        // Set the global initialization flag
+        is_ddup_initialized = true;
```
danielsn, line 169: Does this need to be an atomic write

ddtrace/internal/datadog/profiling/dd_wrapper/src/interface.cpp
```diff
@@ -139,18 +161,26 @@ ddup_init()
...
+        return true;
+    }();
+
+    if (initialized)
+        ++initialized_count;
```
danielsn, line 174: does this need to be atomic?

ddtrace/internal/datadog/profiling/dd_wrapper/src/profile.cpp
```diff
@@ -0,0 +1,233 @@
...
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+using namespace Datadog;
+
+void
+Profile::reset()
+{
+    // Drop the profiles
```
danielsn, line 17: this doesn't need a lock?

ddtrace/internal/datadog/profiling/dd_wrapper/src/profile.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+Profile::~Profile()
+{
+    reset();
+}
+
+bool
+Profile::cycle_buffers()
+{
+    std::lock_guard<std::mutex> lock(profile_mtx);
```
danielsn, line 38: why is this the only thing that needs a mutex

ddtrace/internal/datadog/profiling/cmake/FindCppcheck.cmake
```diff
@@ -0,0 +1,8 @@
+include(ExternalProject)
+
+ExternalProject_Add(cppcheck_project
+  GIT_REPOSITORY https://github.com/danmar/cppcheck.git
+  GIT_TAG "2.13.3"
+  CMAKE_ARGS -DCMAKE_INSTALL_PREFIX=${CMAKE_BINARY_DIR}/cppcheck
```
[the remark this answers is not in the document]
sanchda: That's correct.  Ordinarily, I'd shove a prebuilt cppcheck into a docker container or pull it from the repo, but so far I had managed to avoid adding additional dependencies, and even though this isn't incorporated into CI yet, it's nice not to depend too much on my own system's setup.

I'm not 100% sure how this will get handled in the sequel.  There are a few other things happening around splitting dd-trace-py which may make this easier.

ddtrace/internal/datadog/profiling/dd_wrapper/CMakeLists.txt
```diff
@@ -7,45 +7,64 @@ project(dd_wrapper
...
     src/interface.cpp
 )
 
+# If we were given flags to do sanitization or static analyis, use them
+if (SANITIZE_OPTIONS)
+  if (SANITIZE_OPTIONS MATCHES "fanalyzer")
+    target_compile_options(dd_wrapper PRIVATE -fanalyzer)
+  elseif (SANITIZE_OPTIONS MATCHES "cppcheck")
+    set(DO_CPPCHECK ON)
+  else()
```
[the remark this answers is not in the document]
sanchda: First of all, I have some uncertainty around the future of this precise code.  I believe what we'll end up doing is standardizing all of the logic in cmake and just calling specific arguments from CI rather than having a full-on wrapper to help build things.  That's what cmake is for.

As for your question, the legal values are

- the string "fanalyzer" which is only valid for GCC and enables GCC's `-fanalyze` static analysis flag
- the string "cppcheck" which will not add any sanitizers, but will build cppcheck (if needed) and run it afterward
- everything else is treated as a valid argument to the clang frontend's `-fsanitize=` option (and also as a linker flag); whether or not _those_ are valid depends on what [clang accepts](https://clang.llvm.org/docs/UsersManual.html#controlling-code-generation), which varies from build-to-build (for example, 17 has a leak analyzer which is separate from memsan).

Ultimately, I think we'll pin this to whatever features we guarantee in the build process.

Not sure if I answered the question, hit me back if I missed the point.

ddtrace/internal/datadog/profiling/dd_wrapper/dd_wrapper.md
```diff
@@ -0,0 +1,74 @@
+# dd_wrapper
+
+This is experimental code that uses libdatadog for the collection and propagation of profiling data.  In theory this should just be a thin wrapper, but in practice some more state management.
```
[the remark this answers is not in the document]
sanchda: It seems I accidentally a whole

ddtrace/internal/datadog/profiling/dd_wrapper/dd_wrapper.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Historical context
+
+Traditionally, dd-trace-py maintained three separate sampling mechanisms for the collection of profiling data.  These are called "collectors" in dd-trace-py parlance, but we'll refer to them as "samplers" in conformance with the broader litearture.
+
+### Stack sampler
+
+This is a simple stack sampler that collects stack traces at a dynamic interval.  Supposedly, it's a fixed-interval profiler, but the interval is adjusted dynamically depending on the observed overhead.  This sampler lives in a Python thread.  When it wakes up, it looks at the current threads/tasks and collects stack traces for each of them.  It also looks at the current exceptions and collects similar data for them.
```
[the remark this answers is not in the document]
sanchda: In a sense, yeah.  Note that although this is a description of the "v1" stack sampler, I do have support for propagating this context to echion in the echion PR.

What the noted method does is increase the amount of walltime between samples, which in turn associates more walltime (and almost definitely CPU time) to the subsequent sample.

It's a form of upscaling, but I don't prefer these instantaneous measures--they get really sloppy.  Having dynamically scaling sampling intervals all the time can wreak havoc on data fidelity, and I bet we introduce bias because the sampling rate becomes tightly coupled to some underlying behavior.

I had a vision to replace this mechanism with, at the very least, a PID controller.  That's TBD.

ddtrace/internal/datadog/profiling/dd_wrapper/dd_wrapper.md
```diff
@@ -0,0 +1,74 @@
...
+* Execution resumes from the same point as the parent.  One way to think about this is, execution resumes from the thread that called fork
+* The heap is configured to be copied-on-write.
+
+This means any intermediate heap-stored state managed by a sibling thread (i.e., a thread which did not call `fork()`) may become corrupt in the child.  We need to protect or re-initialize such state when possible.
+
+## Data organization in this code
+
+Roughly, the data is organized in the following way
+
+### Samples
```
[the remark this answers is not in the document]
sanchda: It will emit two separate samples with identical stack traces.  Each individual string will get de-duplicated only one time (global storage), and each stack trace will get unwound separately.

There have been issues where this was a problem for some reason or another, but right now I think the overriding issue is a lost performance opportunity (e.g., only unwind once).

ddtrace/internal/datadog/profiling/dd_wrapper/src/interface.cpp
sanchda: Good question, thank you!

This block is assigned to a variable with [static local storage duration](https://en.cppreference.com/w/cpp/language/storage_duration#Static_local_variables).  The standard guarantees that this section will be initialized only once, and strictly by precisely on thread.  This applies to the body of the lambda as well.  This wasn't your question, and you may know this already, but I wanted to get that piece out of the way just in case.

I don't think it would hurt, but I thought to limit the number of atomic variables for no particular reason.  If this initialization block is running in one thread while _another_ thread is already charging toward an upload, then the upload will deflect.

I didn't think putting the update slightly earlier in the memory pipeline will affect this much--we're not exposing any intermediate states, at least.

ddtrace/internal/datadog/profiling/dd_wrapper/src/interface.cpp
sanchda: It should be, yeah.  The purpose of this value is just to report on the pathological scenario where initialization somehow happens multiple times.  From that perspective, mission accomplished no matter if we shed _extra_ increments every now and again.  However, the precise number could be significant for analysis, in which case we definitely don't want to drop it.

Thank you, I'll fix this.

ddtrace/internal/datadog/profiling/dd_wrapper/test_build.sh
```diff
@@ -0,0 +1,157 @@
...
+      echo "Unknown option: $1"
+      exit 1
+      ;;
+  esac
+else
+  set_gcc
+fi
+
+# If there are two arguments, override build mode
+BUILD_MODE=${2:-Debug}
```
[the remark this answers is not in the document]
sanchda: It is now

ddtrace/internal/datadog/profiling/dd_wrapper/src/global_cache.cpp
sanchda: Mine too.  Removed this since I was able to simplify the global_cache out of this code entirely.

ddtrace/internal/datadog/profiling/dd_wrapper/dd_wrapper.md
```diff
@@ -0,0 +1,74 @@
...
+
+1. `ddup_start_sample()` (every sample begins with a call to `start_sample()`)
+2. adding frames, walltime, tags, etc
+3. ddup_flush_sample() (every sample ends with a call to `flush_sample()`)
+
+The act of flushing a sample stores its data in a ddog_prof_Profile object (which is wrapped by Profile in this code).
+
+There's one wrinkle here.  The navigation through frame data (unwinding) may result in temporary strings.  We need to cache these strings.  In order to minimize overhead, strings are cached (and de-duplicated) in a cache attached to the Profile rather than the Sample.
+
+After `start_sample()`, the current sample in the current thread is saved in `thread_local` storage, removing the need for the thread to call into global cache.
```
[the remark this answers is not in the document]
sanchda: This is a fundamental question and I answered it by dropping the cache entirely.

Here's the deal, as far as I understand it.

On Linux, thread-local storage is just another form of heap allocation.  The threading runtime (NPTL) sets it up when a thread is made and associates that storage to a given thread at runtime.  C++ hijacks this mechanism for `thread_local`.  Post-fork(), nothing special is done to these arenas--it's heap memory, so the user could have sent pointers around in their application, it can't get cleaned up.

On the other hand, provided the user did not do a silly thing (like _not_ propagate pointers to TLS memory), those pages (I believe TLS areas are page-aligned and allocated individually) are CoW, meaning that the marginal overhead to the system for keeping them in the child is nearly zero.

I knew this in my first cut and struggled to form a conviction, but after reading your comment I realized that in the average case a user calling `fork()` will be running a fork-based worker pool with regeneration.  The children won't spawn children of their own.  I think the marginal value of preventing useless TLS copies is very close to zero.  Ignoring this problem simplifies the code a lot.

ddtrace/internal/datadog/profiling/dd_wrapper/src/uploader_builder.cpp
sanchda: Indeed, and doing so would be rather pointless (it's a loser-takes-all situation no matter how you slice the variants here, since we are doing trivial assignment).

I removed these locks.

ddtrace/internal/datadog/profiling/dd_wrapper/src/uploader_builder.cpp
sanchda: I refactored this.  Now we check _all_ the tags and send a big blob of all the failures, with the understanding that the average user will only have to fix one thing, but the _really annoyed_ user will have to fix several things iteratively.

How we log this is still a problem though.

ddtrace/internal/datadog/profiling/dd_wrapper/src/uploader.cpp
sanchda: Blink and you'll miss it!  The `ddog_exporter` member of Uploader is a unique_ptr.  The template allows you to specify an optional destructor which will be called to cleanup the thing the pointer points to when it's done pointing.  That's what this does.

ddtrace/internal/datadog/profiling/dd_wrapper/src/sample_builder.cpp
sanchda: It should definitely log a warning

ddtrace/internal/datadog/profiling/dd_wrapper/src/sample.cpp
sanchda: Good catch!

To the first degree, yes.

To the second degree, yes, we should also _not_ stash the function/filename stuff, but we absolutely should keep track of how many frames we dropped, since we need to create a virtual frame with that information.

ddtrace/internal/datadog/profiling/dd_wrapper/src/global_cache.cpp
sanchda: The main issue was just performance.  Creating a sample takes time, so reduce the lock surface.  This is somewhat silly because bouncing all over the lock is arguably worse than just holding it for some additional amount of time, but since this is a one-time call per thread, maybe not too important.

Either way, I removed all this code.  I don't think we need it.  Added comments in the .md

sanchda:
Something about my build here is goofy, since I'm getting

```
ValueError: Cannot repair wheel, because required library "libdd_wrapper.so" could not be located
```

Will look into it.  I think the PR is safe to review as before, even with that defect, though.

ddtrace/internal/datadog/profiling/dd_wrapper/dd_wrapper.md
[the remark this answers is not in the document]
sanchda: (fixed btw)

ddtrace/internal/datadog/profiling/dd_wrapper/dd_wrapper.md
```diff
@@ -0,0 +1,74 @@
...
+
+This means any intermediate heap-stored state managed by a sibling thread (i.e., a thread which did not call `fork()`) may become corrupt in the child.  We need to protect or re-initialize such state when possible.
+
+## Data organization in this code
+
+Roughly, the data is organized in the following way
+
+### Samples
+
+Samples are stored on the heap.  They are managed by a GlobalCache object.  GlobalCache provides a `get()` method which takes a thread ID and returns a Sample (or creates a sample if needed).  Samples form the basis of a single collection operation.
```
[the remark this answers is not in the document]
sanchda: Yup!  Any and all thread-local storage is leaked after `fork()`.  The idea is that TLS is managed on individual pages, and those pages will never be accessed (unless the old TLS data is referenced in the new process somehow), so CoW will ensure that the minimum copies are maintained.

In reality, it's not quite so nice.  If the parent process mutates its own TLS page, then the copy will occur in the parent, so all the children are left hanging onto dead references.

That said,
* The use of `fork()` is primarily motivated by worker processes in Python, not stacks of children
* We try to keep only one sample per thread

ddtrace/internal/datadog/profiling/dd_wrapper/include/libdatadog_helpers.hpp
```diff
@@ -0,0 +1,131 @@
...
+        errmsg = err_to_msg(&res.err, "Error pushing tag");
+        errmsg += "(val:'" + std::string(val) + "')";
+        ddog_Error_drop(&res.err);
+        return false;
+    }
+    return true;
+}
+
+inline static bool
+add_tag_unsafe(ddog_Vec_Tag& tags, std::string_view key, std::string_view val, std::string& errmsg)
```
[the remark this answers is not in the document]
sanchda: I factored this differently as a result of other comments, but I changed the names without changing the fact that there are two species of `add_tag()`

variant 1:  takes an enum and a value.  This is safer because the enum is something we control and it's maintained in static storage.  The value may be passed by the user though, so I'd hardly call this "safe."  Mistake of naming in the original.

variant 2:  takes two string-like values.  Both are provided by the user, and each may have separate syntax checks in the backend.

I changed this so that 1 uses the inner implementation of 2 and just pulls the key names from static storage.

ddtrace/internal/datadog/profiling/dd_wrapper/include/libdatadog_helpers.hpp
```diff
@@ -0,0 +1,131 @@
...
+        std::cout << errmsg << std::endl;
+        return false;
+    }
+    return true;
+}
+
+inline static std::string_view
+str_from_key(const ExportLabelKey key)
+{
+    constexpr std::array<std::string_view, static_cast<size_t>(ExportLabelKey::_Length)> keys = { EXPORTER_LABELS(
```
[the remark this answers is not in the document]
sanchda: Right--specifically, it's a compile-time constant, so we're returning a chunk of memory that's interned in the binary itself.

ddtrace/internal/datadog/profiling/dd_wrapper/src/uploader.cpp
sanchda: Fixed

ddtrace/internal/datadog/profiling/dd_wrapper/test/forking.cpp
sanchda: I'm not sure.  I think better tests need to be defined, these are very rudimentary and I didn't even try to dress them up as anything but pathetic.

ddtrace/internal/datadog/profiling/dd_wrapper/src/interface.cpp
sanchda: Ah, indeed.  I had originally thought of doing this with a barrier, but I changed it to an exclusive lock to protect new uploads from sneaking in during the `fork()` operation.

ddtrace/internal/datadog/profiling/dd_wrapper/include/uploader.hpp
```diff
@@ -17,16 +20,25 @@ struct DdogProfExporterDeleter
...
+
+  public:
     size_t profile_seq = 0;
     std::string runtime_id;
     std::unique_ptr<ddog_prof_Exporter, DdogProfExporterDeleter> ddog_exporter;
     std::string url;
 
     std::string errmsg;
 
+    static inline ddog_CancellationToken* cancel{ nullptr };
```
danielsn, line 34: The token is static?

ddtrace/internal/datadog/profiling/dd_wrapper/src/interface.cpp
```diff
@@ -25,9 +25,27 @@ thread_local std::optional<Datadog::Sample> g_profile = std::nullopt;
...
+
+// Since we don't control the internal state of libdatadog's exporter and we want to prevent state-tearing
+// after a `fork()`, we just outright prevent uploads from happening during a `fork()` operation.  Since a
+// new upload may be scheduled _just_ as `fork()` is entered, we also need to prevent new uploads from happening.
+// This mutex is released in both the child and the parent.
+void
+ddup_prefork()
+{
+    Datadog::Uploader::lock();
+    Datadog::Uploader::cancel_inflight();
```
danielsn, line 48: Is this synchronous or async?  If async, we might still have state tearing?

ddtrace/internal/datadog/profiling/dd_wrapper/include/libdatadog_helpers.hpp
```diff
@@ -0,0 +1,131 @@
...
+inline std::string_view
+to_string(ExportTagKey key)
+{
+    constexpr size_t num_keys = static_cast<size_t>(ExportTagKey::_Length);
+    constexpr std::array<std::string_view, num_keys> keys = { EXPORTER_TAGS(X_STR) };
+    constexpr std::string_view invalid = "";
+
+    if (static_cast<size_t>(key) >= num_keys)
+        return invalid;
+    return keys[static_cast<size_t>(key)];
```
danielsn, line 84:
```suggestion
    size_t key_idx = static_cast<size_t>(key);

    if (key_idx) >= num_keys)
        return invalid;
    return keys[key_idx];
```

ddtrace/internal/datadog/profiling/dd_wrapper/src/uploader_builder.cpp
```diff
@@ -0,0 +1,141 @@
...
+    if (!_runtime_id.empty())
+        runtime_id = _runtime_id;
+}
+
+std::string
+join(const std::vector<std::string>& vec, const std::string& delim)
+{
+    return std::accumulate(
+      vec.begin(), vec.end(), std::string(), [&delim](const std::string& a, const std::string& b) -> std::string {
+          if (!a.empty()) // Use !a.empty() for clarity
```
danielsn, line 78: What is `b.empty()`?

ddtrace/internal/datadog/profiling/dd_wrapper/src/uploader.cpp
```diff
@@ -0,0 +1,116 @@
...
+        errmsg = err_to_msg(&result.err, "Error serializing pprof");
+        ddog_Error_drop(&result.err);
+        std::cerr << errmsg << std::endl;
+        return false;
+    }
+    ddog_prof_EncodedProfile* encoded = &result.ok;
+
+    // If we have any custom tags, set them now
+    ddog_Vec_Tag tags = ddog_Vec_Tag_new();
+    add_tag(tags, ExportTagKey::runtime_id, runtime_id, errmsg);
```
danielsn, line 33: What if this fails?

ddtrace/internal/datadog/profiling/dd_wrapper/src/sample.cpp
```diff
@@ -0,0 +1,324 @@
...
+    std::fill(std::begin(labels), std::end(labels), ddog_prof_Label{});
+    locations.clear();
+    cur_label = 0;
+    dropped_frames = 0;
+}
+
+bool
+Sample::flush_sample()
+{
+    // Samples are local to threads, meaning they are local to processes.  We can only arrive here if
```
danielsn, line 127: Is this still true

ddtrace/internal/datadog/profiling/dd_wrapper/src/interface.cpp
```diff
@@ -1,300 +1,301 @@
...
+
+    initialized_count.fetch_add(static_cast<int>(initialized));
+    if (initialized_count > 1) {
+        std::cerr << "ddup_init() called " << initialized_count << " times" << std::endl;
     }
 }
 
-void
-ddup_start_sample()
+unsigned int
```
danielsn, line 128: Why not just return the actual enum?  Operating over the enum would prevent most of these casts

ddtrace/internal/datadog/profiling/dd_wrapper/include/uploader.hpp
danielsn: I'm a little nervous about all the static state we're dangling around.  Can make it hard to test

ddtrace/internal/datadog/profiling/dd_wrapper/include/types.hpp
```diff
@@ -0,0 +1,34 @@
...
+    Wall = 1 << 1,
+    Exception = 1 << 2,
+    LockAcquire = 1 << 3,
+    LockRelease = 1 << 4,
+    Allocation = 1 << 5,
+    Heap = 1 << 6,
+    All = CPU | Wall | Exception | LockAcquire | LockRelease | Allocation | Heap
+};
+
+// Rather than wrapping and unwrapping values when they're flushed, we just use a vector.
```
danielsn, line 16: Where is the vector?

ddtrace/internal/datadog/profiling/dd_wrapper/include/sample_manager.hpp
```diff
@@ -0,0 +1,254 @@
...
+    _Length,
+    Invalid,
+};
+
+using SampleStorage = std::vector<Sample>;
+
+class SampleManager
+{
+  private:
+    static inline std::array<std::atomic<bool>, static_cast<size_t>(SampleHandle::_Length)> handler_state{};
```
danielsn, line 31: Do we want everything `static`

ddtrace/internal/datadog/profiling/dd_wrapper/include/libdatadog_helpers.hpp
```diff
@@ -0,0 +1,135 @@
...
+    auto ddog_err = ddog_Error_message(err);
+    std::string err_msg;
+    return std::string{ msg } + " (" + err_msg.assign(ddog_err.ptr, ddog_err.ptr + ddog_err.len) + ")";
+}
+
+inline std::string_view
+to_string(ExportTagKey key)
+{
+    constexpr auto num_keys = static_cast<size_t>(ExportTagKey::Length_);
+    constexpr std::array<std::string_view, num_keys> keys = { EXPORTER_TAGS(X_STR) };
```
r1viollet, line 79: Minor, the invalid key could be part of the tags to help with debug when retrieving the profile.

ddtrace/internal/datadog/profiling/dd_wrapper/include/profile.hpp
```diff
@@ -0,0 +1,90 @@
...
+#include <vector>
+
+extern "C"
+{
+#include "datadog/profiling.h"
+}
+
+namespace Datadog {
+
+constexpr unsigned int g_default_max_nframes = 128;
```
r1viollet, line 21: Is there another python max defined (with a value of 100) ? I am not weighing in on what we should do, just trying to check what standards are defined in python.

ddtrace/internal/datadog/profiling/dd_wrapper/include/profile.hpp
```diff
@@ -0,0 +1,90 @@
...
+  private:
+    // Serialization for static state
+    // - string table
+    // - ddog_profile
+    std::atomic<bool> first_time{ true };
+    std::mutex profile_mtx{};
+    std::atomic<pid_t> current_pid{ 0 };
+
+    // Storage for strings
+    std::deque<std::string> string_storage{};
```
r1viollet, line 39: I think we have already talked about this (sorry if I forgot the conclusion)
Why do we not make the storage the std::set (and put standard strings in there, as well as make string views hashable) ?

ddtrace/internal/datadog/profiling/dd_wrapper/src/profile.cpp
```diff
@@ -0,0 +1,228 @@
...
+        std::cerr << "Error initializing bottom half of profile storage" << std::endl;
+        return;
+    }
+
+    // We're done. Don't do this again.
+    first_time.store(false);
+}
+
+std::string_view
+Profile::insert_or_get(std::string_view sv)
```
r1viollet, line 189: When we sample with echion, we have stable strings at least for the lifetime of that sample ? 
Can some of the responsibility on string hashing be pushed to libdatadog ?
Meaning, we just use string views on the echion strings, then we let libdatadog intern things ?

ddtrace/internal/datadog/profiling/dd_wrapper/src/profile.cpp
```diff
@@ -0,0 +1,228 @@
...
+    // non-RAII heap storage has been dropped, state has been reinitialized
+    // we can pretend it's the first time all over again
+    first_time.store(true);
+}
+
+bool
+Profile::cycle_buffers()
+{
+    const std::lock_guard<std::mutex> lock(profile_mtx);
+    std::swap(last_profile, cur_profile);
```
r1viollet, line 56: several threads can update the cur_profile ?

ddtrace/internal/datadog/profiling/dd_wrapper/src/sample.cpp
```diff
@@ -0,0 +1,323 @@
...
+
+bool
+Sample::push_release(int64_t lock_time, int64_t count)
+{
+    if (type_mask & SampleType::LockRelease) {
+        values[profile_state.val().lock_release_time] += lock_time;
+        values[profile_state.val().lock_release_count] += count;
+        return true;
+    }
+    std::cout << "bad push release" << std::endl;
```
r1viollet, line 203: I know this is a probably a temporary workaround for logging. 
Perhaps we can just plug a log_once to stderr before we figure out how to do the logger compatible with python ?

ddtrace/internal/datadog/profiling/dd_wrapper/src/profile.cpp
r1viollet: obviously this has to work beyond echion.
Though the same can be true for samples ? ++ref on that python string, --ref when we push the sample to libdatadog ?

ddtrace/internal/datadog/profiling/dd_wrapper/src/profile.cpp
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    // We're done. Don't do this again.
+    first_time.store(false);
+}
+
+std::string_view
+Profile::insert_or_get(std::string_view sv)
+{
+    const std::lock_guard<std::mutex> lock(string_table_mtx); // Serialize access
```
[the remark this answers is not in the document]
r1viollet: Ah yeah, I just looked at the code around dirty. Sorry I had not read that yet.

sanchda:
> I was not able to complete the review. I'll come back to this. Thanks for publishing the PR.

This code got far too big for what it did and I struggled to carve it up into smaller components which would be individually correct.  I appreciate your time, thank you.

sanchda:
In the middle of refactoring, needs more work.

ddtrace/internal/datadog/profiling/dd_wrapper/include/profile.hpp
```diff
@@ -0,0 +1,90 @@
...
+
+// Serves to collect individual samples, as well as lengthen the scope of string data
+class Profile
+{
+  private:
+    // Serialization for static state
+    // - string table
+    // - ddog_profile
+    std::atomic<bool> first_time{ true };
+    std::atomic<long> g_profile_seq{ 0 };
```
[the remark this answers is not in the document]
sanchda: Eliminated profile_seq for now

ddtrace/internal/datadog/profiling/dd_wrapper/test/threading.cpp
sanchda: Refactored

ddtrace/internal/datadog/profiling/dd_wrapper/test/api.cpp
sanchda: Refactored

ddtrace/internal/datadog/profiling/dd_wrapper/src/uploader_builder.cpp
sanchda: Removed

ddtrace/internal/datadog/profiling/dd_wrapper/src/uploader.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+bool
+Uploader::upload(ddog_prof_Profile& profile)
+{
+    // Serialize the profile
+    ddog_prof_Profile_SerializeResult result = ddog_prof_Profile_serialize(&profile, nullptr, nullptr, nullptr);
+    if (result.tag != DDOG_PROF_PROFILE_SERIALIZE_RESULT_OK) {
+        errmsg = err_to_msg(&result.err, "Error serializing pprof");
+        ddog_Error_drop(&result.err);
+        std::cerr << errmsg << std::endl;
```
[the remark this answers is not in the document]
sanchda: Good catch, fixed

ddtrace/internal/datadog/profiling/dd_wrapper/src/sample.cpp
sanchda: Fixed even better because now each caller of `start_sample` has its own opaque pointer

ddtrace/internal/datadog/profiling/dd_wrapper/src/profile.cpp
sanchda: Removed

ddtrace/internal/datadog/profiling/dd_wrapper/src/interface.cpp
sanchda: Confirmed that this is sync, added comment indicating such

ddtrace/internal/datadog/profiling/dd_wrapper/dd_wrapper.md
```diff
@@ -0,0 +1,117 @@
...
+This sampler lives in a Python thread.
+When it wakes up, it looks at the current threads/tasks and collects stack traces for each of them.
+It also looks at the current exceptions and collects similar data for them.
+
+Although this sampler operates in its own thread, since its fundamental interaction requires taking the GIL, it never samples at a time when other threads are on-CPU (except if they're off-GIL in C code or in io or whatever, but that isn't an issue).
+
+
+### Lock sampler
+
+This patches some common lock implementations to perform sampling.
```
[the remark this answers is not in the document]
sanchda: It's possible, yes.  Although we instrument the main ones AFAIK.

ddtrace/internal/datadog/profiling/dd_wrapper/dd_wrapper.md
```diff
@@ -0,0 +1,117 @@
...
+As such, the sampling occurs from the active application thread.
```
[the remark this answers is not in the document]
sanchda: Yes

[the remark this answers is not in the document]
sanchda: Will give this readme a small facelift

ddtrace/internal/datadog/profiling/_ddup.pyx
```diff
@@ -39,7 +39,7 @@ IF UNAME_SYSNAME == "Linux":
 
         void ddup_init()
 
-        void ddup_start_sample()
+        int ddup_start_sample(int handle)
```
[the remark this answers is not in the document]
sanchda: Refactored out

ddtrace/internal/datadog/profiling/dd_wrapper/include/interface.hpp
```diff
@@ -1,54 +1,54 @@
...
+    void ddup_config_max_nframes(int max_nframes);
+
+    void ddup_config_user_tag(const char* key, const char* val);
+    void ddup_config_sample_type(unsigned int type);
+
+    bool ddup_is_initialized();
+    void ddup_init();
+
+    unsigned int ddup_start_sample(unsigned int handle);
+    void ddup_push_walltime(unsigned int handle, int64_t walltime, int64_t count);
```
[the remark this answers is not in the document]
sanchda: Factored out

ddtrace/internal/datadog/profiling/dd_wrapper/include/libdatadog_helpers.hpp
sanchda: Good call, factored it out though, sorry

ddtrace/internal/datadog/profiling/dd_wrapper/src/uploader.cpp
sanchda: Factored to now add a more descriptive error, indicating which values failed and why.

ddtrace/internal/datadog/profiling/dd_wrapper/src/sample.cpp
sanchda: No, good catch

ddtrace/internal/datadog/profiling/dd_wrapper/src/interface.cpp
sanchda: Factored, but discussed the challenges of presenting an enum class across an FFI boundary, especially when that enum is used to index as is the case here.  AFAIK, there's no way to avoid the check at the boundaries.

ddtrace/internal/datadog/profiling/dd_wrapper/include/profile.hpp
sanchda: Correct, I did indeed have a value of 64 which would have been overridden by this default.  I've factored it slightly differently
1. Constants are now in a constants header
2. Confirmed that the default in the .py code is 64, added a note about the origin in the constants file.

ddtrace/internal/datadog/profiling/dd_wrapper/src/profile.cpp
sanchda: I think both of the ideas are good and worth pursuing.  I think we can easily solve this in two ways
1. For any interface in libdatadog which pushes a reference (pointer) type, provide a variant which immediately stores the value of the reference (probably the only type we care about is strings)
2. in this code, allow the user to send up a Python string.  Then instead of a string table, all we need to do is maintain a "dequeue" (or even a fixed array) of referenced strings to decrement at the end.  I need to think about this a bit harder though.

Echion maintains its own cache which actually makes 1 irrelevant probably.

ddtrace/internal/datadog/profiling/dd_wrapper/src/sample.cpp
sanchda: I have a short-term todo to just plug into the dd-trace-py logger.  I was going to do it in this PR, but it got way too big.

ddtrace/internal/datadog/profiling/dd_wrapper/src/uploader_builder.cpp
sanchda: Ah, good catch.  empty `b`, nonempty `a` will result in spurious delimiters being appended.

ddtrace/internal/datadog/profiling/dd_wrapper/include/types.hpp
sanchda: Factored this out

ddtrace/internal/datadog/profiling/dd_wrapper/src/profile.cpp
sanchda: Indeed, extended the protection.  In particular, there's one exemplary case where the critical period is defined by the caller.  In that instance (uploading), I added the concept of `borrowing` and then `releasing` the profile, which uses the same lock as the other internal function.

I think this is better now, PTAL

ddtrace/internal/datadog/profiling/dd_wrapper/include/libdatadog_helpers.hpp
sanchda: Indeed!  When I add logging, all of these places will emit a more nuanced/local error.

That said, this specific function _should_ be properly guarded with correct unit tests.  All of the valid tags are passed from literal expressions in the code.  It never gets generated.

ddtrace/internal/datadog/profiling/dd_wrapper/include/libdatadog_helpers.hpp
```diff
@@ -0,0 +1,135 @@
...
+inline bool
+add_tag(ddog_Vec_Tag& tags, std::string_view key, std::string_view val, std::string& errmsg)
+{
+    if (key.empty() || val.empty()) {
+        return false;
+    }
+
+    ddog_Vec_Tag_PushResult res = ddog_Vec_Tag_push(&tags, to_slice(key), to_slice(val));
+    if (res.tag == DDOG_VEC_TAG_PUSH_RESULT_ERR) {
+        errmsg = err_to_msg(&res.err, "");
```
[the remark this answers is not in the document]
sanchda: `err_to_msg` does a full copy, but I agree that when possible the subsequent code should be written to make consumption proximal to invalidation.  I've moved all instances of `err_to_msg/drop/"log"`to do the consumption ("log") before dropping the error.

ddtrace/internal/datadog/profiling/dd_wrapper/include/profile.hpp
sanchda: Sorry--does this conversation answer your question?  https://github.com/DataDog/dd-trace-py/pull/6731/files/be1aac6146cf8cd614e49f93ba960e16d2d3710f#r1304402419

The basic premise is that the version of C++ we have in CI right now doesn't support heterogeneous access for sets, so the incoming `string_view` would need to get converted to `string` for comparison.  I think there are some fun and exciting ways around this, but I'm even more interested in removing the temporary storage entirely if we can help it.

ddtrace/internal/datadog/profiling/dd_wrapper/include/sample_manager.hpp
sanchda: I think we've reduced this a lot in the latest iteration, but we can do better.  It'd require some more refactors, but I think I'd like to tack that on as a followup PR.  For discussion, consider the following

UploaderBuilder has a lot of `static`, but it is actually treated as a singleton.  It doesn't have to be either.  Of its 10 static objects, 9 can be factored into a context class.  The remainder is a lock which we may have to keep.

The 3 statics in SampleManager are the exact same story.  Two can be factored into a single context object, for instance, with the remainder being a lock.

I'm not 100% sure how to factor out the static-ness of the lock or cancellation token in the uploader, nor the static Profile in the Sample, without significant refactors.

Either way, TBD.

ddtrace/internal/datadog/profiling/dd_wrapper/dd_wrapper.md
```diff
@@ -0,0 +1,117 @@
...
+
+### Stack sampler
+
+This is a simple stack sampler that collects stack traces at a dynamic interval.
+Supposedly, it's a fixed-interval profiler, but the interval is adjusted dynamically depending on the observed overhead.
+This sampler lives in a Python thread.
+When it wakes up, it looks at the current threads/tasks and collects stack traces for each of them.
+It also looks at the current exceptions and collects similar data for them.
+
+Although this sampler operates in its own thread, since its fundamental interaction requires taking the GIL, it never samples at a time when other threads are on-CPU (except if they're off-GIL in C code or in io or whatever, but that isn't an issue).
```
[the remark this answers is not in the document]
sanchda: Yes, commenting

ddtrace/internal/datadog/profiling/dd_wrapper/include/profile.hpp
r1viollet: Well at least I knew I was going to be confused again.

sanchda:
@tabgok sorry you were pulled in as a reviewer.  I fat-fingered a rebase, then pushed.  You can verify from the modified files that this no longer touches Framework Integrations code.

PR #8308, opened 2024-02-06
feat(tracing): lazy sampling

ZStriker19:
This PR makes sampling lazy, effectively making our extended sampling API that can take resource and operation name far more useful, since we will now always be sampling after those tags have been added.  Currently, often resource is added afterwards and operation name can change, therefore this PR essentially finishes the extension.

To make this PR simpler for reviewing, changes in this PR can be categorized as such:

1. Do not sample in `_start_span`

2. Sample right before propagation by modifying all of the `HTTPPropagator.inject()` calls to take a tracer and span so we can sample there right before the sampling decision is propagated.

3. Sample right before encoding using `TraceSamplingProcessor`

4. Sample right before  `_DBM_Propagator.inject()`

5. Sample right before `os.fork`

**Design Decisions:**

1. _Why are we running `context._update_tags` when we sample?_
We need the tags and metrics from the context on the root_span when we sample. This unfortunately leads to us running update_tags twice, once when we sample, and once in the TraceTagsProcessor right before we encode the trace chunk._

The alternative was pulling out the `context._meta` and `context._metrics` values to combine them with the span's metrics and tags values in `sampling_rule.py::tags_match` or running sampling on those values in addition to the span's values to come up with a match. Both of these felt messier, and technically the possibility exists that the same tag key could exist on the context and the span, in which case the current span key should take precedence (as is the current behavior with context._update_tags). This could obviously be implemented with a `continue` in the for loop evaluating tag rules against spans, however it adds to the mess. I'm still kind of on the fence about this choice and am happy to discuss.

I've added the update in sample, because there are multiple other places where we need to make sure that we update, before sampling, e.g. before propagation http, before propagation dbm. Therefore I didn't want to make sure we always ran context.update_tags after every egress point correctly before we sample.

2. _It's kind of messy to modify the `HttpPropagator.inject` signature to take a span._

No doubt, but I couldn't find a way around this besides either wrapping inject, with a method that samples before running it, or having to sample every place we call `HttpPropagator.inject`. It seemed safer to me to call sampling inside of it. However it isn't great, because if customers are using inject() int their code, don't pass in a span, then sampling is not run. I will add a rn about that. 
Another option was to always run `ddtrace.tracer.current_root_span()` to grab the span, which we do as a back up if inject is called without a span kwarg value.  However, Emmett and I decided that the ability to pass in an explicit span if possible should be included be the norm in our use.
I'm open to any suggestions regarding this.

3. _Ok so you instrumented os.fork to sample before we fork, but what about all of the other ways you can change execution context? E.g. spawn, pOpen, etc.? Can't that lead to differences in context.sampling_priority between the parent process's spans and the child processes spans?_

Yes, it can. I'm not sure I can add instrumentation to all of the ways you can change execution context in this PR, however I'd be happy to hear advice on how to add it to any easy to reach places. I've also added to [the documentation](https://github.com/DataDog/dd-trace-py/pull/8308/files#diff-151a99faebb7c20650eb815a2821b6613090470c2873994c64334fc9190ee1b0R84-R154) that you should sample manually before changing execution context.

4. _Follow up, to number 3: Does this break gunicorn or gevent since for execution context switching I've only changed os.fork to sample before the context change?_

No it does not, we don’t actually ever start traces in the parent process, the request is just routed to the worker, and the trace starts there. In the event that middleware is added to a gunicorn + gevent application the middleware sits between the parent process routing the request and the worker, so sampling propagation should work fine there as well, as long as the middleware and the application logic are executing in the same execution context.

5. _What about when customers have an egress point that's not automatically instrumented or they don't use `HttpPropagator.inject` for? Won't that break sampling?_

Yes, they'll need to sample manually before the injection. I've added that here in [the docs](https://github.com/DataDog/dd-trace-py/pull/8308/files#diff-151a99faebb7c20650eb815a2821b6613090470c2873994c64334fc9190ee1b0R241-R243).

## Checklist

## Reviewer Checklist

- [x] Title is accurate


ddtrace/_trace/processor/__init__.py
```diff
@@ -143,28 +143,29 @@ class TraceSamplingProcessor(TraceProcessor):
...
     _compute_stats_enabled = attr.ib(type=bool)
+    sampler = attr.ib()
 
     def process_trace(self, trace):
         # type: (List[Span]) -> Optional[List[Span]]
         if trace:
+            chunk_root = trace[0]
+            root_ctx = chunk_root._context
+            chunk_root.sampled = self.sampler.sample(trace)
+            _update_span_tags_from_context(chunk_root, root_ctx)
```
ZStriker19, line 154: We run ` _update_span_tags_from_context` here and in TraceTagsProcessor. I'm a bit unsure if it's necessary. I think it may only be needed here, since I believe this runs before TraceTagsProcessor, and we don't seem modify the context between now and then. 

I am however also wondering if we need to run `sample` before we update the span tags. Potential fixes for this:
1. We look at context when sampling to check everything set on it
2. We update the root span beforehand

ddtrace/span.py
```diff
@@ -168,6 +169,10 @@ def __init__(
 
         # sampling
         self.sampled = True  # type: bool
+        # 0 = no sampling has run
```
ZStriker19, line 172: The code comment essentially explains what my interpretation of the RFC is. We need a way to figure out where `sample` has run, to figure out if we should run the `sample` again. For example, the `sample` call in propagation/http.py should override the sampling decision if `sample` was called in  processor/__init__ (before encoding).  Neither of those sample calls should override manually setting sampling_priority in code on the span.

ZStriker19: I need to write some tests for this overriding behavior.

ddtrace/_trace/processor/__init__.py
ZStriker19: Actually it's probably ok to remove this behavior since we have the sampler now look inside the context object for tags and metrics when sampling, so it doesn't need to all be on the root span when we sample.

ddtrace/propagation/http.py
```diff
@@ -944,7 +949,9 @@ def parent_call():
         if config.propagation_http_baggage_enabled is True and span_context._baggage is not None:
             for key in span_context._baggage:
                 headers[_HTTP_BAGGAGE_PREFIX + key] = span_context._baggage[key]
-
+        if ddtrace.tracer._sampler and span:
```
ZStriker19, line 952: Let's grab the current_root_span if we don't have a span passed in.

ddtrace/_trace/tracer.py
```diff
@@ -227,6 +231,7 @@ def __init__(
         self.context_provider = context_provider or DefaultContextProvider()
         self._user_sampler: Optional[BaseSampler] = None
         self._sampler: BaseSampler = DatadogSampler()
+        register_before_in_child(self.sample_before_fork)
```
ZStriker19, line 234: Do I need to unregister and re-register each time tracer._sampler changes?

ZStriker19: I think the answer is yes

ZStriker19: I think the answer is no

ZStriker19:
Open questions:
1. Does this break gunicorn or gevent since for execution context switching I've only changed `os.fork` to sample before the context change?
2. Do I need to update the advanced section of asyncio like how I did the other context execution changing examples? https://ddtrace.readthedocs.io/en/stable/advanced_usage.html#tracing-across-asyncio-tasks
3. Have I forgotten any other egress points where I should be running the sampler before?

tests/integration/test_integration_snapshots.py
```diff
@@ -152,13 +154,38 @@ def task2(tracer):
...
+        # need to manually sample before Popen
+        tracer._sampler.sample(parent)
         p = multiprocessing.Process(target=task, args=(tracer,))
         p.start()
         p.join()
     tracer.shutdown()
 
 
+@snapshot(async_mode=False)
+def test_tracer_trace_across_only_fork():
```
ZStriker19, line 167: This new test is flakey, sometimes the child span is missing, need to figure this out. Fork is actually instrumented to have sampler run before the fork, and seems to work properly.

tests/snapshots/tests.integration.test_integration_snapshots.test_trace_with_wrong_metrics_types_not_sent.json
```diff
@@ -0,0 +1,25 @@
+[[
```
ZStriker19, line 1: I think that previously this snapshot wasn't generated for some reason?

ddtrace/_trace/processor/__init__.py
```diff
@@ -189,6 +196,10 @@ def on_span_finish(self, span):
             span.set_metric("_dd.top_level", 1)
 
 
+def _update_span_tags_from_context(span, context):
+    span._context._update_tags(span)
```
emmettbutler, line 200:
```suggestion
    context._update_tags(span)
```
I'm taking a guess - was this the logic you intended here?

emmettbutler (changes requested):
This is looking really good. I have lots of questions and edit suggestions. The only one I think blocks the merge is the question about logic in the processor.

tests/tracer/test_propagation.py
```diff
@@ -116,7 +119,7 @@ def test_inject_128bit_trace_id_b3multi():
             HTTPPropagator.inject(span.context, headers)
             trace_id_hex = "{:032x}".format(span.trace_id)
             span_id_hex = "{:016x}".format(span.span_id)
-            assert headers == {"x-b3-traceid": trace_id_hex, "x-b3-spanid": span_id_hex}
+            assert headers == {"x-b3-traceid": trace_id_hex, "x-b3-spanid": span_id_hex, "x-b3-sampled": "1"}
```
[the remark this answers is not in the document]
ZStriker19: Yeah, Munir pointed those out to me as well, should be resolved with this: https://github.com/DataDog/system-tests/pull/2240 trying it out on this test PR: https://github.com/DataDog/dd-trace-py/pull/8715

tests/internal/test_database_monitoring.py
```diff
@@ -97,6 +97,10 @@ def test_dbm_propagation_full_mode():
 
     dbspan = tracer.trace("dbname", service="orders-db")
 
+    # since inject() below will call the sampler we just call the sampler here
+    # so sampling priority will align in the traceparent
+    tracer.sampler.sample(dbspan._local_root)
```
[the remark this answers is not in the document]
ZStriker19: Gotcha, for this case, in the real world it should always covered by the dbm inject code: https://github.com/DataDog/dd-trace-py/pull/8308/files#diff-b86d5de5f005bdd3a34c46580f7bb4310b4a2925cdb907a3e1c35deffc88c6a2R65-R69. We just have to call it early in this test so that when we do `sqlcomment = dbm_popagator._get_dbm_comment(dbspan)` it grabs a comment with a traceparent that has a priority. 

Another option would be to sample inside of `_get_dbm_comment` but that felt worse to me initially.

ddtrace/internal/forksafe.py
```diff
@@ -50,6 +51,20 @@ def ddtrace_after_in_child():
             log.exception("Exception ignored in forksafe hook %r", hook)
 
 
+def ddtrace_before_in_child():
```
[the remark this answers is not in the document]
ZStriker19: Agreed better name!

mabdinur (commented):
I am not a big fan of moving sampling from one central point (at start span) to multiple points each of which could fail in to run in different scenarios. I understand we have already discussed the tradeoffs of this design and are comfortable moving forward but we need to make sure this change is easy to debug and maintain. We should be generous in our use of warnings, debug logs and inline comments. We need ensure all scenarios where sampling could be misconfigured notify users and all user facing sampling documentation is detailed and up to date

ddtrace/_trace/tracer.py
```diff
@@ -317,6 +323,29 @@ def deregister_on_start_span(self, func: Callable) -> Callable:
         self._hooks.deregister(self.__class__.start_span, func)
         return func
 
+    def sample_before_fork(self) -> None:
+        span = self.current_root_span()
+        if span is not None and span.context.sampling_priority is None:
+            self.sampler.sample(span)
+
+    @property
+    def sampler(self):
```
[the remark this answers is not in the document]
emmettbutler: We discussed this [here](https://github.com/DataDog/dd-trace-py/pull/8572#issuecomment-1986176538). I echo @mabdinur's concern about adding to the public API.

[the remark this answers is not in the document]
ZStriker19: Ok cool, let me put it behind a sample method

ddtrace/_trace/tracer.py
```diff
@@ -317,6 +323,29 @@ def deregister_on_start_span(self, func: Callable) -> Callable:
...
+        return self._sampler_current
+
+    @sampler.setter
+    def sampler(self, value):
+        self._sampler_current = value
+        # we need to update the processor that uses the sampler
+        if getattr(self, "_deferred_processors", None):
+            for aggregator in self._deferred_processors:
```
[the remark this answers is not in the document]
ZStriker19: That seems kind of more dangerous to always assume, and I don't think we should have to be looping like this very often. If you think that's off, then I can make the change though.

ddtrace/contrib/aiohttp/patch.py
```diff
@@ -88,7 +88,7 @@ async def _traced_clientsession_request(aiohttp, pin, func, instance, args, kwar
             span.service = url.host
 
         if pin._config["distributed_tracing"]:
-            HTTPPropagator.inject(span.context, headers)
+            HTTPPropagator.inject(span.context, headers, span)
```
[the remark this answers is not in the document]
ZStriker19: I doubt many users are calling HTTPPropagator.inject to begin with, but if they are, without passing in the span, this line will almost certainly grab the correct span regardless: https://github.com/DataDog/dd-trace-py/pull/8308/files#diff-db5c2c06582af0a924e93e08a60bea035ff4de9bf97a08bc12b7098c3546c8dcR958-R960

Noting other option as well:
We could just completely remove this kwarg and rely on the above referenced code to always grab us the root span. When I spoke with @emmettbutler  about this he thought it would be better if we could be explicit when it was an option, and therefore I stuck to that. 
I don't know in what circumstances `tracer.current_root_span` would fail us. I think I'm fine with either.
```
 def current_root_span(self) -> Optional[Span]:
        span = self.current_span()
        if span is None:
            return None
        return span._local_root
```

ddtrace/_trace/tracer.py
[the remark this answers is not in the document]
ZStriker19: https://github.com/DataDog/dd-trace-py/pull/8722

tests/opentelemetry/test_trace.py
```diff
@@ -162,6 +162,8 @@ def test_otel_start_current_span_without_default_args(oteltracer):
 @pytest.mark.snapshot(ignores=["metrics.net.peer.port", "meta.traceparent", "meta.flask.version"])
 def test_distributed_trace_with_flask_app(flask_client, oteltracer):  # noqa:F811
     with oteltracer.start_as_current_span("test-otel-distributed-trace") as otel_span:
+        # need to manually run sampling before grabbing otel_span._ddspan.context._tracestate
+        oteltracer._tracer.sampler.sample(otel_span._ddspan)
```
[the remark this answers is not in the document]
ZStriker19: Oddly in this test, we are manually pulling out values from the otel_span and creating headers. Sampling and propagation actually work fine automatically if we just remove manually generating and adding the headers. I think that still keeps the spirit of the test, so I'll do that instead.

ddtrace/contrib/aiohttp/patch.py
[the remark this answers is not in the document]
ZStriker19: Spoke with Kyle, just rely on tracer.current_root_span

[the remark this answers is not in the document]
emmettbutler: Makes sense. I hadn't really thought about the functional "breaking change" incurred by expecting a span argument without a default. The real issue is that `current_root_span` must be the default whether or not there's the option to pass in a span.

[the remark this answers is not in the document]
ZStriker19: Also down the line, would be good to have inject be a tracer method, since it already has all of the info.

ddtrace/_trace/tracer.py
```diff
@@ -317,6 +323,29 @@ def deregister_on_start_span(self, func: Callable) -> Callable:
...
+
+    @property
+    def sampler(self):
+        return self._sampler_current
+
+    @sampler.setter
+    def sampler(self, value):
+        self._sampler_current = value
+        # we need to update the processor that uses the sampler
+        if getattr(self, "_deferred_processors", None):
```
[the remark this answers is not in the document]
ZStriker19: Good idea, I'll add one

ZStriker19:
I think I need to change the system-tests start_span method a bit to get sampling and propagation tests to pass

ddtrace/_trace/processor/__init__.py
```diff
@@ -135,30 +135,55 @@ def unregister(self):
 
 @attr.s
 class TraceSamplingProcessor(TraceProcessor):
-    """Processor that keeps traces that have sampled spans. If all spans
-    are unsampled then ``None`` is returned.
+    """Processor that runs both trace and span sampling rules.
```
ZStriker19, line 138: This change was necesarry to have trace sampling run before span sampling, I like having all of the sampling run in one processor, but if people have other opinions, happy to discuss.

tests/tracer/test_processors.py
```diff
@@ -491,16 +509,16 @@ def test_single_span_sampling_processor_w_tracer_sampling(
 
 
 def test_single_span_sampling_processor_w_tracer_sampling_after_processing():
-    """Test that single span sampling tags and tracer sampling context are applied to spans
-    if the trace sampling is changed after the span is processed.
+    """Since the root span has MANUAL_KEEP_KEY set and the child span has not yet run through
```
ZStriker19, line 512: Previously, since the SpanSamplingProcessor ran on_span_finish, it would run, while  `root.context.sampling_priority == AUTO_REJECT`. Therefore causing span sampling to apply span sampling tags to the child span. Then doing  `root.set_tag(MANUAL_KEEP_KEY)` would cause the child span's context to update, leading to: `child.context.sampling_priority == USER_KEEP`. Therefore causing the child span to have both single span and trace sampling tags on it. 

For current behavior please read the doc string, I think this is actually better now.

ddtrace/_trace/processor/__init__.py
```diff
@@ -135,30 +135,55 @@ def unregister(self):
...
+            if self.single_span_rules:
+                for span in trace:
+                    if span.context.sampling_priority is not None and span.context.sampling_priority <= 0:
+                        for rule in self.single_span_rules:
+                            if rule.match(span):
+                                rule.sample(span)
+                                # If stats computation is enabled, we won't send all spans to the agent.
+                                # In order to ensure that the agent does not update priority sampling rates
+                                # due to single spans sampling, we set all of these spans to manual keep.
+                                if config._trace_compute_stats:
```
mabdinur, line 183: Wait a minute, so if single span sampling rules are set and stats are computed all spans are set to manual keep? Wouldn't this override sampling decisions where we should drop a span? This seems like a bug

ddtrace/_trace/tracer.py
```diff
@@ -298,7 +301,10 @@ def _atexit(self) -> None:
         self.shutdown(timeout=self.SHUTDOWN_TIMEOUT)
 
     def sample(self, span):
-        self._sampler.sample(span)
+        if self._sampler is not None:
+            self._sampler.sample(span)
+        else:
+            log.debug("No sampler available to sample span")
```
mabdinur, line 307: Should this be a warning? Is there a valid scenario where no samplers are set

ddtrace/_trace/tracer.py
```diff
@@ -341,6 +347,29 @@ def deregister_on_start_span(self, func: Callable) -> Callable:
         self._hooks.deregister(self.__class__.start_span, func)
         return func
 
+    def sample_before_fork(self) -> None:
```
mabdinur, line 350: Should we make this this an internal method? Would a user ever call this manually

ddtrace/appsec/_api_security/api_manager.py
```diff
@@ -128,8 +128,12 @@ def _schema_callback(self, env):
 
         try:
             # check both current span and root span for sampling priority
+            # if sampling has not yet run for the span, we default to treating it as sampled
+            if root.context.sampling_priority is None and env.span.context.sampling_priority is None:
+                priorities = (1,)
```
mabdinur, line 133: ?
```suggestion
                priorities = (1,1)
```

ddtrace/contrib/algoliasearch/patch.py
```diff
@@ -125,7 +125,7 @@ def _patched_search(func, instance, wrapt_args, wrapt_kwargs):
         span.set_tag_str(SPAN_KIND, SpanKind.CLIENT)
 
         span.set_tag(SPAN_MEASURED_KEY)
-        if span.context.sampling_priority is None or span.context.sampling_priority <= 0:
+        if span.context.sampling_priority is not None and span.context.sampling_priority <= 0:
```
mabdinur, line 128: Why do we even need to check the sampling priority in an integration?

ddtrace/propagation/http.py
```diff
@@ -950,6 +966,18 @@ def parent_call():
             for key in span_context._baggage:
                 headers[_HTTP_BAGGAGE_PREFIX + key] = span_context._baggage[key]
 
+        if hasattr(ddtrace, "tracer") and hasattr(ddtrace.tracer, "sample"):
```
mabdinur, line 969: We should log an error if this if statement returns False, these fields should always be set

ddtrace/propagation/http.py
```diff
@@ -950,6 +966,18 @@ def parent_call():
...
+            if non_active_span is not None:
+                if _is_top_level(non_active_span):
```
mabdinur, line 971: top level will return True for service entry spans. Shouldn't we only sample on root spans

ddtrace/propagation/http.py
```diff
@@ -950,6 +966,18 @@ def parent_call():
             for key in span_context._baggage:
                 headers[_HTTP_BAGGAGE_PREFIX + key] = span_context._baggage[key]
 
+        if hasattr(ddtrace, "tracer") and hasattr(ddtrace.tracer, "sample"):
+            if non_active_span is not None:
```
mabdinur, line 970: Can we simplify this logic to: 

- If tracer.current_context == span_context THEN call tracer.current_root_span().sample() 
- If tracer.current_context != span_context then call non_active.sample() 
   -  if non_active is None then log an actionable warning

mabdinur: This way users will be notified right away if httpinject is not sampling some correcty

releasenotes/notes/lazy_sampling-93057adeaccbb46f.yaml
```diff
@@ -0,0 +1,22 @@
+---
+prelude: >
```
mabdinur, line 2: This release note is a bit verbose. Can we move this info to our library docs and then link the docs in the release notes.

mabdinur (approved):
Left some nits and questions. Once those are addressed this PR should be mergeable. I don't want to block this change for another week. I trust your judgement

ddtrace/_trace/processor/__init__.py
ZStriker19: Nah, this is the desired behavior and is consistent with what it was before. Basically, if we set the single span sampling tags, the trace-agent will take the span, despite the span having `_sampling_prioirity_v1=-1`. Which is fine, however, in the case that we don't send the entire trace (`config._trace_compute_stats=True`), the agent would update priority sampling rates incorrectly, therefore to avoid that, we set all single span sampled spans to manual_keep, so they don't count towards the agent's trace level sample rate calculations.

ddtrace/_trace/tracer.py
ZStriker19: You're right, warning is probably better.

ddtrace/appsec/_api_security/api_manager.py
ZStriker19: In order to make it a tuple, you ned to put a comma after the first member, however we never need to priorities to have anything besides just a single 1 in it, since all we do is check if various values are inside of it, having multiple 1s is the same as having one 1.

ddtrace/contrib/algoliasearch/patch.py
ZStriker19: I think the idea was to avoid unnecessary and possibly expensive operations (mostly getting and setting tags) on a span if we knew it was going to be dropped. This brings up some additional questions as to if we should do this anymore now that we know single span sampling could choose to keep etc. However, I think that's out of the scope of this PR, so I'd prefer just trying to maintain behavior as well as possible. 

Although with lazy sampling the extra tag setting code will run more often, since if sampling hasn't run yet, or there isn't an upstream decision, we'll add the tags, so there may be a small performance impact.

I think a follow up PR could remove this behavior all together, but that would come a with a worse performance impact on average, since we'd always set the tags, even if there was an upstream drop.

I guess another thing to think about is if  single span sampling is used to keep these spans, after they've decided to be dropped, and didn't have tags added that could be confusing. I'll add this as a discussion topic for sync to discuss what's best, but don't think this should block this PR.

ddtrace/propagation/http.py
ZStriker19: If you look down a few lines, we check `if root_span is not None and root_span.context.sampling_priority is None` which essentially makes sure we're only sampling on a root span.

ZStriker19: I did change the logic a bit though, to just grab span._local_root, since if you look at the start_span logic, we always set the span._local_root to the span, if there's no parent for it.

ddtrace/propagation/http.py
ZStriker19: I changed the code a bit to rely on span._local_root, which simplifies the logic well I think.

releasenotes/notes/lazy_sampling-93057adeaccbb46f.yaml
ZStriker19: I agree that the note is a bit verbose, and it's somewhat covered in the doc updates I did, still I think I'd prefer being verbose and informative in this case, and see little downside to the choice. However, if you feel strongly I can change.

ZStriker19: Actually thought about this a bit more and I added more to our actual docs around this, but going to keep the rn the same if that's alright.

ZStriker19:
This parametric system tests run shows we pass all of the sampling test cases this PR should pass (the pr is just this one, but running a system-tests branch that enables all of the sampling tests, and has a modified inject method which will be merged once this is released. https://github.com/DataDog/dd-trace-py/actions/runs/8412705554/job/23033894434?pr=8715

tests/opentelemetry/test_trace.py
```diff
@@ -161,12 +161,8 @@ def test_otel_start_current_span_without_default_args(oteltracer):
 )
 @pytest.mark.snapshot(ignores=["metrics.net.peer.port", "meta.traceparent", "meta.flask.version"])
 def test_distributed_trace_with_flask_app(flask_client, oteltracer):  # noqa:F811
-    with oteltracer.start_as_current_span("test-otel-distributed-trace") as otel_span:
-        headers = {
-            "traceparent": otel_span._ddspan.context._traceparent,
-            "tracestate": otel_span._ddspan.context._tracestate,
-        }
-        resp = flask_client.get("/otel", headers=headers)
```
emmettbutler, line 169: Why did you make this change?

ddtrace/propagation/http.py
```diff
@@ -940,7 +944,29 @@ def parent_call():
 
         :param Context span_context: Span context to propagate.
         :param dict headers: HTTP headers to extend with tracing attributes.
+        :param Span non_active_span: Only to be used if injecting a non-active span.
         """
+        if non_active_span is not None and non_active_span.context is not span_context:
+            log.error(
+                "span_context and non_active_span.context are not the same, but should be, non_active_span.context "
```
emmettbutler, line 951:
```suggestion
                "span_context and non_active_span.context are not the same, but should be. non_active_span.context "
```

emmettbutler (approved):

tests/opentelemetry/test_trace.py
ZStriker19: The way it was originally written we generated headers, and then sent them through manually. However, if you just rely on the inject method, it generates these headers anyways, and then in addition runs sampling. 

The alternative is that we could manually run sampling and then manually inject that header, but it made more sense to me to test more of how this would actually be used in the wild.

ddtrace/llmobs/_integrations/base.py
```diff
@@ -73,13 +73,17 @@ def llmobs_enabled(self) -> bool:
         return config._llmobs_enabled
 
     def is_pc_sampled_span(self, span: Span) -> bool:
-        if span.context.sampling_priority is None or span.context.sampling_priority <= 0:
-            return False
+        if span.context.sampling_priority is not None:
```
mabdinur, line 76: https://github.com/DataDog/dd-trace-py/pull/8695/files#r1537936773

Yun-Kim (approved):
LGTM from ml_obs!

zarirhamza (approved):

gnufede (approved):
LGTM from ASM perspective

docs/advanced_usage.rst
```diff
@@ -226,7 +237,11 @@ To trace requests across hosts, the spans on the secondary hosts must be linked
 - On the server side, it means to read propagated attributes and set them to the active tracing context.
 - On the client side, it means to propagate the attributes, commonly as a header/metadata.
 
-`ddtrace` already provides default propagators but you can also implement your own.
+`ddtrace` already provides default propagators but you can also implement your own. Note that `ddtrace`` makes
```
duncanista, line 240:
```suggestion
`ddtrace` already provides default propagators but you can also implement your own. Note that `ddtrace` makes
```

duncanista (approved):
LGTM – on the guild side, docs and release note.

PR #8316, opened 2024-02-07
feat: use backup entity identifier when container ID is unavailable [AIT-9282]

emmettbutler:
This change adds the ``Datadog-Entity-ID`` header to payloads sent to the Datadog Agent. This header can be set to the container ID or the container's cgroup node inode, and serves as a unique identifier for containers running under Linux ``cgroupv2``.

Black-box testing for this functionality is implemented [here](https://github.com/DataDog/system-tests/blob/a0064c4012f19fbec5b9f0583efc80ab5f37c35e/tests/test_data_integrity.py#L147).

## Checklist

## Reviewer Checklist

- [x] Title is accurate


sanchda (approved):
This precise technique may not work in cgroupsv2 (IIRC, the way this metadata is presented is a bit different than the sysfs interfaces we know from v1), but given that v2 isn't widely rolled out, I don't think that's a problem.  Disclaimer:  didn't read the RFC.

Otherwise glad to see a failover, and good to see this consolidated!  Looks good from the Profiling side.

emmettbutler:
@sanchda it's actually pretty important that this works against cgroupsv2. Can you share any more information about why it won't work?

sanchda:
> Can you share any more information about why it won't work?

Checking the docs real quick, I may have commented too hastily.

sanchda:
Ah, yup--according to the [docs](https://raw.githubusercontent.com/torvalds/linux/master/Documentation/admin-guide/cgroup-v2.rst)

```
/proc/$PID/cgroup" lists a process's cgroup membership.  If legacy
cgroup is in use in the system, this file may contain multiple lines,
one for each hierarchy.  The entry for cgroup v2 is always in the
format "0::$PATH"::
```

So as long as the mountpoint for v2 is anything other that `/sys/fs/cgroup` (which will be the case on mixed systems), then the additional check into the mountpoint for the cgroup2 will be necessary.

I imagine that in pure-v2 configurations the mountpoint is indeed `/sys/fs/cgroup` and my comments are irrelevant.

emmettbutler:
@sanchda thanks again for the help. The approach taken here is straight from the RFC, which seems to accept an approach that doesn't work on mixed systems. I could be wrong about that, but this seems like a reasonable first step anyway.

ZStriker19 (approved):

majorgreys (approved):

P403n1x87 (approved):

jessicagamio:
@emmettbutler to make sure i understand this one, we should expect an alternate container ID in trace data if the actual one is not available aka accessible?

emmettbutler:
Yup. It will be transmitted as the `datadog-entity-id` header. @jessicagamio

jessicagamio:
Gotcha. @emmettbutler one last question cause i had a little more time to dig deeper on the subject today.  Would this essentially resolve the issue where the Tracer can't identify its own container ID in cgroupv2 [ - Documented here ?](https://datadoghq.atlassian.net/wiki/spaces/TS/pages/3189670032/Traces+missing+container+tags+with+cgroup+V2#Impact)
The Solution to those scenarios has been to switch over to UDS transport. Would this update then essentially make that switch unnecessary?

emmettbutler:
Yup, that's exactly what this feature is for. @jessicagamio

jessicagamio:
@emmettbutler Sorry for the noise on this. Share this with the team and  was asked if this would be dependent on any specific version of the Agent?

emmettbutler:
As far as I can tell, this change doesn't have any effect unless it's talking to an Agent version 7.50.0 or later. ([here's where support was added in the agent](https://github.com/DataDog/datadog-agent/pull/21236))

PR #8339, opened 2024-02-08
feat(llmobs): submit span events rather than llmobs records

Yun-Kim:
## Summary

This PR does a few things:
- Replaces LLMObs Records with LLMObs Span events (to conform to LLMObs intake payload structure changes)
- `LLMObsWriter` class switches from submitting LLMObs records to submit Span events.
- `LLMObsWriter` is now a single class owned by the `LLMObs` service instance rather than each `LLMIntegration` classes owning an instance of `LLMObsWriter`.
- `LLMIntegration` classes now mark LLM (completion/chat) spans and set temporary ml_obs tags to be extracted by the LLMObs service `LLMObsTraceProcessor` to create span events and submit them to the `LLMObsWriter` to be written to LLMObs intake.
- Removes the now unnecessary `DD_APP_KEY` as a config option.
- Introduces a new span type `llm` which the `LLMObsTraceProcessor` uses to identify spans which should also be converted to LLMObs span events and submitted to LLMObs.

### LLMObs Records --> LLMObs Span Event
The `LLMIntegration` classes currently generate LLMObs records and pass it to their `LLMObsWriter` instance to submit it to LLMObs' record intake. However, the LLMObs intake is being updated to accept Span events, which this PR creates new support for (and removes support for submitting LLMObs records). The new span event structure looks something like this:
```
class LLMObsEvent:
    span_id: str
    trace_id: str
    parent_id: str
    session_id: str
    tags: List[str]
    name: str
    start_ns: int
    duration: float
    error: int
    meta: Dict[str, Any]
    metrics: Dict[str, Any]
```

This also includes some changes to the LLMObs writer:
- Submits payloads to a new intake url `https://llmobs-intake.{DD_SITE}/api/v2/llmobs`.
- Removes all previous workarounds to submit only one record per interval, as the new span event structure does not have any of the previous limitations.
- Removes `DD_APP_KEY` as a config option, as the new endpoint does not need it (only needs API key).

### New Workflow (who owns the LLMObsWriter now?)

Currently, the workflow for LLMObs involves something like this:
```
1. OpenAI integration traces a LLM call
2. If `DD_LLMOBS_ENABLED=1`, then the OpenAIIntegration instance will generate a LLMObs record, then pass it to its `LLMObsWriter` thread to submit to LLMObs
```

The proposed workflow looks something like this:
```
1. If `DD_LLMOBS_ENABLED=1`, then the LLMObs service is enabled automatically on ddtrace-run/openai/bedrock integration load.
2. The LLMObs service starts its own LLMObsWriter thread, (instead of LLMIntegrations owning their own writer threads).
3. The LLMObs service also adds a LLMObsTraceProcessor to the active tracer.
4. The OpenAIIntegration traces (using same tracer) a LLM call, and marks that span as being an LLM-type. It also sets temporary `ml_obs.*` tags on the span.
5. Once the span is finished, it triggers the tracer's trace processors.
6. The LLMObsTraceProcessor takes all traces containing LLM-type spans, removes the `ml_obs.*` tags to create a LLMObs Span event, and submits to LLMObs via its LLMObsWriter thread.
```
Note: in this approach, we're creating temporary `ml_obs.*` tags on the APM spans before removing them to populate the LLMObs span event to submit to LLMObs. These temporary tags will only be created if LLMObs is enabled, and will never be submitted on the APM span object.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


ddtrace/settings/config.py
```diff
@@ -552,7 +552,6 @@ def __init__(self):
         self._telemetry_install_time = os.getenv("DD_INSTRUMENTATION_INSTALL_TIME", None)
 
         self._dd_api_key = os.getenv("DD_API_KEY")
-        self._dd_app_key = os.getenv("DD_APP_KEY")
```
emmettbutler, line 555: This is a significant enough change to the public interface that I think it deserves a deprecation warning.

emmettbutler (changes requested):

ddtrace/settings/config.py
Yun-Kim: @emmettbutler While I would agree, there's some context here that I think makes this actually not so significant and justifies no release notes:
- `DD_APP_KEY` was a private config option for the openai/bedrock integrations for LLMObs users to submit data to LLMObs. This was later refactored to being a global config object in #8254, but since this was only for LLMObs users (currently private beta phase) we didn't add any release notes to add it as part of our public interface.
- With the recent change to LLMObs intake, we actually don't need `DD_APP_KEY` anymore to submit LLMObs payloads. This means this env var is unnecessary (unfortunately only found this out after #8254 was merged 😢)
- LLMObs is still in a private design partnership phase right now, meaning we don't need to worry about deprecating this change since there aren't any active users submitting LLMObs data.

The takeaway for me here is to be more careful adding to our public interface in the future, but in the meantime I don't see any reason that we shouldn't delete this config entry which wasn't public to begin with. WDYT?

emmettbutler: Makes sense to me - no deprecation necessary. Thanks for explaining.

ddtrace/contrib/openai/_endpoint_hooks.py
```diff
@@ -203,6 +205,7 @@ class _CompletionHook(_BaseCompletionHook):
 
     def _record_request(self, pin, integration, span, args, kwargs):
         super()._record_request(pin, integration, span, args, kwargs)
+        span.span_type = SpanTypes.LLM
```
Yun-Kim, line 208: This is being set manually here because OpenAI spans are created by the same patched method, but the underlying endpoints can be non-LLM (ex: embeddings/images/audio vs completion/chat)

mabdinur (approved):

emmettbutler (approved):

PR #8431, opened 2024-02-15
feat(llmobs): set parent ID as the nearest LLMObs span in ancestor tree

Yun-Kim:
This PR changes generated LLMObs span events to ensure that a given LLMObs span event has a parent ID which belongs to the nearest LLMObs span in its ancestor tree rather than always its parent span.

This is because LLMObs doesn't store non-LLMObs span events, meaning that in the case of traces with non-LLMObs spans, i.e. `LLMObs-span --> APM span --> LLMObs-span` the last span's parent ID will be that of the second (APM) span which isn't sent to LLMObs, meaning there will be a gap/error when we try to recreate the trace. The solution is to only use LLMObs spans to set the parent ID, i.e. the nearest span in the ancestor tree which is also of type `LLM`. 

This PR also adds tests for the `LLMObsTraceProcessor` and refactors some commonly used fixtures/utils.

### Limitations and Future Steps
In cases of distributed traces, multithreading, or forked processes, a span will not have access to all of the spans in its ancestor tree beyond its immediate parent span (usually through a context). This means that we will likely submit multiple trace fragments to LLMObs each with its own "root" span that doesn't have a LLMObs parent ID. This is fine for now as LLMObs does not currently support distributed tracing, but for when we want to, we'll need to rethink how we're submitting LLMObs spans.

The alternative approaches are:
- Submit all spans in a trace which contains a LLMObs span, which means we won't have to find the nearest LLMObs span in a span's parent tree (which will always be its immediate parent). Requires changes on LLMObs backend/intake to accept non-LLM type spans.
- Propagate LLMObs parent ID to all spans, as well as on distributed request/response headers. Not ideal as we will have to add tags to potentially every span.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


ddtrace/llmobs/_llmobs.py
```diff
@@ -108,22 +108,22 @@ def _llmobs_tags(span: Span, meta: Dict[str, Any]) -> List[str]:
...
         self._writer.enqueue(span_event)
 
-    def _llmobs_span_event(self, span, meta, metrics):
+    def _llmobs_span_event(self, span: Span, meta: Dict[str, Any], metrics: Dict[str, Any]) -> Dict[str, Any]:
         """Span event object structure."""
         return {
             "trace_id": "{:x}".format(span.trace_id),
             "span_id": str(span.span_id),
-            "parent_id": str(span.parent_id or ""),
+            "parent_id": str(self._get_llmobs_parent_id(span) or ""),
```
lievan, line 126: If `self._get_llmobs_parent_id(span)` returns `None`, i think `parent_id` will be set to `str(None)` instead of `""`. 

Is it intended for it to be `""` when there's no parent?

lievan (approved):
Looks good to me just one comment

ddtrace/llmobs/_llmobs.py
Yun-Kim: Good question! Using the `or` here returns `""` if `self._get_llmobs_parent_id(span)` returns `None`. If there's no parent, then this will always evaluate to `""`.

PR #8438, opened 2024-02-16
feat(tracing): deprecate `[]` notation for `DD_TRACE_METHODS`

erikayasuda:
## Description
This PR deprecates the `[]` notation for `DD_TRACE_METHODS` in favor of the `:` notation.

When customers tried to add `django_q` methods to `DD_TRACE_METHODS`, calling `__import__` during bootstrapping to figure out whether the provided method was a module or class method caused boostrapping errors. This is because modules like `django_q` attempt to do configurations upon attempting to import, which we don't want happening during bootstrapping.

The notation change for `DD_TRACE_METHODS` is primarily to help us distinguish between the legacy/deprecated and new semantics for `DD_TRACE_METHODS`. With the deprecated semantics, we needed to guess what the base module was by importing, since only the method name is within the brackets even if it's actually a class method. With the new semantics, we are asking customers to include the class with the method name if it's a class method.

Example:
```
DD_TRACE_METHODS="mymodule.mysubmodule.myclass[myfunc,otherfunc];..."
```
Here we would attempt to `__import__(mymodule.mysubmodule.myclass)`, fail because the class is included, then re-attempt with `__import__(mymodule.mysubmodule)`, which would succeed and indicate the base module is only `mymodule.mysubmodule`, and the functions are a class function called `myclass.myfunc` and `myclass.otherfunc`.

With the new semantics, this guess work would be taken out.
```
DD_TRACE_METHODS="mymodule.mysubmodule:myclass.myfunc,myclass.otherfunc;..."
```
Here we know for sure the base module is just `mymodule.mysubmodule`, and the function names are also explicitly listed.

## Risks
This is a customer facing change, so respective updates to customer facing documentation are included. After the next major release, customers will be required to switch to the updated semantics.

## Checklist

## Reviewer Checklist

- [x] Title is accurate

[APMS-10881]: https://datadoghq.atlassian.net/browse/APMS-10881?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ddtrace/internal/tracemethods.py
```diff
@@ -56,34 +95,38 @@ def _parse_trace_methods(raw_dd_trace_methods):
...
+        for qualified_method in _parse_legacy_trace_methods(raw_dd_trace_methods):
+            # We don't know if the method is a class method or a module method, so we need to assume it's a module
+            # and if the import fails then go a level up and try again.
+            base_module_guess = ".".join(qualified_method.split(".")[:-1])
+            method_name = qualified_method.split(".")[-1]
+            module = None
+
+            while base_module_guess:
+                try:
+                    module = __import__(base_module_guess)
```
P403n1x87, line 114: If this import is likely triggered at bootstrap time, then perhaps the better thing to do would be to register an after import hook, like we do with integrations. The problem of force-loading a module at boot is, besides the potential side-effects, the incompatibility with the module cloning mechanism for supporting things such as `gevent`.

erikayasuda: We do register an after import hook, but this import happens first to figure out whether the method is a module or class method. The goal for this PR is to deprecate the current syntax of DD_TRACE_METHODS so that this distinction is clear from the the env var itself, and we don't need to try these imports to figure it out during bootstrapping. That way we can just parse the string and make the after import hooks immediately.

ddtrace/internal/tracemethods.py
```diff
@@ -1,19 +1,68 @@
...
-    # type: (str) -> typing.List[str]
-    """Return the methods to trace based on the specification of DD_TRACE_METHODS.
 
-    DD_TRACE_METHODS is specified to be FullyQualifiedClassOrModuleName[comma-separated-methods]
+def _parse_trace_methods(raw_dd_trace_methods: str) -> List[Tuple[str, str]]:
+    """Return a list of the module,methodname tuples to trace based on the
+    specification of DD_TRACE_METHODS.
+
+    DD_TRACE_METHODS is specified to be FullyQualifiedModuleName:comma-separated-methods;...
+    The square bracket notation will be deprecated in favor of this new syntax
```
emmettbutler, line 15: This line doesn't make sense in the context of the rest of this file. It's a true statement, but `_parse_trace_methods` doesn't know anything about "square bracket notation" - that's `_parse_legacy_trace_methods`' concern.

ddtrace/internal/tracemethods.py
```diff
@@ -1,19 +1,68 @@
...
 
     Note that support for wildcard methods ([*]) is not implemented.
```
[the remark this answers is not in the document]
erikayasuda: Is this not relevant to call out? Or are we just mentioning it in the public docs and not in this method comment?

tests/contrib/django/test_django.py
```diff
@@ -2424,3 +2425,19 @@ def test_django_base_handler_failure(client, test_spans):
...
+    "dd_trace_methods,error_expected",
+    [("django_q.tasks[async_task]", True), ("django_q.tasks:async_task", False)],  # legacy syntax, updated syntax
+)
+def test_djangoq_dd_trace_methods(dd_trace_methods, error_expected):
+    with daphne_client(
+        "application",
+        additional_env={"DD_TRACE_METHODS": dd_trace_methods, "_DD_TRACE_WRITER_ADDITIONAL_HEADERS": ""},
+        django_settings_module="tests.contrib.django.django_app.settings",
+    ) as (client, proc):
+        assert client.get("simple/").status_code == 200
```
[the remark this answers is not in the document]
erikayasuda: Sounds relatively lightweight, so I think it makes sense to add that here. I'll probably do something similar to this? https://github.com/DataDog/dd-trace-py/blob/main/tests/contrib/django/test_django_patch.py#L20

tests/contrib/django/test_django_snapshots.py
```diff
@@ -19,7 +19,7 @@
 
 
 @contextmanager
-def daphne_client(django_asgi, additional_env=None):
+def daphne_client(django_asgi, additional_env=None, django_settings_module="tests.contrib.django.django_app.settings"):
```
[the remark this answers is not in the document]
erikayasuda: Nope, I needed this in the previous iteration but I ended up consolidating the configs to the same django app. I'll remove this!

ddtrace/internal/tracemethods.py
```diff
@@ -1,19 +1,68 @@
...
+                % qualified_methods
+            )
 
+        # Store the prefix and the methods  (eg. for "foo.bar.baz:qux,quux",
+        # this is "foo.bar.baz" for the prefix and "qux,quux" for the methods)
+        qualified_method_prefix, methods = qualified_methods.split(":")
+
+        if qualified_method_prefix == "__main__":
+            # __main__ cannot be used since the __main__ that exists now is not the same as the __main__ that the user
+            # application will have. __main__ when sitecustomize module is run is the builtin __main__.
```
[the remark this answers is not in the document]
erikayasuda: I think it would be helpful, I'll add some of this context to the error message

tests/contrib/django/test_django.py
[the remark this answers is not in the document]
erikayasuda: Actually if I'm gonna do that, I'll make it a snapshot test and move it to `test_django_snapshots.py`

erikayasuda:
@DataDog/apm-tees This change is the first step towards deprecating the legacy `DD_TRACE_METHODS` syntax that was causing some problems for customers trying to trace `django_q` methods. Details for the change are outlined in the PR description.

releasenotes/notes/fix-dd-trace-methods-2c1b4af30a7bf3ef.yaml
```diff
@@ -0,0 +1,5 @@
+---
+deprecations:
```
ZStriker19, line 2: I think we could make this two notes, one is a feature with the new notation, and one is the deprecation.

tests/contrib/django/test_django_snapshots.py
```diff
@@ -55,7 +55,7 @@ def daphne_client(django_asgi, additional_env=None):
     client.wait()
 
     try:
-        yield client
+        yield (client, proc)
```
ZStriker19, line 58: should we just create two daphne clients? One that yields (client,proc) and one that yields just client (like the original)? That way you don't have to touch all these other tests because the client is returning something we don't care about in most cases?
E.g.
```
def daphne_client_1
    yield client, proc

def daphne_client_2()
    client, proc = daphne_client_1()
    yield client
```

releasenotes/notes/fix-dd-trace-methods-2c1b4af30a7bf3ef.yaml
ZStriker19: Perhaps we could reference the docs in the feature rn?

scripts/run-test-suite
```diff
@@ -5,6 +5,8 @@ RIOT_PATTERN=${1}
 DDTRACE_FLAG=$([ -v _CI_DD_API_KEY ] && echo '--ddtrace')
 COVERAGE_FLAG=$([[ "${2:-false}" == false ]] && echo '--no-cov')
 DDTEST_CMD=$([[ ${3} == "1" ]] && echo "./scripts/ddtest")
+# temporary workaround for pytest issues in CI jobs
+DDTRACE_FLAG=$([[ "${4:-false}" == true ]] && echo '--no-ddtrace')
```
ZStriker19, line 9: Could you explain why we need this in the description?

erikayasuda: This was due to the pytest incompatibility that's getting worked on in this PR: https://github.com/DataDog/dd-trace-py/pull/8357

If it gets merged tomorrow, I'll remove this workaround and comment. If it looks like it's gonna take longer, I'll just merge this with this workaround and once the other PR is merged, I'll remove this workaround.

ddtrace/internal/tracemethods.py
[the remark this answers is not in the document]
emmettbutler: It's sort of out of place in the function that parses the new format, which doesn't allow any square brackets. If you want to keep it around, maybe move it to the legacy parser function?

scripts/run-test-suite
romainkomorndatadog: @erikayasuda , I would pin `pytest` to `~=7.0` in the relevant environments rather than disable the `ddtrace` plugin.

erikayasuda: Removed the workaround instead since it looks like the other PR merged. If there are still issues I'll pin `pytest` to `~=7.0` for the `django` test suites

ddtrace/internal/tracemethods.py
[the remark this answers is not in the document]
erikayasuda: Oh I see what you're pointing out. In that case, I might just remove the square brackets from this comment since wild cards aren't supported with for this new notation either.

tests/contrib/django/test_django_snapshots.py
erikayasuda: I think I'm gonna keep it as is for now, since I'm already yielding both anyways (and I can't really separate them). It does feel a little gross that I'm touching those other tests, but I figured I could clean this up later if it becomes excessive.

emmettbutler:
🎉

Issue #8206, opened 2024-01-29
ddtrace list.remove(x) x not in list while using subprocess.check_call


ya-pekatoros:
Hi there!

I've faced some problem with ddtrace, while calling subprocess.check_call

The example:
```
def chrome_print_pdf(
        content: str,
        output_file: str,
        app: str = PDF_DEFAULT_APP,
        debug: bool = False,
):
    if app is not None:
        content = content.replace('/static/', os.path.join(settings.BASE_DIR, app, 'static/'))

    ifd = tempfile.NamedTemporaryFile(suffix='.html', delete=(not debug))
    if debug:
        print(ifd.name)

    ifd.write(content.encode())
    ifd.flush()

    subprocess.check_call([
        'chromium', '--headless', '--disable-gpu', '--virtual-time-budget=10000',
        '--print-to-pdf=%s' % output_file,
        ifd.name,
    ])
```

During the process periodically I unexpectedly receive ValueError: list.remove(x): x not in list by ddtrace (seems like it happens during subprocess shutdown):

```
26 {4}
method : check_call
code : retcode = call(*popenargs, **kwargs)
lineno : 364
filename : /usr/local/lib/python3.10/subprocess.py

27 {4}
method : call
code : return p.wait(timeout=timeout)
lineno : 347
filename : /usr/local/lib/python3.10/subprocess.py

28 {4}
method : sentry_patched_popen_wait
code : return old_popen_wait(self, *a, **kw)
lineno : 235
filename : /packages310/lib/python3.10/site-packages/sentry_sdk/integrations/stdlib.py

29 {4}
method : wait
code : return self._wait(timeout=timeout)
lineno : 1209
filename : /usr/local/lib/python3.10/subprocess.py

30 {4}
method : _wait
code : (pid, sts) = self._try_wait(0)
lineno : 1959
filename : /usr/local/lib/python3.10/subprocess.py

31 {4}
method : _try_wait
code : (pid, sts) = os.waitpid(self.pid, wait_flags)
lineno : 1917
filename : /usr/local/lib/python3.10/subprocess.py

32 {4}
method : wrap_signals
code : f(*args, **kwargs)
lineno : 17
filename : /packages310/lib/python3.10/site-packages/ddtrace/internal/utils/signals.py

33 {4}
method : handle_exit
code : f()
lineno : 68
filename : /packages310/lib/python3.10/site-packages/ddtrace/internal/atexit.py

34 {4}
method : _atexit
[...]

PR #8449, opened 2024-02-16
fix(tracing): avoid exploding on valueerror during _atexit

emmettbutler:
This change fixes https://github.com/DataDog/dd-trace-py/issues/8206 by catching the `ValueError` that can sometimes be raised by tracer shutdown during `_atexit` handling. This makes sense because the tracer needs to shut itself down as quietly as possible to avoid polluting application logs.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


ddtrace/_trace/tracer.py
```diff
@@ -291,7 +291,10 @@ def _atexit(self) -> None:
             self.SHUTDOWN_TIMEOUT,
             key,
         )
-        self.shutdown(timeout=self.SHUTDOWN_TIMEOUT)
+        try:
+            self.shutdown(timeout=self.SHUTDOWN_TIMEOUT)
+        except ValueError:  # hook cleanup can sometimes raise this
```
mabdinur, line 296: The value error is likely raise here: https://github.com/DataDog/dd-trace-py/blob/9319ecf64d9cb0fb35a24339abec19151f942f06/ddtrace/internal/forksafe.py#L68. I think this behavior is intentional but I am not sure why we need to crash applications if we attempt to register a hook that doesn't exist.

Should we change this behavior for all ddtrace products?

emmettbutler: Seems worth a shot.

PR #8461, opened 2024-02-20
chore: make span.sampled reference span.context.sampling_priority and deprecate it

ZStriker19:
With `span.sampled` being outdated, and no longer reliable as a way to tell if a span is actually being sampled, we make this change to make its value reference `span.context.sampling_priority` which is what actually matters when it comes to sampling. This deprecates span.sampled and removes its use internally, replacing it with checks for span.context.sampling_priority to mimic behavior.

If the value of `span.context.sampling_priority`, is `None` we know that sampling hasn't yet run, and then maintain the original behavior of `span.sampled` being `True` until sampling is run, if it's >0 we know that the span has been sampled, if <=0 we know the span has not been sampled.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


ZStriker19:
Or do I make everything reference _sampling_priority_v1? and then deprecate span.sampled?

ddtrace/_trace/span.py
```diff
@@ -260,6 +260,23 @@ def duration(self, value):
         # type: (float) -> None
         self.duration_ns = int(value * 1e9)
 
+    @property
+    def sampled(self):
+        # type: () -> bool
+        if not self.context.sampling_priority:
+            return False
+        return self.context.sampling_priority > 0
```
mabdinur, line 268: We can simplify this a bit:
```suggestion
        return self.context.sampling_priority != None and self.context.sampling_priority > 0
```

ddtrace/_trace/span.py
```diff
@@ -260,6 +260,23 @@ def duration(self, value):
...
+
+    @sampled.setter
+    def sampled(self, value):
+        # type: (bool) -> None
+        # only use sampled to set this there isn't a parent
+        if self.context.sampling_priority is None:
+            if value is True:
+                self.context.sampling_priority = 1
+            elif value is False:
+                self.context.sampling_priority = 0
```
mabdinur, line 278:
```suggestion
            self.context.sampling_priority = int(value) # int(True) == 1, int(False) == 0
```

ddtrace/_trace/span.py
```diff
@@ -168,10 +166,11 @@ def __init__(
         self.parent_id = parent_id  # type: Optional[int]
         self._on_finish_callbacks = [] if on_finish is None else on_finish
 
+        self._context = context._with_span(self) if context else None  # type: Optional[Context]
+
```
ZStriker19, line 170: need to set `_context` before setting `sampled` now

ddtrace/_trace/span.py
```diff
@@ -82,8 +82,6 @@ class Span(object):
         "span_type",
         "start_ns",
         "duration_ns",
-        # Sampler attributes
```
ZStriker19, line 85: remove from slots so we don't cause an error between the slot and the property

emmettbutler (commented):
Want a review before fixing the test failures?

ZStriker19:
nope sorry, thought it was there, but realized a critical flaw.

ZStriker19:
def test_sampling_with_rate_sampler_with_tiny_rate(writer, tracer):
    # this fails because we never actually touch _sampling_priority_v1 and
    # we default to setting sampled to true therefore setting _sampling_priority_v1 to 1 by default
    # I guess I can just get rid of that in span.py and see what happens.
    # however that'll mess with whether or not we do certain behavior in some integrations
    
    # thinking of this as well, making sampled rely on _sampling_priority_v1 
    # won't totally work, because once we move sampling to the end of a span's life cycle
    # sampled will be false for the life of the span. So this original change 
    # won't quite work, or will need to be modified.

ZStriker19:
# previously this test passed because every span
    # started as sampled, and the ratesampler never actually changes if a span is sampled though
    def test_sample_rate_deviation(self):

ddtrace/_trace/span.py
```diff
@@ -260,6 +256,13 @@ def duration(self, value):
         # type: (float) -> None
         self.duration_ns = int(value * 1e9)
 
+    @property
+    def sampled(self):
+        # type: () -> Optional[bool]
+        if self.context.sampling_priority is None:
+            return None
+        return self.context.sampling_priority > 0
```
mabdinur, line 264:
```suggestion
        return self.context.sampling_priority > 0
      
     @sampled.settr
     def sampled(self, value):
          @deprecated("pls don't do this, use set_tag(Manual keep/drop), this is a no-op"")
```

We should include a setter but make sure it logs a deprecation warning. This will maintain backwards compatibility. We should make sure we never use this internally

ddtrace/_trace/span.py
```diff
@@ -260,6 +258,29 @@ def duration(self, value):
         # type: (float) -> None
         self.duration_ns = int(value * 1e9)
 
+    @property
+    def sampled(self):
+        # type: () -> Optional[bool]
+        deprecate(
+            "span.sampled is deprecated and will be removed in a future version of the tracer.",
+            message="""span.sampled references the state of `span.context.sampling_priority`.
```
ZStriker19, line 266:
```suggestion
            message="""span.sampled references the state of span.context.sampling_priority.
```

ddtrace/_trace/span.py
```diff
@@ -260,6 +258,29 @@ def duration(self, value):
         # type: (float) -> None
         self.duration_ns = int(value * 1e9)
 
+    @property
+    def sampled(self):
+        # type: () -> Optional[bool]
+        deprecate(
+            "span.sampled is deprecated and will be removed in a future version of the tracer.",
+            message="""span.sampled references the state of span.context.sampling_priority.
+            Please use `span.context.sampling_priority` instead to check if a span is sampled.""",
```
ZStriker19, line 267:
```suggestion
            Please use span.context.sampling_priority instead to check if a span is sampled.""",
```

ddtrace/_trace/span.py
```diff
@@ -260,6 +258,29 @@ def duration(self, value):
...
+            # this maintains original span.sampled behavior, where all spans would start
+            # with span.sampled = True until sampling runs
+            return True
+        return self.context.sampling_priority > 0
+
+    @sampled.setter
+    def sampled(self, value):
+        deprecate(
+            "span.sampled is deprecated and will be removed in a future version of the tracer.",
+            message="This is a no-op setter. Please use span.set_tag(MANUAL_KEEP/MANUAL_DROP) to keep or drop spans.",
```
ZStriker19, line 280:
```suggestion
            message="This is a no-op setter. Please use span.set_tag("manual.keep"/"manual.drop") to keep or drop spans.",
```

emmettbutler (approved):

mabdinur (approved):

Yun-Kim (approved):
LGTM for ml_obs 👍

PR #8471, opened 2024-02-21
feat(profiling): add experimental outside-in stack sampling method (stack v2)

sanchda:
This allows users to optionally use a new stack sampling technique.  Briefly,

* Avoids issues in Python 3.11 and later where the C struct backing Python frame objects may contain invalid pointers (causing segfaults within standard API calls)
* Addresses a bias in the "normal" stack sampler where we had a tendency to sample threads only at the points where they transitioned off-GIL
* Hopefully removes some serialization overhead, although in the first few versions the overall consumption of this technique may be higher.

There are a few tests around the new code, those will be added to CI in a later update.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


ddtrace/internal/datadog/profiling/stack_v2/src/stack_renderer.cpp
```diff
@@ -0,0 +1,89 @@
...
+StackRenderer::render_message(std::string_view msg)
+{
+    // This is unused for now, since the dd-trace-py profiler doesn't support emitting arbitrary messages.
+    (void)msg;
+}
+
+void
+StackRenderer::render_thread_begin(PyThreadState* tstate,
+                                   std::string_view name,
+                                   microsecond_t wall_time_us,
```
danielsn, line 16: We sometime use this type, and sometimes just int.  Is there a reason for which/when?

ddtrace/internal/datadog/profiling/stack_v2/src/stack_renderer.cpp
```diff
@@ -0,0 +1,89 @@
...
+                                   uintptr_t thread_id,
+                                   unsigned long native_id)
+{
+    (void)tstate;
+    sample = ddup_start_sample();
+    if (sample == nullptr) {
```
danielsn, line 22: Should we log an error?

ddtrace/internal/datadog/profiling/stack_v2/src/stack_renderer.cpp
```diff
@@ -0,0 +1,89 @@
...
+        return;
+    }
+    ddup_push_threadinfo(sample, static_cast<int64_t>(thread_id), static_cast<int64_t>(native_id), name.data());
```
danielsn, line 25: stringview is not null terminated.  Shouldn't we also pass the length?

ddtrace/internal/datadog/profiling/stack_v2/src/stack_renderer.cpp
```diff
@@ -0,0 +1,89 @@
...
+        return;
+    }
+    static const std::string_view invalid = "<invalid_utf8>";
+    if (!utf8_check_is_valid(name.data(), name.size())) {
+        name = invalid;
+    }
+    if (!utf8_check_is_valid(file.data(), file.size())) {
+        file = invalid;
+    }
+    ddup_push_frame(sample, name.data(), file.data(), 0, line);
```
danielsn, line 48: String views are not null terminated

ddtrace/internal/datadog/profiling/stack_v2/src/stack_renderer.cpp
```diff
@@ -0,0 +1,89 @@
...
+    ddup_drop_sample(sample);
+    sample = nullptr;
+}
+
+void
+StackRenderer::render_native_frame(std::string_view name, std::string_view file, uint64_t line)
+{
+    if (sample == nullptr) {
+        return;
+    }
```
danielsn, line 58: This is an error.  Should we log it somewhere?

ddtrace/internal/datadog/profiling/stack_v2/src/stack_renderer.cpp
sanchda: Good call.  @r1viollet made the same observation.

Refactored interface.cpp to accept a `string_view` instead of a naked C string.

ddtrace/internal/datadog/profiling/stack_v2/include/util/cast_to_pyfunc.hpp
```diff
@@ -0,0 +1,11 @@
+#pragma once
+#include <Python.h>
+
+// Abiding by the contract required in the Python interface requires abandoning certain checks in one's main project.
+// Rather than abandoning discipline, we can include this header in a way that bypasses these checks.
```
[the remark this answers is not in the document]
sanchda: This was an overly terse description.  I've fleshed out the problem and rational in more detail.

ddtrace/internal/datadog/profiling/dd_wrapper/test/forking.cpp
```diff
@@ -14,8 +14,8 @@ upload_in_thread()
 [[noreturn]] void
 profile_in_child(unsigned int num_threads, unsigned int run_time_ns, std::atomic<bool>& done)
 {
-    // Assumes setup has been called. Launch some samplers, wait, upload, exit.
-    // Managing interleaved execution is tricky, so the fork exits--rather than returns--when it is done.
+    // Assumes setup has been called. Launch some samplers, wait, upload, return.
+    // Exit because we don't want this process to exit
```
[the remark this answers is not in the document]
sanchda: Not sure why I clobbered the old one.  Reverted.

ddtrace/internal/datadog/profiling/ddup/_ddup.pyx
```diff
@@ -13,6 +16,18 @@ from .utils import sanitize_string
...
+        cdef cppclass string_view:
+            string_view(const char* s, size_t count)
+
+    cdef pystr_to_sv(object s):
+        cdef bytes s_bytes = b""
+        try:
+            s_bytes = ensure_binary(s)
+        except Exception:
+            pass
+        return string_view(<const char*>s_bytes, len(s_bytes))
```
danielsn, line 29: Is there any thing special we need to do to handle unicode?

ddtrace/internal/datadog/profiling/dd_wrapper/src/uploader_builder.cpp
```diff
@@ -115,8 +115,10 @@ UploaderBuilder::build()
 
     for (const auto& [tag, data] : tag_data) {
         std::string errmsg;
-        if (!add_tag(tags, tag, data, errmsg)) {
-            reasons.push_back(std::string(to_string(tag)) + ": " + errmsg);
+        if (!data.empty()) {
```
danielsn, line 118: Log an error here

ddtrace/internal/datadog/profiling/setup_custom.sh
```diff
@@ -0,0 +1,275 @@
...
+highest_clang=""
+highest_clangxx=""
+
+# Function to find the highest version of compilers
+find_highest_compiler_version() {
+  local base_name=$1
+  local highest_var_name=$2
+  local highest_version=0
+
+  for version in {20..5}; do
```
danielsn, line 22: Where does this range come from?

ddtrace/internal/datadog/profiling/stack_v2/include/stack_renderer.hpp
```diff
@@ -0,0 +1,39 @@
+#pragma once
+
+// Makes up for shortcomings in my echion PR
```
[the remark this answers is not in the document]
sanchda: Oops, got rid of that section, but kept the comment.  Removed

ddtrace/internal/datadog/profiling/stack_v2/src/sampler.cpp
```diff
@@ -0,0 +1,112 @@
...
+        for_each_interp([&](PyInterpreterState* interp) -> void {
+            for_each_thread(interp, [&](PyThreadState* tstate, ThreadInfo& thread) {
+                thread.sample(interp->id, tstate, wall_time_us);
+            });
+        });
+
+        // If we've been asked to stop, then stop
+        if (!is_profiling.load()) {
+            break;
+        }
```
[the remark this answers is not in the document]
sanchda: Indeed

ddtrace/internal/datadog/profiling/stack_v2/src/sampler.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+        // Sleep for the remainder of the interval, get it atomically
+        // Generally speaking system "sleep" times will wait _at least_ as long as the specified time, so
+        // in actual fact the duration may be more than we indicated.  This tends to be more true on busy
+        // systems.
+        std::this_thread::sleep_until(sample_time_now + microseconds(sample_interval_us.load()));
+    }
+
+    // Release the profiling mutex
+    profiling_mutex.unlock();
```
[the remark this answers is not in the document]
sanchda: Removed this whole mess in favor of a simpler approach.  Whenever a thread is started or stopped, a transaction number is incremented.  When a thread is started, a copy of that number is given to the thread as an identifier.  Whenever the thread's identifier does not match the global value, then the sampling thread is invalidated.

sanchda:
Marking as manual merge.  I forced the affected extensions to always build to see how packaging works in CI.

ddtrace/internal/datadog/profiling/stack_v2/src/stack_renderer.cpp
sanchda: Agree, logging to cerr for now as we did in the previous PR, then we'll incorporate all of those lines into the logger in the sequel.

ddtrace/internal/datadog/profiling/stack_v2/src/stackv2.cpp
```diff
@@ -0,0 +1,62 @@
...
+    double min_interval_s = g_default_sampling_period_s;
+    double max_nframes = g_default_max_nframes;
+
+    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "|dd", kwlist, &min_interval_f, &max_frames)) {
+        return NULL; // If an error occurs during argument parsing
+    }
+
+    Sampler::get().set_interval(min_interval_s);
+    Sampler::get().set_max_nframes(max_nframes);
+    Sampler::init(min_interval_s, max_frames);
```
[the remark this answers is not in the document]
sanchda: Agree, this was silly.  Fixed!

ddtrace/internal/datadog/profiling/stack_v2/include/sampler.hpp
```diff
@@ -3,34 +3,38 @@
...
     // The sampling interval is atomic because it needs to be safely propagated to the sampling thread
-    std::atomic<unsigned long> sample_interval_us{ g_default_sampling_period_us };
+    std::atomic<microsecond_t> sample_interval_us{ g_default_sampling_period_us };
 
-    std::atomic<bool> is_profiling{ false };
-    std::mutex profiling_mutex;
+    // This is not a running total of the number of launched threads; it is a sequence for the
+    // transactions upon the sampling threads (usually starts + stops). This allows threads to be
+    // stopped or started in a straightforward manner without finer-grained control (locks)
+    std::atomic<unsigned long> thread_seq_num{ 0 };
```
danielsn, line 23: `uint64_t`

ddtrace/internal/datadog/profiling/stack_v2/src/sampler.cpp
```diff
@@ -24,33 +24,29 @@ Sampler::sampling_thread()
...
-
-    // Release the profiling mutex
-    profiling_mutex.unlock();
 }
 
 void
 Sampler::set_interval(double new_interval_s)
 {
-    unsigned int new_interval_us = static_cast<unsigned int>(new_interval_s * 1e6);
+    microsecond_t new_interval_us = static_cast<unsigned int>(new_interval_s * 1e6);
```
danielsn, line 38: Should this be `static_cast<microsecond_t>`?

ddtrace/internal/datadog/profiling/stack_v2/src/sampler.cpp
```diff
@@ -7,12 +7,12 @@
...
 
 void
-Sampler::sampling_thread()
+Sampler::sampling_thread(unsigned long seq_num)
 {
     using namespace std::chrono;
     auto sample_time_prev = steady_clock::now();
 
-    while (true) {
+    while (seq_num == thread_seq_num.load()) {
```
danielsn, line 15: I like this solution much better :)

ddtrace/internal/datadog/profiling/stack_v2/src/sampler.cpp
```diff
@@ -66,47 +62,32 @@ Sampler::get()
...
+Sampler::start()
+{
+    static std::once_flag once;
+    std::call_once(once, [this]() { this->one_time_setup(); });
 
-    // OK, now we can start the profiler.
-    is_profiling.store(true);
-    std::thread t(&Sampler::sampling_thread, this);
+    // OK, now we can start the profiler.  When we launch the profiler, we mutate the sequence number and give it to the
+    // thread.  This will (eventually) terminate any outstanding thread.
```
danielsn, line 82: I find this comment confusing

ddtrace/internal/datadog/profiling/stack_v2/src/sampler.cpp
```diff
@@ -24,33 +24,29 @@ Sampler::sampling_thread()
             });
         });
 
-        // If we've been asked to stop, then stop
-        if (!is_profiling.load()) {
-            break;
-        }
-
         // Sleep for the remainder of the interval, get it atomically
```
danielsn, line 27: Would it be worth doing another check for changed sequence number right before the sleep?  Or is the sample itself fast enough relative to the sleep that this wouldn't help much?

ddtrace/internal/datadog/profiling/stack_v2/src/sampler.cpp
```diff
@@ -24,33 +24,29 @@ Sampler::sampling_thread()
...
-    unsigned int new_interval_us = static_cast<unsigned int>(new_interval_s * 1e6);
+    microsecond_t new_interval_us = static_cast<unsigned int>(new_interval_s * 1e6);
     sample_interval_us.store(new_interval_us);
 }
 
 void
 Sampler::set_max_nframes(unsigned int _max_nframes)
 {
-    max_nframes = _max_nframes;
+    if (thread_seq_num.load() == 0) {
```
danielsn, line 45: log an error otherwise

ddtrace/internal/datadog/profiling/stack_v2/src/sampler.cpp
```diff
@@ -24,33 +24,29 @@ Sampler::sampling_thread()
...
 void
 Sampler::set_interval(double new_interval_s)
 {
-    unsigned int new_interval_us = static_cast<unsigned int>(new_interval_s * 1e6);
+    microsecond_t new_interval_us = static_cast<unsigned int>(new_interval_s * 1e6);
     sample_interval_us.store(new_interval_us);
 }
 
 void
 Sampler::set_max_nframes(unsigned int _max_nframes)
```
danielsn, line 43: Why is this not part of one-time-setup?

ddtrace/internal/datadog/profiling/stack_v2/include/sampler.hpp
sanchda: I can dig it.  Fixed

ddtrace/internal/datadog/profiling/stack_v2/src/sampler.cpp
sanchda: Me too.  Fixed, I think.

ddtrace/internal/datadog/profiling/stack_v2/src/sampler.cpp
sanchda: We can, it's a cheap check.  Fixed.

ddtrace/internal/datadog/profiling/stack_v2/src/sampler.cpp
sanchda: Ah, this isn't straightforward.  I factored it out differently, which removes the propagation of max nframes from this context.

sanchda:
I have an outstanding cmake issue which broken IAST, but otherwise this is good to review.

ddtrace/internal/datadog/profiling/stack_v2/src/sampler.cpp
```diff
@@ -7,7 +7,7 @@
 using namespace Datadog;
 
 void
-Sampler::sampling_thread(unsigned long seq_num)
+Sampler::sampling_thread(uint64_t seq_num)
```
danielsn, line 10: ⛏️ `const uint64_t`

danielsn (approved):
LGTM

ddtrace/internal/datadog/profiling/stack_v2/src/sampler.cpp
sanchda: Absolutely, fixed.  I'll run this through my cppcheck harness.

ddtrace/internal/datadog/profiling/stack_v2/include/util/cast_to_pyfunc.hpp
```diff
@@ -0,0 +1,11 @@
...
+inline PyCFunction
```
P403n1x87, line 6:
```suggestion
inline static PyCFunction
```
?

ddtrace/internal/datadog/profiling/stack_v2/include/vendored/utf8_validate.hpp
```diff
@@ -0,0 +1,34 @@
+#pragma once
+
+// From https://gist.github.com/ichramm/3ffeaf7ba4f24853e9ecaf176da84566
+inline static bool
+utf8_check_is_valid(const char* str, int len)
```
P403n1x87, line 5: Can we verify that this is needed for CPython >= 3.7? In theory, as far as stacks are concerned,  filenames, and names/qualnames should be instances of `str`

ddtrace/internal/datadog/profiling/stack_v2/src/stack_v2.cpp
```diff
@@ -0,0 +1,71 @@
...
+    static char** kwlist = const_cast<char**>(const_kwlist);
+    double min_interval_s = g_default_sampling_period_s;
+
+    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "|d", kwlist, &min_interval_s)) {
+        return NULL; // If an error occurs during argument parsing
+    }
+
+    Sampler::get().set_interval(min_interval_s);
+
+    return PyLong_FromLong(1);
```
P403n1x87, line 21: What feedback would we lose if we returned `None` instead?

ddtrace/profiling/collector/memalloc.py
```diff
@@ -182,8 +179,8 @@ def collect(self):
                     # DEV: This might happen if the memalloc sofile is unlinked and relinked without module
                     #      re-initialization.
                     LOG.debug("Invalid state detected in memalloc module, suppressing profile")
-
-        if self._export_py_enabled:
+            return []
```
P403n1x87, line 182:
```suggestion
            return tuple()
```
for type consistency?

ddtrace/settings/profiling.py
```diff
@@ -40,7 +40,12 @@ def _derive_default_heap_sample_size(heap_config, default_heap_sample_size=1024
 
 def _is_valid_libdatadog():
     # type: () -> bool
-    return platform.machine() in ["x86_64", "aarch64"] and "glibc" in platform.libc_ver()[0]
+    return platform.system() == "Linux"
+
+
+def _is_valid_v2_stack():
+    # type: () -> bool
+    return platform.python_version_tuple() >= ("3", "7") and _is_valid_libdatadog()
```
P403n1x87, line 48:
```suggestion
    return sys.version_info >= (3, 7) and _is_valid_libdatadog()
```

ddtrace/internal/datadog/profiling/stack_v2/include/vendored/utf8_validate.hpp
sanchda: Ah, yeah--underdocumented procedure, sorry.

I noticed very rarely during testing that sometimes non-unicode strings would be passed to libdatadog.  My hunch is that those are moments where one of two things happen
1. We're in the middle of reading a frame (well, code...) struct in echion and it gets over-written in a way that replaces the string with a valid pointer, but the target is a non-string.
2. We're in the middle of reading from a string location, but that string had been cleared and re-allocated to some other struct (maybe this is more likely).

This is rare, but currently in libdatadog it drops the whole sample.  Or it panics?  I forgot TBH.  What I'd like to do eventually is handle the difference in libdatadog, since I'm anxious about doing _multiple_ levels of encoding-validation in the first place.  For now, having a check like this burns some CPU, but ultimately allows us to retain more of the stack.

I'm fine either way (drop the sample efficiently or try to keep the sample inefficiently).  What do you think?

ddtrace/internal/datadog/profiling/stack_v2/src/stack_v2.cpp
sanchda: Ah, good point.  Originally this returned a bool, but right now there's nothing to do about it.  Agree.

ddtrace/settings/profiling.py
sanchda: Oof.  Thank you.

ddtrace/profiling/collector/memalloc.py
sanchda: I think that's right.  Yoink!

ddtrace/internal/datadog/profiling/stack_v2/include/vendored/utf8_validate.hpp
P403n1x87: Ah ok thanks for the extra context. I think it's fine to keep this for now then if we get the benefit of retaining more data.

releasenotes/notes/profiling-add-stack-v2-975ebe845cdadc8f.yaml
```diff
@@ -0,0 +1,9 @@
+---
+features:
+  - |
+    profiling: implement an experimental stack sampling feature, which can be enabled by setting
+    DD_PROFILING_STACK_V2_ENABLED=true. This new sampler should resolve segfault issues on Python
```
majorgreys, line 5:
```suggestion
    ``DD_PROFILING_STACK_V2_ENABLED=true``. This new sampler should resolve segfault issues on Python
```

releasenotes/notes/profiling-add-stack-v2-975ebe845cdadc8f.yaml
```diff
@@ -0,0 +1,9 @@
...
+    3.11 and later, while also decreasing the latency contribution of the profiler in many
+    situations, and also improving the accuracy of stack-sampling data. This feature is currently
+    only available on Linux using cpython 3.8 or greater. Requires
+    DD_PROFILING_EXPORT_LIBDD_ENABLED=true to be set.
```
majorgreys, line 9:
```suggestion
    ``DD_PROFILING_EXPORT_LIBDD_ENABLED=true`` to be set.
```

releasenotes/notes/profiling-add-stack-v2-975ebe845cdadc8f.yaml
```diff
@@ -0,0 +1,9 @@
+---
+features:
+  - |
+    profiling: implement an experimental stack sampling feature, which can be enabled by setting
+    DD_PROFILING_STACK_V2_ENABLED=true. This new sampler should resolve segfault issues on Python
+    3.11 and later, while also decreasing the latency contribution of the profiler in many
+    situations, and also improving the accuracy of stack-sampling data. This feature is currently
+    only available on Linux using cpython 3.8 or greater. Requires
```
majorgreys, line 8:
```suggestion
    only available on Linux using CPython 3.8 or greater. Requires
```

setup.py
```diff
@@ -454,25 +453,30 @@ def get_exts_for(name):
...
                 cmake_args=[
                     "-DPY_MAJOR_VERSION={}".format(sys.version_info.major),
                     "-DPY_MINOR_VERSION={}".format(sys.version_info.minor),
                     "-DPY_MICRO_VERSION={}".format(sys.version_info.micro),
-                    "-Ddd_wrapper_INSTALL_DIR={}".format(DDUP_DIR),
                 ],
             )
         )
 
+        # One of the stack v2 dependencies doesn't quite work on 3.7, so only support later for now
```
majorgreys, line 471: Have you made a note already about which dependency is the issue?

majorgreys (changes requested):
only formatting of release note

setup.py
sanchda: Will make this explicit

sanchda: Fixed in [6d56227](https://github.com/DataDog/dd-trace-py/pull/8471/commits/6d5622744a3a5e6472bb26bc15a52b97348bfa47)

ddtrace/internal/datadog/profiling/stack_v2/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+
+    LOG = get_logger(__name__)
+    LOG.debug("Failed to import _stack_v2: %s", e)
+
+    def is_available():
+        # type: () -> bool
+        return False
+
+    # Decorator for not-implemented
+    def not_implemented(func):
```
P403n1x87, line 19: Perhaps we can define this once in `ddtrace.internal.compat`?

ddtrace/internal/datadog/profiling/stack_v2/src/stack_v2.cpp
```diff
@@ -0,0 +1,76 @@
...
+PyCFunction stack_v2_start = cast_to_pycfunction(_stack_v2_start);
+
+static PyObject*
+stack_v2_stop(PyObject* self, PyObject* args)
+{
+    (void)self;
+    (void)args;
+    Sampler::get().stop();
+    Py_INCREF(Py_None);
+    return Py_None;
```
P403n1x87, line 39:
```suggestion
    Py_RETURN_NONE
```
https://docs.python.org/3.10/c-api/none.html#c.Py_RETURN_NONE

ddtrace/internal/datadog/profiling/stack_v2/src/stack_v2.cpp
```diff
@@ -0,0 +1,76 @@
...
+{
+    // Assumes the interval is given in fractional seconds
+    (void)self;
+    double new_interval;
+    if (!PyArg_ParseTuple(args, "d", &new_interval)) {
+        return NULL; // If an error occurs during argument parsing
+    }
+    Sampler::get().set_interval(new_interval);
+    Py_INCREF(Py_None);
+    return Py_None;
```
P403n1x87, line 53:
```suggestion
    Py_RETURN_NONE;
```

ddtrace/internal/datadog/profiling/stack_v2/__init__.py
```diff
@@ -0,0 +1,33 @@
+try:
+    from ._stack_v2 import *  # noqa: F401, F403
+
+    def is_available():
```
P403n1x87, line 4: If we make this a module attribute instead we could probably save a function call

ddtrace/settings/profiling.py
```diff
@@ -40,7 +41,12 @@ def _derive_default_heap_sample_size(heap_config, default_heap_sample_size=1024
 
 def _is_valid_libdatadog():
     # type: () -> bool
-    return platform.machine() in ["x86_64", "aarch64"] and "glibc" in platform.libc_ver()[0]
+    return platform.system() == "Linux"
+
+
+def _is_valid_v2_stack():
+    # type: () -> bool
+    return sys.version_info >= (3, 7) and _is_valid_libdatadog()
```
P403n1x87, line 49:
```suggestion
    return _is_valid_libdatadog()
```
We no longer support < 3.7 so this condition can never be true?

sanchda: Ah, good catch.  First of all, this check is wrong.  Second, it's no longer needed because of how we check later.  Will fix.

ddtrace/internal/datadog/profiling/stack_v2/__init__.py
sanchda: Good call, although fixing this up would be a slight detour (I made testing annoying by fencing this failure path off).  I think we might consolidate these interfaces differently in the near-term anyway.

PR #8476, opened 2024-02-21
feat(llmobs): inline annotations for LLMObs spans

Yun-Kim:
This PR adds an LLMObs SDK to allow users to instrument their applications with LLMObs spans.
Additionally, this PR does the following:
- Fixes the internal temporary tag system of storing LLMObs data on APM spans.
- Adjusts the OpenAI/Bedrock LLMObs integrations to submit spans using the new SDK method.
- Refactors the OpenAI tests to use a `_expected_llmobs_tags()` helper.

### LLMObs Span Types

There are 5 types of operations currently traced by the SDK:
- Agent: entity that interacts with, initiates, and/or orchestrates actions in the LLM chain
- Workflow: series of static steps executed given an input to return an output
- LLM: interaction with a large language model
- Tool: software/interfaces used for interacting with or supporting the LLM
- Task: specific objective or function performed with the LLM. (e.g. preprocessing, language translation, analysis, postprocessing)

## Tracing an operation

The `ddtrace.llmobs.LLMObs` service provides the following methods for users to trace an operation that a given code block entails: `LLMObs.agent(), LLMObs.workflow(), LLMObs.tool(), LLMObs.task(), LLMObs.llm()`. These methods accept the following arguments:
- `name (Optional)`: name of the operation. If not provided, the operation type will be set as the name of the span by default.
- `model_name (Required; only for llm())`: name of the invoked LLM.
- `model_provider (Optional; only for llm())`: name of LLM provider. If not provided, a value of "custom" will be set by default.
- `session_id (Optional)`: ID of underlying user session. Required for tracking sessions. Note: session ID only needs to be set once per trace/process.

These methods can be used just like `ddtrace.tracer.trace()`, i.e. returns a Datadog span object, both as a context manager:
```python

def sanitize_prompt_automatic_context_management(prompt):
	with LLMObs.task(name="sanitize_prompt") as task_span:
	# user application logic to sanitize a prompt
	return sanitized_prompt
```
Or the span can be started and finished manually:
```python

def sanitize_prompt_but_manually(prompt):
	task_span = LLMObs.task(name="sanitize_prompt")
	# user application logic to sanitize a prompt
	task_span.finish()
	return sanitized_prompt

```
### OOTB Integrations
**Note**: for OpenAI Python SDK and AWS Bedrock (using boto3/botocore) users, you do not need to use the SDK to trace LLM calls as the OpenAI/Bedrock integrations automatically trace and annotate the LLM calls.

## Annotating a span

The SDK provides a method `LLMObs.annotate()` to annotate spans with inputs, outputs, and metadata. The LLMObs.annotate() method accepts the following arguments: 
- span: The span to annotate. If span is not provided, the SDK will tag the current active span instead.
- parameters: A dictionary of input parameter key-value pairs such as max tokens and model temperature.
- input_data: Either a raw string (for workflow/tool/task spans) or a list of dictionaries with this format: {“role”: “...”, “content”: “...”} (for llm/agent spans). 
- output_data: Same as input_data. 
- tags: A dictionary of key-value pairs that users can add as arbitrary tags on the span. 
- metrics: A dictionary of ke-value pairs that users can add as metrics on the span (i.e. `prompt_tokens, completion_tokens, total_tokens`)

An example use case of the `LLMObs.annotate()` method is shown below:
```python

def llm_call_inline(prompt):
	with LLMObs.llm(name="llm_call", model="model_name", model_provider="model_provider") as llm_span:
		resp = ... # llm call here
		LLMObs.annotate(
  			span=llm_span,
parameters={"temperature": 0, "max_tokens": 200},
input_data=[{"role": "user", "content": "Hello world!"}],
output_data=[{"role": "assistant", "content": "How can I help?"}],
tags={"host": "host_name"},
metrics={"prompt_tokens": 30, "completion_tokens": 20, "total_tokens": 50},
)
	return resp
```

## Internal Design
Previously we would store LLMObs data on APM spans using temporary tags: `_ml_obs.meta, _ml_obs.metrics`. However, this was problematic and led to increasing complexity as the `ml_obs.meta` dict is a large dictionary with arbitrary value types, and since spans can only store tags as strings, we would need to json dump/load the temporary `_ml_obs.meta` tag every time we wanted to modify it.

The new approach is to split out the `_ml_obs.meta` field into multiple tags, i.e. `_ml_obs.meta.model_name`, `_ml_obs.meta.model_provider`, `_ml_obs.meta.input.messages`, `_ml_obs.meta.input_value`, `_ml_obs.meta.input.parameters`, `_ml_obs.meta.output.value`, `_ml_obs.meta.output.messages`, `_ml_obs.meta.span.kind`, `_ml_obs.session_id`, and `_ml_obs.tags`. This way, we know exactly what types of data each tag should map to, and we don't need to repeatedly json dump/load tags to modify them as before.

### Future Steps
This PR only covers inline tracing. The next step is to offer function decorator support to minimize difficulty of instrumentation.
Additionally, we need to add support for OpenAI streamed responses, as we don't currently extract/store response data.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


ddtrace/llmobs/_llmobs.py
```diff
@@ -75,6 +89,164 @@ def disable(cls) -> None:
...
+    def annotate(
+        cls,
+        span: Optional[Span] = None,
+        parameters: Optional[Dict[str, Any]] = None,
+        input_data: Optional[Union[List[Dict[str, Any]], str]] = None,
+        output_data: Optional[Union[List[Dict[str, Any]], str]] = None,
+        tags: Optional[Dict[str, Any]] = None,
+    ):
+        """
+        Sets the parameter, input, and output tags for a given LLMObs span.
```
sabrenner, line 164: Nice docstring! I think it's very descriptive & informative

ddtrace/llmobs/_llmobs.py
```diff
@@ -75,6 +89,164 @@ def disable(cls) -> None:
...
+        tags: Dictionary of key-value pairs to tag the LLMObs span with.
+        """
+        if cls.enabled is False or cls._instance is None:
+            raise Exception("LLMObs.annotate() cannot be used while LLMObs is disabled.")
+        if span is None:
+            span = cls._instance.tracer.current_span()
+            if span is None:
+                raise ValueError("No span provided and no active span found.")
+        if span.span_type != SpanTypes.LLM:
+            raise ValueError("span must be an LLM-type span.")
```
sabrenner, line 184: More of a curiosity question, def not blocking: is there a reason for raising exceptions over logging + gracefully returning? I know in the Node tracer we opt for the latter, in favor of a smooth user experience (except for instrumented library errors, which we make sure get thrown). 

I see you do a mix of both (as in the lines below), so I was just wondering if it's a best-practice Python thing, or something else. 

**Thought:** since this a new aspect of the user experience for the Python tracer, maybe logging could provide a better end experience?

ddtrace/llmobs/_llmobs.py
```diff
@@ -75,6 +89,164 @@ def disable(cls) -> None:
...
+        span.set_tag_str(SPAN_KIND, operation_kind)
+        if session_id is not None:
+            span.set_tag_str(SESSION_ID, session_id)
+        if model_name is not None:
+            span.set_tag_str(MODEL_NAME, model_name)
+        if model_provider is not None:
+            span.set_tag_str(MODEL_PROVIDER, model_provider)
+        return span
+
+    @classmethod
```
sabrenner, line 112: Should these methods also have docstrings? I know there are actual docs themselves in the works/that can be referred to, but wondering if a copypasta/short version here could be beneficial to user experience with the SDK as well.

sabrenner (approved):
Super cool! Had one question about docstrings, and another about throwing errors vs logging & returning, but that one's really more of a curiosity thing. Thanks for the informative PR description, it was very helpful in putting these changes into context WRT how the SDK is designed to be used. Actual implementation here LGTM too!

ddtrace/llmobs/_llmobs.py
Yun-Kim: Yes they should, good catch!

ddtrace/llmobs/_llmobs.py
```diff
@@ -75,6 +89,164 @@ def disable(cls) -> None:
...
+        if not model_name:
+            raise ValueError("model_name must be the specified name of the invoked model.")
+        if model_provider is None:
+            model_provider = "custom"
+        return cls._instance._start_span(
+            "llm", name, model_name=model_name, model_provider=model_provider, session_id=session_id
+        )
+
+    @classmethod
+    def tool(cls, name: Optional[str] = None, session_id: Optional[str] = None) -> Span:
```
Yun-Kim, line 131: Note to self: need to propagate session_id downwards so that users don't need to set it explicitly for every single span

ddtrace/llmobs/_llmobs.py
Yun-Kim: Good point, will think about logging rather than crashing the user app - we should aim to be as minimally disruptive as possible.

ddtrace/llmobs/_llmobs.py
```diff
@@ -75,6 +89,164 @@ def disable(cls) -> None:
...
+        """
+        Tags input/output for a given LLMObs span.
+        io_type: "input" or "output".
+        meta: Span's meta dictionary.
+        data: String or dictionary of key-value pairs to tag as the input or output.
+        """
+        if io_type not in ("input", "output"):
+            raise ValueError("io_type must be either 'input' or 'output'.")
+        if not isinstance(data, (str, list)):
+            raise TypeError(
```
Yun-Kim, line 214: Note: probably should just warn and return here to avoid disrupting user app

PR #8530, opened 2024-02-27
feat(llmobs): add ml_app tag to LLMObs spans

Yun-Kim:
This PR adds a `DD_LLMOBS_APP_NAME` env var which configures the value of the `ml_app` tag to set on LLMObs span events.
Note that this env var is for private use only, so this doesn't need a release note.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


ddtrace/llmobs/_llmobs.py
```diff
@@ -382,10 +382,11 @@ def _llmobs_span_event(self, span: Span) -> Dict[str, Any]:
     @staticmethod
     def _llmobs_tags(span: Span) -> List[str]:
         tags = [
-            "version:%s" % (config.version or ""),
-            "env:%s" % (config.env or ""),
-            "service:%s" % (span.service or ""),
+            "version:{}".format(config.version or ""),
```
mabdinur, line 385: But: We could use f strings here

tests/contrib/openai/utils.py
```diff
@@ -36,6 +36,7 @@ def _expected_llmobs_tags(error=None):
         "env:",
         "service:",
         "source:integration",
+        "ml_app:",
```
mabdinur, line 39: Should this have a default value? In the tracer we we set service names to  "unnamed_python_service"

ddtrace/settings/config.py
```diff
@@ -552,6 +552,7 @@ def __init__(self):
 
         self._llmobs_enabled = asbool(os.getenv("DD_LLMOBS_ENABLED", False))
         self._llmobs_sample_rate = float(os.getenv("DD_LLMOBS_SAMPLE_RATE", 1.0))
+        self._llmobs_ml_app = os.getenv("DD_LLMOBS_ML_APP", "")
```
mabdinur, line 555: If we don't expect customers to set this envar we should prefix it with an underscore
```suggestion
        self._llmobs_ml_app = os.getenv("_DD_LLMOBS_ML_APP", "")
```

mabdinur (approved):
Left some nuts. Feel free to resolve. Overall it looks great!

ddtrace/settings/config.py
Yun-Kim: Good point - I'll just remove this from the config for now, as users might use it but I don't want to visibly add this to our public API.

tests/contrib/openai/utils.py
Yun-Kim: Thanks for the catch!

ddtrace/llmobs/_llmobs.py
Yun-Kim: Will do these as a bigger cleanup effort in a future PR

alexbarksdale (approved):

PR #8532, opened 2024-02-27
ci(sqlite3): upgrade tests to support latest versions of sqlite3

zarirhamza:
Adds testing for internal package `sqlite3` without upgrading python versions. Note the comment further down in the PR about potential issues on running CI locally. Right now, we are only testing `sqlite3` v3.27 where latest is v3.45. Seeing as how this is a widely used integration, we need to improve testing across a range of values rather than just one

## Checklist

## Reviewer Checklist

- [x] Title is accurate


zarirhamza:
`sqlite3` is an internal package meaning that its version cannot be changed without changing python itself. The testrunner runs python 3.10 which correspondingly has sqlite3 v3.27.

In order to test an actual range of values, we use `pysqlite3-binary` which are precompiled binaries of different sqlite3 versions and replace the `sqlite3` modules to use them in test when available.

The issue arises that these binaries do not build/install/compile on Mac. This means that both running tests via `riot run` as well as creating riotfiles via `scripts/compile-and-prune-test-requirements` will not work unless run in the testrunner or another linux env. Note that these are only run when working on this specific integration and it can be noted for whoever works on the future to make necessary adjustments.

mabdinur (approved):
We will need to revisit the issues with running pysqlite on macos. Ideally all tests should run locally without issue.

PR #8572, opened 2024-03-01
feat(tracing): make tracer._sampler public

ZStriker19:
In anticipation of lazy sampling, users may create egress points that we cannot account for, and therefore miss sampling the span on. For example, an integration that where a user is manually injecting trace headers into the outgoing request will miss calling tracer.sampler.sample() which could lead to broken traces. To avoid that we'll make sure to advise users when to call sampler in the lazy sampling PR: https://github.com/DataDog/dd-trace-py/pull/8308

## Checklist

## Reviewer Checklist

- [x] Title is accurate


ZStriker19:
I did think of that but didn't try the implementation. I'll give it a go to investigate if it's better.

releasenotes/notes/sampler_public-0570e2ce18bc604b.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    tracing: Makes ``tracer.sampler`` public, now named ``tracer.sampler``. This should not be called
```
ZStriker19, line 4: this needs to be updated if we decide to pull the sampler out of the tracer.

emmettbutler:
@ZStriker19 I think that for this change to remain small enough to merge, the Tracer should continue to hold a Sampler instance. My suggestion about removing that attribute was meant more as a long-term possibility than a suggestion for right now.

The point I wanted and still want to get across is that exposing sampling functionality via `tracer.sampler` rather than `tracer.do_sampling` limits our ability to change the underlying implementation in the future. Of course, deprecations are cheap, so this isn't a dealbreaker. Instead it's just a suggestion about a readily available way to futureproof the new public API surface you're considering here.

emmettbutler (approved):
I think an interface to the specific functionality you want to make public would be more futureproof than having the interface be unrestricted access to the `.sampler` attribute as you've written here. With that said, if you want to go ahead with this approach I'm ok with it.

PR #8574, opened 2024-03-01
fix(tracing): remove allow_false from samplers

mabdinur:
- Removes allow_false parameter from Sampler.sample(....), this parameters adds unnecessary complexity to samplers.
- Removes usages of the `_sample_rate` tag (from snapshot tests and samplers). This tag is deprecated. From internal docs:
   - "[_sample_rate should be set] when the tracer drops entirely (not sent to the Agent) a configured rate of local traces, independently of any other mechanism. It ignores the sampling_priority attribute. This was the very first implementation of sampling, and was called client sampling. This has been replaced with priority sampling."
   - "The Agent upscales all APM Metrics when counting spans from a tracer applying random client sampling"
   - "This mechanism is deprecated and should get removed from all tracers."
- Updates tracer tests to align with the expected sampling behavior

NOTE: This could be seen as a breaking change but I think it is a necessary one. `allow_false` should NEVER be used. It defeats the purpose of sampling and introduces unexpected side effects.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


tests/tracer/test_sampler.py
```diff
@@ -124,19 +120,17 @@ def test_sample_rate_deviation(self):
     def test_deterministic_behavior(self):
         """Test that for a given trace ID, the result is always the same"""
         tracer = DummyTracer()
-
-        tracer._sampler = RateSampler(0.5)
+        tracer._sampler = DatadogSampler(default_sample_rate=0.5, rate_limit=-1)
```
mabdinur, line 123: @ZStriker19 RateSampler doesn't actually sample spans (it does not set the sampling priority it just returns True/False). So I had to use the DatadogSampler instead. Should we fix this behavior?

tests/tracer/test_sampler.py
```diff
@@ -124,19 +120,17 @@ def test_sample_rate_deviation(self):
...
 
             samples = tracer.pop()
             assert (
-                len(samples) <= 1
-            ), "evaluating sampling rules against a span should result in either dropping or not dropping it"
-            # sample rate metric is only set on sampled spans
-            sampled = 1 == len([sample for sample in samples if sample.get_metric(SAMPLE_RATE_METRIC_KEY) is not None])
+                len(samples) == 1
+            ), f"DummyTracer should always store a single span, regardless of sampling decision {samples}"
+            sampled = 1 == len([sample for sample in samples if sample.context.sampling_priority > 0])
```
mabdinur, line 133:
```suggestion
            sampled = len(samples) == 1 and samples[0].context.sampling_priority > 0
```

tests/tracer/test_sampler.py
ZStriker19: We could just check to make sure the value returned by the RateSampler is the same for each span it samples with the same id.

mabdinur, line 123:
```suggestion
        # Since RateSampler does not set the sampling priority on a span, we will use a DatadogSampler
        tracer._sampler = DatadogSampler(default_sample_rate=0.5, rate_limit=-1)
```

tests/tracer/test_sampler.py
```diff
@@ -94,7 +93,7 @@ def test_sample_rate_deviation(self):
         for sample_rate in [0.1, 0.25, 0.5, 1]:
             tracer = DummyTracer()
 
-            tracer._sampler = RateSampler(sample_rate)
+            tracer._sampler = DatadogSampler(default_sample_rate=sample_rate, rate_limit=-1)
```
mabdinur, line 96:
```suggestion
            # Since RateSampler does not set the sampling priority on a span, we will use a DatadogSampler
            # with rate limiting disabled.
            tracer._sampler = DatadogSampler(default_sample_rate=sample_rate, rate_limit=-1)
```

ddtrace/constants.py
```diff
@@ -1,3 +1,5 @@
+# This key allows the agent to ignore the sampling_priority set on a span. This
```
ZStriker19, line 1: When you say sampling_priority you mean `_sampling_priority_v1` which is set on the context? Could we clarify the comment?

ZStriker19 (approved):

ddtrace/constants.py
```diff
@@ -1,3 +1,5 @@
...
+# legacy behavior is deprecated and SAMPLE_RATE_METRIC_KEY will be removed.
```
mabdinur, line 2:
```suggestion
# TODO: Deprecate and remove the SAMPLE_RATE_METRIC_KEY constant.
# This key enables legacy trace sampling support in the Datadog agent.
```

emmettbutler:
@mabdinur This is great. I'm glad you discovered that `allow_false` had become unnecessary. It was the main wrinkle in the sampling code and removing it does a huge amount for that code's clarity.

PR #8613, opened 2024-03-05
fix(starlette): fixes incorrect handling of path/root_path in older starlette versions

zarirhamza:
`root_path` now correctly contains the duplicated mounted path found in `path` for `starlette` > v0.34 as per ASGI standards

Previously, there was no duplication when mounting applications on `starlette`. This required us to set tags using both `root_path` and `path`. However, this was later fixed in later versions of starlette leading to duplication.

Logic is now implemented such that the ASGI middleware follows the specifications and only takes the `path` into consideration for tracing and older versions of `starlette` have their spans edited within the `starlette `integration

Changes are also made to the testing snapshots to account for the TestClient change after v0.20 in `starlette`

## Checklist

## Reviewer Checklist

- [ ] Title is accurate


ddtrace/contrib/starlette/patch.py
```diff
@@ -163,9 +168,53 @@ def traced_handler(wrapped, instance, args, kwargs):
     if core.get_item(HTTP_REQUEST_BLOCKED):
         raise trace_utils.InterruptException("starlette")
 
+    # If we are on a version of starlette that is less than 0.34, we need to fix the span url
+    # since starlette sets the incorrect values for path/root_path
```
mabdinur, line 172: Can we link to the starlette issue here. This will make this problem easier to track
```suggestion
    # If we are on a version of starlette that is less than 0.34, we need to fix the span url
    # since starlette sets the incorrect values for path/root_path
```

zarirhamza: The original issue is [here](https://github.com/encode/starlette/issues/1336)

PR #8629, opened 2024-03-06
fix(asm): fix cpu leak

christophe-papazian:
This PR fixes a cpu leak in ASM instrumentation, that would add additional overhead to all requests if ASM, API Security and Telemetry were enabled on Flask instrumentation.

Related to incident 25768.

Add one regression unit test for all ASM supported frameworks.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


ddtrace/appsec/_asm_request_context.py
```diff
@@ -151,7 +151,7 @@ def finalise(self):
             callbacks = GLOBAL_CALLBACKS.get(_CONTEXT_CALL, []) if env.must_call_globals else []
             env.must_call_globals = False
             if env is not None and env.callbacks is not None and env.callbacks.get(_CONTEXT_CALL):
```
P403n1x87, line 153: `env.callbacks.get(_CONTEXT_CALL)` is being called twice. Perhaps the first call can be assigned and re-used?

christophe-papazian: it's a get on a native dict, so very fast. I prefer to not introduce performance improvement in this PR, I opened another PR for that to be merged later: https://github.com/DataDog/dd-trace-py/pull/8617
This is only bug fix.

emmettbutler (approved):
This is up there with avoiding the use of mutable keyword arguments on the list of Python toe-stubbers
https://stackoverflow.com/a/2347423/735204

ddtrace/appsec/_asm_request_context.py
```diff
@@ -151,7 +151,7 @@ def finalise(self):
...
-                callbacks += env.callbacks.get(_CONTEXT_CALL)
+                callbacks = callbacks + env.callbacks.get(_CONTEXT_CALL)
```
Julio-Guerra, line 154: It's worth mentioning for future devs going through this code why `+=` isn't equivalent.

Julio-Guerra: Quoting [this SO answer](https://stackoverflow.com/questions/2347265/why-does-behave-unexpectedly-on-lists/2347423#2347423):
> For types that support both __iadd__ and __add__ you therefore have to be careful which one you use. a += b will call __iadd__ and mutate a, whereas a = a + b will create a new object and assign it to a. They are not the same operation!

Julio-Guerra: I also wonder if static analyzer could help warn here about an unintended side effect.

gnufede (approved):

Issue #8651, opened 2024-03-08
ddtrace-run 2.7.2 raise free(): invalid pointer


wadinj:
### Which version of dd-trace-py are you using?
2.7.2
### Which version of pip are you using?
24.0

<details>
  <summary>`pip freeze`</summary>
colorama==0.4.6
comm==0.2.0
ConfigArgParse==1.7
contourpy==1.2.0
countryinfo==0.1.2
coverage==7.3.4
crashtest==0.4.1
cryptography==42.0.5
cycler==0.12.1
dataclasses-json==0.6.4
ddsketch==2.0.4
ddtrace==2.7.1
debugpy==1.8.0
decorator==5.1.1
deepdiff==6.7.1
defusedxml==0.7.1
Deprecated==1.2.14
dill==0.3.7
distlib==0.3.8
dparse==0.6.4b0
duckdb==0.10.0
dulwich==0.21.7
envier==0.5.0
et-xmlfile==1.1.0
execnet==2.0.2
executing==2.0.1
fakeredis==2.21.1
fastapi==0.110.0
fastapi-cache2==0.2.1
fastjsonschema==2.19.0
filelock==3.13.1
Flask==3.0.0
Flask-Cors==4.0.0
Flask-Login==0.6.3
fonttools==4.47.0
fqdn==1.5.1
freezegun==1.4.0
frictionless==4.40.8
frozenlist==1.4.1
fsspec==2024.2.0
gevent==23.9.1
geventhttpclient==2.0.11
google_trans_new_tmp==1.2.2
greenlet==3.0.3
h11==0.14.0
httpcore==1.0.2
httpx==0.27.0
identify==2.5.33
idna==3.6
importlib-metadata==6.11.0
iniconfig==2.0.0
installer==0.7.0
ipykernel==6.29.3
ipython==8.22.1
isodate==0.6.1
isoduration==20.11.0
isort==5.13.2
itsdangerous==2.1.2
ixbrl-viewer==1.4.18
jaraco.classes==3.3.1
jedi==0.19.1
jeepney==0.8.0
Jinja2==3.1.3
jmespath==1.0.1
json5==0.9.14
jsonpointer==2.4
jsonschema==4.20.0
jsonschema-specifications==2023.11.2
jupyter-events==0.9.0
jupyter-lsp==2.2.2
jupyter_client==8.6.0
jupyter_core==5.5.1
jupyter_server==2.12.1
jupyter_server_terminals==0.5.0
[...]

PR #8675, opened 2024-03-12
ci: display backtrace from core dumps

P403n1x87:
We make it easier to debug segmentation faults and other crashes in CI that can generate a core dump. We rely on gdb to display the complete backtraces available from any generated core dumps during test runs in Circle CI.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


sanchda:
Quick question:  does `ulimit -c` work across our CI runners?  It can be tricky to guarantee the generation of corefiles (indeed, that they even land in a consistent place).  If you feel pretty good about this, please ignore the next part--I'm trying to be helpful, but I am ignorant.


I have an alternative workflow I've been using lately with some success.  It has several disadvantages over corefile analysis--it doesn't use a debugger, so you can't interactively inspect anything.  I also think its backtrace capabilities are a little less powerful than gdb.  It does have a few nice features though

* Prints the backtrace, mappings, and registers to stderr
* Can be instrumented to trigger not just no SIGSEGV, but also SIGABRT (nice for catching those nasty libc errors like double-free)
* Is straightforward to translate to any glibc-based system (for customer incidences)
* Prints immediately, without having to wait for other forks or children to terminate (usually not a problem in CI)

I describe this workflow in a [recent issue](https://github.com/DataDog/dd-trace-py/issues/8651#issuecomment-1991788957).  It's annoying because the glibc maintainers have stopped providing this utility, but I have a suggestion for how to get it from old .deb archives.  No pressure to use it or anything--I love GDB and I think the proposal will work great barring anything unlucky with how the CI runners work.  Just throwing out an idea for consideration.

.circleci/config.templ.yml
```diff
@@ -141,12 +141,22 @@ commands:
             - start_docker_services:
                 env: SNAPSHOT_CI=1
                 services: testagent << parameters.docker_services >>
+            - run: ulimit -c unlimited
```
sanchda, line 144: I left a top-level comment in the conversation, but is this generally respected?  There are lots of ways to interfere with corefile generation, and modern Linux distros basically try them all...

.circleci/config.templ.yml
```diff
@@ -141,12 +141,22 @@ commands:
...
             - run:
                 environment:
                   DD_TRACE_AGENT_URL: << parameters.trace_agent_url >>
                   RIOT_RUN_RECOMPILE_REQS: "<< pipeline.parameters.riot_run_latest >>"
                 command: |
                   ./scripts/run-test-suite '<<parameters.pattern>>' <<pipeline.parameters.coverage>> 1
+            - run:
+                command: |
+                  mkdir -p /tmp/core_dumps
+                  cp core.* /tmp/core_dumps
```
sanchda, line 154: In general, is it possible for any of the test processes to modify the cwd so as to move the corefiles elsewhere?  If not, please ignore.  If so, then we probably have limited ability to modify the corepattern etc, but maybe we could do a broader search with `find`

.circleci/config.templ.yml
```diff
@@ -141,12 +141,22 @@ commands:
...
+                  ./scripts/bt
+                when: on_fail
+            - store_artifacts:
+                name: "Store core dumps"
+                path: /tmp/core_dumps
```
sanchda, line 159: I like this, but we'd need some way to also extract the precise versions of system and artifact .so files (or at least build things in a reliable way).

Please ignore if the intent isn't to extract corefiles for further analysis.

scripts/bt
```diff
@@ -0,0 +1,12 @@
...
+
+set -eo pipefail
+
+for core in core.*
+do
+    echo "========  Core dump: ${core}  ========"
+    binary=$(echo ${core} | sed -e 's/core\.[0-9]*\.//' -e 's/!/\//g')
+    echo "Binary: ${binary}"
+    echo "Backtrace:"
+    gdb -q -batch -ex "bt full" -ex q ${binary} ${core}
```
sanchda, line 11: Sounds good to me!

sanchda (approved):
Left some comments, but looks good.  I can take another look when it leaves draft, but I think this is fairly important.

P403n1x87:
> Quick question: does `ulimit -c` work across our CI runners? It can be tricky to guarantee the generation of corefiles (indeed, that they even land in a consistent place). If you feel pretty good about this, please ignore the next part--I'm trying to be helpful, but I am ignorant.

Yep I have already tested this in #7659 (in fact this is where the code comes from). More on this from the [Circle CI docs](https://circleci.com/docs/artifacts/#uploading-core-files).

.circleci/config.templ.yml
P403n1x87: Ah this is a good point. I suppose it would be possible for this to happen. We could experiment with trying to set an absolute path just to be sure

.circleci/config.templ.yml
P403n1x87: Good point. I had this here from earlier experiments. But indeed if we want to keep these, we should collect all the binaries as well. Another complication might be the platform, meaning the analysis might have to be done on a similar architecture

majorgreys (approved):

PR #8691, opened 2024-03-14
feat(llmobs): make DD_LLMOBS_APP_NAME required

Yun-Kim:
Previously DD_LLMOBS_APP_NAME was a strongly recommended (but not required) config for LLMObs users. However, in the case users do not specify the env var, the ML Obs product will not be able to differentiate between spans from potentially different ML applications and show clustering/prompt data all together under the previous default `unnamed-ml-app` name.

By making this a required configuration, we will avoid this behavior. However we need to ensure that this is documented clearly for user setup docs.

## Checklist

## Reviewer Checklist

- [x] Title is accurate

### DIFF
--- a/ddtrace/llmobs/_llmobs.py
+++ b/ddtrace/llmobs/_llmobs.py
@@ -71,6 +71,9 @@ class LLMObs(Service):
         if not config._dd_api_key:
             cls.enabled = False
             raise ValueError("DD_API_KEY is required for sending LLMObs data")
+        if not config._llmobs_ml_app_name:
+            cls.enabled = False
+            raise ValueError("DD_LLMOBS_APP_NAME is required for sending LLMObs data")
 
         cls._instance = cls(tracer=tracer)
         cls.enabled = True
@@ -388,7 +391,7 @@ class LLMObsTraceProcessor(TraceProcessor):
             "env:{}".format(config.env or ""),
             "service:{}".format(span.service or ""),
             "source:integration",
-            "ml_app:{}".format(self._get_ml_app_name(span)),
+            "ml_app:{}".format(config._llmobs_ml_app or "unnamed-ml-app"),
             "ddtrace.version:{}".format(ddtrace.__version__),
             "error:%d" % span.error,
         ]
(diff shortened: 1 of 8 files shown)


ddtrace/settings/config.py
```diff
@@ -552,6 +552,7 @@ def __init__(self):
 
         self._llmobs_enabled = asbool(os.getenv("DD_LLMOBS_ENABLED", False))
         self._llmobs_sample_rate = float(os.getenv("DD_LLMOBS_SAMPLE_RATE", 1.0))
+        self._llmobs_ml_app = os.getenv("DD_LLMOBS_APP_NAME")
```
mabdinur, line 555: Can the llmobs app name default to DD_SERVICE?

ddtrace/llmobs/_llmobs.py
```diff
@@ -71,6 +71,9 @@ def enable(cls, tracer=None):
         if not config._dd_api_key:
             cls.enabled = False
             raise ValueError("DD_API_KEY is required for sending LLMObs data")
+        if not config._llmobs_ml_app_name:
+            cls.enabled = False
+            raise ValueError("DD_LLMOBS_APP_NAME is required for sending LLMObs data")
```
mabdinur, line 76: Can we make this a little more actionable? Ex:
```suggestion
            raise ValueError("DD_LLMOBS_APP_NAME is required for sending LLMObs data. Ensure this configuration is set before running your application.")
```

mabdinur (approved):
👍 💯 🥮

ddtrace/settings/config.py
Yun-Kim: We are currently doing this right now, but we want ml-app to be a required configuration (a service may make calls to different ml apps). Defaulting it to the service name is problematic because the ML Obs product is ml-app centric (i.e. data is shown based on apps rather than individual services)

lievan (approved):

ddtrace/settings/config.py
lievan, line 555: Maybe we should give users the ability to set ml-app in code as well for specific spans. For example if they have 3 different agent spans running within one environment, and they want each one to be a different ml app just so they can get clustering analysis at each of those levels

lievan: Not saying that they should do this, but it would be interesting to see if they _do_ do it if we give them that option

Yun-Kim: Screenshot 2024-03-15 at 8 43 17 AM

lievan: Oops didn't see that, sorry!

emmettbutler (approved):
This looks to me like a change that would benefit from a release note. If there's a good reason not to include one, that's fine.

PR #8695, opened 2024-03-15
feat(llmobs): submit span events for the langchain integration

sabrenner:
## Summary
This PR makes the LangChain integration submit LLMObs Span Events **for LLM and chat model calls** if LLMObs is enabled, and if the LLMObs span is sampled. It accomplishes this by:

- Setting the span type `SpanTypes.LLM` on `langchain` APM spans so they are properly processed by the trace processor service in the `LLMObs` service
- Tagging each (in this case, llm or chat model) span with additional `_ml_obs.*` tags, which get popped from the trace when submitting the data they represent to LLMObs intake through the LLMObs writer

This PR is the first PR of three separate PRs for fully supporting sending span events from the LangChain integration. The following PRs are a WIP and will be opened shortly:

1. Submitting span events from chains (might require a bit more work)
2. Supporting streaming for the LangChain integration, as subsequently, making sure those submit span events too (the latter part might not require as much work as the former)

### For Reviewers
Most of the files touched are snapshot files to account for the span type being changed. Feel free to ignore these files (everything else is relevant for review).

Additionally, no release notes/changelog, as this is an internal change for submitting span events to LLMObs intake.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


tests/contrib/langchain/test_langchain.py
```diff
@@ -1256,3 +1258,90 @@ def test_vectorstore_logs_error(langchain, ddtrace_config_langchain, mock_logs,
             "documents": [],
         }
     )
+
+
+@pytest.mark.parametrize("ddtrace_global_config", [dict(_llmobs_enabled=True, _llmobs_sample_rate=1.0)])
```
sabrenner, line 1263: will update with the `_llmobs_ml_app` option once #8691 is merged, or in one of the follow up PRs for langchain span events

tests/contrib/langchain/test_langchain.py
```diff
@@ -1256,3 +1258,93 @@ def test_vectorstore_logs_error(langchain, ddtrace_config_langchain, mock_logs,
...
+
+        with request_vcr.use_cassette(cassette_name):
+            generate_trace("Can you explain what an LLM chain is?")
+        span = mock_tracer.pop_traces()[0][0]
+
+        expected_llmobs_writer_calls = cls._expected_llmobs_calls(span, input_role, output_role)
+        assert mock_llmobs_writer.enqueue.call_count == 1
+        mock_llmobs_writer.assert_has_calls(expected_llmobs_writer_calls)
+
+    def test_llmobs_llm(self, langchain, mock_llmobs_writer, mock_tracer, request_vcr):
```
Yun-Kim, line 1328: Could we add some additional test coverage for more model providers? I.e. cohere, ai21) similar to how we do it in `test_bedrock.py --> TestLLMObsBedrock`? Thanks!

tests/contrib/langchain/test_langchain.py
```diff
@@ -1256,3 +1258,93 @@ def test_vectorstore_logs_error(langchain, ddtrace_config_langchain, mock_logs,
...
+                        "max_tokens": int(span.get_tag("langchain.request.openai.parameters.max_tokens")),
+                    },
+                    token_metrics={
+                        "prompt_tokens": span.get_metric("langchain.tokens.prompt_tokens"),
+                        "completion_tokens": span.get_metric("langchain.tokens.completion_tokens"),
+                        "total_tokens": span.get_metric("langchain.tokens.total_tokens"),
+                    },
+                    tags={
+                        "ml_app": "unnamed-ml-app",
+                        "service": "",
```
Yun-Kim, line 1295: Don't need to specify `service/ml_app` here, the default values are already `""/"unnamed-ml-app"` respectively

tests/contrib/langchain/test_langchain.py
Yun-Kim: Let's update here, thanks!

ddtrace/llmobs/_integrations/langchain.py
```diff
@@ -20,6 +28,114 @@
...
+
+    def _llmobs_set_metrics_tags(
+        self,
+        span: Span,
+    ) -> None:
+        metrics = {}
+        prompt_tokens = span.get_metric("langchain.tokens.prompt_tokens")
+        completion_tokens = span.get_metric("langchain.tokens.completion_tokens")
+        # grab off tag directly to minimize differences in reported total tokens and just summing,
+        # in case of discrepancies
```
Yun-Kim, line 127: no need to comment here, your code is self-explanatory here

ddtrace/llmobs/_integrations/langchain.py
```diff
@@ -20,6 +28,114 @@
...
+
+        # output messages
+        output_messages = [{"content": ""}]
+        if not err:
+            output_messages = []
+            for message_set in chat_completions.generations:
+                for chat_completion in message_set:
+                    output_messages.append(
+                        {
+                            "content": self.trunc(chat_completion.text),
```
Yun-Kim, line 113: No truncation here as well, also same concerns about type checking and checking for the `type` field

ddtrace/llmobs/_integrations/langchain.py
```diff
@@ -20,6 +28,114 @@
...
+            for message in message_set:
+                input_messages.append(
+                    {
+                        "content": self.trunc(message.content),
+                        "role": str(message.type),
+                    }
+                )
+        span.set_tag_str(INPUT_MESSAGES, json.dumps(input_messages))
+
+        # output messages
```
Yun-Kim, line 105: Same here

ddtrace/llmobs/_integrations/langchain.py
```diff
@@ -20,6 +28,114 @@
...
+        err: bool = False,
+    ) -> None:
+        # input messages
+        input_messages = []
+        for message_set in chat_messages:
+            for message in message_set:
+                input_messages.append(
+                    {
+                        "content": self.trunc(message.content),
+                        "role": str(message.type),
```
Yun-Kim, line 100: Just curious, do messages have a type field? (If it's a recent API addition, we should also keep supporting older versions which may not have that field)

ddtrace/llmobs/_integrations/langchain.py
```diff
@@ -20,6 +28,114 @@
...
+        chat_completions: Any,
+        err: bool = False,
+    ) -> None:
+        # input messages
+        input_messages = []
+        for message_set in chat_messages:
+            for message in message_set:
+                input_messages.append(
+                    {
+                        "content": self.trunc(message.content),
```
Yun-Kim, line 99: Let's remove the truncation here as well.
Also a heads up that with #8681, we added type checking for different types of messages (dictionary vs langchain objects) so we'll need to add type checking here too

ddtrace/llmobs/_integrations/langchain.py
```diff
@@ -20,6 +28,114 @@
...
+        span.set_tag_str(OUTPUT_MESSAGES, json.dumps(message_content))
+
+    def _llmobs_set_meta_tags_from_chat_model(
+        self,
+        span: Span,
+        chat_messages: List[List[Any]],
+        chat_completions: Any,
+        err: bool = False,
+    ) -> None:
+        # input messages
```
Yun-Kim, line 93: Same here

ddtrace/llmobs/_integrations/langchain.py
```diff
@@ -20,6 +28,114 @@
...
+        if isinstance(prompts, str):
+            prompts = [prompts]
+        span.set_tag_str(INPUT_MESSAGES, json.dumps([{"content": str(prompt)} for prompt in prompts]))
+
+        # output messages
+        message_content = [{"content": ""}]
+        if not err:
+            message_content = [{"content": self.trunc(completion[0].text)} for completion in completions.generations]
```
Yun-Kim, line 83: For regular APM langchain/openai integrations, truncation is required for prompt/responses for storage/privacy limits, but for LLMObs we need the entire prompt/response so let's remove the truncation here

Also, need to double check why we do `completion[0].text` instead of `completion` (for my own sanity)

ddtrace/llmobs/_integrations/langchain.py
```diff
@@ -20,6 +28,114 @@
...
+        prompts: List[Any],
+        completions: Any,
+        err: bool = False,
+    ) -> None:
+        # input messages
+        if isinstance(prompts, str):
+            prompts = [prompts]
+        span.set_tag_str(INPUT_MESSAGES, json.dumps([{"content": str(prompt)} for prompt in prompts]))
+
+        # output messages
```
Yun-Kim, line 80: Same here

ddtrace/llmobs/_integrations/langchain.py
```diff
@@ -20,6 +28,114 @@
...
+        self._llmobs_set_metrics_tags(span)
+
+    def _llmobs_set_meta_tags_from_llm(
+        self,
+        span: Span,
+        prompts: List[Any],
+        completions: Any,
+        err: bool = False,
+    ) -> None:
+        # input messages
```
Yun-Kim, line 75: I think this code block is self-explanatory, no need for commenting here

ddtrace/llmobs/_integrations/langchain.py
```diff
@@ -20,6 +28,114 @@
...
+                input_parameters["max_tokens"] = int(max_tokens)
+            if input_parameters:
+                span.set_tag_str(INPUT_PARAMETERS, json.dumps(input_parameters))
+
+        if operation == "llm":
+            self._llmobs_set_meta_tags_from_llm(span, inputs, response, error)
+        elif operation == "chat_model":
+            self._llmobs_set_meta_tags_from_chat_model(span, inputs, response, error)
+        elif operation == "chain":
+            # TODO to be added as a follow-up PR
```
Yun-Kim, line 64: For cleanliness sake, let's avoid leaving TODOs in code, and keep track of this task on our JIRA instead

Yun-Kim (commented):
Looks mostly great! Just a couple small questions/concerns

ddtrace/contrib/langchain/patch.py
```diff
@@ -158,6 +160,7 @@ def traced_llm_generate(langchain, pin, func, instance, args, kwargs):
         model=model,
         api_key=_extract_api_key(instance),
     )
+    span.span_type = SpanTypes.LLM
```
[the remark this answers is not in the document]
sabrenner: ooh good idea, will do this

ddtrace/llmobs/_integrations/langchain.py
sabrenner: yeah from what I've seen, they do (for an 'AIMessage' --> 'ai', for 'HumanMessage' --> 'human'). didn't think about backwards compatibility, so I will take a look at how this translates to older versions

ddtrace/llmobs/_integrations/langchain.py
sabrenner: ah yeah that totally makes sense, will change here (and the other places).

and yeah, will do!

sabrenner: so it looks like completions is of type `List[List[Generation]]`, even on older versions of LangChain (debugging `langchain==0.0.192`, the older version that we test with, it is also that type). I copypasta'd this from `patch`, but I think we do it this way since it's a nested list of `Generation` objects, so we have to access the `.text` field. I suppose we take the first one for simplicity in matching 1:1 for prompts/responses for LLMs, but correct me if I'm wrong!

ddtrace/llmobs/_integrations/langchain.py
```diff
@@ -44,43 +44,52 @@ def llmobs_set_tags(
...
             self._llmobs_set_meta_tags_from_chat_model(span, inputs, response, error)
         elif operation == "chain":
-            # TODO to be added as a follow-up PR
             pass
         self._llmobs_set_metrics_tags(span)
 
+    def _llmobs_set_input_parameters(
+        self,
+        span: Span,
+        model_provider: Optional[str],
```
Yun-Kim, line 60: Should specify `model_provider: Optional[str] = None` for an optional arg

ddtrace/llmobs/_integrations/langchain.py
```diff
@@ -44,43 +44,52 @@ def llmobs_set_tags(
...
+                or 0.0  # huggingface
+            )
+            max_tokens = (
+                span.get_tag(f"langchain.request.{model_provider}.parameters.max_tokens")
+                or span.get_tag(f"langchain.request.{model_provider}.parameters.maxTokens")
+                or span.get_tag(f"langchain.request.{model_provider}.parameters.model_kwargs.max_tokens")  # ai21
+                or 0  # huggingface
+            )
+            input_parameters["temperature"] = float(temperature)
+            input_parameters["max_tokens"] = int(max_tokens)
```
Yun-Kim, line 76: Can we only specify the temp/max_tokens if it's a provided value? i.e. we don't want to specify default max_tokens = 0 if it's not set as a parameter

ddtrace/llmobs/_integrations/langchain.py
sabrenner: jk, it actually looks like it could be a bit more descriptive: https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/callbacks/promptlayer_callback.py#L142-L155

these lines seem to correspond to more traditional analysis for role ("user", "system", "assistant", ...). Additionally, for `ChatMessage` prompts, the role is specified on the object as "role". changes pushed here if you want to check it out

ddtrace/llmobs/_integrations/langchain.py
Yun-Kim: Good point, we'll have to update the bedrock integration as well (not in this PR)

ddtrace/llmobs/_integrations/base.py
```diff
@@ -99,7 +100,7 @@ def _set_base_span_tags(self, span: Span, **kwargs) -> None:
         """Set default LLM span attributes when possible."""
         pass
 
-    def trace(self, pin: Pin, operation_id: str, **kwargs: Dict[str, Any]) -> Span:
+    def trace(self, pin: Pin, operation_id: str, span_type: Optional[str] = None, **kwargs: Dict[str, Any]) -> Span:
```
Yun-Kim, line 103:
```suggestion
    def trace(self, pin: Pin, operation_id: str, submit_to_llmobs: bool = False, **kwargs: Dict[str, Any]) -> Span:
```
Sorry for last suggestion, I think it makes the most sense if we just mark this as a boolean flag instead of having to specify the LLMObs span type, since we don't really care about other span types

ddtrace/llmobs/_integrations/base.py
```diff
@@ -113,6 +114,8 @@ def trace(self, pin: Pin, operation_id: str, **kwargs: Dict[str, Any]) -> Span:
         # Enable trace metrics for these spans so users can see per-service openai usage in APM.
         span.set_tag(SPAN_MEASURED_KEY)
         self._set_base_span_tags(span, **kwargs)
+        if span_type and span_type == SpanTypes.LLM:
+            span.span_type = span_type
```
Yun-Kim, line 118:
```suggestion
        if submit_to_llmobs:
            span.span_type = SpanTypes.LLM
```

Yun-Kim (approved):
One last suggestion but otherwise LGTM, great work!

ddtrace/contrib/langchain/patch.py
```diff
@@ -348,6 +368,14 @@ def traced_chat_model_generate(langchain, pin, func, instance, args, kwargs):
         integration.metric(span, "incr", "request.error", 1)
         raise
     finally:
+        if integration.is_pc_sampled_llmobs(span):
```
mabdinur, line 371: Side note: I found a bug in `is_pc_sampled_llmobs`. This method does not always return the correct sampling decision. `BaseLLMIntegration._llmobs_pc_sampler.sample(span)` only returns a boolean, it does not make a sampling decision. Sampling decisions are recorded by setting the `_sampling_priority_v1` tag on a span. Right now all LLM spans are being sampled by the global tracer and this tracer could be using a different sampler (ex: Rate by Service Sampler).

I am not sure if this is a bug in LLM integrations or the RateSampler.

@ZStriker19 should the RateSampler make a sampling decision?

cc: @Yun-Kim

tests/contrib/langchain/conftest.py
```diff
@@ -72,6 +72,16 @@ def mock_tracer(langchain, mock_logs, mock_metrics):
...
 
 
+@pytest.fixture
+def mock_llmobs_writer():
+    patcher = mock.patch("ddtrace.llmobs._llmobs.LLMObsWriter")
+    LLMObsWriterMock = patcher.start()
+    m = mock.MagicMock()
+    LLMObsWriterMock.return_value = m
+    yield m
+    patcher.stop()
```
mabdinur, line 82: We should wrap this in a `try-finally`. We should always stop the patcher even if the test fails unexpectedly.

mabdinur (approved):
The core change looks good. Could we update the PR description to include what span events are being sent.

I also left a comment about a sampling bug this is unrelated to this change.

ddtrace/contrib/langchain/patch.py
ZStriker19: It can make a decision, but it doesn't assign tags. Up for discussion if it should, but we'd have to rework some other code.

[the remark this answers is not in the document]
Yun-Kim: We should rename `is_pc_sampled_llmobs()` to `is_sampled_llmobs()`, it's a bit misleading name right now since the purpose of it is to determine whether or not we submit the span to LLMObs. We do this by setting temporary tags on the span and then we'll extract them onto the LLMObs span event, so technically setting tags is a mechanism for us to submit an LLMObs span event.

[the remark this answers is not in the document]
Yun-Kim: For reference, we do this already in the [OpenAI](https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/contrib/openai/_endpoint_hooks.py#L230) and [Bedrock](https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/contrib/botocore/services/bedrock.py#L109) integrations as well.

Yun-Kim: For what it's worth, `is_pc_sampled_llmobs()` isn't meant to interfere in any sampling decisions for APM spans submitted to the trace agent, it's designed for determining whether or not to create and submit an LLMObs span event from the corresponding APM span. It should be independent from APM sampling in general (we are planning on becoming less dependent on the RateSampler for this)

Yun-Kim: Will resolve this for now and we can have this as an offline discussion.

PR #8711, opened 2024-03-19
chore(ci_visibility): experimental CI Visibility manual testing API

romainkomorndatadog:
Introducing an API that will eventually supplant a lot of the code that currently lives in the `pytest` and `unittest` integrations.

It is currently in an incomplete state to minimize the size of the introductory change and allow for smaller iterative updates. Feature parity with existing integrations is a non-goal. Some of the features intended to be supported by some of the flags introduced by the API are also unimplemented.

### `ddtrace.ext.ci_visibility.api`

This is the user-facing API that contains the various methods, structures (eg: dataclasses, named tuples, etc) to be used when interacting with the CIVisibility service. It is all in one file to give a single-point-of-reference, rather than having to look across multiple modules for information.

It is meant to be used in a stateless way (without users needing to keep track of objects representing test items). The API emits events (eg: "session start", "test discovered" , "test started", "test finished") on a best-effort basis. Classes and classmethods are only used for namespacing purposes.

### `ddtrace.internal.ci_visibility.api`

Introduces classes to represent CI Visibility items, and stores them on the CIVisibility singleton instance's `_session_data` attribute. The code allows for concurrent sessions, though that is not an expected or supported behavior.

To support non-linear progression through modules and suites, spans are started with their parent spans manually specified, and without the core execution contexts API (subject to change).

### Other changes

Several setter/getter classmethods are introduced in the `CIVsibility` class to respect the use of `_instance` as a private attribute.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


ddtrace/ext/ci_visibility/api.py
```diff
@@ -0,0 +1,435 @@
...
+
+log = get_logger(__name__)
+
+
+class CITestStatus(Enum):
+    PASS = TestStatus.PASS.value
+    FAIL = TestStatus.FAIL.value
+    SKIP = TestStatus.SKIP.value
+    XFAIL = TestStatus.XFAIL.value
+    XPASS = TestStatus.XPASS.value
```
romainkomorndatadog, line 46: This is definitely goofy and will need cleaning up at some point (the `internal` pieces will likely import this instead of the other way around).

anmarchenko (approved):
Given I don't understand python well, it looks correct and clean

ddtrace/ext/ci_visibility/api.py
```diff
@@ -0,0 +1,435 @@
...
+
+class CITest(_CIVisibilityAPIBase):
+    class DiscoverArgs(NamedTuple):
+        test_id: CITestId
+        codeowners: Optional[List[str]] = None
+        source_file_info: Optional[CISourceFileInfo] = None
+
+    @staticmethod
+    @_catch_and_log_exceptions
+    def discover(
```
anmarchenko, line 336: just out of curiosity: at what stage does attachment of test to test suite happens?

ddtrace/ext/ci_visibility/api.py
```diff
@@ -0,0 +1,435 @@
...
+        item_id: CITestId,
+        codeowners: Optional[List[str]] = None,
+        source_file_info: Optional[CISourceFileInfo] = None,
+        is_early_flake_detection: bool = False,
```
anmarchenko, line 340: interesting point here: do API consumers decide if early flake detection is enabled or should the library itself decide it?

ddtrace/ext/ci_visibility/api.py
romainkomorndatadog: Part of the stateless aspect is that there is actually no "attachment" from the consumer side of the API: the user of the API has to create consistent ID objects. While that does arguably open the door for bugs (eg: if the ID objects aren't consistently generated), I think it's pretty easy to get right.

"Inside" CI Visibility, the attachment is done at discovery time. I do plan on adding a "non-discovery" mode where the attachment will happen at test start time, but even in those cases, the correctness of the IDs passed in is what determines said attachment.

I preferred that over the idea of having the user have to stash objects somewhere (eg: as attributes on the testrunner's objects).

anmarchenko: interesting approach, thank you for explanation!

tests/ci_visibility/test_ci_visibility.py
```diff
@@ -102,6 +115,32 @@ def test_ci_visibility_service_enable():
             CIVisibility.disable()
 
 
+def test_ci_visibility_service_enable_without_service():
```
emmettbutler, line 118: In what sense is this "without service"?

tests/ci_visibility/api/fake_runner_all_fail.py
```diff
@@ -0,0 +1,129 @@
+"""Fake test runner where all tests pass"""
```
emmettbutler, line 1:
```suggestion
"""Fake test runner where all tests fail"""
```

tests/ci_visibility/api/fake_runner_all_fail.py
```diff
@@ -0,0 +1,129 @@
...
+    except ValueError:
+        return api.CIExcInfo(*sys.exc_info())
+
+
+def main():
+    take_over_logger_stream_handler()
+
+    api.enable_ci_visibility()
+
+    # START DISCOVERY
```
emmettbutler, line 22: This might be a good use case for a function

ddtrace/ext/ci_visibility/_ci_visibility_base.py
```diff
@@ -0,0 +1,98 @@
...
+    def set_tags(item_id: _CIVisibilityRootItemIdBase, tags: Dict[str, Any], recurse: bool = False):
+        log.debug("Setting tags for item %s: %s", item_id, tags)
+
+    @staticmethod
+    def delete_tag(item_id: _CIVisibilityRootItemIdBase, tag_name: str, recurse: bool = False):
+        log.debug("Deleting tag for item %s: %s", item_id, tag_name)
+
+    @staticmethod
+    def delete_tags(item_id: _CIVisibilityRootItemIdBase, tag_names: List[str], recurse: bool = False):
+        log.debug("Deleting tags for item %s: %s", item_id, tag_names)
```
emmettbutler, line 98: This logging might be confusing as a default implementation that doesn't actually do the thing it's logging about.

ddtrace/ext/ci_visibility/util.py
```diff
@@ -0,0 +1,26 @@
...
+
+    No uncaught errors should ever reach the integration-side, and potentially cause crashes.
+    """
+
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        # try:
+        return func(*args, **kwargs)
+
+    # except Exception:
```
emmettbutler, line 22: Seems like this whole function might be dead code?

emmettbutler (approved):

tests/ci_visibility/api/fake_runner_all_fail.py
romainkomorndatadog: Yeah. I kinda wanted to leave myself flexibility to change the tests individually. Tests is the one place where I don't mind not being DRY.

ddtrace/ext/ci_visibility/util.py
romainkomorndatadog: Ah crud. Thanks for noticing. Forgot I undid the change in [4ca35a5](https://github.com/DataDog/dd-trace-py/pull/8711/commits/4ca35a55e5851a9c6e45ed4ac11e6f7c002184cc).

ddtrace/ext/ci_visibility/_ci_visibility_base.py
romainkomorndatadog: Yeah this falls in the "not implemented yet" set of things in the API.

tests/ci_visibility/test_ci_visibility.py
romainkomorndatadog: That `CIVisibility.enable()` is called without `service='test-service'`.

gnufede (approved):

PR #8714, opened 2024-03-19
feat(llmobs): support generators in LLMObs decorators

lievan:
Support decorating generators in llm obs python sdk for generator llm use cases e.g. streaming response tokens

**notes**
1. ~When break is used while calling a decorated generator, the span released will be an error span since the tracer catches builtins.GeneratorExit and annotates the span with that error. This might be confusing for users since they'll see a red error span in the front-end~
2.  The span is only finished when the generator exits/the generator object is garbage collected
```
@tool 
def do_work():
    for i in range(100)
        yield i

f = do_work()
next(f)
next(f)
// goes on to do other stuff
// the span is still active at this point
```

## Checklist

## Reviewer Checklist

- [ ] Title is accurate


Yun-Kim (commented):
Reviewing the two notes in your description, I wasn't able to recreate the scenario where an unfinished generator would lead to the span never being finished. I tried your repro case, but led to getting a `GeneratorExit` error and the span finishing. Can you confirm that the span does never finish?

I'm also slightly concerned about the implications of your first note (getting a `GeneratorExit` error even if a user explicitly exits the unfinished generator on purpose. Technically it's not a user error and will not be a positive user experience if they see a bunch of false positive error spans. What do you think?

ddtrace/llmobs/decorators.py
```diff
@@ -9,14 +10,30 @@
...
+        span_name = name
+        if span_name is None:
+            span_name = func.__name__
+
+        @wraps(func)
+        def gen_wrapper(*args, **kwargs):
+            if not LLMObs.enabled or LLMObs._instance is None:
+                log.warning("LLMObs.llm() cannot be used while LLMObs is disabled.")
+                for ret in func(*args, **kwargs):
+                    yield ret
```
P403n1x87, line 22: If this branch is hit, then it looks like the generator will be called twice. Is this intentional? Also I think a generator wrapper needs to handle `StopIteration` `GeneratorExit`, `throw` etc...

lievan: Thanks for pointing this out!

PR #8727, opened 2024-03-20
chore(ci_visibility): introduce experimental internal coverage collector

romainkomorndatadog:
Introduces the `ModuleCodeCollector` which collects coverage and executable lines for imported modules.

The collector has two modes, one that stores executed lines on the singleton instance, and one that uses context variables and a context manager.

This also introduces changes to the `pytest` integration as well as the `CIVisibility` service's use of coverage to feature-flag using the new module collector.

The features are gated behind `_DD_USE_INTERNAL_COVERAGE` and `_DD_COVER_SESSION` (which introduces a new `coverage run` like behavior).

There are no unit tests though the overall use of the feature flags has been tested quite extensively in the process of collecting performance data.

There are no release notes since this is an entirely undocumented feature for the moment.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


romainkomorndatadog:
For posterity's sake, I insist on mentioning that the overwhelming majority of this work comes from @P403n1x87 , and I'm just piggybacking my changes on top.

Code review should mainly focus on the safety of the feature flags (eg: that we don't break things unless people are using undocumented flags from unsupported features).

ddtrace/internal/coverage/code.py
```diff
@@ -0,0 +1,264 @@
...
+        # mutating the parent code objects, the hashes maintained by the
+        # generator will be invalidated.
+        for nested_code, parent_code in list(collect_code_objects(code)):
+            # Instrument the code object
+            new_code = self.instrument_code(nested_code)
+
+            # If it has a parent, update the parent's co_consts to point to the
+            # new code object.
+            if parent_code is not None:
+                replace_in_tuple(parent_code.co_consts, nested_code, new_code)
```
P403n1x87, line 218: For anybody wondering why we do this hacky thing instead of recursively scanning the arguments of the abstract code instruction, looking for `CodeType` objects to replace, it turns out that the current approach is much faster 🤷

ZStriker19 (approved):

ddtrace/internal/coverage/code.py
```diff
@@ -0,0 +1,341 @@
...
+_original_exec = exec
+
+ctx_covered = ContextVar("ctx_covered", default=None)
+ctx_coverage_enabed = ContextVar("ctx_coverage_enabled", default=False)
+
+
+def collapse_ranges(numbers):
+    # This function turns an ordered list of numbers into a list of ranges.
+    # For example, [1, 2, 3, 5, 6, 7, 9] becomes [(1, 3), (5, 7), (9, 9)]
+    # WARNING: Written by Copilot
```
P403n1x87, line 26:
```suggestion
    # This function turns an ordered list of numbers into a list of ranges.
    # For example, [1, 2, 3, 5, 6, 7, 9] becomes [(1, 3), (5, 7), (9, 9)]
```

ddtrace/internal/coverage/code.py
```diff
@@ -0,0 +1,341 @@
...
+
+CWD = Path.cwd()
+
+_original_exec = exec
+
+ctx_covered = ContextVar("ctx_covered", default=None)
+ctx_coverage_enabed = ContextVar("ctx_coverage_enabled", default=False)
+
+
+def collapse_ranges(numbers):
```
P403n1x87, line 23:
```suggestion
def collapse_ranges(numbers: t.List[int]) -> t.List[t.Tuple[int, int]]:
```

ddtrace/internal/coverage/code.py
```diff
@@ -0,0 +1,341 @@
...
+    # This function turns an ordered list of numbers into a list of ranges.
+    # For example, [1, 2, 3, 5, 6, 7, 9] becomes [(1, 3), (5, 7), (9, 9)]
+    # WARNING: Written by Copilot
+    if not numbers:
+        return ""
```
P403n1x87, line 28:
```suggestion
        return []
```

ddtrace/internal/coverage/code.py
```diff
@@ -0,0 +1,341 @@
...
+                # This line has already been covered
+                lines.add(line)
+
+        if ctx_coverage_enabed.get():
+            ctx_lines = ctx_covered.get()[path]
+            if line not in ctx_lines:
+                ctx_lines.add(line)
+
+    @classmethod
+    def report(cls, ignore_nocover: bool = False):
```
P403n1x87, line 105: There is a fair bit of logic here now. Perhaps moving this into a `report` submodule might make this look a bit more tidy?

ddtrace/internal/module.py
```diff
@@ -53,6 +63,19 @@ def _patch_run_code() -> None:
...
+
+
+def unregister_run_module_transformer(transformer: TransformerType) -> None:
+    """Unregister a run module transformer.
+
+    If the transformer was not registered, a ``ValueError`` exception is raised.
+    """
+    _run_module_transformers.remove(transformer)
+
+
```
P403n1x87, line 78: I think this is just dead code for now. This would be useful for a stand-alone coverage tool whereby we would have to make sure that we instrument the main module that's being run. With pytest we don't have this problem.

riotfile.py
```diff
@@ -561,7 +561,7 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
             # Default Python 3 (3.10) collections package breaks with kombu/vertica, so specify Python 3.9 instead.
             pys="3.9",
             create=True,
-            skip_dev_install=True,
+            skip_dev_install=True,
```
P403n1x87, line 564:
```suggestion
            skip_dev_install=True,
```

ddtrace/internal/module.py
romainkomorndatadog: 🤔 , what about in the `ddtrace-run` context? eg: with our `unittest` integration?

ddtrace/internal/coverage/code.py
romainkomorndatadog: I give you `report.py` and even throw in `util.py` for free.

ddtrace/internal/module.py
P403n1x87: What would the main module be in that case? Is it not `unittest` itself? If so, I think we wouldn't need this code

romainkomorndatadog: I think it depends (eg: someone could have a wrapper that loads the class and does some discovery).

I guess I'll cross that bridge when I get to it. I removed it in 05b2686fd .

P403n1x87 (approved):

ddtrace/internal/ci_visibility/coverage.py
```diff
@@ -16,12 +18,18 @@
...
 from ddtrace.internal.logger import get_logger
+from ddtrace.internal.utils.formats import asbool
 
 
 log = get_logger(__name__)
 _global_relative_file_paths_for_cov: Dict[str, Dict[str, str]] = {}
 
+# This feature-flags experimental collection of code coverage via our internal ModuleCodeCollector.
+# It is disabled by default because it is not production-ready.
+USE_DD_COVERAGE = asbool(os.environ.get("_DD_USE_INTERNAL_COVERAGE", "false"))
```
P403n1x87, line 31: If we expect this feature to have a set of configuration env vars perhaps we can create a sub-class of `envier.En` in `ddtrace/settings`

romainkomorndatadog: I don't expect it to, at the moment. This is only a feature flag used for internal testing.

I suppose there may be some set of releases where we want to provide the ability to switch between our internal coverage and the current solution, but I was planning on addressing that at the point where our internal coverage collector becomes the default for ITR.

juan-fernandez (approved):

emmettbutler (approved):

Issue #5043, opened 2023-02-04
Please support Dramatiq


santigandolfo:
### Summary of problem

Please consider adding [dramatiq](https://github.com/Bogdanp/dramatiq) to the [list of supported libraries](https://docs.datadoghq.com/tracing/trace_collection/compatibility/python/#library-compatibility). It currently has around 3.4k stars on github, so it is certainly a popular and widely used library.

### Which version of dd-trace-py are you using?

1.7.4

### Which version of pip are you using?

23.0

### Which libraries and their versions are you using?

<details>
  <summary>`pip freeze`</summary>
dramatiq==1.13.0
</details>

### How can we reproduce your problem?

When you currently use ddtrace like with dramatiq like so:

```
$ ddtrace-run dramatiq app.main
```

### What is the result that you get?

Nothing really happens.

### What is the result that you expected?

The app should how up under Datadog / APM / Services.

oav17:
It would be nice to support Dramatiq. Do you think it's possible?

PR #8779, opened 2024-03-26
feat(dramatiq): add `dramatiq` integration

erikayasuda:
## Overview
This PR adds the [`dramatiq`](https://github.com/Bogdanp/dramatiq/tree/master) library as a supported integration for `dd-trace-py`. This addresses https://github.com/DataDog/dd-trace-py/issues/5043.

### In Scope for this PR
- Instrumenting the `dramatiq.actor.Actor.send_with_options()` method. This is the method called when a function with the `@dramatiq.actor` decorator is called asynchronously. See example below:

```python
# app.py
import dramatiq
from flask import Flask

app = Flask(__name__)

@dramatiq.actor
def my_func():
   return "response"

@dramatiq.actor
def my_other_func(a: int, b: int) -> int:
   return a + b

@app.route('/')
def index():
   my_func.send() # this calls send_with_options() under the hood
   my_other_func.send_with_options(args=(1, 1), options={"max_retries": 3})
   return 'hello world'

if __name__ == '__main__':
    app.run(host='0.0.0.0', port=6060)
```

Running the above example Flask app with `ddtrace-run flask run` and hitting the `/` endpoint would generate a trace in the UI like the following:

Screenshot 2024-03-26 at 3 28 21 PM

The detailed span info for `my_other_func.send_with_options(...)` would look something like this:

```python
{
   actor: {
      name: my_other_func
      options: {"max_retries": 3}
   }
   env: test
   language: python
   span: {
      kind: producer
   }
}
```

#### NOTE
- The duration of the span is of the `send_with_options()` call itself, and not reflective of the execution duration of the function being asynchronously completed.
- When calling a function asynchronously with `send()`, it will also display on the UI as a `send_with_options()` span. This is because `send()` calls `send_with_options()` with empty options, so tracing both of these functions would create two spans for every instance that `send()` is called, with the exact same span information. 

### Out of Scope for this PR
- All other `dramatiq` methods.
- Supporting the actual duration of the function executed asynchronously by `send_with_options()`

The above and additional features can be investigated and added to this initial iteration at a later time.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


tests/contrib/dramatiq/test_snapshots.py
```diff
@@ -0,0 +1,102 @@
...
+        patch()
+
+        @dramatiq.actor
+        def fn_task():
+            return "hi"
+
+        fn_task.send()
+        fn_task.send_with_options(options={"max_retries": 1})
+
+    @snapshot(wait_for_num_traces=0)
```
erikayasuda, line 29: This and line 42 for `test_fn_task_synchronous()`. This isn't actually doing what I want it to do, which is to confirm that there are 0 traces generated. It looks like `wait_for_num_traces` could accidentally prematurely consider this correct by meeting the loop exit condition, when in reality the call did generate a trace. Leaving this blank causes an indexing error though, so maybe this IS ok? Doesn't feel right though, so looking for suggestions on making this better.

tests/contrib/dramatiq/test_snapshots.py
```diff
@@ -0,0 +1,117 @@
...
+        @dramatiq.actor
+        def fn_task():
+            raise ValueError("test error")
+
+        fn_task.send()
+
+    # Ignoring these two values due to variance in method name
+    # Python 3.7 - 3.9 -> send_with_options
+    # Python 3.10+ -> Actor.send_with_options
+    @snapshot(ignores=["meta.error.message", "meta.error.stack"], wait_for_num_traces=1)
```
erikayasuda, line 85: This feels a little clunky, but not sure if there's a cleaner way to do this?

tests/contrib/dramatiq/test_snapshots.py
emmettbutler: If you're wanting to check that no traces are generated, a snapshot test might not be the best fit. Instead, you could set up a `DummyTracer` and then verify that its list of traces was empty, similarly to [this example](https://github.com/DataDog/dd-trace-py/blob/fedf88e7161e7eff19038231c2b2fa8f8f75ddec/tests/contrib/cassandra/test.py#L441-L449).

tests/contrib/dramatiq/test_snapshots.py
emmettbutler: The fact that `error.type` isn't ignored means this test still meaningfully verifies that an error was reflected in the span. I think it's fine to ignore the more specific details of that error.

tests/contrib/dramatiq/test_snapshots.py
```diff
@@ -0,0 +1,102 @@
...
+        # it should catch exceptions in task functions
+        @dramatiq.actor
+        def fn_task():
+            raise ValueError("test error")
+
+        fn_task.send()
+
+
+@snapshot(wait_for_num_traces=1)
+def test_fn_retry_exception(stub_broker, stub_worker):
```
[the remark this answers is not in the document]
emmettbutler: I've used the exact same approach. I can't find evidence in documentation, but fixtures to instance methods might be something that Pytest doesn't support.

emmettbutler (approved):

tests/contrib/dramatiq/test_snapshots.py
erikayasuda: `Pin`!! 📌

PR #8796, opened 2024-03-27
fix(setuptools): use guess-next-dev schema

tabgok:
A backport (#8367) changed our setuptools_scm schema from `guess-next-dev` to `release-branch-semver` for the 2.5 and 2.6 releases; but this change was never put into mainline.  This change fixes the issue.

Prior to this change, setuptools_scm reported the version for patch releases as the next minor release.  Afterwards, versions are patch releases (as expected).

## Checklist

## Reviewer Checklist

- [ ] Title is accurate

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -69,7 +69,7 @@ Homepage = "https://github.com/DataDog/dd-trace-py"
 "Source Code" = "https://github.com/DataDog/dd-trace-py/"
 
 [tool.setuptools_scm]
-version_scheme = "release-branch-semver"
+version_scheme = "get-next-dev"
 write_to = "ddtrace/_version.py"
 
 [tool.cython-lint]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -69,7 +69,7 @@ Homepage = "https://github.com/DataDog/dd-trace-py"
 "Source Code" = "https://github.com/DataDog/dd-trace-py/"
 
 [tool.setuptools_scm]
-version_scheme = "get-next-dev"
+version_scheme = "guess-next-dev"
 write_to = "ddtrace/_version.py"
 
 [tool.cython-lint]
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -69,7 +69,7 @@ Homepage = "https://github.com/DataDog/dd-trace-py"
 "Source Code" = "https://github.com/DataDog/dd-trace-py/"
 
 [tool.setuptools_scm]
-version_scheme = "release-branch-semver"
+version_scheme = "guess-next-dev"
```
P403n1x87, line 72: Let's add a comment above this property explaining why this needs to be `release-branch-semver` on the main branch

christophe-papazian (changes requested):
**I am requesting this PR to be closed.**

## Explanations:
we have 2 possible policies for setup tools, none of them is ok for all cases.
Setuptools is looking back from current commit and looking for the first version tag in ancestor commit.
Then either it’s using that version with `release-branch-semver` or using the next minor dev version with `guess-next-dev`

Currently we need the next minor dev version to be able to enable new system tests for new “just merged in main” features. But we need the same version policy for all backports, otherwise we will enable tests in system tests for features merged in the next version.

So for now, we are using “guess-next-dev” in main and as soon as it’s needed we change it into “release-branch-semver” for minor version branches for all backports.

(This is a known problem that I already explained to a guild meeting, while sponsoring for one of the following solution, but this was never supported by the guild)

What are doing other big python projects on that problem ?
- Do as we are doing (not really satisfying)
- Use always “release-branch-semver” but add a new minor dev tag to the main branch on an empty commit after each creation of a new minor/major branch. We would need to integrate that step to our release todo list
- Drop setuptools for version computation and have the version as constants in a file that would need to be updated for each release. Require more work but with less dependency, simpler build process and more control.

PR #8798, opened 2024-03-27
feat(botocore): add support for tagging bedrock message input/outputs

Yun-Kim:
This PR adds support to the botocore integration's bedrock service to correctly tag input/output messages from Anthropic calls.

Previously bedrock's models only supported raw text prompts and returned text outputs. However, Anthropic's newest claude 3 supports a chat message API, which means we need to support that as well.

This change also switches to using `tracer.trace()` instead of `tracer.start_span(..., activate=False)` for bedrock spans, because the latter meant that bedrock spans would always be root spans (messing up parenting for traces containing non-root bedrock spans). 
Additionally by getting rid of the `activate=False` argument, this means that bedrock spans will now continue to be the active span until the stream/body is completely consumed. Previously we allowed bedrock spans to not be active, but if other downstream operations happen in the bedrock span then they would not correctly be child spans of the bedrock spans.

## Checklist

## Reviewer Checklist

- [x] Title is accurate

### DIFF
--- a/ddtrace/contrib/botocore/services/bedrock.py
+++ b/ddtrace/contrib/botocore/services/bedrock.py
@@ -143,8 +143,10 @@ def _extract_request_params(params: Dict[str, Any], provider: str) -> Dict[str, 
             "stop_sequences": text_generation_config.get("stopSequences", []),
         }
     elif provider == _ANTHROPIC:
+        prompt = request_body.get("prompt", "")
+        messages = request_body.get("messages", "")
         return {
-            "prompt": request_body.get("prompt"),
+            "prompt": prompt or messages,
             "temperature": request_body.get("temperature", ""),
             "top_p": request_body.get("top_p", ""),
             "top_k": request_body.get("top_k", ""),
@@ -189,7 +191,7 @@ def _extract_response(span: Span, body: Dict[str, Any]) -> Dict[str, List[str]]:
             text = body.get("results")[0].get("outputText")
             finish_reason = body.get("results")[0].get("completionReason")
         elif provider == _ANTHROPIC:
-            text = body.get("completion")
+            text = body.get("completion", "") or body.get("content", "")
             finish_reason = body.get("stop_reason")
         elif provider == _COHERE:
             text = [generation["text"] for generation in body.get("generations")]
@@ -226,8 +228,15 @@ def _extract_streamed_response(span: Span, streamed_body: List[Dict[str, Any]]) 
             text = "".join([chunk["outputText"] for chunk in streamed_body])
             finish_reason = streamed_body[-1]["completionReason"]
         elif provider == _ANTHROPIC:
-            text = "".join([chunk["completion"] for chunk in streamed_body])
-            finish_reason = streamed_body[-1]["stop_reason"]
+            for chunk in streamed_body:
+                if "completion" in chunk:
+                    text += chunk["completion"]
+                    if chunk["stop_reason"]:
+                        finish_reason = chunk["stop_reason"]
+                elif "delta" in chunk:
+                    text += chunk["delta"].get("text", "")
+                    if "stop_reason" in chunk["delta"]:
+                        finish_reason = str(chunk["delta"]["stop_reason"])
         elif provider == _COHERE and streamed_body:
             if "is_finished" in streamed_body[0]:  # streamed response
                 if "index" in streamed_body[0]:  # n >= 2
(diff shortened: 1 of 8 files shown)


ddtrace/contrib/botocore/services/bedrock.py
```diff
@@ -325,12 +334,11 @@ def patched_bedrock_api_call(original_func, instance, args, kwargs, function_var
     pin = function_vars.get("pin")
     endpoint_name = function_vars.get("endpoint_name")
     integration = function_vars.get("integration")
-    # This span will be finished separately as the user fully consumes the stream body, or on error.
-    bedrock_span = pin.tracer.start_span(
+    # This span will be finished once the user fully consumes the stream body, or on error.
```
lievan, line 337: What changed so that we want to use `trace` instead of `start_span` with `activation=False`?

ddtrace/llmobs/_integrations/bedrock.py
```diff
@@ -55,3 +55,39 @@ def _llmobs_metrics(span: Span, formatted_response: Optional[Dict[str, Any]]) ->
...
         return metrics
+
+    @staticmethod
+    def _extract_input_message(prompt):
+        """Extract input messages from the stored prompt.
+        Anthropic allows for messages and multiple texts in a message, which requires some special casing.
+        """
+        if isinstance(prompt, str):
+            return [{"content": prompt}]
+        if not isinstance(prompt, list):
```
lievan, line 66: Is it conventional for the python tracer to add a warning log in cases like this?

Might be helpful for users if the spec for prompt changes again and their data suddenty doesn't show up

ddtrace/llmobs/_integrations/bedrock.py
```diff
@@ -55,3 +55,39 @@ def _llmobs_metrics(span: Span, formatted_response: Optional[Dict[str, Any]]) ->
...
+            return [{"content": ""}]
+        input_messages = []
+        for p in prompt:
+            content = p.get("content", "")
+            if isinstance(content, list) and isinstance(content[0], dict):
+                for entry in content:
+                    if entry.get("type") == "text":
+                        input_messages.append({"content": entry.get("text"), "role": str(p.get("role"))})
```
lievan, line 74: nit, maybe should do a `str(p.get("role") or "")` here

ddtrace/llmobs/_integrations/bedrock.py
```diff
@@ -55,3 +55,39 @@ def _llmobs_metrics(span: Span, formatted_response: Optional[Dict[str, Any]]) ->
...
+                    elif entry.get("type") == "image":
+                        # Store a placeholder for potentially enormous binary image data.
+                        input_messages.append({"content": "{IMAGE DETECTED}", "role": str(p.get("role"))})
```
lievan, line 77: should we mention this in the release notes or docs somewhere? 

A bit hesitant about this since an LLM can also just generate `{IMAGE DETECTED}`. If we do set a placeholder, maybe we should consider making the same as how we surface sensitive information scanning `([IMAGE DETECTED])` so the way we display placeholder text is standardized. I can also see an argument for wanting these two placeholders to have separate displays though

lievan (approved):
Looks good, just have a question about the placeholder text for images

ZStriker19 (approved):

ddtrace/contrib/botocore/services/bedrock.py
Yun-Kim: This is a subtle but pretty important change and has to do with trace parenting. Basically revolves around the idea of what span is "active" (which spans will automatically then be children from).

This change switches to using `tracer.trace()` instead of `tracer.start_span(..., activate=False)` for bedrock spans, because the latter method meant that bedrock spans would always be root spans (messing up parenting for traces containing non-root bedrock spans). Now bedrock spans can correctly be created as children of whatever span is active.
Also, by getting rid of the activate=False argument, this means that bedrock spans will now continue to be the active span until the stream/body is completely consumed. Previously we allowed bedrock spans to not be active, meaning other downstream spans would be children of the bedrock span's parent span (or in a new trace).

ddtrace/llmobs/_integrations/bedrock.py
Yun-Kim: Good point, will add a warning

ddtrace/llmobs/_integrations/bedrock.py
Yun-Kim: We can move forward with being consistent with the sensitive information scanning placeholder.
For regular APM users, this placeholder won't matter (since they will still get the truncated image blob tag anyway) so we don't need to add into a release note. For LLMObs customers, we'll just have to add this to our internal/onboarding docs so customers are aware we don't support submitting entire image blobs.

PR #8809, opened 2024-03-29
chore(tracer): use importlib metadata to infer flask version as the old method is deprecated

christophe-papazian:
Change the method we use to infer Flask version, as the current method is deprecated (remove associate warnings from Flask framework).
This PR will ensure the Flask support in the tracer won't break with next Flask 3.1 release.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


ddtrace/contrib/flask/patch.py
```diff
@@ -71,7 +72,10 @@
 
 def get_version():
     # type: () -> str
-    return getattr(flask, "__version__", "")
+    try:
+        return importlib_metadata.version("flask")
```
brettlangdon, line 76: 🚲 🏠 (closest to "bike shed" I can do).

Should we offer an API in `ddtrace.internal.packages` for getting a single package's version? Then we could also cache it like the others.


Definitely not urgent to address, just a thought.

ddtrace/internal/packages.py
```diff
@@ -10,6 +10,12 @@
 from ddtrace.internal.utils.cache import callonce
 
 
+try:
```
brettlangdon, line 13: 🤔 why didn't we do this before?

Is there a high overhead from importing `importlib_metadata`?

ddtrace/contrib/flask/patch.py
christophe-papazian: Yes, it would be great!

ddtrace/internal/packages.py
```diff
@@ -10,6 +10,12 @@
...
+    import importlib.metadata as importlib_metadata
+except ImportError:
+    import importlib_metadata  # type: ignore[no-redef]
```
erikayasuda, line 16: Why are we moving this out to the module scope instead of keeping it within the function scope (for this specific function)? I see it's still imported at the function level in some others

ddtrace/internal/packages.py
christophe-papazian: I don't know, I would love to have feedback on it. I don't think that it can be a problem with the regular setup.
But I was wondering if it could be a serverless issue.

brettlangdon: Good question, locally it seems to load pretty quick for me, so maybe importing isn't the issue?

ddtrace/internal/packages.py
christophe-papazian: it was in a local scope, I moved it to the module scope to make it available in the two other files that need to access the current Flask version.

ddtrace/internal/packages.py
christophe-papazian: I'll wait for the gitlab benchmark comments on that PR, it should signal if there is a problem.

christophe-papazian: I'd like to have some review from serverless. It seems there is some regression on that one.
But we will need a fix to still support next Flask version.

emmettbutler (approved):

gnufede (approved):

ddtrace/internal/packages.py
christophe-papazian: I changed the solution to use local imports. There is no significant benchmark changes now.

PR #7618, opened 2023-11-15
feat(starlette/fastapi): trace background tasks using span links

mabdinur:
Resolves https://github.com/DataDog/dd-trace-py/issues/7132 and https://github.com/DataDog/dd-trace-py/issues/3851 , by ensuring background tasks are not nested under fastapi/starlette request spans. Instead traced background tasks linked to request spans via span links. 

This PR introduces the first usage of span links in a ddtrace integration. H̶o̶w̶e̶v̶e̶r̶ ̶s̶p̶a̶n̶ ̶l̶i̶n̶k̶s̶ ̶a̶r̶e̶ ̶n̶o̶t̶ ̶y̶e̶t̶ ̶s̶u̶p̶p̶o̶r̶t̶e̶d̶ ̶b̶y̶ ̶t̶h̶e̶ ̶d̶a̶t̶a̶d̶o̶g̶ ̶a̶g̶e̶n̶t̶.̶ ̶S̶p̶a̶n̶ ̶l̶i̶n̶k̶s̶ ̶g̶e̶n̶e̶r̶a̶t̶e̶d̶ ̶b̶y̶ ̶f̶a̶s̶t̶a̶p̶i̶ ̶a̶n̶d̶ ̶s̶t̶a̶r̶l̶e̶t̶t̶e̶ ̶i̶n̶t̶e̶g̶r̶a̶t̶i̶o̶n̶s̶ ̶w̶i̶l̶l̶ ̶n̶o̶t̶ ̶a̶p̶p̶e̶a̶r̶ ̶i̶n̶ ̶t̶h̶e̶ ̶D̶a̶t̶a̶d̶o̶g̶ ̶U̶I̶ ̶u̶n̶t̶i̶l̶ ̶t̶h̶e̶ ̶n̶e̶x̶t̶ ̶d̶a̶t̶a̶d̶o̶g̶ ̶a̶g̶e̶n̶t̶ ̶r̶e̶l̶e̶a̶s̶e̶.̶ ̶

## Checklist

- [x] Change(s) are motivated and described in the PR description.
- [x] Testing strategy is described if automated tests are not included in the PR.
- [x] Risk is outlined (performance impact, potential for breakage, maintainability, etc).
- [x] Change is maintainable (easy to change, telemetry, documentation).
- [x] [Library release note guidelines](https://ddtrace.readthedocs.io/en/stable/releasenotes.html) are followed. If no release note is required, add label `changelog/no-changelog`.
[...]

PR #8812, opened 2024-04-01
fix(starlette): don't crash on functions without `__name__`

ostrowr:
Don't crash when encountering functions without `__name__` or `__module__` attributes.

Starlette background tasks can be started with any `callable`, and not all `callable`s (annoyingly) are guaranteed to have `__name__` attributes. Specifically, `functools.partial` generates a function without a `__name__`. There is some debate (see https://github.com/python/cpython/issues/91002) about whether this is the right behavior (it's clearly confusing, since it caused a bug here!) but right now, this makes the whole background task crash due to an attribute error.

This PR just tries to get the name/module, and if not available, replaces them with the string `<unknown>` rather than crashing the whole task. (Happy to switch to a different string here, `<unknown>` is just a placeholder that kind of matches python's `__name__` for lambdas (`<lambda>`)

## Checklist

## Reviewer Checklist

- [x] Title is accurate

### DIFF
--- a/ddtrace/contrib/starlette/patch.py
+++ b/ddtrace/contrib/starlette/patch.py
@@ -177,10 +177,12 @@ def traced_handler(wrapped, instance, args, kwargs):
 def _trace_background_tasks(module, pin, wrapped, instance, args, kwargs):
     task = get_argument_value(args, kwargs, 0, "func")
     current_span = pin.tracer.current_span()
+    module_name = getattr(module, "__name__", "<unknown>")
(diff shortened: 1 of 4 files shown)


ostrowr:
Still need to update the changelog; `riot run reno new` is giving me problems.

ostrowr:
Doesn't look like I have permissions to add backport labels but it looks like this bug was introduced in https://github.com/DataDog/dd-trace-py/pull/7618

emmettbutler (approved):

PR #8830, opened 2024-04-02
ci: suppress failures from post-job docker-compose logs call

emmettbutler:
This pull request silences ephemeral failures like [this one](https://github.com/DataDog/dd-trace-py/actions/runs/8524577528/job/23349480264), which have been observed a few times recently on main. The gathering of logs after the job succeeds is not a critical part of the job itself, so it's ok to ignore when it fails.

## Checklist

## Reviewer Checklist

- [ ] Title is accurate

### DIFF
--- a/.github/workflows/lib-injection.yml
+++ b/.github/workflows/lib-injection.yml
@@ -106,4 +106,4 @@ jobs:
       - name: Output app logs (LOOK HERE IF THE JOB FAILS)
         if: success() || failure()
         run: |
-          docker-compose logs
+          docker-compose logs || true


brettlangdon:
Will this actually fix anything? It seems the example build failed at an earlier step because `docker-compose` wasn't present, so the job is going to still fail since it is only the logs job you are adding `|| true` to (do I understand that correctly?)

.github/workflows/lib-injection.yml
```diff
@@ -106,4 +106,4 @@ jobs:
       - name: Output app logs (LOOK HERE IF THE JOB FAILS)
         if: success() || failure()
         run: |
-          docker-compose logs
+          docker-compose logs || true
```
erikayasuda, line 109: I saw that the issue is from `docker-compose: command not found` from the link you shared. Is `docker-compose` not guaranteed to be installed every time?

emmettbutler:
Ah yeah, good call @brettlangdon. We can't really skip all of the docker-compose stuff - instead we'll have to make sure it's always available or xfail this test when it's not.

PR #8835, opened 2024-04-02
fix(langchain): type check chain.invoke() input argument name

Yun-Kim:
This PR fixes the langchain integration's patched chain method to check for the correct input argument name.

In `LangChain<0.1`, we patch `langchain.Chain.__call__()`, which uses `inputs: Union[Dict[str, Any], str]` as the argument to the chain invocation. However in `LangChain>=0.1`, we patch `langchain.Chain.invoke()`, which uses `input: Dict[str, Any]` as the argument to the chain invocation. We use the same traced function to patch both methods, but this subtle change broke our argument parsing, which expected the name `inputs` instead of `input`.

## Checklist

## Reviewer Checklist

- [x] Title is accurate
- [X] Release note makes sense to a user of the library

### DIFF
--- a/ddtrace/contrib/langchain/patch.py
+++ b/ddtrace/contrib/langchain/patch.py
@@ -581,7 +581,10 @@ def traced_chain_call(langchain, pin, func, instance, args, kwargs):
     span = integration.trace(pin, "%s.%s" % (instance.__module__, instance.__class__.__name__), interface_type="chain")
     final_outputs = {}
     try:
-        inputs = get_argument_value(args, kwargs, 0, "inputs")
+        if SHOULD_PATCH_LANGCHAIN_COMMUNITY:
+            inputs = get_argument_value(args, kwargs, 0, "input")
+        else:
+            inputs = get_argument_value(args, kwargs, 0, "inputs")
         if not isinstance(inputs, dict):
             inputs = {instance.input_keys[0]: inputs}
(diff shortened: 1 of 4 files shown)


ddtrace/contrib/langchain/patch.py
```diff
@@ -629,9 +632,12 @@ async def traced_chain_acall(langchain, pin, func, instance, args, kwargs):
...
     final_outputs = {}
     try:
-        inputs = get_argument_value(args, kwargs, 0, "inputs")
-        if not isinstance(inputs, dict):
-            inputs = {instance.input_keys[0]: inputs}
+        if SHOULD_PATCH_LANGCHAIN_COMMUNITY:
+            inputs = get_argument_value(args, kwargs, 0, "input")
+        else:
+            inputs = get_argument_value(args, kwargs, 0, "inputs")
+            if not isinstance(inputs, dict):
```
sabrenner, line 639: should this `if` apply in both cases? or just when it's an older langchain version (seems in the non async traced method, it's done in either case)?

sabrenner (approved):
lgtm, one nonblocking question

erikayasuda (approved):

ddtrace/contrib/langchain/patch.py
Yun-Kim: Yeah good catch, I missed this

PR #8900, opened 2024-04-08
chore(sampling): add rc for trace sampling rules

ZStriker19:
This PR implements remote config for `DD_TRACE_SAMPLING_RULES`:
1. Add the rc
2. Add `provenance` which we parse from rc and use to give a new decision maker `_dd.p.dm` to either `-11` for `customer` configuration or `-12` for `dynamic` configuration.
3. The most confusing part of this implementation takes place in[ tracer._on_global_config_update](https://github.com/DataDog/dd-trace-py/pull/8900/files#diff-e2ff2c401c4b927861c9fc104deb21aee510e2b273cc4569315bd611a64ff3baL1123-R1164).  Essentially implementing the logic for choosing the correct sample_rate and sampling_rules depending on rc input.

In addition to the sample_rate we already had I added one for sampling_rules and for the interaction between them. This PR also passes the newly added [system-tests](https://github.com/DataDog/system-tests/blob/main/tests/parametric/test_dynamic_configuration.py#L559-L683) for this behavior.

Since this feature is in internal beta, there's no release note.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


ddtrace/internal/sampling.py
```diff
@@ -278,8 +280,14 @@ def is_single_span_sampled(span):
 def _set_sampling_tags(span, sampled, sample_rate, priority_category):
     # type: (Span, bool, float, str) -> None
     mechanism = SamplingMechanism.TRACE_SAMPLING_RULE
-    if priority_category == "rule":
+    if priority_category == "rule_default":
```
emmettbutler, line 283: You could use the constants here

ddtrace/settings/config.py
```diff
@@ -802,3 +814,62 @@ def enable_remote_configuration(self):
         remoteconfig_poller.register("APM_TRACING", remoteconfig_pubsub)
         remoteconfig_poller.register("AGENT_CONFIG", remoteconfig_pubsub)
         remoteconfig_poller.register("AGENT_TASK", remoteconfig_pubsub)
+
+    def _remove_invalid_rules(self, rc_rules: List) -> List:
+        """Remove invalid sampling rules from the JSON string."""
```
emmettbutler, line 819:
```suggestion
        """Remove invalid sampling rules from the given list"""
```

ddtrace/internal/sampling.py
emmettbutler, line 283: You could use the constants here

ddtrace/settings/config.py
```diff
@@ -802,3 +814,62 @@ def enable_remote_configuration(self):
...
+                ("service" not in rule and "name" not in rule and "resource" not in rule and "tags" not in rule)
+                or "sample_rate" not in rule
+                or "provenance" not in rule
+            ):
+                log.debug("Invalid sampling rule from remoteconfig found, rule will be removed: %s", rule)
+                rc_rules.remove(rule)
+
+        return rc_rules
+
+    def _tags_to_dict(self, tags):
```
emmettbutler, line 832:
```suggestion
    def _tags_to_dict(self, tags:List[Dict]):
```

ddtrace/settings/config.py
```diff
@@ -802,3 +814,62 @@ def enable_remote_configuration(self):
...
+        """
+        Converts a list of tag dictionaries to a single dictionary.
+        """
+        if isinstance(tags, list):
+            return {tag["key"]: tag["value_glob"] for tag in tags}
+        return tags
+
+    def convert_rc_trace_sampling_rules(self, rc_rules: List[Dict[str, Any]]) -> Optional[str]:
+        """Example of an incoming rule:
+        [
```
emmettbutler, line 842: This is a great use case for doctest https://docs.python.org/3/library/doctest.html

ddtrace/settings/config.py
```diff
@@ -802,3 +814,62 @@ def enable_remote_configuration(self):
...
+                '[{"sample_rate":1.0,"service":"my-service","resource":"*","name":"web.request","tags":{"care_about":"yes","region":"us-*"},provenance":"customer"}]'
+        """
+        rc_rules = self._remove_invalid_rules(rc_rules)
+        for rule in rc_rules:
+            # Convert tags to dictionary
+            tags = rule.get("tags")
+            if tags:
+                rule["tags"] = self._tags_to_dict(tags)
+        if rc_rules:
+            # Convert JSON to string
```
emmettbutler, line 872: These comments are redundant because they say what's being done rather than why it's being done.

ddtrace/sampler.py
```diff
@@ -283,13 +284,23 @@ def _parse_rules_from_env_variable(self, rules):
...
+                    name=name,
+                    resource=resource,
+                    tags=tags,
+                    provenance=provenance,
                 )
             except ValueError as e:
                 raise ValueError("Error creating sampling rule {}: {}".format(json.dumps(rule), e))
             sampling_rules.append(sampling_rule)
+
+        provenance_order = ["customer", "dynamic", "default"]
```
emmettbutler, line 301: This might be good as a global constant

ddtrace/_trace/tracer.py
```diff
@@ -286,7 +287,7 @@ def __init__(
         self._shutdown_lock = RLock()
 
         self._new_process = False
-        config._subscribe(["_trace_sample_rate"], self._on_global_config_update)
+        config._subscribe(["_trace_sample_rate", "_trace_sampling_rules"], self._on_global_config_update)
```
emmettbutler, line 290: This makes me wonder why we call `_subscribe` more than once.

ddtrace/_trace/tracer.py
```diff
@@ -1125,18 +1126,44 @@ def _is_span_internal(span):
 
     def _on_global_config_update(self, cfg, items):
         # type: (Config, List) -> None
-        if "_trace_sample_rate" in items:
-            # Reset the user sampler if one exists
-            if cfg._get_source("_trace_sample_rate") != "remote_config" and self._user_sampler:
-                self._sampler = self._user_sampler
+
+        # sampling configs always come as a pair
+        if "_trace_sample_rate" in items and "_trace_sampling_rules" in items:
```
emmettbutler, line 1131: Putting this complex logic in a `sampling.py` helper function might underline the fact that it's complex and not directly related to the tracer. It would also reduce the cognitive load incurred on editors of this file.

emmettbutler (approved):

ddtrace/_trace/tracer.py
ZStriker19: I like that idea too, but it leads to a circular import because I need to import DatadogSampler into sampling.py. I guess I can just return the sampling_rules and sample_rate though.

ZStriker19: Ah, can't do that either, I will break it into a separate method though.

PR #8948, opened 2024-04-11
chore(internal): add agentless mode for telemetry writer

romainkomorndatadog:
This enables the telemetry writer to work properly when agentless mode is used (mainly during CI Visibility use cases).

The format is identical, with the only changes being:
* the path (not using the agent proxy prefix)
* the URL (based on `DD_SITE`)
* adding the API key as a header

This also removes the "skip if agentless" decorator which only existed because the writer didn't support agentless mode.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


romainkomorndatadog:
I need to work on relevant tests.

ddtrace/internal/telemetry/writer.py
```diff
@@ -121,38 +136,53 @@ def __eq__(self, other):
...
+
     @property
     def url(self):
-        return parse.urljoin(self._agent_url, self._endpoint)
+        return parse.urljoin(self._telemetry_url, self._endpoint)
 
     def send_event(self, request: Dict) -> Optional[httplib.HTTPResponse]:
         """Sends a telemetry request to the trace agent"""
+        if self._is_disabled:
+            log.debug("Telemetry is disabled, not sending event.")
```
brettlangdon, line 167: this might be a bit noisy, no?

brettlangdon: missing an early return?

ddtrace/internal/telemetry/writer.py
```diff
@@ -121,38 +136,53 @@ def __eq__(self, other):
...
+        endpoint_str = "Datadog endpoint" if self._is_agentless else "Datadog Agent"
         try:
             rb_json = self._encoder.encode(request)
             headers = self.get_headers(request)
             with StopWatch() as sw:
-                conn = get_connection(self._agent_url)
+                conn = get_connection(self._telemetry_url)
                 conn.request("POST", self._endpoint, rb_json, headers)
                 resp = get_connection_response(conn)
+                # print("sent %d in %.5fs to %s. response: %s", len(rb_json), sw.elapsed(), self.url, resp.status)
```
brettlangdon, line 179:
```suggestion
```

tests/conftest.py
```diff
@@ -401,7 +401,9 @@ def filter_heartbeat_events():
 
 @pytest.fixture
 def telemetry_writer():
```
brettlangdon, line 403: You could do something like:

```python
@pytest.fixture
def telemetry_writer_is_agentless():
    return False

@pytest.fixture
def telemetry_writer(telemetry_writer_is_agentless):
    telemetry_writer = TelemetryWriter(is_periodic=False, agentless=telemetry_writer_is_agentless)
```


This way in the future if we ever want an individual test to be agentless it'll be easy to overwrite the fixture value.

brettlangdon: not a strong opinion, an option/pattern we have used elsewhere.

tests/telemetry/test_telemetry_metrics.py
```diff
@@ -20,6 +20,7 @@ def _assert_metric(
 ):
     test_agent.telemetry_writer.periodic()
     events = test_agent.get_events()
+    # print(f"EVENTS EVETNTS {events=}")
```
brettlangdon, line 23:
```suggestion
```

brettlangdon (commented):
Should we have tests that validate that we send payloads to the proper url when in agentless?

anmarchenko (approved):

tests/telemetry/test_telemetry_metrics.py
romainkomorndatadog: Damnit. I thought I got all those. 😅 

Can you tell I got frustrated because we leak our external CI env into our tests, and then we have that funny fixture that re-instantiates the writer? 😛

ddtrace/internal/telemetry/writer.py
romainkomorndatadog: Fixed in [af51bd3](https://github.com/DataDog/dd-trace-py/pull/8948/commits/af51bd374a4aba7bc6393a2390d231fb62a54664) .

ddtrace/internal/telemetry/writer.py
romainkomorndatadog: Delete in [af51bd3](https://github.com/DataDog/dd-trace-py/pull/8948/commits/af51bd374a4aba7bc6393a2390d231fb62a54664).

tests/telemetry/test_telemetry_metrics.py
romainkomorndatadog: Deleted in [af51bd3](https://github.com/DataDog/dd-trace-py/pull/8948/commits/af51bd374a4aba7bc6393a2390d231fb62a54664).

brettlangdon (approved):

romainkomorndatadog:
> Should we have tests that validate that we send payloads to the proper url when in agentless?

Sorry, @brettlangdon , totally missed this question. I just added fbffd2217 to validate the `dd-api-key` header on top of validating that the agentless setup properly sets the endpoint and telemetry URL.

The original battery of tests that ensure the telemetry actually gets sent. Even though it goes to the agent, the only practical difference between the regular and agentless behaviors are the URL and the API key header, and if they pass the agent-based tests (some of which check headers), then I think that's enough.

PR #8961, opened 2024-04-12
feat(tracer): isolated tracer flare logic

erikayasuda:
## Overview
This PR adds the tracer flare logic to dd-trace-py without wiring it up to the RC subscriber. Therefore, these changes will not be visible on the customer end yet. There will be a follow-up PR to wire this up.

A tracer flare is a tar file containing tracer logs from the last 5 minutes, as well as a JSON file of the current tracer configurations. This pair of files will be generated per tracer instance that is connected to the agent, and the tar containing all of these files will be sent to a Zendesk ticket via an endpoint on the ddagent. Users can trigger the tracer flare the same way they trigger agent flares. See Agent Flare documentation [here](https://docs.datadoghq.com/agent/troubleshooting/send_a_flare/?tab=agentv6v7).

## Checklist

## Reviewer Checklist

- [x] Title is accurate


ddtrace/internal/logger.py
```diff
@@ -177,3 +178,11 @@ def handle(self, record):
             # Increment the count of records we have skipped
             # DEV: `self.buckets[key]` is a tuple which is immutable so recreate instead
             self.buckets[key] = DDLogger.LoggingBucket(logging_bucket.bucket, logging_bucket.skipped + 1)
+
+
+def _get_handler(logger: logging.Logger, handler_name: str) -> Optional[logging.Handler]:
```
erikayasuda, line 183: Originally I had this type hint as `DDLogger`, but looks like the `ddtrace` logger is a `logging.Logger` type and not `DDLogger`? Not sure if this is intended

docker-compose.yml
```diff
@@ -124,7 +124,7 @@ services:
         volumes:
           - ddagent:/tmp/ddagent:rw
     testagent:
-        image: ghcr.io/datadog/dd-apm-test-agent/ddapm-test-agent:v1.16.0
+        image: ghcr.io/datadog/dd-apm-test-agent/ddapm-test-agent:v1.17.0
```
erikayasuda, line 127: Need to update this in order for the POST requests to `tracer_flare/v1` endpoint to properly be checked by the testagent, and return a 200

riotfile.py
```diff
@@ -427,6 +427,9 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
         ),
         Venv(
             name="internal",
+            env={
+                "DD_TRACE_AGENT_URL": "http://localhost:8126",
+            },
```
emmettbutler, line 432:
```suggestion
```
I believe this is the default, so it's not necessary to make it explicit here.

tests/internal/test_tracer_flare.py
```diff
@@ -0,0 +1,176 @@
...
+                "hostname": "myhostname",
+                "user_handle": "user.name@datadoghq.com",
+            },
+            "task_type": "tracer_flare",
+            "uuid": "d53fc8a4-8820-47a2-aa7d-d565582feb81",
+        },
+    ]
+    response = mock.MagicMock()
+    response.status = 200
+    flare = Flare()
```
emmettbutler, line 33: It might be less error-prone to do this in `setUp` to avoid unexpectedly sharing a `Flare` instance across tests.

tests/internal/test_tracer_flare.py
```diff
@@ -0,0 +1,176 @@
...
+            "args": {
+                "case_id": "1111111",
+                "hostname": "myhostname",
+                "user_handle": "user.name@datadoghq.com",
+            },
+            "task_type": "tracer_flare",
+            "uuid": "d53fc8a4-8820-47a2-aa7d-d565582feb81",
+        },
+    ]
+    response = mock.MagicMock()
```
emmettbutler, line 31: This looks unused as far as I can tell.

tests/internal/test_tracer_flare.py
```diff
@@ -0,0 +1,176 @@
...
+        self.config_file_path = f"{TRACER_FLARE_DIRECTORY}/tracer_config_{self.pid}.json"
+
+    def tearDown(self):
+        self.confirm_cleanup()
+
+    def test_single_process_success(self):
+        """
+        Validate that the baseline tracer flare works for a single process
+        AGENT_CONFIG expected behavior:
+        - file handler added to logger
```
emmettbutler, line 47: I like including this sort of thing in the `assert` text. For example `assert file_handler is not None, "file handler is successfully added to logger"`.

tests/internal/test_tracer_flare.py
```diff
@@ -0,0 +1,176 @@
...
+        p.start()
+        for p in processes:
+            p.join()
+
+    def test_no_app_logs(self):
+        """
+        Validate that app logs are not being added to the
+        file, just the tracer logs
+        """
+        app_logger = Logger(name="my-app", level=10)  # DEBUG level
```
emmettbutler, line 158:
```suggestion
        app_logger = Logger(name="my-app", level=logging.DEBUG)
```

ddtrace/internal/logger.py
emmettbutler: I think from the perspective of type hints' static analysis, it's just a `Logger` because it is passed through `logging.getLogger`.

ddtrace/internal/flare.py
```diff
@@ -0,0 +1,184 @@
...
+        self.original_log_level = 0
+        self.timeout = timeout_sec
+
+    def _prepare(self, configs: List[dict]):
+        """
+        Update configurations to start sending tracer logs to a file
+        to be sent in a flare later.
+        """
+        os.makedirs(TRACER_FLARE_DIRECTORY, exist_ok=True)
+        for c in configs:
```
emmettbutler, line 41: This variable is used in enough lines that naming it with a single letter might become a little cumbersome.

ddtrace/internal/flare.py
```diff
@@ -0,0 +1,184 @@
...
+
+log = get_logger(__name__)
+
+
+class Flare:
+    def __init__(self, timeout_sec: int = DEFAULT_TIMEOUT_SECONDS):
+        self.original_log_level = 0
+        self.timeout = timeout_sec
+
+    def _prepare(self, configs: List[dict]):
```
emmettbutler, line 35: This function signature couples the `Flare` class to the particular format of the "agent config" dictionary. You might consider decoupling them and providing an interface to the `Flare` class that could be used outside the context of having a well-formatted config dictionary available.

ddtrace/internal/flare.py
```diff
@@ -0,0 +1,184 @@
...
+        """
+        for c in configs:
+            # AGENT_TASK is currently being used for multiple purposes
+            # We only want to generate the tracer flare if the task_type is
+            # 'tracer_flare'
+            if type(c) != dict or c.get("task_type") != "tracer_flare":
+                continue
+            args = c.get("args", {})
+
+            self._revert_configs()
```
emmettbutler, line 84: All of this is happening inside of a loop. Is it your intention to do the revert once per `config`? This could be simplified by decoupling `Flare` from the config format.

emmettbutler (approved):
Looking good! None of my comments need to block merge. They're mostly suggestions about cleaner separation of concerns, which will become more important as you build on this foundation.

ddtrace/internal/flare.py
```diff
@@ -0,0 +1,184 @@
...
+    def __init__(self, timeout_sec: int = DEFAULT_TIMEOUT_SECONDS):
+        self.original_log_level = 0
+        self.timeout = timeout_sec
+
+    def _prepare(self, configs: List[dict]):
+        """
+        Update configurations to start sending tracer logs to a file
+        to be sent in a flare later.
+        """
+        os.makedirs(TRACER_FLARE_DIRECTORY, exist_ok=True)
```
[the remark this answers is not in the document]
brettlangdon: (this really isn't that important)

ddtrace/internal/flare.py
```diff
@@ -0,0 +1,184 @@
...
+            headers["DD-API-KEY"] = config._dd_api_key
+        return headers, body.getvalue().encode()
+
+    def _get_valid_logger_level(self, flare_log_level: int) -> int:
+        valid_original_level = 100 if self.original_log_level == 0 else self.original_log_level
+        return min(valid_original_level, flare_log_level)
+
+    def _clean_up_files(self):
+        try:
+            shutil.rmtree(TRACER_FLARE_DIRECTORY)
```
brettlangdon, line 187: should you add a `return` after this?

if we delete the directory then we'll also delete the lock file, so the next `os.remove(TRACER_FLARE_LOCK)` should always fail?

ddtrace/internal/flare.py
```diff
@@ -0,0 +1,184 @@
...
+            self.original_log_level = ddlogger.level
+
+            # Set the logger level to the more verbose between original and flare
+            # We do this valid_original_level check because if the log level is NOTSET, the value is 0
+            # which is the minimum value. In this case, we just want to use the flare level, but still
+            # retain the original state as NOTSET/0
+            valid_original_level = 100 if self.original_log_level == 0 else self.original_log_level
+            logger_level = min(valid_original_level, flare_log_level_int)
+            ddlogger.setLevel(logger_level)
+            _add_file_handler(ddlogger, flare_file_path, flare_log_level, TRACER_FLARE_FILE_HANDLER_NAME)
```
brettlangdon, line 70: if you have `_add_file_handler` return the instance of the handler, it would make removing later easier.

```python
self.log_handler = _add_file_handler(...)

...


ddlogger.removeHandler(self.log_handler)
```

you also wouldn't need `_get_handler` either ?

ddtrace/internal/flare.py
```diff
@@ -0,0 +1,184 @@
...
+        config_file = f"{TRACER_FLARE_DIRECTORY}/tracer_config_{pid}.json"
+        try:
+            with open(config_file, "w") as f:
+                tracer_configs = {
+                    "configs": config.__dict__,
+                }
+                json.dump(
+                    tracer_configs,
+                    f,
+                    default=lambda obj: obj.__repr__() if hasattr(obj, "__repr__") else obj.__dict__,
```
brettlangdon, line 134: do you have an example of what this config ends up looking like?

I am wondering if there is anything in there we definitely do not want to include, or if there is anything that gets rendered weird with the `__dict__` + `__repr__` usage.

ddtrace/internal/flare.py
```diff
@@ -0,0 +1,184 @@
...
+            log.debug("ddtrace logs will not be routed to the %s file handler anymore", TRACER_FLARE_FILE_HANDLER_NAME)
+        else:
+            log.debug("Could not find %s to remove", TRACER_FLARE_FILE_HANDLER_NAME)
+        ddlogger.setLevel(self.original_log_level)
+
+    def _generate_payload(self, params: Dict[str, str]) -> Tuple[dict, bytes]:
+        tar_stream = io.BytesIO()
+        with tarfile.open(fileobj=tar_stream, mode="w") as tar:
+            for file_name in os.listdir(TRACER_FLARE_DIRECTORY):
+                flare_file_name = f"{TRACER_FLARE_DIRECTORY}/{file_name}"
```
brettlangdon, line 156: same here re: path separator.

ddtrace/internal/flare.py
```diff
@@ -0,0 +1,184 @@
...
+        if handler_to_remove:
+            ddlogger.removeHandler(handler_to_remove)
+            log.debug("ddtrace logs will not be routed to the %s file handler anymore", TRACER_FLARE_FILE_HANDLER_NAME)
+        else:
+            log.debug("Could not find %s to remove", TRACER_FLARE_FILE_HANDLER_NAME)
+        ddlogger.setLevel(self.original_log_level)
+
+    def _generate_payload(self, params: Dict[str, str]) -> Tuple[dict, bytes]:
+        tar_stream = io.BytesIO()
+        with tarfile.open(fileobj=tar_stream, mode="w") as tar:
```
brettlangdon, line 154: how big does this usually end up being? do we have any concerns over holding this in memory?

if it is like <5mb, ignore me. if we think it could grow 20+mb, I wonder if we'll have memory usage issues?

I am probably making mountain from a mole hill, but thought I'd check your thoughts.

ddtrace/internal/flare.py
```diff
@@ -0,0 +1,194 @@
...
+            body.write(f"{value}{newline}")
+
+        body.write("--" + boundary + newline)
+        body.write(f'Content-Disposition: form-data; name="flare_file"; filename="flare.tar"{newline}')
+        body.write(f"Content-Type: application/octet-stream{newline}{newline}")
+        body.write(tar_stream.getvalue().decode() + newline)
+        body.write("--" + boundary + "--")
+        headers = {
+            "Content-Type": "multipart/form-data; boundary=" + boundary,
+            "Content-Length": str(len(body.getvalue().encode())),
```
brettlangdon, line 175: TIL: `str(body.getbuffer().nbytes)` should do the same thing.

And that way you don't have to encode the body twice.

the alternative could be:

```python
body_value = body.getvalue().encode()
headers = {
    "Content-Type": "multipart/form-data; boundary=" + boundary,
    "Content-Length": str(len(body_value)),
}

return headers, body_value
```

riotfile.py
erikayasuda: It's actually not set as a default, so I had to specify for this test suite. Would it make sense to set it as a default?

ddtrace/internal/flare.py
```diff
@@ -0,0 +1,184 @@
...
+
+            # Validate the flare log level
+            flare_log_level = c.get("config", {}).get("log_level").upper()
+            flare_log_level_int = logging.getLevelName(flare_log_level)
+            if type(flare_log_level_int) != int:
+                raise TypeError("Invalid log level provided: %s", flare_log_level_int)
+
+            ddlogger = get_logger("ddtrace")
+            pid = os.getpid()
+            flare_file_path = f"{TRACER_FLARE_DIRECTORY}/tracer_python_{pid}.log"
```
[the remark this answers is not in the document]
erikayasuda: Thanks for catching that, I'll use `pathlib` to be safe!

ddtrace/internal/flare.py
erikayasuda: Ah, looks like I can do that with `io.BytesIO` but not `io.StringIO`. `io.StringIO` only has `getvalue()`, which doesn't have `nbytes`. Let me swap it over to `io.BytesIO`

ddtrace/internal/flare.py
erikayasuda: We do limit individual files to 15mb by default, but you're right we don't have a bound on how big the directory can become. One idea off the top of my head is we check the size of the directory in its current state, and if it's above some threshold we set, we don't add more flare files to it. This would be further leaning on the "best effort" nature of the tracer flare, but would be pretty easy to implement.

ddtrace/internal/flare.py
erikayasuda: Yeah I do have an example locally. Would it be beneficial to have a snippet as a method comment?

ddtrace/internal/flare.py
erikayasuda: Ah, good catch. No this should just happen once. I had to add a loop because each RC product _could_ have multiple configs included, and only one of them is the tracer_flare request. Either way, we already have the expectation that we can only process one tracer flare request at a time. I'll add a return statement in the `finally` block so that we just process one and exit.

erikayasuda: I guess the one thing with having the return is that in the very odd case that there are two tracer flare configs, we wouldn't be logging the warning about skipping the second one, but I don't think it's a huge deal.

ddtrace/internal/flare.py
brettlangdon: I'd probably prefer it in a test to a comment

ddtrace/internal/flare.py
brettlangdon: when you iterate over the files, start with the configs, then add as many log files up until you hit the limit?

but I am wondering if there is a way to include that log files were skipped due to size.... or maybe you take them pid by pid, so you always have config + log for a given pid... 🤔

ddtrace/internal/flare.py
erikayasuda: I agree, I'll make a follow up PR to implement this 😄

ddtrace/internal/flare.py
erikayasuda: Ooh, nice. I'll do this instead.

erikayasuda: Actually, it's kinda handy for tests, so I'll just move it to the test file 😅

ddtrace/internal/flare.py
erikayasuda: Yeah I think I have to do it in the pairs. If anything, the log files will already be there anyways when I add the file handler to the logger... So the reverse would be more doable

PR #8969, opened 2024-04-12
chore(tracer): add dataclass for tracer flare RC configs

erikayasuda:
## Overview
Adds a `FlareSendRequest` dataclass that get ingested by the `Flare.send()` method, ensuring that we always get a valid RC config when this method is called.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


ddtrace/internal/flare.py
```diff
@@ -29,13 +27,28 @@
...
+            raise TypeError("Invalid log level provided: %s", log_level)
+        self.log_level = log_level
+
+
+class FlareSendRequest:
+    def __init__(self, case_id: str, hostname: str, email: str):
+        self.case_id = case_id
+        self.source = "tracer_python"
+        self.hostname = hostname
+        self.email = email
```
brettlangdon, line 42: you can probably do this as an actual dataclass

```suggestion
@dataclasses.dataclass
class FlareSendRequest:
    case_id: str
    hostname: str
    email: str
    source: str = "tracer_python"
```

Same for the one above, you could use a [__post_init__](https://docs.python.org/3/library/dataclasses.html#dataclasses.__post_init__) to do the checking/validation of the level.

ddtrace/internal/flare.py
```diff
@@ -29,13 +27,28 @@
 log = get_logger(__name__)
 
 
+class FlarePrepRequest:
+    def __init__(self, log_level: str):
+        if logging.getLevelName(log_level) == f"Level {log_level}":
+            raise TypeError("Invalid log level provided: %s", log_level)
+        self.log_level = log_level
```
brettlangdon, line 34: do you want to just store this as the integer rather than the name?

you could do something like this:

```
log_level: int

def __init__(self, log_level_name: str):
    log_level: int | str = logging.getLevelName(log_level_name)
    if not isinstance(log_level, int):
        raise ValueError(f"Unknown log level: {log_level_name}")
    self.log_level = log_level
```

brettlangdon (commented):
nice, I like this change, a few minor nits/suggestions

ddtrace/internal/flare.py
```diff
@@ -29,13 +28,33 @@
...
+@dataclasses.dataclass
+class FlarePrepRequest:
+    log_level: int
+
+    def __init__(self, log_level_name: str):
+        log_level: int | str = logging.getLevelName(log_level_name)
+        if type(log_level) == int:
+            self.log_level = log_level
+        else:
+            raise TypeError("Invalid log level provided: %s", log_level)
```
brettlangdon, line 40: Rather than overwriting the `__init__` it could be better to add a classmethod that creates a proper `FlarePrepRequest` from a log level name instead. Otherwise the typing/autocomplete for the dataclass might get weird? e.g. it only has an int but only takes a str (maybe just my own biased opinion?)


```suggestion
    @classmethod
    def from_log_level_name(cls, log_level_name: str) -> 'FlarePrepRequest':
        log_level: int | str = logging.getLevelName(log_level_name)
        if type(log_level) == int:
            return cls(log_level)
        else:
            raise TypeError("Invalid log level provided: %s", log_level)
```

ddtrace/internal/flare.py
```diff
@@ -47,90 +66,66 @@ def prepare(self, configs: List[dict]):
...
+        ddlogger = get_logger("ddtrace")
+        pid = os.getpid()
+        flare_file_path = TRACER_FLARE_DIRECTORY / pathlib.Path(f"tracer_python_{pid}.log")
+        self.original_log_level = ddlogger.level
+
+        # Set the logger level to the more verbose between original and flare
+        # We do this valid_original_level check because if the log level is NOTSET, the value is 0
+        # which is the minimum value. In this case, we just want to use the flare level, but still
+        # retain the original state as NOTSET/0
+        valid_original_level = 100 if self.original_log_level == 0 else self.original_log_level
```
brettlangdon, line 81: what is `100`?

It might be better to use `logging.NOTSET` / `logging.CRITICAL` / etc instead of the numeric values so it is easier to read.

ddtrace/internal/flare.py
```diff
@@ -47,90 +66,66 @@ def prepare(self, configs: List[dict]):
...
+            except Exception as e:
+                log.error("Failed to create %s file", lock_path)
+                raise e
+            try:
+                client = get_connection(config._trace_agent_url, timeout=self.timeout)
+                headers, body = self._generate_payload(flare_send_req.__dict__)
+                client.request("POST", TRACER_FLARE_ENDPOINT, body, headers)
+                response = client.getresponse()
+                if response.status == 200:
+                    log.info("Successfully sent the flare")
```
brettlangdon, line 113: if we keep this as an info, is there more we can give users? like what zendesk ticket or flare id or something?

especially since this might be triggered by the UI, letting someone who might not have known a flare was sent how they can follow-up, could be good.

this could be a follow-up change/improvement, doesn't need to happen here.

ddtrace/internal/flare.py
```diff
@@ -47,90 +66,66 @@ def prepare(self, configs: List[dict]):
...
+                else:
+                    log.error(
+                        "Upload failed with %s status code:(%s) %s",
```
brettlangdon, line 116: same here, this doesn't make mention of a flare at all.

brettlangdon (commented):
minor nits and suggestions that could be handled as follow-ups.

ddtrace/internal/flare.py
erikayasuda: Good call, I'll fix these

ddtrace/internal/flare.py
erikayasuda: Hmm I think I'm a bit confused about the typing concern? Are you saying we won't know whether we have a string or int log level to pass into the `FlarePrepRequest` init? I don't think that should be a concern given the structure of the `AGENT_CONFIG` product (for now).

erikayasuda: Chatted offline. Decided this was probably unnecessary, and we can just pass in the str form and let the class method do the conversions.

erikayasuda:
@brettlangdon I just took a look at what's generating that codecov output saying I have 0% coverage on my changes. It looks like it's generated from running `diff-cover --compare-branch $(git rev-parse --abbrev-ref origin/HEAD) coverage.xml || true`, which returns 100% when I run this locally. Trying to figure out why this isn't what's showing up in CircleCI...

ddtrace/internal/flare.py
```diff
@@ -50,11 +38,11 @@ class FlareSendRequest:
 
 class Flare:
     def __init__(self, timeout_sec: int = DEFAULT_TIMEOUT_SECONDS):
-        self.original_log_level = 0  # NOTSET
+        self.original_log_level = logging.NOTSET
         self.timeout = timeout_sec
         self.file_handler: Optional[RotatingFileHandler] = None
 
-    def prepare(self, flare_prep_req: FlarePrepRequest):
+    def prepare(self, log_level: str):
```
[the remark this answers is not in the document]
erikayasuda: Oops I missed this. I think either way I would have to do some validation, so I don't have strong feelings for either or. Unless there's a way for me to type hint this as a log level int/str?

ddtrace/internal/flare.py
```diff
@@ -29,13 +28,21 @@
...
+
 class Flare:
     def __init__(self, timeout_sec: int = DEFAULT_TIMEOUT_SECONDS):
-        self.original_log_level = 0  # NOTSET
+        self.original_log_level = logging.NOTSET
         self.timeout = timeout_sec
         self.file_handler: Optional[RotatingFileHandler] = None
 
-    def prepare(self, configs: List[dict]):
+    def prepare(self, log_level: str):
```
brettlangdon, line 45: I forgot, we could help with typing here too:

```python
LogLevelString = typing.Literal["DEBUG"] | typing.Literal["WARNING"] | ...

def prepare(self, log_level: LogLevelString):
    pass
```


this is just a 💡, you don't need to incorporate. since this is a private api, we can get away with the slightly looser typing.

brettlangdon (approved):

Issue #8281, opened 2024-02-04
gevent could not work with ddtrace in pytest


zdyj3170101136:
code 
```
import ddtrace.auto
import gevent.monkey
gevent.monkey.patch_all()

from ddtrace.profiling import Profiler
import os
import time
prof = Profiler(url="http://127.0.0.1:7070")
prof.start()
time.sleep(3)
print(os.getpid())
prof.stop()
print("wait for stop")
```

work great when use pprof:
```
DD_PROFILING_OUTPUT_PPROF=pprof python3.7 test.py
20142
wait for stop
```

deadlock when use pytest:
```
DD_PROFILING_OUTPUT_PPROF=pprof python3.7 -m pytest -s abc.py
========================================================================= test session starts =========================================================================
platform linux -- Python 3.7.7, pytest-7.1.3, pluggy-1.2.0
rootdir: /root
plugins: ddtrace-1.20.16
collecting ... 20216

^CKeyboardInterrupt
2024-02-04T09:36:35Z
collected 0 items / 1 error

=============================================================================== ERRORS ================================================================================
_______________________________________________________________________ ERROR collecting abc.py _______________________________________________________________________
<frozen importlib._bootstrap>:983: in _find_and_load
    ???
<frozen importlib._bootstrap>:967: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:677: in _load_unlocked
    ???
/usr/lib/python3.7/site-packages/ddtrace/internal/module.py:224: in _exec_module
    self.loader.exec_module(module)
/usr/lib/python3.7/site-packages/_pytest/assertion/rewrite.py:168: in exec_module
    exec(co, module.__dict__)
abc.py:12: in <module>
    prof.stop()
/usr/lib/python3.7/site-packages/ddtrace/profiling/profiler.py:78: in stop
    self._profiler.stop(flush)
/usr/lib/python3.7/site-packages/ddtrace/internal/service.py:85: in stop
    self._stop_service(*args, **kwargs)
/usr/lib/python3.7/site-packages/ddtrace/profiling/profiler.py:329: in _stop_service
    self._scheduler.join()
/usr/lib/python3.7/site-packages/ddtrace/internal/periodic.py:148: in join
    self._worker.join(timeout)
/usr/lib/python3.7/threading.py:1044: in join
    self._wait_for_tstate_lock()
/usr/lib/python3.7/threading.py:1060: in _wait_for_tstate_lock
[...]

Issue #4845, opened 2022-12-29
Deadlock with `pytest`+`ddtrace` in py 3.11 that `-p no:ddtrace` does not solve


olivierlefloch:
### Summary of problem

When adding Python 3.11 support to an existing codebase that uses `ddtrace`, we encounter deadlocks at the end of the test suite (after `pytest` reports its results) in CI. This does not occur on previous versions of Python.

Furthermore, disabling the `ddtrace` plugin using `-p no:ddtrace` does not resolve the issue.

Specifying `DD_TRACE_ENABLED=false pytest …` does work around the issue, however.

### Which version of dd-trace-py are you using?

`1.6.3`

### Which version of pip are you using?

`22.3.1`

### Which libraries and their versions are you using?

```
datadog==0.44.0
[…]
ddtrace==1.6.3 
[…]
gevent==22.10.2
[…]
pytest==7.2.0
```

### How can we reproduce your problem?

I have not (yet?) put together a minimal reproduction case as there seem to potentially be several issues at play here ; happy to help drill down as appropriate.

### What is the result that you get?

Github Actions times out after our self imposted timeout.

### What is the result that you expected?

1. Test suite does not deadlock even with the `ddtrace` plugin enabled, on Python 3.11,
2. Specifying `-p no:ddtrace` fully disables the plugin, including starting a global tracer.

### Related issues

- https://github.com/DataDog/dd-trace-py/pull/3674
- https://github.com/DataDog/dd-trace-py/issues/4149
- https://github.com/DataDog/dd-trace-py/issues/4179

mabdinur:
Hey @olivierlefloch, 

Thanks for brining this issue to our attention 😄

> Specifying -p no:ddtrace fully disables the plugin, including starting a global tracer.

I believe `no:ddtrace` does not disable the global tracer, it should only stop the ddtrace pytest plugin from being configured.  When you set `no:ddtrace` do you still see test runs in the Datadog CI product?  


> Test suite does not deadlock even with the ddtrace plugin enabled, on Python 3.11,

[...]

PR #8999, opened 2024-04-16
chore(ci_visibility): only import ddtrace when pytest ddtrace module is enabled

romainkomorndatadog:
This defers importing ddtrace in the `pytest`, `pytest-bdd`, and `pytest-benchmark` plugins until after it's been determined that the plugin is enabled with (eg: by running `pytest --ddtrace`).

The fixtures that use `ddtrace` defer importing until the feature is actually requested due to the fact that fixtures imported later (eg: after plugins have loaded) are not usable.

Even though this doesn't actually address the issues reported in https://github.com/DataDog/dd-trace-py/issues/8281 or https://github.com/DataDog/dd-trace-py/issues/4845 (due to the side effects of `ddtrace/__init__.py`), this change is still helpful in other ways (in particular since I'll be introducing a new version of the `pytest` plugin shortly).

Fixtures were confirmed to still be working by:
* using `--ddtrace-patch-all`
* adding tags to spans using the `ddspan` fixture
* seeing that the `ddtracer` fixture returns the tracer object.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


ddtrace/contrib/pytest/plugin.py
```diff
@@ -11,398 +11,22 @@
...
-from ddtrace.internal.ci_visibility.utils import _generate_fully_qualified_test_name
-from ddtrace.internal.ci_visibility.utils import get_relative_or_absolute_path_for_path
-from ddtrace.internal.ci_visibility.utils import take_over_logger_stream_handler
-from ddtrace.internal.constants import COMPONENT
-from ddtrace.internal.logger import get_logger
-
 
+DDTRACE_HELP_MSG = "HELP DDTRACE"
+DDTRACE_INCLUDE_CLASS_HELP_MSG = "HELP INCLUDE DDTRACE"
+NO_DDTRACE_HELP_MSG = "NO HELP"
```
juan-fernandez, line 21: :question: what are these used for?

ddtrace/contrib/pytest/_plugin_v1.py
```diff
@@ -0,0 +1,861 @@
...
+        return pytest_module_item.nodeid
+
+    @staticmethod
+    @pytest.hookimpl(trylast=True)
+    def pytest_ddtrace_get_item_test_name(item):
+        """Extract name from item, prepending class if desired"""
+        if hasattr(item, "cls") and item.cls:
+            if item.config.getoption("ddtrace-include-class-name") or item.config.getini("ddtrace-include-class-name"):
+                return "%s.%s" % (item.cls.__name__, item.name)
+        return item.name
```
juan-fernandez, line 861: :question: is this file mostly what we had in `ddtrace/contrib/pytest/plugin.py`? Or do we have any other change?

juan-fernandez (approved):

ddtrace/contrib/pytest/_plugin_v1.py
romainkomorndatadog: It's almost exactly the same thing.

ddtrace/contrib/pytest/plugin.py
romainkomorndatadog: OMG. Thank you for catching this. 😱 

They used to be imported from another module and I temporary declared them locally to avoid the import and make sure it worked, but never gave them their original values.

ddtrace/contrib/pytest/_plugin_v1.py
romainkomorndatadog: The practical difference is that this defines the plugin as a class, and moves the module-level plugin hook functions to be static methods on the class, but aside from that, there should be no functionality difference.

emmettbutler (approved):

PR #9016, opened 2024-04-17
fix(ci_visibility): use the default tracer when enabling CI Visibility

romainkomorndatadog:
This switches back to using the default tracer when using CI Visiblity, and fixes an issue where non-test traces were getting lost due to:
* going through a non-existent local agent when using `DD_CIVISIBILITY_AGENTLESS_ENABLED=true`
* being in a different tracer object which broke trace IDs

The historical change behind switching to a separate tracer was so that we would be able to use CI Visibility in this repo, which was causing CI Visibility traces to get mixed up with other traces. Since that would remain a problem the behavior to use a separate tracer with a separate context provider, via the use of `_DD_CIVISIBILITY_USE_CI_CONTEXT_PROVIDER`, is maintained.

This also tweaks the `TraceCiVisibilityFilter` to no longer force the `service` of spans to match that of the original under which the tracer was started, which allows for test-level traces with non-test spans (eg: with distributed tracing or `--ddtrace-patch-all` enabled) to properly reflect the right services.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


romainkomorndatadog:
A race condition remains before `CIVisibility.enable()` get called and [reconfigures the writer](https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/internal/ci_visibility/recorder.py#L228) where the tracer could end up attempting to send traces to the agent, but the context in which `CIVisibility` is used makes it pretty unlikely that that would matter.

The other option would be to move agentless behavior to the default tracer, and I'm not sure that is worth the effort.

juan-fernandez (approved):

emmettbutler (approved):

romainkomorndatadog:
Converting this back to draft and dismissing reviews since it looks like it causes a performance regression.

romainkomorndatadog:
Getting back to "for review", the performance impact is expected: roughly 50% overhead is what we see in tracing overhead, but tracing is typically disabled for our `pytest` plugin unless `--ddtrace-patch-all` is used.

PR #9023, opened 2024-04-17
chore(botocore): refactor the bedrock integration

emmettbutler:
This pull request applies some basic refactors to the Bedrock integration that use the Core API, improving the separation of concerns between tracing and this integration.

Existing tests cover the changed functionality.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


ddtrace/_trace/trace_handlers.py
```diff
@@ -666,6 +733,7 @@ def listen():
         "botocore.patched_kinesis_api_call",
         "botocore.patched_sqs_api_call",
         "botocore.patched_stepfunctions_api_call",
+        "botocore.patched_bedrock_api_call",
```
erikayasuda, line 736: General Q... So our ideal end state for all integrations is for each one to be registered here instead of directly accessing the `datadog_integration` product? And then all of the span management would happen in the core as well?

erikayasuda (approved):
One general Q, otherwise LGTM!

ddtrace/_trace/trace_handlers.py
emmettbutler: The ideal end state I have in mind with this change is that given a module in contrib, it contains only instrumentation and event dispatches. The tracing code that reacts to those events could live in `ddtrace/_trace`. Does that answer your question?

emmettbutler:
@Yun-Kim your review is the only thing blocking this pull request. Please take a look when you can!

ddtrace/contrib/botocore/services/bedrock.py
```diff
@@ -292,61 +262,63 @@ def _extract_streamed_response_metadata(span: Span, streamed_body: List[Dict[str
...
-        service=schematize_service_name("{}.{}".format(pin.service, endpoint_name)),
-        resource=operation,
+    model_provider, model_name = params.get("modelId").split(".")
+    with core.context_with_data(
+        "botocore.patched_bedrock_api_call",
+        pin=pin,
+        span_name=function_vars.get("trace_operation"),
+        service=schematize_service_name("{}.{}".format(pin.service, function_vars.get("endpoint_name"))),
+        resource=function_vars.get("operation"),
         span_type=SpanTypes.LLM,
```
Yun-Kim, line 309: Just curious, how does this ensure the span.span_type gets set? I thought `context_with_data(..., **kwargs) only sets kwargs as part of the data dict

ddtrace/_trace/trace_handlers.py
```diff
@@ -610,6 +611,68 @@ def _on_botocore_update_messages(ctx, span, _, trace_data, __, message=None):
...
+    metadata: Dict[str, Any],
+    body: Dict[str, List[Dict]],
+    should_set_choice_ids: bool,
+) -> None:
+    text = formatted_response["text"]
+    dd_span = ctx[ctx["call_key"]]
+    if should_set_choice_ids:
+        for i in range(len(text)):
+            dd_span.set_tag_str("bedrock.response.choices.{}.id".format(i), str(body["generations"][i]["id"]))
+    datadog_integration = ctx["bedrock_integration"]
```
Yun-Kim, line 658:
```suggestion
    integration = ctx["bedrock_integration"]
```
Just to be consistent with the other handlers

ddtrace/_trace/trace_handlers.py
```diff
@@ -610,6 +611,68 @@ def _on_botocore_update_messages(ctx, span, _, trace_data, __, message=None):
...
+
+def _on_botocore_bedrock_process_response(
+    ctx: core.ExecutionContext,
+    formatted_response: Dict[str, Any],
+    metadata: Dict[str, Any],
+    body: Dict[str, List[Dict]],
+    should_set_choice_ids: bool,
+) -> None:
+    text = formatted_response["text"]
+    dd_span = ctx[ctx["call_key"]]
```
Yun-Kim, line 654:
```suggestion
    span = ctx[ctx["call_key"]]
```
Just to be consistent with the other handlers

ddtrace/contrib/botocore/services/bedrock.py
```diff
@@ -240,23 +214,18 @@ def _extract_streamed_response(span: Span, streamed_body: List[Dict[str, Any]])
...
                     text = "".join([chunk["text"] for chunk in streamed_body[:-1]])
                     finish_reason = streamed_body[-1]["finish_reason"]
             else:
                 text = [chunk["text"] for chunk in streamed_body[0]["generations"]]
                 finish_reason = [chunk["finish_reason"] for chunk in streamed_body[0]["generations"]]
-                for i in range(len(text)):
-                    span.set_tag_str(
-                        "bedrock.response.choices.{}.id".format(i),
-                        str(streamed_body[0]["generations"][i].get("id", None)),
-                    )
```
Yun-Kim, line 259: Just to confirm, was this (streamed case) and the non-streamed case combined to set `body` and handle both in the same place in `botocore.bedrock.process_response`?

ddtrace/contrib/botocore/services/bedrock.py
```diff
@@ -81,44 +85,22 @@ def __iter__(self):
...
-            self._datadog_span.finish()
+            metadata = _extract_streamed_response_metadata(self._execution_ctx, self._body)
+            formatted_response = _extract_streamed_response(self._execution_ctx, self._body)
+            core.dispatch(
+                "botocore.bedrock.process_response",
+                [
+                    self._execution_ctx,
+                    formatted_response,
+                    metadata,
+                    self._body,
```
Yun-Kim, line 96: We're passing in `self._body` which is a list of dictionaries so I'm worried this will break `_on_botocore_bedrock_process_response()` which expects a dictionary.

Yun-Kim (commented):
Looks mostly great, just a couple clarifying questions. Thanks!

ddtrace/contrib/botocore/services/bedrock.py
emmettbutler: @Yun-Kim The `span_type` gets pulled out of the context object [here](https://github.com/DataDog/dd-trace-py/blob/3ecf22943c4f9f0ab74b69242297b9ac3090129d/ddtrace/_trace/trace_handlers.py#L95) and is used as a kwarg to `start_span`[here](https://github.com/DataDog/dd-trace-py/blob/3ecf22943c4f9f0ab74b69242297b9ac3090129d/ddtrace/_trace/trace_handlers.py#L109). That `_start_span` handler is entered because there's a listener registered [here](https://github.com/DataDog/dd-trace-py/blob/3ecf22943c4f9f0ab74b69242297b9ac3090129d/ddtrace/_trace/trace_handlers.py#L674) for an event that gets fired on context creation [here](https://github.com/DataDog/dd-trace-py/blob/3ecf22943c4f9f0ab74b69242297b9ac3090129d/ddtrace/internal/core/__init__.py#L181).

ddtrace/contrib/botocore/services/bedrock.py
emmettbutler: Yup, I was able to deduplicate some similar bits of code by the implementation of `_on_botocore_bedrock_process_response`.

ddtrace/contrib/botocore/services/bedrock.py
Yun-Kim: Just an FYI, I double checked the bedrock API and we only care about tagging choice ID (only available when) the provider is COHERE and it is not streamed, so we should be fine
i.e. this line should always evaluate to false since we're in the `__iter__` stream handler.

ddtrace/contrib/botocore/services/bedrock.py
```diff
@@ -292,61 +262,63 @@ def _extract_streamed_response_metadata(span: Span, streamed_body: List[Dict[str
...
-    except Exception:
-        _handle_exception(bedrock_span, integration, prompt, sys.exc_info())
-        raise
+        call_key="instrumented_bedrock_call",
+        call_trace=True,
+        bedrock_integration=function_vars.get("integration"),
+        params=params,
+        model_provider=model_provider,
+        model_name=model_name,
+    ) as ctx:
```
Yun-Kim, line 316: For LLMObs spans, there are some things we like to do on span startup ([right now not much](https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/llmobs/_integrations/base.py#L121-L122), but in the near future we're planning on adding a couple temporary tags and processing). 

I was planning on implementing this by starting the span using `bedrock_integration.trace(...)` but obviously this won't be feasible if we move over to using `core.context_with_data()`. Do you know if we will be able to extract the same LLMObs-specific logic using the new core dispatch/handler framework?

Yun-Kim (approved):
LGTM, just one last question!

brettlangdon (approved):

ddtrace/contrib/botocore/services/bedrock.py
emmettbutler: Yes, absolutely you can. Sounds like the approach I might take would be to have the LLMObs code listen to the context-started event fired from `context_with_data`, and start its own span in the event handler. I'd be happy to do a deeper review and/or pair with you on your code if you decide to take a similar approach.

PR #9041, opened 2024-04-18
chore(opentelemetry): avoid unnecessary use of core API in opentelemetry module

### LINKED ISSUE #9029: DDTraceDeprecationWarning: The 'span' keyword argument on ExecutionContext methods is deprecated
### Summary of problem

Using opentelemetry-api, logs spammed with deprecation warnings about the `span` keyword arg. We're not using that anywhere in our code; I think it's coming through https://github.com/DataDog/dd-trace-py/blob/b1dab58296355786c258bcd1b1c489a6f21f33aa/ddtrace/opentelemetry/_span.py#L98 and others in that module.

2.8.0

opentelemetry_api-1.24.0

<details>
  
</details>

We're invoking with:
```
ENV DD_TRACE_OTEL_ENABLED=True
CMD [ "ddtrace-run", "python", "app.py" ]
```

Logs getting spammed with warnings:
```
/function/.venv/lib/python3.11/site-packages/ddtrace/internal/core/__init__.py:153: DDTraceDeprecationWarning: The 'span' keyword argument on ExecutionContext methods is deprecated and will be removed in a future version.: Please store contextual data on the ExecutionContext object using other kwargs and/or set_item()
```

emmettbutler:
This change resolves #9029 by removing the use of the Core API from the opentelemetry module. This is based on the insight that `core.set_item` was being used here to set a span tag on a span to which the existing class holds a reference, and that the reference lives exactly as long as the class does without being reassigned. Therefore, storing a boolean flag in that span's tags is equivalent to storing it in the classes' attribute list. This means that the core's context bookkeeping is nothing but overhead in this case and can be removed.

Changed functionality is covered by existing tests.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


ddtrace/opentelemetry/_span.py
```diff
@@ -77,9 +73,9 @@ def __init__(
 
         self._ddspan = datadog_span
         if record_exception is not None:
-            self._record_exception = record_exception
+            self._record_exception = record_exception if record_exception is not None else True
```
brettlangdon, line 76: @mabdinur 

Do these properties exist on `OtelSpan`?

Are we supposed to be using these private attributes?

brettlangdon (commented):
I am rather surprised or wary that this doesn't change the behavior...... were these attributes not being used?

ddtrace/opentelemetry/_span.py
mabdinur: Yes these fields should be private. The OTEL API does not define any fields for the [Span](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-api/src/opentelemetry/trace/span.py#L57) object. These fields are defined in the [span implementation](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-sdk/src/opentelemetry/sdk/trace/__init__.py#L774) in the OTEL SDK.

mabdinur: In the Datadog API error messages and statuses are always recorded on spans. In the OTEL API this functionality is configurable.

brettlangdon (approved):

ddtrace/opentelemetry/_span.py
```diff
@@ -77,9 +73,9 @@ def __init__(
...
         if set_status_on_exception is not None:
-            self._set_status_on_exception = set_status_on_exception
+            self._set_status_on_exception = set_status_on_exception if set_status_on_exception is not None else True
```
mabdinur, line 78: The otelspan object should be stateless. All fields set on the OtelSpan should reference a value on the underlying spans (ie `_set_status_on_exception` should set a tag/context field on `OtelSpan._ddspan`).  This way a Datadog Span can always be converted to an OTEL Span without losing information. This property ensures the OpenTelemetry and Datadog APIs interoperable. [Here](https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/opentelemetry/_context.py#L62) is one place where we convert Datadog Spans to Otel Spans. 

I will add a regression test to capture this scenario.

emmettbutler: What I'm understanding from this is that the setting of a tag on the DD span is a functional requirement. If that's the case, I'll change the code to do so without using the core API, since using a span as a "core context" is an implementation detail of the core, and a deprecated one at that.

PR #9086, opened 2024-04-24
feat(botocore): tracing support for bedrock embeddings

Yun-Kim:
This PR adds tracing support for AWS bedrock-runtime embeddings operations, which were added recently (after tracing support for llm/chat models was added).

Replaces #9075 due to large changes from #9023 being merged.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


ddtrace/_trace/trace_handlers.py
```diff
@@ -632,8 +632,9 @@ def _on_botocore_patched_bedrock_api_call_exception(ctx, exc_info):
     span = ctx[ctx["call_key"]]
     span.set_exc_info(*exc_info)
     prompt = ctx["prompt"]
+    model_name = ctx["model_name"]
```
lievan, line 635: should we use `.get` to be more resistent to key errors here?

ddtrace/contrib/botocore/services/bedrock.py
```diff
@@ -42,19 +42,16 @@ def read(self, amt=None):
...
                 core.dispatch(
                     "botocore.bedrock.process_response",
-                    [
-                        self._execution_ctx,
-                        formatted_response,
-                        None,
-                        self._body[0],
-                        self._execution_ctx["model_provider"] == _COHERE,
-                    ],
+                    [self._execution_ctx, formatted_response, None, self._body[0], should_set_choice_ids],
```
lievan, line 50: is body guaranteed to have length >= 1?

ddtrace/_trace/trace_handlers.py
```diff
@@ -632,8 +632,9 @@ def _on_botocore_patched_bedrock_api_call_exception(ctx, exc_info):
...
     integration = ctx["bedrock_integration"]
-    if integration.is_pc_sampled_llmobs(span):
+    if integration.is_pc_sampled_llmobs(span) and "embed" not in model_name:
```
lievan, line 637: Is "embed" a string we set in the model name? Or does bedrock have "embed" in all their embedding models?

I wonder if there's a better way to check if something is an embedding model, this logic seems brittle

Yun-Kim: Unfortunately Bedrock uses the same `invoke_models` API method to call both LLM and embeddings models. The only pattern I could find was the model IDs of the embedding models all contain `embed.`. ([ref](https://docs.aws.amazon.com/bedrock/latest/userguide/model-ids.html#model-ids-arns)). I can't see any other way to check if an embedding model is being invoked.

ddtrace/_trace/trace_handlers.py
Yun-Kim: ctx is a ddtrace internal object which extract model_name from the bedrock request, so we shouldn't worry about this key being contained or not.

lievan (approved):

ddtrace/_trace/trace_handlers.py
lievan: ok, perhaps we should maintain a list of model name prefixes that we explicitly support, and check if model name adheres to any of those prefixes.

that method is a little tedious but it allows us to be clear in the documentation on what models are supported for tracing -- the contract of "if you use a model with any of these prefixes, it will be traced" sounds a bit better then "if you use a model without the string 'embed' in it" 

this is probably not worth it in this case since, but it might be a good pattern to have long term with all our integrations given how fast new models are coming out (e.g. an image model comes out with a different i/o structure that is incompatible with our tracing)

erikayasuda (approved):

PR #9087, opened 2024-04-24
chore(botocore): abstract away propagation header extraction code

emmettbutler:
This change adds a layer of abstraction between the botocore integration and the extraction of distributed tracing information from request data by using the Core API, increasing the separation of concerns between instrumentation and products.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


ddtrace/contrib/botocore/services/sqs.py
```diff
@@ -161,17 +161,23 @@ def patched_sqs_api_call(original_func, instance, args, kwargs, function_vars):
...
-                    [ctx, e.response, botocore.exceptions.ClientError, set_response_metadata_tags],
+                    [
+                        ctx,
+                        e.response,
+                        botocore.exceptions.ClientError,
+                        config.botocore.operations[ctx[ctx["call_key"]].resource].is_error_code,
+                    ],
                 )
                 raise
+        parent_ctx.end()
```
brettlangdon, line 180: do we need this in a finally?

or... outside of the `if should_instrument` block? would it cause us to have a leaky/unclosed context?

brettlangdon: same with the one above?

what happens if we never end an `ExecutionContext`?

ddtrace/_trace/trace_handlers.py
```diff
@@ -676,6 +680,31 @@ def _on_botocore_bedrock_process_response(
...
+        if propagate:
+            ctx.set_safe("distributed_context", extract_DD_context_from_messages(result["Messages"], message_parser))
+
+
+def _on_botocore_kinesis_getrecords_post(
+    ctx: core.ExecutionContext,
+    _,
+    __,
+    ___,
+    ____,
```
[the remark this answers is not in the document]
emmettbutler: Nah, you're not allowed to have multiple parameters with the same name. I agree it's kind of weird.

ddtrace/contrib/botocore/services/sqs.py
emmettbutler: good catch. I think when I wrote it I was telling myself that it would get closed when it left scope, but thinking about it again I'm not so sure. Better to close explicitly with a `finally`.

ddtrace/contrib/botocore/services/sqs.py
```diff
@@ -133,52 +133,54 @@ def patched_sqs_api_call(original_func, instance, args, kwargs, function_vars):
...
-                    "botocore.patched_sqs_api_call.exception",
-                    [
-                        ctx,
-                        e.response,
-                        botocore.exceptions.ClientError,
-                        config.botocore.operations[ctx[ctx["call_key"]].resource].is_error_code,
-                    ],
-                )
-                raise
+    try:
```
brettlangdon, line 136: Should we just be doing `with parent_ctx:` ?

Or maybe even do that around the whole function?

```python
def wrap_with_parent_ctx(name):
    def _outer(func):
        def _inner(*args, **kwargs):
            with core.ExecutionContext(name) as parent_ctx:
                return func(parent_ctx, *args, **kwargs)
        return _inner
    return _outer


@wrap_with_parent_ctx("botocore.patched_sqs_api_call.propagated")
def patched_sqs_api_call(parent_ctx, original_func, instance, args, kwargs, function_vars):
    ...
```


I don't _love_ this suggestion, just trying to think about how do we get the safety without adding to the code complexity (# levels of indentation/branches/etc)

brettlangdon (approved):

PR #9097, opened 2024-04-25
chore(internal): add dispatch_async function to core API

emmettbutler:
This change adds `core.dispatch_async`, which does the same thing as `core.dispatch` but is callable from `async def` functions. This is necessary to support asynchronous integrations like Redis in #9096.

## Checklist

## Reviewer Checklist

- [ ] Title is accurate


brettlangdon (changes requested):
We were able to remove the need for this change in the Redis PR.

I'd like to try and avoid adding an async dispatch if we can. Since the concept is meant to be synchronously taking action based on a given event, making the listeners/handlers async would break that implicit contract (I think that implicit contract is necessary for the different products?)

Issue #9135, opened 2024-04-30
ddtrace breaks Bedrock Cohere Command R models 


tstadel:
### Summary of problem
When invoking requests to Bedrock Cohere Command R models, ddtrace breaks the system by trying to iterate over the `generations` key of the bedrock response. For Command R models however the response key has changed to `text` and doesn't return an iteration but the response string.

### Which version of dd-trace-py are you using?
2.8.1
Looked into main and the issues is here when iterating over `None`: https://github.com/DataDog/dd-trace-py/blob/f1beaaefc2e627d5a25285df568b2c1c9242457f/ddtrace/contrib/botocore/services/bedrock.py#L183

### Which version of pip are you using?
24.0

### Which libraries and their versions are you using?
boto3==1.34.84
boto3-extensions==0.23.0
botocore==1.34.94

### How can we reproduce your problem?
Run code under https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-cohere-command-r-plus.html with ddtrace

### What is the result that you get?
TypeError: 'NoneType' object is not iterable

```Traceback (most recent call last):
  File "/opt/venv/lib/python3.10/site-packages/ddtrace/contrib/botocore/services/bedrock.py", line 56, in read
    formatted_response = _extract_response(self._datadog_span, self._body[0])
  File "/opt/venv/lib/python3.10/site-packages/ddtrace/contrib/botocore/services/bedrock.py", line 197, in _extract_response
    text = [generation["text"] for generation in body.get("generations")]
TypeError: 'NoneType' object is not iterable
```

### What is the result that you expected?
Tracing works and ddtrace doesn't break the code.

PR #9137, opened 2024-04-30
fix: botocore bedrock integration breaks for command R requests

tstadel:
fixes: https://github.com/DataDog/dd-trace-py/issues/9135

This PR does not add support for Command R models as this is more complex and would require:
- distingiushing Command models from Command R models
- trace inputs of Command R models
- trace outputs of Command R models

Instead this PR just ensures that ddtrace doesn't break Command R invocation.


sabrenner:
👋 Hi! Thanks for the assist here - I'm cherry picking this commit into #9173 as part of some other error handling we're adding into our bedrock integration. Makes CI a bit easier to run as well. Once that PR goes through, I'll close this one out.

Yun-Kim:
Closing as #9173 should resolve the issue addressed by this PR. Thanks for raising the issue @tstadel!

PR #9150, opened 2024-05-02
feat(tracer): update RC pollers to act on `AGENT_CONFIG` and `AGENT_TASK` products for the tracer flare

erikayasuda:
## Checklist

## Reviewer Checklist

- [ ] Title is accurate


ddtrace/internal/flare/flare.py
```diff
@@ -12,7 +12,6 @@
 from typing import Optional
 from typing import Tuple
 
-from ddtrace import config
```
erikayasuda, line 15: Had to remove this because of circular dependency issues

ddtrace/internal/flare/flare.py
```diff
@@ -161,20 +172,20 @@ def _generate_payload(self, params: Dict[str, str]) -> Tuple[dict, bytes]:
             encoded_key = key.encode()
             encoded_value = value.encode()
             body.write(b"--" + boundary + newline)
-            body.write(b'Content-Disposition: form-data; name="{%s}"{%s}{%s}' % (encoded_key, newline, newline))
-            body.write(b"{%s}{%s}" % (encoded_value, newline))
+            body.write(b'Content-Disposition: form-data; name="%s"%s%s' % (encoded_key, newline, newline))
```
erikayasuda, line 175: The `{}` brackets were typos from the previous way that the string substitution was happening, and this was breaking the header format.

ddtrace/internal/flare/flare.py
```diff
@@ -36,14 +35,28 @@ class FlareSendRequest:
...
+        api_key: Optional[str] = None,
+        timeout_sec: int = DEFAULT_TIMEOUT_SECONDS,
+        flare_dir: str = TRACER_FLARE_DIRECTORY,
+    ):
         self.original_log_level: int = logging.NOTSET
         self.timeout: int = timeout_sec
         self.flare_dir: pathlib.Path = pathlib.Path(flare_dir)
         self.file_handler: Optional[RotatingFileHandler] = None
+        self.url: str = trace_agent_url
+        self._api_key: Optional[str] = api_key
```
erikayasuda, line 57: If this is a concern, I can have the API key get passed into the `send()` method instead

ddtrace/internal/flare/_subscribers.py
```diff
@@ -0,0 +1,67 @@
+from datetime import datetime
```
erikayasuda, line 1: If it feels more appropriate to have this in the same file as the other RC stuff (`ddtrace/internal/remoteconfig/_subscribers.py`) I can move it there. I originally had it there, but then realized this would ping the remoteconfig approver group each time. Wasn't sure if that was necessary.

ddtrace/internal/flare/flare.py
```diff
@@ -36,14 +35,28 @@ class FlareSendRequest:
     source: str = "tracer_python"
 
 
+class TracerFlareSendError(Exception):
+    def __init__(self, message: str):
+        super().__init__(message)
+        self.message = message
```
erikayasuda, line 41: Not necessary, but none of the default error types seemed to really fit the failure type here. If there's a generic type that works better, I'm more than happy to clean this up

ddtrace/internal/flare/flare.py
```diff
@@ -96,35 +109,33 @@ def send(self, flare_send_req: FlareSendRequest):
...
+                log.error("Failed to send tracer flare to Zendesk ticket %s: %s", flare_send_req.case_id, e)
                 raise e
             finally:
                 client.close()
                 # Clean up files regardless of success/failure
                 self.clean_up_files()
-                return
 
-    def _generate_config_file(self, pid: int):
+    def _generate_config_file(self, config: dict, pid: int):
```
erikayasuda, line 133: @brettlangdon You previously mentioned having some kind of validation method for the config. Is it safe to assume here that the main keys in the `Config` object won't change much, so I could safely write a more "rigid" validation method (ex - make a static list of the top level keys, verify they exist in the JSON)? Or, are we just looking for possible human unfriendly rendering of the file (ex - make sure none of the values are just object references)?

ddtrace/internal/flare/_subscribers.py
emmettbutler: I think it makes sense to live in this file.

ddtrace/internal/flare/flare.py
emmettbutler: It looks to me like the entire `__init__` you've got defined is redundant. I'm pretty sure this code would work the same if lines 39-41 were replaced with `pass`.

ddtrace/settings/config.py
```diff
@@ -805,15 +809,92 @@ def _format_tags(self, tags: List[Union[str, Dict]]) -> Dict[str, str]:
             pairs = [t.split(":") for t in tags]  # type: ignore[union-attr,misc]
         return {k: v for k, v in pairs}
 
+    def _tracerFlarePubSub(self):
```
emmettbutler, line 812: These four functions don't actually have much to do with "config" as we use that term in ddtrace. Moving them to `internal/flare` and importing them here might reduce the cognitive load required to understand what this file is doing.

tests/internal/test_tracer_flare.py
```diff
@@ -168,3 +172,114 @@ def test_no_app_logs(self):
...
+        assert original_request_start is not None
+
+        # Generate another AGENT_CONFIG product to trigger a request
+        # This should not be processed
+        with mock.patch("ddtrace.internal.flare.flare.Flare.prepare") as mock_flare_prep:
+            self.generate_agent_config()
+            mock_flare_prep.assert_not_called()
+
+        # Nothing should have changed, and we should still be processing
+        # only the original request
```
emmettbutler, line 284: Putting comments like these in the `assert` text (`assert foo == bar, "foo should be bar"`) adds a layer of usability to the tests.

tests/internal/test_tracer_flare.py
```diff
@@ -168,3 +172,114 @@ def test_no_app_logs(self):
...
+                "hostname": "myhostname",
+                "user_handle": "user.name@datadoghq.com",
+            },
+            "task_type": "tracer_flare",
+            "uuid": "d53fc8a4-8820-47a2-aa7d-d565582feb81",
+        },
+    ]
+
+    def setUp(self):
+        from ddtrace import config
```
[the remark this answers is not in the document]
emmettbutler: I think what you're trying to do here is reset `config.Config.flare`? I'm not sure importing does what you want. Would it work to be more explicit about the state you're trying to reset, like `config.Config = config.Config()`?

ddtrace/internal/flare/flare.py
brettlangdon: this is the library's config (`ddtrace.config` thing)?

or is this something we receive from remote config?

erikayasuda: Yep the `ddtrace.config`

tests/internal/test_tracer_flare.py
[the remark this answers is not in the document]
erikayasuda: I'm importing here because I want to use the `_handle_tracer_flare` method and test that that handles the RC products correctly 😕 I could just reproduce the code here, but that means we have to track when those changes happen and update here too, which feels bad

ddtrace/settings/config.py
erikayasuda: Sounds good, I'll go ahead and move them over

ddtrace/settings/config.py
```diff
@@ -808,12 +808,17 @@ def _format_tags(self, tags: List[Union[str, Dict]]) -> Dict[str, str]:
     def enable_remote_configuration(self):
         # type: () -> None
         """Enable fetching configuration from Datadog."""
+        from ddtrace.internal.flare.flare import Flare
+        from ddtrace.internal.flare.handler import _handle_tracer_flare
+        from ddtrace.internal.flare.handler import _tracerFlarePubSub
```
erikayasuda, line 813: Had to move these to be in-line to fix `ModuleWatchdog is already installed` errors.

erikayasuda:
> Found 4 performance improvements and 46 performance regressions! Performance is the same for 159 metrics, 9 unstable metrics.

🙃 Will look into the top few worst regressions

emmettbutler:
@erikayasuda they're failing on main, so no need to worry about them here.

ddtrace/settings/config.py
```diff
@@ -803,12 +803,17 @@ def _format_tags(self, tags: List[Union[str, Dict]]) -> Dict[str, str]:
...
         from ddtrace.internal.remoteconfig.worker import remoteconfig_poller
 
         remoteconfig_pubsub = self._remoteconfigPubSub()(self._handle_remoteconfig)
+        flare = Flare(trace_agent_url=self._trace_agent_url, api_key=self._dd_api_key)
+        tracerflare_pubsub = _tracerFlarePubSub()(_handle_tracer_flare, flare)
         remoteconfig_poller.register("APM_TRACING", remoteconfig_pubsub)
-        remoteconfig_poller.register("AGENT_CONFIG", remoteconfig_pubsub)
```
emmettbutler, line 810: I'm surprised that this change doesn't break anything. I'd expect changing the handler for `AGENT_CONFIG` to have a meaningful effect, because it's presumably already used, separately from the flare. Just a flag - please tell me if I'm missing information that makes this change ok.

tests/internal/test_tracer_flare.py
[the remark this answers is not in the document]
emmettbutler: It looks like this import isn't used anywhere in the tests. If that's right, I recommend removing it.

emmettbutler (approved):

ddtrace/settings/config.py
erikayasuda: I think this switch isn't breaking because these changes won't go into effect until the app is restarted/ddtrace is updated? I haven't actually tried to test an ongoing app upgrading to a version that has this change, but I didn't think it would be a risk because `enable_remote_configuration` only runs when ddtrace is starting up. If that's an incorrect assumption, then it might be something I need to look into though

tests/internal/test_tracer_flare.py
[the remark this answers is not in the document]
erikayasuda: Yep I'll clean this up, thanks for catching it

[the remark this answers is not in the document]
erikayasuda: Oh wait I do use it for `flare=Flare(trace_agent_url=config._trace_agent_url)` 🤦‍♀️ I think I can set it as a constant and hard code it though, since that's the only place it gets used here.

ddtrace/settings/config.py
brettlangdon: I don't think I understand this change. are we no longer going to call `remoteconfig_pubsub` on `AGENT_CONFIG` changes? is that a problem?

I am having trouble tracing the code to understand what we do from `AGENT_CONFIG` events.

brettlangdon: Could settings changes come in through `APM_TRACING` or `AGENT_CONFIG`?

brettlangdon (commented):
sorry, only a partial review for now.

ddtrace/internal/flare/_subscribers.py
```diff
@@ -0,0 +1,71 @@
...
+    def has_stale_flare(self) -> bool:
+        if self.current_request_start:
+            curr = datetime.now()
+            flare_age = (curr - self.current_request_start).total_seconds()
+            stale_age = self.stale_tracer_flare_num_mins * 60
+            return flare_age >= stale_age
+        return False
+
+    def _get_data_from_connector_and_exec(self):
+        if self.current_request_start is not None and self.has_stale_flare():
```
brettlangdon, line 39:
```suggestion
        if self.has_stale_flare():
```

I think this is the same thing, since you check `if self.current_request_start:` and return False if it is

ddtrace/internal/flare/_subscribers.py
```diff
@@ -0,0 +1,71 @@
...
+                "Tracer flare request started at %s is stale, reverting "
+                "logger configurations and cleaning up resources now",
+                self.current_request_start,
+            )
+            self.current_request_start = None
+            self._callback(self.flare, {}, True)
+            return
+
+        data = self._data_connector.read()
+        product_type = data.get("metadata", [{}])[0].get("product_name")
```
brettlangdon, line 50: is `metadata` always going to be at least 1 element?

it could be safer to do:

```python
metadata = data.get("metdata")
if not metadata:
    # Log?
    return

product_type = metadata[0].get("product_name")
```

since `not metadata` will also confirm that the list is not empty.

ddtrace/internal/flare/handler.py
```diff
@@ -0,0 +1,90 @@
...
+        return
+
+    if "config" not in data:
+        log.warning("Unexpected tracer flare RC payload %r", data)
+        return
+    if len(data["config"]) == 0:
+        log.warning("Unexpected number of tracer flare RC payloads %r", data)
+        return
+
+    product_type = data.get("metadata", [{}])[0].get("product_name")
```
brettlangdon, line 44: same here, we can probably be a little more verbose to be more clear and try to avoid the defaults of `[{}]`.

Could be the same for `configs` too, since if we don't have any configs, then we don't need to call the prepare functions.

releasenotes/notes/add-tracer-flare-65e275bca27631dd.yaml
```diff
@@ -0,0 +1,5 @@
+---
+features:
+  - |
+    tracer: This introduces the tracer flare functionality. Currently the tracer flare includes the
```
brettlangdon, line 4: do we have any public docs we can link to on what this is/how to use it?

brettlangdon (commented):
minor nits only.

releasenotes/notes/add-tracer-flare-65e275bca27631dd.yaml
erikayasuda: As far as I'm aware, no 😢 I think now that the other tracers are implementing the flare, there's been some talk about having a centralized doc for this

ddtrace/internal/flare/_subscribers.py
erikayasuda: Hmmm.. From my manual testing yes, but that's just my limited exposure. I'll add this to be safe

PR #9152, opened 2024-05-03
wip: propagate llmobs parent

Yun-Kim:
## Checklist

## Reviewer Checklist

- [ ] Title is accurate


ddtrace/llmobs/_llmobs.py
```diff
@@ -30,11 +30,14 @@
...
+from ddtrace.llmobs._constants import PROPAGATED_PARENT_ID_KEY
 from ddtrace.llmobs._constants import SESSION_ID
 from ddtrace.llmobs._constants import SPAN_KIND
 from ddtrace.llmobs._constants import SPAN_START_WHILE_DISABLED_WARNING
 from ddtrace.llmobs._constants import TAGS
 from ddtrace.llmobs._trace_processor import LLMObsTraceProcessor
+from ddtrace.llmobs._utils import _get_llmobs_parent_id
 from ddtrace.llmobs._utils import _get_ml_app
 from ddtrace.llmobs._utils import _get_session_id
 from ddtrace.llmobs._writer import LLMObsEvalMetricWriter
```
Yun-Kim, line 43: I'm debating on if we should add `requests/httpx/urllib3/...` integrations to the list of supported integrations, since users will need to have these enabled in order for distributed tracing to work correctly. Thoughts @lievan ?

emmettbutler (approved):

ddtrace/llmobs/_llmobs.py
lievan: Let's patch these in enable() but not list it as a supported integration?

Yun-Kim: As per offline discussion - we will have the same functionality as `ddtrace-run` where we call `patch_all()` for non-LLMObs integrations. For users that use libraries that aren't patched by default by `patch_all()` (ex: urllib3), we will add documentation in our corp docs to explicitly call `ddtrace.patch(urllib3=True)`.

Yun-Kim:
Note: The http propagation benchmark tests have reported a significant increase in memory usage from this commit: https://github.com/DataDog/dd-trace-py/pull/9152#:~:text=Comparing%20candidate%20commit-,8786fb5,-in%20PR%20branch. However the changes from that commit do not justify such a memory increase, and `LLMObs` should not be enabled in benchmark tests. Additionally, the benchmarks have not run on this branch since that commit. This will need more investigation but I do not believe that this PR is responsible for such a memory increase.

PR #9157, opened 2024-05-03
fix(grpc): resolve issue with incorrect gRPC response being returned

brettlangdon:
The usage of the core API was not correct in the gRPC integration. It was working on accident based on how the listener was written.

Since the listener mutates the underlying response class and not the response instance, there is nothing that needs to be returned, so we can change the `dispatch_with_results` to `dispatch`.

As well, the previous usage of the `dispatch_with_results` API was not correct, the response was always replaced with the result from the ASM listener even when the response was `None`.

## Changelog

No changelog entry is added since this is a fix for an unreleased change so it is not yet a "bug".

## Checklist

## Reviewer Checklist

- [x] Title is accurate

### DIFF
--- a/ddtrace/appsec/_handlers.py
+++ b/ddtrace/appsec/_handlers.py
@@ -410,7 +410,6 @@ def _on_grpc_response(response):
 
     msg_cls = type(response)
     _patch_protobuf_class(msg_cls)
-    return response
 
 
 def listen():
@@ -425,4 +424,4 @@ core.on("django.patch", _on_django_patch)
 core.on("flask.patch", _on_flask_patch)
 
 core.on("asgi.request.parse.body", _on_asgi_request_parse_body, "await_receive_and_body")
-core.on("grpc.response_message", _on_grpc_response, "response")
+core.on("grpc.response_message", _on_grpc_response)
--- a/ddtrace/contrib/grpc/client_interceptor.py
+++ b/ddtrace/contrib/grpc/client_interceptor.py
@@ -81,14 +81,7 @@ def _handle_response(span, response):
     # google-api-core which has its own future base class
     # https://github.com/googleapis/python-api-core/blob/49c6755a21215bbb457b60db91bab098185b77da/google/api_core/future/base.py#L23
     if hasattr(response, "_response"):
-        result = core.dispatch_with_results(
-            "grpc.response_message",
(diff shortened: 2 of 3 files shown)


erikayasuda (approved):

ddtrace/appsec/_handlers.py
```diff
@@ -425,4 +424,4 @@ def listen():
 core.on("flask.patch", _on_flask_patch)
 
 core.on("asgi.request.parse.body", _on_asgi_request_parse_body, "await_receive_and_body")
-core.on("grpc.response_message", _on_grpc_response, "response")
+core.on("grpc.response_message", _on_grpc_response)
```
mabdinur, line 427: Why we need to make this change? Does the name of the callback matter?

mabdinur (approved):
Nice catch. I left one small nit. It's interesting that none of our tests detected the mutated response. Should we update our grpc+iast tests: https://github.com/DataDog/dd-trace-py/blob/434f71188c9374cd6892f774fd262cd2bc181f56/tests/appsec/iast/test_grpc_iast.py#L25?

ddtrace/appsec/_handlers.py
brettlangdon: The name (third parameter) is the key used to store the return value of the listener when calling `dispatch_with_results`, since we are not calling with `dispatch_with_results` anymore, the name is not necessary.

No harm in leaving it, but I figured it might help to signify that it is not supposed to return a value.

brettlangdon:
> Nice catch. I left one small nit. It's interesting that none of our tests detected the mutated response. Should we update our grpc+iast tests:
> 
> https://github.com/DataDog/dd-trace-py/blob/434f71188c9374cd6892f774fd262cd2bc181f56/tests/appsec/iast/test_grpc_iast.py#L25
> 
> ?

Yeah, agreed. It _seems_ like this should always cause an issue when using gRPC + ASM (should happen regardless of IAST being enabled or disabled).

juanjux (approved):

juanjux:
@brettlangdon @mabdinur it seems `main` is broken right because of last friday's "fix" now and this PR really fix the failure so I am merging  it (CI is green) and will do a further PR to add a test checking that the message is not changed.

PR #9162, opened 2024-05-06
chore(internal): fix CI Visibility usage of the core API

brettlangdon:
Fixes some small inconsistencies with the core api usage in CI Visibility.

1. A call to `dispatch_with_results` when no value is expected to be returned (small performance optimization to avoid building a results dictionary when it isn't needed).
2. A call to `dispatch_with_results` which would not return the underlying listener value
    a. (I _think_ it would always return a falsey value since it should have been returning the `core._MissingEvent` `EventResult` which is falsey and which has a value of `None`... ?)

I am not sure which tests should be evaluated/updated, since I am surprised there weren't any failing tests before, there _should_ have been.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


emmettbutler (approved):
Oh yeah, it's not clear at all to me why this isn't causing `test_is_item_itr_skippable_test_level` to fail in current main. This change does look more correct, though.

ddtrace/ext/ci_visibility/api.py
```diff
@@ -239,7 +239,7 @@ def finish(
     @_catch_and_log_exceptions
     def get_settings(item_id: Optional[CISessionId] = None):
         log.debug("Getting settings for session %s", item_id)
-        core.dispatch_with_results("ci_visibility.session.get_settings", (item_id,))
+        core.dispatch("ci_visibility.session.get_settings", (item_id,))
```
romainkomorndatadog, line 242: The actual fix here is to actually return the settings.

romainkomorndatadog (changes requested):
These (the settings request) aren't in use yet so they're not tested (and, clearly, not working either).

edit: the external API `is_item_itr_skippable()` static method (that ends up calling `core.dispatch_with_results("ci_visibility.itr.is_item_skippable")` itself) is also not tested yet.

`test_is_item_itr_skippable_test_level` is directly testing the `CIVisibility.is_item_itr_skippable()` function directly (which is why it's in `test_ci_visibility.py`), so any changes in this PR wouldn't affect its results.

I'll take over the PR.

ddtrace/ext/ci_visibility/api.py
romainkomorndatadog: Actually, since I don't have a use case for this at the moment, I'm just going to delete it.

romainkomorndatadog:
Renaming this to a `chore` because this API isn't released yet, so it's not a "fix".

romainkomorndatadog (approved):

emmettbutler (approved):

PR #9172, opened 2024-05-06
feat(llmobs): support in-code config for llmobs

lievan:
Support in-code configuration for LLMObs users, to enable LLMObs and specify the following configurations that currently require environment variable configuration.

- ml_app
- list of integrations to patch (will patch all LLMObs integrations by default)
- dd_llmobs_no_apm (turn off APM, telemetry, remote config, metrics)
- DD site, DD env, DD service (will override config/env vars)

```

LLMObs.enable(
    ml_app="comms/langchain", 
    integrations=["openai"],
    llmobs_agentless_enabled=True,
    # api_key =...
    # site=...
    # env=...
    # service=...
    # _tracer=None
)
```
Allowing in-code setup also improves the dev experience for people tracing experimental apps with LLMObs. It also abstracts away a long list of environment variables non-APM customers are required to set to turn off all APM related features.

This PR should not break _any_ previous way of setting up the Python SDK (e.g. using env vars and `ddtrace-run`). 

Arguments passed to enable() should take precedence over environment variables, with the exception of `DD_LLMOBS_ENABLED`. 

This PR also does a couple minor things:
- If `DD_LLMOBS_NO_APM` env var is detected or configured through LLMObs.enable(), the OpenAI and LangChain integrations will disable submitting metrics unless the corresponding env vars `DD_{OPENAI,LANGCHAIN}_METRICS_ENABLED` is set to True.
- We also automatically disable both telemetry writer and remote config pollers if `DD_LLMOBS_NO_APM` is detected or configured through LLMObs.enable().
- We automatically patch the LLMObs integrations on LLMObs.enable().
- Removes all LLMObs.enable() references in individual integration patch code (openai, botocore, langchain)

Note:
- This change (only for LLMObs users) will override `config.service, config.env` if these are passed in to `LLMObs.enable()`.  
- If a user runs via `ddtrace-run`, they cannot use `LLMObs.enable()` to configure their settings.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


ddtrace/llmobs/_llmobs.py
```diff
@@ -75,30 +95,103 @@ def _stop_service(self) -> None:
...
+                        "%s is unsupported - LLMObs currently supports %s", integration, str(llmobs_integrations)
+                    )
+
         cls._instance = cls(tracer=tracer)
         cls.enabled = True
         cls._instance.start()
         atexit.register(cls.disable)
         log.debug("%s enabled", cls.__name__)
 
+    @classmethod
```
lievan, line 174: introduce these functions so that customers can manually turn on llmobs integrations with the LLMObs service without importing any other ddtrace related code

ddtrace/llmobs/_llmobs.py
```diff
@@ -75,30 +97,127 @@ def _stop_service(self) -> None:
...
+        # since our integrations depend on config._llmobs_enabled to set llmobs data
+        config._llmobs_enabled = switch
+        cls.enabled = switch
+
+    @classmethod
+    def enable(
+        cls,
+        ml_app: Optional[str] = None,
+        integrations: Optional[List[str]] = None,
+        dd_apm_enabled=True,
```
lievan, line 111: thoughts on this name? @Yun-Kim @sabrenner

sabrenner: I think it could be confusing to have this config option named `dd_apm_enabled` while the "corresponding" env var is `DD_LLMOBS_NO_APM`. It feels like if both have the same value, they should mean the same thing, whereas now I think if both are set to `True` they mean opposite things.

lievan: renaming to `dd_llmobs_no_apm` for consistency with env var and also going to update the enable() function so that all the env vars non-APM customers need to set are switched to the correct value when `dd_llmobs_no_apm` is true

ddtrace/llmobs/_llmobs.py
```diff
@@ -75,30 +92,127 @@ def _stop_service(self) -> None:
...
             return
 
+        if os.getenv("DD_LLMOBS_ENABLED") and not asbool(os.getenv("DD_LLMOBS_ENABLED")):
+            LLMObs._toggle(False)
+            log.debug("LLMObs.enable() called when DD_LLMOBS_ENABLED is set to false, not starting LLMObs service")
+            return
+
+        # grab required values for LLMObs
+        # reading from environment variables is needed in case preload script wasn't called.
+        config._dd_site = config._dd_site or os.getenv("DD_SITE") or dd_site
```
sabrenner, line 135: for a future pr - i'm wondering if as we add more configuration options it'd be better to have our own config coalescing in a separate module, similar as to how it's done in `config.py`, instead of doing it all here. would make it a bit cleaner and probably easier to test too. wdyt?

ddtrace/llmobs/_llmobs.py
```diff
@@ -75,30 +92,127 @@ def _stop_service(self) -> None:
...
 
+        # grab optional values
+        config.env = os.getenv("DD_ENV") or config.env or dd_env
+        config.service = os.getenv("DD_SERVICE") or config.service or dd_service
+
+        if dd_llmobs_no_apm or asbool(os.getenv("DD_LLMOBS_NO_APM", "false")):
+            for k, v in cls._no_apm_env_config.items():
+                # don't override environment variables are are explicitly set.
+                if not os.getenv(k):
+                    os.environ[k] = v
```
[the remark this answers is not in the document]
sabrenner: additionally, whether we're modifying internal config options, or we end up going with modifying env vars, it might be helpful to emit a log saying something like
```
'dd_llmobs_no_apm' was set to True, setting DD_OPENAI_METRICS_ENABLED to False
```
etc., just so it's clear to a user with debug logs turned on or a support engineer that this is what happened under the hood. wdyt?

ddtrace/llmobs/_llmobs.py
sabrenner, line 135: Just a note - taking a look back for reference, it actually does seem like other tracers (ie Node) prioritize config options over environment variables (this one's very much on me, when we discussed this last Friday I had it backwards). I think what we have here, prioritizing environment variables makes sense, and is fine as long as we are consistent with it. But, we should just confirm between ourselves this is the behavior we want, as if we want to swap it in the future I think it could be considered breaking (maybe?). wdyt? cc @Yun-Kim

sabrenner (commented):
Some non-blocking comments, but one important one about just verifying that we're OK with prioritizing environment variables over config options. This is on me, as I had it backwards when we discussed it last week. Node, for example, does prioritize config options over environment variables. I think the rationale is config options could always be set by environment variables, ie:

```python
LLMObs.enable(
  ml_app=os.environ["DD_LLMOBS_APP_NAME"],
  # etc...
)
```
So grabbing the config option first could be seen as more relevant (depends on the developer practices of the end user too I guess).

I do think that what we currently have (env var over config) could be fine, but only as long as we're consistent with it. Otherwise, I think swapping the order of precedence could be considered a breaking change.

It's just difficult to interpret intention 😅

ddtrace/llmobs/_llmobs.py
lievan: I think prioritizing env vars makes sense in this case since `LLMObs.enable` is only expected to be called once to globally configure LLMObs, and also we want to support people turning LLMObs on/off without having to make in code changes

ddtrace/llmobs/_llmobs.py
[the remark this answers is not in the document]
lievan: i like the debug log idea, and agree with not changing the remote config env var

unfortunately `DD_OPENAI_METRICS_ENABLED` isn't set at the config object level, the openai integration directly reads from that env var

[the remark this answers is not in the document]
sabrenner: ahh ok, i think this is fine then

Yun-Kim (commented):
Some comments

ddtrace/llmobs/_llmobs.py
```diff
@@ -75,26 +83,88 @@ def _stop_service(self) -> None:
...
+
+        # enable LLMObs integations
+        llmobs_integrations = {
+            LLMObs.langchain: lambda: patch(langchain=True),
+            LLMObs.openai: lambda: patch(openai=True),
+            LLMObs.botocore: lambda: patch(bedrock=True),
+        }
+        if integrations:
+            for integration in integrations:
+                if integration in llmobs_integrations:
```
Yun-Kim, line 159:
```suggestion
                if integration.lower() in llmobs_integrations:
```

ddtrace/llmobs/_llmobs.py
```diff
@@ -75,26 +83,88 @@ def _stop_service(self) -> None:
...
+                    llmobs_integrations[integration]()
+                else:
+                    log.warning(
+                        "%s is unsupported - LLMObs currently supports %s", integration, str(llmobs_integrations)
```
Yun-Kim, line 163:
```suggestion
                        "%s is unsupported - LLMObs currently supports %s", integration, str(llmobs_integrations.keys())
```

ddtrace/llmobs/_llmobs.py
Yun-Kim: Agreed. Config should take precedence over env vars - I've made some [suggestions](https://github.com/DataDog/dd-trace-py/pull/9172/files#r1598982879) to make the config override env vars. Let me know if that's not what you mean

ddtrace/llmobs/_llmobs.py
```diff
@@ -75,26 +80,109 @@ def _stop_service(self) -> None:
...
+
+        # grab required values for LLMObs
+        # reading from environment variables is needed in case preload script wasn't called.
+        config._dd_site = dd_site or config._dd_site
+        config._dd_api_key = dd_api_key or config._dd_api_key
+        config._llmobs_ml_app = ml_app or config._llmobs_ml_app
+
+        # validate required values for LLMObs
         if not config._dd_api_key:
-            cls.enabled = False
```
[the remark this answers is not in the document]
Yun-Kim: Turns out this was unnecessary (good removal) since we're not toggling the enabled value until after we raise the error.

Yun-Kim:
@lievan I've made a small refactor based on some comments I made. However it looks like there's an issue somewhere in this PR that subtly enabling LLMObs even for tests that don't involve LLMObs (CI test snapshots show that `_ml_obs.*` tags are present on tags, and that `LLMObs.enabled` evaluates to True. We need to figure this out before we can merge this in.

ddtrace/llmobs/_llmobs.py
```diff
@@ -43,6 +47,13 @@
 log = get_logger(__name__)
 
 
+SUPPORTED_INTEGRATIONS = {
+    "bedrock": lambda: patch(botocore=True),
+    "langchain": lambda: patch(langchain=True),
+    "openai": lambda: patch(openai=True),
```
Yun-Kim, line 53: I replaced `LLMObs.openai, LLMObs.bedrock, LLMObs.langchain` with this to allow users to specify string arguments (would be simpler than to make them have to use yet another `LLMObs` SDK object). Will need to change documentation accordingly

ddtrace/llmobs/_llmobs.py
```diff
@@ -88,7 +99,29 @@ def _stop_service(self) -> None:
...
-    def enable(cls, tracer=None):
+    def enable(
+        cls,
+        ml_app: Optional[str] = None,
+        integrations: Optional[List[str]] = None,
+        dd_llmobs_no_apm: bool = False,
+        dd_site: Optional[str] = None,
+        dd_api_key: Optional[str] = None,
+        dd_env: Optional[str] = None,
+        dd_service: Optional[str] = None,
```
Yun-Kim, line 110:
```suggestion
        llmobs_no_apm: bool = False,
        site: Optional[str] = None,
        api_key: Optional[str] = None,
        env: Optional[str] = None,
        service: Optional[str] = None,
```
Can we remove the `dd` prefixes? Just to keep things clean (happy to hear otherwise if you had reasons for adding the prefix)

Yun-Kim (approved):
LGTM, great work!

tabgok (approved):

PR #9211, opened 2024-05-09
feat(asm): standalone asm

gnufede:
## Description
This PR allows ASM customers to disable APM, and use only ASM features. To have the full picture, see also #9444 and #9445. 

## Details
It involves changes when opting out from APM with `DD_EXPERIMENTAL_APPSEC_STANDALONE_ENABLED=true` and `DD_APPSEC_ENABLED=true`:
- Doesn't disable Span Processors.
- Rate limiting of APM traces is reduced to minimal levels (1 trace per minute).
- `Datadog-Client-Computed-Stats` header is sent to the agent so it doesn't compute stats on its end, but they aren't computed on the library side either.
- Disables metric reporting.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


ddtrace/_trace/tracer.py
```diff
@@ -232,7 +233,18 @@ def __init__(
...
+        self._asm_enabled = asm_config._asm_enabled
         self._sampler: BaseSampler = DatadogSampler()
+        self._apm_opt_out = self._asm_enabled and not self.enabled
+        if self._apm_opt_out:
+            # If ASM is enabled but tracing is disabled,
+            # we need to set the rate limiting to 1 trace per minute
+            # for the backend to consider the service as alive.
+            from ddtrace.internal.rate_limiter import RateLimiter
+
+            self._sampler.limiter = RateLimiter(rate_limit=1, time_window=60e9)  # 1 trace per minute
```
ZStriker19, line 245: Technically we can't rely on the _sampler always having a rate limiter in the event that customers set their own custom sampler (deprecated and is going to be removed in next major release). So we may want to try except around this and log something.

ddtrace/_trace/tracer.py
```diff
@@ -232,7 +233,18 @@ def __init__(
...
+            # Disable compute stats
+            config._trace_compute_stats = False
```
ZStriker19, line 247: Why do we need to disable compute stats?

gnufede: Well, it's part of the RFC. I think because we're disabling "APM" features basically

ddtrace/_trace/tracer.py
gnufede: Changing that in the setter of the sampler instead, to deactivate that feature if Standalone ASM is enabled

ddtrace/_trace/tracer.py
```diff
@@ -250,14 +250,14 @@ def __init__(
...
-                headers={"Datadog-Client-Computed-Stats": "yes"} if self._compute_stats else {},
+                headers={"Datadog-Client-Computed-Stats": "yes"} if (self._compute_stats or self._apm_opt_out) else {},
+                report_metrics=not self._apm_opt_out,
                 response_callback=self._agent_response_callback,
             )
         self._single_span_sampling_rules: List[SpanSamplingRule] = get_span_sampling_rules()
         self._writer: TraceWriter = writer
         self._partial_flush_enabled = config._partial_flush_enabled
         self._partial_flush_min_spans = config._partial_flush_min_spans
-        self._asm_enabled = asm_config._asm_enabled
```
gnufede, line 260: n.b: already done in line 236

ddtrace/_trace/tracer.py
```diff
@@ -332,6 +342,9 @@ def sampler(self, value):
     https://ddtrace.readthedocs.io/en/stable/configuration.html#DD_TRACE_SAMPLING_RULES""",
             category=DDTraceDeprecationWarning,
         )
+        if self._apm_opt_out:
+            log.error("Cannot set a custom sampler with Standalone ASM mode")
```
christophe-papazian, line 346: isn't a warning enough? Application is not crashing here.

gnufede: yeah, true

christophe-papazian (approved):

tests/tracer/test_propagation.py
```diff
@@ -304,6 +304,82 @@ def test_extract(tracer):  # noqa: F811
...
+            assert child_span.context._meta == {
+                "_dd.origin": "synthetics",
+                "_dd.p.dm": "-3",
+                "_dd.p.test": "value",
+            }
+    tracer.configure(appsec_enabled=False, appsec_standalone_enabled=False)
+
+
+def test_extract_asm_standalone_present_appsec_tag_sampling_priority_honored(tracer):  # noqa: F811
+    tracer.configure(appsec_enabled=True, appsec_standalone_enabled=True)
```
juanjux, line 346: nit: maybe we could use `tracer.configure(appsec_enabled=True, apm_opt_out=True)` so use the same naming than inside the tracer code.

juanjux (approved):

ddtrace/internal/writer/writer.py
```diff
@@ -454,7 +458,7 @@ def __init__(
         max_payload_size=None,  # type: Optional[int]
         timeout=None,  # type: Optional[float]
         dogstatsd: Optional[DogStatsd] = None,
-        report_metrics=False,  # type: bool
+        report_metrics=True,  # type: bool
```
Julio-Guerra, line 461: intended? 🤔

tests/appsec/appsec_utils.py
```diff
@@ -77,6 +87,8 @@ def appsec_application_server(
         env["_DD_REMOTE_CONFIGURATION_ADDITIONAL_HEADERS"] = "X-Datadog-Test-Session-Token:%s," % (token,)
     if appsec_enabled is not None:
         env["DD_APPSEC_ENABLED"] = appsec_enabled
+    if appsec_standalone_enabled is not None:
+        env["DD_EXPERIMENTAL_APPSEC_STANDALONE_ENABLED"] = appsec_standalone_enabled
```
Julio-Guerra, line 91:
```suggestion
        env["DD_EXPERIMENTAL_APPSEC_STANDALONE_ENABLED"] = appsec_standalone_enabled # TODO: leverage APM disablement once available with standalone ASM enablement being equivalent to `appsec_enabled and apm_tracing_enabled`
```

ddtrace/internal/writer/writer.py
gnufede: Yes, this parameter wasn't actually used before

ddtrace/pin.py
```diff
@@ -143,7 +143,7 @@ def override(
     def enabled(self):
         # type: () -> bool
         """Return true if this pin's tracer is enabled."""
-        return bool(self.tracer) and self.tracer.enabled
+        return bool(self.tracer) and (self.tracer.enabled or self.tracer._apm_opt_out)
```
emmettbutler, line 146: Hiding this complexity behind `tracer.enabled`, possibly by making that attribute into a property, would reduce the cognitive load incurred by future editors of code in `Pin` and elsewhere.

ddtrace/_trace/tracer.py
```diff
@@ -402,7 +415,7 @@ def get_log_correlation_context(self) -> Dict[str, str]:
         If there is no active span, a dictionary with an empty string for each value will be returned.
         """
         active: Optional[Union[Context, Span]] = None
-        if self.enabled:
+        if self.enabled or self._asm_enabled:
```
emmettbutler, line 418: Counterintuitively, this approach couples `Tracer` to the ASM product. You might consider adding a layer of abstraction that makes `Tracer` hold a generic list of "products". This might be too much of a scope expansion for your work at the moment, so I'd probably still approve this approach. In the long run, though, the `Tracer` mustn't know specifics about other products.

ddtrace/pin.py
gnufede: Hmm... I was trying to avoid that. I'm actually setting tracer.enabled to False at some point just to be explicit about APM being disabled, not because we need to.

My point was to highlight the places where the "library" needs to be enabled regardless of the APM product being enabled.

I'd rather remove that `tracer.enabled = False`, if you prefer

ddtrace/_trace/tracer.py
gnufede: Would this be better?

```suggestion
        if self.enabled or self._apm_opt_out:
```

emmettbutler (approved):

PR #9215, opened 2024-05-09
fix(iast): avoid crashing grpc patch if MessageMapContainer import fails

gnufede:
Code Security: Fix an issue at GRPC patching where if `google._upb` module wasn't available, an unhandled ImportError would be raised.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


juanjux (approved):

ddtrace/appsec/_handlers.py
```diff
@@ -376,13 +374,18 @@ def _custom_protobuf_getattribute(self, name):
...
     elif isinstance(ret, MutableMapping):
-        if isinstance(ret, MessageMapContainer) and len(ret):
-            # Patch the message-values class
-            first_key = next(iter(ret))
-            value_type = type(ret[first_key])
-            _patch_protobuf_class(value_type)
-        else:
-            ret = taint_structure(ret, OriginType.GRPC_BODY, OriginType.GRPC_BODY)
+        try:
+            from google._upb._message import MessageMapContainer
```
brettlangdon, line 378: can we move this to the module level?

trying this import + try/except on every getattribute is going to be slow.


```python
MessageMapContainer = None
try:
    from google._upb._message import MessageMapContainer
except ImportError:
    pass


...

def _custom_protobuf_getattribute(self, name):
    ...
    if MessageMapContainer is None:
        return ret

    ...
```


wdyt?

avara1986 (approved):

ddtrace/appsec/_handlers.py
```diff
@@ -376,6 +380,9 @@ def _custom_protobuf_getattribute(self, name):
             source_origin=OriginType.GRPC_BODY,
         )
     elif isinstance(ret, MutableMapping):
+        if MessageMapContainer is None:
+            return ret
+
```
brettlangdon, line 385:
```suggestion
    elif MessageMapContainer is not None, and isinstance(ret, MutableMapping):
```

don't feel like you have to accept this, this is a pedantic performance related improvement.

if in failure scenarios we can avoid the `isinstance` call it'll help with performance.

again, pedantic, feel free to just resolve.

tests/appsec/iast/test_grpc_iast.py
```diff
@@ -72,3 +73,21 @@ def test_taint_iast_last(self):
                     res = stub1.SayHelloLast(requests_iterator)
                     assert hasattr(res, "message")
                     _check_test_range(res.message)
+
+    def test_taint_iast_patching_import_error(self):
+        with mock.patch.dict("sys.modules", {"google._upb._message": None}), override_env({"DD_IAST_ENABLED": "True"}):
```
brettlangdon, line 78: do we know the conditions this is actually not available? certain versions of protobuf? when using pure-python version vs not? when a certain env var is set or something?

I'd almost rather see us have a riot/hatch matrix item to make this be true for all of our tests so we can verify there aren't any other code paths which assume this is available.

wdyt? I am happy to consider this as a follow-up, but would love to see an investigation into it.

brettlangdon (approved):

ddtrace/appsec/_handlers.py
gnufede: I'd accept if you remove the `,` before the `and`

brettlangdon: updated 🤦🏻

tests/appsec/iast/test_grpc_iast.py
gnufede: Yeah, I expected to be able to reproduce in a higher level test, but I couldn't and don't know the conditions where it's not available, so I'd at least have a simple regression test and investigate further once we unlocked the release

brettlangdon: I remember digging through the code, and I cannot remember the scenarios when it isn't available.

it might just be older versions don't have it? maybe we aren't testing with old enough versions of protobuf?

brettlangdon: again, as a follow-up could be good to look into, and maybe document the outcome (even if we cannot get it to work) at least

gnufede: Yeah, we definitely should investigate what is going on. Let's do a follow-up

PR #9222, opened 2024-05-10
chore(internal): bytecode wrapping context

P403n1x87:
We introduce a new mechanism of wrapping functions via a special context manager that is capable of capturing return values as well. The goal is to allow observability into the called functions, to have access to local variables on exit. This approach has the extra benefit of not introducing any extra frames in the call stack of the wrapped function.

## Checklist

## Reviewer Checklist

- [x] Title is accurate

### DIFF
--- a/ddtrace/internal/wrapping/context.py
+++ b/ddtrace/internal/wrapping/context.py
@@ -531,14 +531,12 @@ class _UniversalWrappingContext(BaseWrappingContext):
 
             # Remove the try blocks
             i = 0
-            last_begin = None
             while i < len(bc):
                 instr = bc[i]
                 if isinstance(instr, bytecode.TryBegin) and instr.target is except_label:
-                    last_begin = bc.pop(i)
-                elif isinstance(instr, bytecode.TryEnd) and last_begin is not None and instr.entry is last_begin:
                     bc.pop(i)
-                    last_begin = None
+                elif isinstance(instr, bytecode.TryEnd) and instr.entry.target is except_label:
+                    bc.pop(i)
                 else:
                     i += 1
 
--- a/ddtrace/internal/wrapping/context.py
+++ b/ddtrace/internal/wrapping/context.py
@@ -368,21 +368,21 @@ class _UniversalWrappingContext(BaseWrappingContext):
     def __init__(self, f: FunctionType) -> None:
         super().__init__(f)
 
-        self._contexts: t.Dict[t.Type[WrappingContext], WrappingContext] = {}
+        self._contexts: t.List[WrappingContext] = []
 
     def register(self, context: WrappingContext) -> None:
         _type = type(context)
-        if _type in self._contexts:
+        if any(isinstance(c, _type) for c in self._contexts):
             raise ValueError("Context already registered")
 
-        self._contexts[_type] = context
+        self._contexts.append(context)
+        self._contexts.sort(key=lambda c: c.__priority__)
 
     def unregister(self, context: WrappingContext) -> None:
-        _type = type(context)
-        if _type not in self._contexts:
+        try:
+            self._contexts.remove(context)
+        except ValueError:
             raise ValueError("Context not registered")
-
-        del self._contexts[_type]
 
         if not self._contexts:
             self.unwrap()
@@ -391,7 +391,10 @@ class _UniversalWrappingContext(BaseWrappingContext):
         return type(context) in self._contexts
 
     def registered(self, context_type: t.Type[WrappingContext]) -> WrappingContext:
-        return self._contexts[context_type]
+        for context in self._contexts:
+            if isinstance(context, context_type):
+                return context
+        raise KeyError(f"Context {context_type} not registered")
 
     def __enter__(self) -> "_UniversalWrappingContext":
         super().__enter__()
@@ -399,7 +402,7 @@ class _UniversalWrappingContext(BaseWrappingContext):
         # Make the frame object available to the contexts
         self.set("__frame__", sys._getframe(1))
 
-        for context in sorted(self._contexts.values(), key=lambda c: c.__priority__):
+        for context in self._contexts:
             context.__enter__()
 
         return self
@@ -413,13 +416,13 @@ class _UniversalWrappingContext(BaseWrappingContext):
             # normally
             return
 
-        for context in sorted(self._contexts.values(), key=lambda c: -c.__priority__):
+        for context in self._contexts[::-1]:
             context.__exit__(*exc)
 
         super().__exit__(*exc)
 
     def __return__(self, value: T) -> T:
-        for context in sorted(self._contexts.values(), key=lambda c: -c.__priority__):
+        for context in self._contexts[::-1]:
             context.__return__(value)
 
         return super().__return__(value)
--- a/ddtrace/internal/wrapping/context.py
+++ b/ddtrace/internal/wrapping/context.py
@@ -1,7 +1,9 @@
+from contextvars import ContextVar
 from inspect import iscoroutinefunction
 import sys
 from types import FrameType
 from types import FunctionType
+from types import TracebackType
 import typing as t
 
 
@@ -9,12 +11,6 @@ try:
     from typing import Protocol  # noqa:F401
 except ImportError:
     from typing_extensions import Protocol  # type: ignore[assignment]
-
-try:
-    from typing import override  # type: ignore[attr-defined]
-except ImportError:
-    from typing_extensions import override
-
 
 import bytecode
 from bytecode import Bytecode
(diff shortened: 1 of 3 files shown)


ddtrace/internal/wrapping/context.py
```diff
@@ -0,0 +1,453 @@
+import sys
```
brettlangdon, line 1: docstring with usage/rationale would be great.

brettlangdon: technical background on how it works could be nice too, especially for people who struggle to read the bytecode.

ddtrace/internal/wrapping/context.py
```diff
@@ -0,0 +1,686 @@
...
+
+    def is_registered(self, context: WrappingContext) -> bool:
+        return type(context) in self._contexts
+
+    def registered(self, context_type: t.Type[WrappingContext]) -> WrappingContext:
+        return self._contexts[context_type]
+
+    def __enter__(self) -> "_UniversalWrappingContext":
+        frame = sys._getframe(1)
+        for context in sorted(self._contexts.values(), key=lambda c: c.__priority__):
```
brettlangdon, line 381: should we just store these in a sorted dict and we can resort on insert to avoid this sort here?

also, instead of a dict, it might just be easier to store as a list or tuple, and on insert you iterate the whole list to see if the type is already present.

_might_ be lower overhead than using a dict here? I mostly wonder if we can get the per-call overhead down as low as possible, especially since the registering should be called very infrequently, at least compared to `__enter__`/`__exit__`/`__return__`.

ddtrace/internal/wrapping/context.py
```diff
@@ -0,0 +1,686 @@
...
+        raise NotImplementedError
+
+    def unwrap(self) -> None:
+        raise NotImplementedError
+
+
+# This is the public interface exported by this module
+class WrappingContext(BaseWrappingContext):
+    @override
+    def __enter__(self, _frame: FrameType) -> "WrappingContext":
```
brettlangdon, line 309: should we also provide a base implementation of `__exit__`.

`AbstractContextManager` makes you implement it even though there is a base implementation which is just `return None`, we could do the same here?

If we don't, then we enforce that people acknowledge and handle errors, which isn't a bad thing?

ddtrace/internal/wrapping/context.py
```diff
@@ -0,0 +1,686 @@
...
+    def wrapped(cls, f: FunctionType) -> "BaseWrappingContext":
+        if cls.is_wrapped(f):
+            context = cls.extract(f)
+            assert isinstance(context, cls)  # nosec
+        else:
+            context = cls(f)
+            context.wrap()
+        return context
+
+    def __return__(self, value):
```
brettlangdon, line 288: Should we try to combine `__exit__` and `__return__`?

I am thinking about cases like finishing spans, we now will always have two code paths needed to close a span or context for these wrappers.

e.g. doesn't seem like we have a `try/finally` type setup here which we could get with the other wrapping we do.


`def __exit__(self, return_value, (exc_type, exc_val, exc_tb)): -> return_value`


It does seem to conflate things a bit, but wondering if we can consolidate a bit.


```python
class MyWrapper(WrappingContext):
    def __enter__(self, frame):
        self.span = tracer.trace("name")

    def __exit__(self, exc_type, exc_val, exc_tb):
        self.span.set_exc_info(exc_type, exc_val, exc_tb)
        self.span.finish()

    def __return__(self, rv):
        self.span.finish()
        return rv
```




```python
class MyWrapper(WrappingContext):
    def __enter__(self, _frame):
        self.span = tracer.trace("name")

    # __finally__ get return value and exception info
    # __finally__'s return value is ignored
    def __finally__(self, rv, (exc_type, exc_val, exc_tb)):
        self.span.set_exc_info(exc_type, exc_val, exc_tb)
        self.span.finish()


    # Only define `__return__` if you want to intercept and modify the return value
```


Not sure I love this, just starting conversation to get your thoughts.

ddtrace/internal/wrapping/context.py
P403n1x87: My thinking was that we shouldn't really have that many things registered here, so sorting and traversing should be quick enough. If we are really concerned about performance then the better data structure would probably be a heap, but we would have to back that with a dict anyway, to keep track of what's registered

ddtrace/internal/wrapping/context.py
P403n1x87: We could but I thought it wouldn't make much sense to use a wrapping context manager if you didn't have a need to implement `__exit__`/`__return__`. So in this sense, I'd view them as mandatory to implement by any subclass

tests/internal/test_wrapping.py
```diff
@@ -491,3 +493,198 @@ def f(a, b, c=None):
...
+
+    _type, exc, _ = wc.exc_info
+    assert _type == ValueError
+    assert exc.args == ("foo",)
+
+
+def test_wrapping_context_priority():
+    class HighPriorityWrappingContext(DummyWrappingContext):
+        def __enter__(self, frame):
+            nonlocal mutated
```
brettlangdon, line 609: TIL

tests/internal/test_wrapping.py
```diff
@@ -491,3 +493,198 @@ def f(a, b, c=None):
...
+            super().__exit__(exc_type, exc_value, traceback)
+            raise RuntimeError("broken")
+
+    def foo():
+        raise ValueError("foo")
+
+    wc = BrokenExitWrappingContext(foo)
+    wc.wrap()
+
+    with pytest.raises(RuntimeError):
```
brettlangdon, line 594: should we suppress these? is there an instance where we want our exception to raise?

tests/internal/test_wrapping.py
```diff
@@ -491,3 +493,198 @@ def f(a, b, c=None):
 
     assert closure(1, 2, 3) == (1, 2, 3, 42)
     assert channel == [((42,), {}), closure, ((1, 2, 3), {}), (1, 2, 3, 42)]
+
+
+NOTSET = object()
+
+
+class DummyWrappingContext(WrappingContext):
```
brettlangdon, line 501: trying to think of other scenarios.... a function wrapped with a decorator?

a function that is a context manager?

a function that is a generator?

in theory those shouldn't matter, but maybe easy enough to verify and codify their behavior here.

brettlangdon (commented):
mostly looks good, just some minor conversation pieces.

ddtrace/internal/wrapping/context.py
P403n1x87: I was thinking that one could do something like
```
def _close_context(self, return_value=None, exc_info=None):
    ...

def __return__(self, value):
    self._close_context(return_value=value)

def __exit__(self, *exc_info):
    self._close_context(exc_info=exc_info)
```
in a subclass of `WrappingContext` if there is a lot that can be DRYed in general

tests/internal/test_wrapping.py
P403n1x87: Yep good point, we should add a few more examples in the tests 👍

ddtrace/internal/wrapping/context.py
brettlangdon: yeah, we can always subclass, ok... makes sense.

tests/internal/test_wrapping.py
brettlangdon: This is another place where subclasses could help I guess... we can make a `RobustWrappingContext` that can never crash or fail on the wrapped call (lofty request, but maybe)

brettlangdon (approved):

tests/internal/test_wrapping.py
```diff
@@ -491,3 +494,283 @@ def f(a, b, c=None):
...
+    wc = RecursiveWrappingContext(factorial)
+    wc.wrap()
+
+    assert factorial(5) == 120
+    assert values == [5, 4, 3, 2, 1, 0, 0, 1, 2, 3, 4, 5]
+
+
+def test_wrapping_context_generator():
+    def foo():
+        yield from range(10)
```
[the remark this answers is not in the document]
P403n1x87: We should be able to add support for that if we need to. This is an interesting topic, as a `yield` cannot close the context, so it would have to be treated differently from a return. More interestingly, what happens for generators that never return?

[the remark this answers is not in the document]
brettlangdon: I am happy to yield on the yield topic for later, as long as you think it'll be fine to support later.

generators that never return is at least an existing problem, so I don't think we need a solution today.

designing this context wrapper so people do the right thing in those cases will be great, but not sure we can really detect/determine how best to approach that... ?

ddtrace/internal/wrapping/context.py
```diff
@@ -0,0 +1,687 @@
...
+#   def foo():
+#       with wrapping_context:
+#           # Original function code
+#
+# Because we also want to capture the return value, our context manager extends
+# the Python one by implementing a __return__ method that will be called with
+# the return value of the function. The __exit__ method is only called if the
+# function raises an exception.
+#
+# Because CPython 3.11 introduced zero-cost exceptions, we cannot nest try
```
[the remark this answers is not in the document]
P403n1x87: We would have to use different opcodes anyway, so this shouldn't be much of a deal. The code is already written, so there is no need to change that 🙂

PR #9232, opened 2024-05-10
chore(internal): move rate limiter to PyO3

brettlangdon:
This PR sets up a new `ddtrace.internal._core` module which is a Rust module defined in `src/core/` and uses PyO3.

The first component being moved over is the `ddtrace.internal.rate_limiter.RateLimter`.

This is a well isolated component which has minimal need to be in Python. There are clear performance gains from moving this component to native.

The main motivation from this change is to start to build the basis for adding/moving performance critical components to PyO3.

## Risks

This introduces a requirement on the rust compiler for building from source. We have built this into our dev image for awhile, but there are other places where we do not yet have the proper setup and so processes may fail.

For example, the benchmarking platform image does not have rust compiler yet.

## Testing

Since we kept the same public interface as the original Python module, we are using the existing Python tests as the validation of this change. They should be comprehensive enough to validate the new native version.

## Benchmarks

The benchmarking image is not yet updated to include rust compiler, so the following results are from running locally on my machine only.

The new `rate_limiter` benchmark shows a roughly 4x performance improvement.

| Benchmark                   | main | rate_limiter |
|-----------------------------|------|--------------|
| ratelimiter-defaults        | 732 ns                                                                                             | 178 ns: 4.12x faster                                                                               |
| ratelimiter-no_rate_limit   | 267 ns                                                                                             | 174 ns: 1.54x faster                                                                               |
| ratelimiter-low_rate_limit  | 724 ns                                                                                             | 175 ns: 4.13x faster                                                                               |
| ratelimiter-high_rate_limit | 870 ns                                                                                             | 178 ns: 4.89x faster                                                                               |
| ratelimiter-short_window    | 726 ns                                                                                             | 177 ns: 4.10x faster                                                                               |
| ratelimiter-long_window     | 737 ns                                                                                             | 177 ns: 4.16x faster                                                                               |
| Geometric mean              | (ref)                                                                                              | 3.60x faster                                                                                        |

This is a great improvement, however, the rate limiter is such a small portion of an actual trace, because right now it's biggest impact is on starting a new trace and making a sampling decision. So applications with the biggest possible improvement are those which start a lot of small traces at high volume.

Benchmarks for the `span` suite show a minor improvement, this is because the rate limiter today only takes up a small portion of the total lifecycle of a span.

| Benchmark                    | main | rate_limiter |
|------------------------------|------|--------------|
| span-add-metrics             | 39.2 ms                                                                                    | 38.3 ms: 1.02x faster                                                                      |
| span-start-finish            | 16.6 ms                                                                                    | 15.9 ms: 1.05x faster                                                                      |
| span-start-finish-traceid128 | 17.8 ms                                                                                    | 16.5 ms: 1.08x faster                                                                      |
| Geometric mean               | (ref)                                                                                      | 1.02x faster                                                                               |

Benchmarks for the `tracer` suite showing similar results to `span`.

| Benchmark                    | main | rate_limiter |
|------------------------------|------|--------------|
| tracer-small   | 94.0 us                                                                                      | 91.2 us: 1.03x faster                                                                        |
| tracer-medium  | 844 us                                                                                       | 818 us: 1.03x faster                                                                         |
| Geometric mean | (ref)                                                                                        | 1.02x faster                                                                                 |

`tracer-large` results are not show a statistically significant enough difference in overhead.

Benchmarks for the `flask_simple` suite showing almost no improvement at all.

| Benchmark                    | main | rate_limiter |
|------------------------------|------|--------------|
| flasksimple-tracer           | 1.13 ms                                                                                            | 1.15 ms: 1.02x slower                                                                              |
| flasksimple-debugger         | 566 us                                                                                             | 573 us: 1.01x slower                                                                               |
| flasksimple-appsec-post      | 1.03 ms                                                                                            | 1.05 ms: 1.02x slower                                                                              |
| flasksimple-appsec-telemetry | 1.15 ms                                                                                            | 1.17 ms: 1.01x slower                                                                              |
| Geometric mean               | (ref)                                                                                              | 1.01x slower                                                                                       |

## Follow-up future work

- Move the RateLimiter tests over to rust
- Add Rust formatting, static analysis, testing steps to the CI process

## Checklist

## Reviewer Checklist

- [x] Title is accurate


brettlangdon:
I am having an issue with a test, I cannot figure out how to resolve.

https://app.circleci.com/pipelines/github/DataDog/dd-trace-py/61223/workflows/603631bb-3f6b-4b64-8f12-de4bc2a5f46a/jobs/3836322


```
FAILED tests/internal/test_serverless.py::test_slow_imports - ImportError: PyO3 modules compiled for CPython 3.8 or older may only be initialized once per interpreter process
ERROR tests/internal/test_forksafe.py::test_double_fork - RuntimeError: ModuleWatchdog is already installed
ERROR tests/internal/test_forksafe.py::test_rlock_fork - RuntimeError: ModuleWatchdog is already installed
ERROR tests/internal/test_forksafe.py::test_rlock_basic - RuntimeError: ModuleWatchdog is already installed
ERROR tests/internal/test_forksafe.py::test_event_basic - RuntimeError: ModuleWatchdog is already installed
ERROR tests/internal/test_forksafe.py::test_event_fork - RuntimeError: ModuleWatchdog is already installed
ERROR tests/internal/test_forksafe.py::test_forksafe - RuntimeError: ModuleWatchdog is already installed
ERROR tests/internal/test_forksafe.py::test_method_usage - RuntimeError: ModuleWatchdog is already installed
ERROR tests/internal/test_forksafe.py::test_duplicates - RuntimeError: ModuleWatchdog is already installed
ERROR tests/internal/test_forksafe.py::test_lock_fork - RuntimeError: ModuleWatchdog is already installed
ERROR tests/internal/test_forksafe.py::test_hook_exception - RuntimeError: ModuleWatchdog is already installed
ERROR tests/internal/test_forksafe.py::test_lock_basic - RuntimeError: ModuleWatchdog is already installed
ERROR tests/internal/test_forksafe.py::test_registry - RuntimeError: ModuleWatchdog is already installed
```

.gitlab/benchmarks.yml
```diff
@@ -119,6 +119,11 @@ benchmark-http-propagation-inject:
   variables:
     SCENARIO: "http_propagation_inject"
 
+benchmark-rate-limiter:
+  extends: .benchmarks
+  variables:
+    SCENARIO: "rate_limiter"
```
brettlangdon, line 125: this is failing, saying:

> cannot stat '/app/baseline/benchmarks/rate_limiter/.': No such file or directory

probably because it is looking for the scenario on `main` instead of on this branch? 🤔

benchmarks/rate_limiter/scenario.py
```diff
@@ -0,0 +1,18 @@
...
+
+    def run(self):
+        from ddtrace.internal.compat import time_ns
+        from ddtrace.internal.rate_limiter import RateLimiter
+
+        rate_limiter = RateLimiter(rate_limit=self.rate_limit, time_window=self.time_window)
+
+        def _(loops):
+            for _ in range(loops):
+                rate_limiter.is_allowed(time_ns())
```
brettlangdon, line 16: would a fixed set of time stamps make more sense here?

pyproject.toml
```diff
@@ -182,10 +182,12 @@ exclude_dirs = [
 ]
 
 [tool.cibuildwheel]
-build = ["cp27-*", "cp35-*", "cp36-*", "cp37-*", "cp38-*", "cp39-*", "cp310-*", "cp311-*"]
+build = ["cp37-*", "cp38-*", "cp39-*", "cp310-*", "cp311-*", "cp312-*"]
 test-command = ["python {project}/tests/smoke_test.py"]
 # Skip trying to test arm64 builds on Intel Macs
 test-skip = "*-macosx_universal2:arm64"
+# Rust does not provide Cargo for musllinux 32-bit, so that needs to be skipped
+skip = ["*-musllinux_i686"]
```
brettlangdon, line 190:
```suggestion
```

tests/.suitespec.json
```diff
@@ -34,6 +34,7 @@
         ],
         "core": [
             "ddtrace/internal/__init__.py",
+            "ddtrace/internal/_core.pyi",
```
brettlangdon, line 37: do we also need to add `src/core` into here as well? 🤔

pyproject.toml
```diff
@@ -182,7 +182,7 @@ exclude_dirs = [
 ]
 
 [tool.cibuildwheel]
-build = ["cp27-*", "cp35-*", "cp36-*", "cp37-*", "cp38-*", "cp39-*", "cp310-*", "cp311-*"]
+build = ["cp37-*", "cp38-*", "cp39-*", "cp310-*", "cp311-*", "cp312-*"]
```
P403n1x87, line 185: 😮 this probably deserves its own PR

brettlangdon: oh, probably, this is all ignored anyways since we configure cibuildwheel via env variables, I was surprised that this config is out of date and everything was still working... I wonder if we just drop this 🤷🏻

brettlangdon, line 185:
```suggestion
build = ["cp27-*", "cp35-*", "cp36-*", "cp37-*", "cp38-*", "cp39-*", "cp310-*", "cp311-*"]
```

.gitlab/benchmarks.yml
emmettbutler: that makes sense because benchmarks compare against the base branch

benchmarks/rate_limiter/scenario.py
emmettbutler: Randomness can be problematic in benchmarks, so I'd say if you're seeing an opportunity to reduce randomness it's worth exploring.

pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools_scm[toml]>=4", "cython", "cmake>=3.24.2,<3.28; python_version>='3.7'"]
+requires = ["setuptools_scm[toml]>=4", "cython", "cmake>=3.24.2,<3.28; python_version>='3.7'", "setuptools-rust"]
```
emmettbutler, line 2: Pinning the version of `setuptools-rust` may save a future headache.

src/core/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
```
emmettbutler, line 1: Having these files live in `internal/core` would follow a pattern that's used for other subpackages containing native code, making this new code easier to learn. Is that technically straightforward enough to try?

emmettbutler (approved):

src/core/Cargo.toml
brettlangdon: Easy enough to try.

I think the general pattern other repositories take is:

```
- package_name
  -  __init__.py
  - _native.so
- src
  - native.c
- pyproject.toml
```

https://github.com/pola-rs/polars/tree/main/py-polars

Just because others are doing it, doesn't mean we need to.

Eventually I'd like to see core stuff broken out into it's own package, rather than module.

brettlangdon: Something feels weird about it to me, but yes we can. Let me know if you'd prefer me to update it to nest under `internal/core/`

sanchda: FWIW, I tried the native-code-in-the-same-place-as-the-Python-code layout for [some recent changes to the Profiler](https://github.com/DataDog/dd-trace-py/tree/main/ddtrace/internal/datadog/profiling).  I actually don't love how it turned out, although I wasn't able to predict what would happen at the onset.  Knowing what I know now about this exact code, when I refactor this I'd break it up into a `src` pattern unless we have style guidance to the contrary.

Here are some thoughts.
1. Can't ever rely 100% on setup.py and associated fixtures like PyO3 or Pybind or whatever.  In my specific case, I have a native shared library I need to build and distribute (at load-time) to multiple extensions.  That means I need to carry detritus from a build system along.  Keeping the code together is really elegant and beautiful.  Keeping build cruft, scripts for the build, metadata, and other crap just makes it really hairy.
2. I like strong guarantees that I'm distributing only the native artifacts.  Over the course of extension development, I got basically every part of this wrong.  There's still some wrongness.  Having in-tree, non-distributable files is tricky.  In particular, I like _not_ having to tell the build system to exclude things.  In fact, I'd rather tell the build system precisely what I *do* want, since I presume failure to do so will be an error in testing.
3. I like .pyi quite a lot.  I think there's something to be said about specifying the *interface* to native code alongside the Python code, but putting the implementation itself somewhere else.  I think it sets an impractically high bar to assume everyone contributing to the repo is going to be able to grok the native-language-du-jour.  On the other hand, I think setting up explicit contracts at the interface (and documenting them in a .pyi) helps bridge this gap, while also helping us frame precisely what it is the native code is doing in terms of its interactions with Python.
4. Somewhat unrelated, but I've been radicalized to the idea that native components should be very careful in describing any off-GIL behavior they might bring.  In our current universe, we *will* be mixing forks and threads on Linux.  This is a horrible thing.


Opinions weakly held.  I'm only sharing my experiences.

benchmarks/rate_limiter/scenario.py
```diff
@@ -0,0 +1,28 @@
+import bm
+import math
```
brettlangdon, line 2:
```suggestion
import math

import bm
```

src/core/Cargo.toml
brettlangdon: @emmettbutler what do you think?

emmettbutler: I'm convinced on `src`

PR #9233, opened 2024-05-10
fix grpc segfault

wconti27:
Grpc integration was causing segfaults when using async streaming. The segfault was caused by the incorrect pin object use.

## Checklist

## Reviewer Checklist

- [ ] Title is accurate


ddtrace/contrib/grpc/aio_client_interceptor.py
```diff
@@ -160,13 +152,9 @@ async def _wrap_stream_response(
         span: Span,
     ) -> ResponseIterableType:
         try:
+            _handle_add_callback(call, _done_callback(span))
```
majorgreys, line 155: iiuc this change would mean that in the successful iterable response case, the span would not include the status code and details. I would expect us to have a test case that covers that behavior, which we would be breaking. Can we avoid this change to the generated span?

wconti27: I do not think so, after the last response is yielded when iterating the generator, the code after the loop is never executed in our stream wrapper. That means, we can't call the async functions of `code()` and `details()`. 

Our only option is adding the callback before looping through the response generator, but we can't add code and details since they are async and the callback is sync. 

As far as i can tell, there are no streaming spans currently being created (for async client interceptors atleast) within **ALL** of Org 2, so this should not be a breaking change.

wconti27: yes, updated

ddtrace/contrib/grpc/aio_client_interceptor.py
```diff
@@ -52,11 +53,15 @@ def _handle_add_callback(call, callback):
         callback(call)
 
 
-def _done_callback(span, code, details):
-    # type: (Span, grpc.StatusCode, str) -> Callable[[aio.Call], None]
+def _done_callback(span, code=None, details=None):
```
majorgreys, line 56: do we need to even need to pass in code and details? being this is internal, we can change the function signature.

ddtrace/contrib/grpc/aio_client_interceptor.py
```diff
@@ -52,11 +53,15 @@ def _handle_add_callback(call, callback):
...
     def func(call):
-        # type: (aio.Call) -> None
+        nonlocal code, details
```
majorgreys, line 58: if we don't pass these in, we can drop this

ddtrace/contrib/grpc/aio_client_interceptor.py
```diff
@@ -52,11 +53,15 @@ def _handle_add_callback(call, callback):
...
         try:
+            # type: (aio.Call) -> None
+            if not code and not details:
+                # we need to call __repr__ since we cannot call code() or details() since they are both async
+                code, details = parse_rpc_repr_string(call.__repr__())
```
majorgreys, line 63: keep changes internal. I would also catch an exception from this and log the error.

```suggestion
                code, details = _parse_rpc_repr_string(call.__repr__())
```

ddtrace/contrib/grpc/aio_client_interceptor.py
```diff
@@ -95,6 +100,32 @@ async def _handle_cancelled_error(call, span):
     span.finish()
 
 
+def parse_rpc_repr_string(rpc_string):
```
majorgreys, line 103: This might be better moved to utils. Also, we should add unit tests.

ddtrace/contrib/grpc/aio_client_interceptor.py
```diff
@@ -160,13 +191,9 @@ async def _wrap_stream_response(
         span: Span,
     ) -> ResponseIterableType:
         try:
+            _handle_add_callback(call, _done_callback(span))
```
majorgreys, line 194: As we discussed, we could also check if we are done with the iteration, similar to what we have here:

https://github.com/DataDog/dd-trace-py/blob/5148d18ff13f25914ea3bf5efbae9f6695ee3d64/ddtrace/contrib/grpc/client_interceptor.py#L146-L150

If we do that, I wonder if we could check get the code and details when we catch the `StopIteration`. What do you think?

ddtrace/contrib/grpc/aio_client_interceptor.py
wconti27: yes, there are other wrappers that call this function.

ddtrace/contrib/grpc/aio_client_interceptor.py
wconti27: AFAIK only direct reads will result in a `StopIteration` so we cannot always rely on it for code and details. But i can add that behavior at least to catch the direct read case.

ddtrace/contrib/grpc/aio_client_interceptor.py
majorgreys: Ah, I think then it makes more sense here to leave the existing done callback as-is for the unary case, and add the necessary function for the broken streaming case.

wconti27: ok

ddtrace/contrib/grpc/aio_client_interceptor.py
wconti27: ok

tests/contrib/grpc_aio/test_grpc_aio.py
```diff
@@ -1005,3 +1035,130 @@ async def test_client_streaming(server_info, tracer):
...
+    # No error because cancelled after execution
+    _check_client_span(
+        client_span, "grpc-aio-client", "sayHello", "server_streaming", "hellostreamingworld.MultiGreeter"
+    )
+    _check_server_span(
+        server_span, "grpc-aio-server", "sayHello", "server_streaming", "hellostreamingworld.MultiGreeter"
+    )
+
+
+repr_test_cases = [
```
majorgreys, line 1101: good! perhaps useful to also cases where the right-hand side of each equal is missing

tests/contrib/grpc_aio/test_grpc_aio.py
```diff
@@ -1005,3 +1035,130 @@ async def test_client_streaming(server_info, tracer):
...
+            while True:
+                response = await hello_stream.read()
+                if response == grpc.aio.EOF:
+                    break
+                assert response.message == "Hello number {}, will!".format(i)
+                i += 1
+
+
+@pytest.mark.asyncio
+@pytest.mark.skip("Bug/error from grpc when adding an async streaming client interceptor throws StopAsyncIteration")
```
majorgreys, line 1069: Can we link to the upstream issue for this?

tests/contrib/grpc_aio/test_grpc_aio.py
wconti27: ok done

tests/contrib/grpc_aio/test_grpc_aio.py
wconti27: ok done

majorgreys (approved):

PR #9298, opened 2024-05-16
feat(otel): support mapping otel environment variables to ddtrace environment variables

ddmclav:
## Description
This provides mapping support between values set for OTEL environment variables to Datadog equivalent tracing environment variables.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


releasenotes/notes/otel-env-var-support-for-dd-trace-py-d4378445a1405bab.yaml
```diff
@@ -0,0 +1,14 @@
+---
+features:
+  - |
+    otel: This implementation establishes a mapping mechanism between OpenTelemetry (OTEL) environment variables and 
+    Datadog (DD) environment variables with DD variables taking precedence in cases where both are configured:
```
mabdinur, line 5: We can simplify this a bit: 
```suggestion
    tracing: Ensures the following OpenTelemetry environment variables are mapped to an equivalent Datadog configuration (datadog environment variables taking precedence in cases where both are configured):
```

tests/opentelemetry/test_config.py
```diff
@@ -0,0 +1,304 @@
...
+# # MAPPINGS
+#   "OTEL_SERVICE_NAME": "DD_SERVICE",
+#   "OTEL_LOG_LEVEL": "DD_TRACE_DEBUG",
+#   "OTEL_PROPAGATORS": "DD_TRACE_PROPAGATION_STYLE",
+#   "OTEL_TRACES_SAMPLER": "DD_TRACE_SAMPLE_RATE",
+#   "OTEL_TRACES_EXPORTER": "DD_TRACE_ENABLED",
+#   "OTEL_METRICS_EXPORTER": "DD_RUNTIME_METRICS_ENABLED",
+#   "OTEL_LOGS_EXPORTER": "none",  # This should be ignored
+#   "OTEL_RESOURCE_ATTRIBUTES": "DD_TAGS",
+#   "OTEL_SDK_DISABLED": "DD_TRACE_OTEL_ENABLED",
```
mabdinur, line 13: remove
```suggestion
```

tests/opentelemetry/test_config.py
```diff
@@ -0,0 +1,304 @@
...
+
+
+# Passing in DD ENV VARS and OTEL VARS Tests. Should pass with DD Variables.
```
mabdinur, line 16:
```suggestion
```

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+ENV_VAR_MAPPINGS = {
+    "OTEL_SERVICE_NAME": ("DD_SERVICE", _remap_default),
+    "OTEL_LOG_LEVEL": ("DD_TRACE_DEBUG", _remap_otel_log_level),
+    "OTEL_PROPAGATORS": ("DD_TRACE_PROPAGATION_STYLE", _remap_otel_propagators),
+    "OTEL_TRACES_SAMPLER": ("DD_TRACE_SAMPLE_RATE", _remap_traces_sampler),
+    "OTEL_TRACES_EXPORTER": ("DD_TRACE_ENABLED", _remap_traces_exporter),
+    "OTEL_METRICS_EXPORTER": ("DD_RUNTIME_METRICS_ENABLED", _remap_metrics_exporter),
+    "OTEL_LOGS_EXPORTER": ("none", _remap_logs_exporter),  # This should be ignored
```
mabdinur, line 146: This mapping feels weird. Would it make sense to map logs exporter to an empty string. We do not want to set "none" accidentally.  Also can we update the doc string in `_remap_logs_exporter` to state that it does not remap to a Datadog config.

```suggestion
    "OTEL_LOGS_EXPORTER": ("", _remap_logs_exporter),  # This should be ignored
```

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,190 @@
...
+            "deployment.environment",
+            "service.name",
+            "service.version",
+        ]:
+            dd_tags.append("{}:{}".format(key, value))
+
+    if len(otel_user_tag_dict.items()) > 10:
+        ten_tags = dd_tags[:10]
+        log.warning(
+            "To preserve metrics cardinality, only the following first 10 tags have been processed %s",
```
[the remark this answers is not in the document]
ddmclav: Yup! This there's a test case with 11 tags printing out the first 10 that were included.

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,190 @@
...
+            otel_value,
+        )
+    new_value = "False"
+    return new_value
+
+
+def _remap_logs_exporter(otel_value):
+    """Remaps the otel logs exporter to ddtrace logs exporter"""
+    if otel_value != "none":
+        log.warning("Unsupported logs exporter detected.")
```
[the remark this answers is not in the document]
ddmclav: Included this but also added the log line for supported value of none.

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,190 @@
...
+    for tag in otel_tags:
+        tag_pair = tag.split("=")
+        otel_user_tag_dict[tag_pair[0]] = tag_pair[1]
+
+    for otel_key, dd_key in OTEL_UNIFIED_TAG_MAPPINGS.items():
+        if otel_key in otel_user_tag_dict.keys():
+            dd_tags.append("{}:{}".format(dd_key, otel_user_tag_dict[otel_key]))
+
+    for key, value in otel_user_tag_dict.items():
+        if len(dd_tags) < 11 and key not in [
```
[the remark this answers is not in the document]
ddmclav: Added along with logging discarded tags from having more than 10 otel tags

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,190 @@
...
+            "service.name",
+            "service.version",
+        ]:
+            dd_tags.append("{}:{}".format(key, value))
+
+    if len(otel_user_tag_dict.items()) > 10:
+        ten_tags = dd_tags[:10]
+        log.warning(
+            "To preserve metrics cardinality, only the following first 10 tags have been processed %s",
+            ten_tags,
```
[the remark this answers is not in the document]
ddmclav: Logged remaining left out tags

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,190 @@
...
+
+def _remap_otel_sdk_config(otel_value):
+    """Remaps the otel sdk config to ddtrace sdk config"""
+    if otel_value == "false":
+        new_value = "True"
+    elif otel_value == "true":
+        new_value = "False"
+    else:
+        log.warning("Unexpected value for OTEL_SDK_DISABLED.")
+        return otel_value
```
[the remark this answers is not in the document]
ddmclav: From their docs it seems like it strictly takes true or false but will throw in the support for 0 or 1 just in case!:

> Boolean value. If “true”, a no-op SDK implementation will be used for all telemetry signals. Any other value or absence of the variable will have no effect and the SDK will remain enabled. This setting has no effect on propagators configured through the OTEL_PROPAGATORS variable

mabdinur (approved):

tests/opentelemetry/test_config.py
```diff
@@ -0,0 +1,286 @@
...
+    assert config.service == "Test", config.service
+    assert config._debug_mode is True, config._debug_mode
+    assert config._propagation_style_extract == ["tracecontext", "b3"], config._propagation_style_extract
+    assert config._trace_sample_rate == 1.0, config._trace_sample_rate
+    assert config._tracing_enabled is False, config._tracing_enabled
+    assert config._runtime_metrics_enabled is False, config._runtime_metrics_enabled
+    assert config.tags == {
+        "DD_ENV": "prod",
+        "DD_SERVICE": "bleh",
+        "DD_VERSION": "1.0",
```
mabdinur, line 72:
```suggestion
        "env": "prod",
        "service.name": "bleh",
        "version": "1.0",
```

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,184 @@
+import os
+
+from ..internal.logger import get_logger
+
```
mabdinur, line 4:
```suggestion
from ..internal.constants import ENV_KEY
from ..internal.constants import SERVICE_KEY  
from ..internal.constants import VERSION_KEY

```

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,184 @@
...
+
+log = get_logger(__name__)
+
+
+OTEL_UNIFIED_TAG_MAPPINGS = {
+    "deployment.environment": "DD_ENV",
+    "service.name": "DD_SERVICE",
+    "service.version": "DD_VERSION",
```
mabdinur, line 12:
```suggestion
    "deployment.environment":  ENV_KEY,
    "service.name": SERVICE_KEY,
    "service.version": VERSION_KEY,
```

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,184 @@
...
+            log.debug(
+                "Datadog configuration %s is already set. OpenTelemetry configuration will be ignored: %s=%s",
+                dd_env,
+                otel_env,
+                otel_value,
+            )
+            continue
+
+        if otel_env not in ["OTEL_RESOURCE_ATTRIBUTES", "OTEL_SERVICE_NAME"]:
+            # Resource attributes and service name are case-insensitive
```
zacharycmontoya, line 173: It's possible I'm misunderstanding the comment, but the comment and the if check look like they're doing opposite things to me. Can you confirm whether we should be setting `lower()` only when the key is not OTEL_RESOURCE_ATTRIBUTES or OTEL_SERVICE_NAME?

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,184 @@
...
+            if style not in accepted_styles:
+                accepted_styles.append(style)
+        else:
+            log.warning("Following style not supported by ddtrace: %s.", style)
+    return ",".join(accepted_styles)
+
+
+def _remap_traces_sampler(otel_value):
+    """Remaps the otel trace sampler to ddtrace trace sampler"""
+    if otel_value == "always_on" or otel_value == "parentbased_always_on":
```
zacharycmontoya, line 47: There's a subtle add-on we'll need to perform here: `always_on`, `always_off`, and `traceidratio` will also set a _brand new_ config flag called `DD_TRACE_SAMPLE_IGNORE_PARENT` to the value of `true`. There's two changes here that will need to happen (maybe we can do it in a separate PR)

1. The one OTEL env can set multiple DD envs
2. We need to implement and document the new DD config `DD_TRACE_SAMPLE_IGNORE_PARENT`

Let me know if I can help with these

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,184 @@
...
+        return otel_value
+
+
+def _remap_traces_exporter(otel_value):
+    """Remaps the otel trace exporter to ddtrace trace exporter"""
+    if otel_value != "none":
+        log.warning("An unrecognized trace exporter '%s' is being used; setting DD_TRACE_ENABLED to False.", otel_value)
+        return "False"
+    else:
+        return "True"
```
zacharycmontoya, line 63: This appears to be a bug, if the value is `none` we should still be returning `"False"` since the user configured that they wanted to turn off traces

ddtrace/settings/_otel_remapper.py
ddmclav: @zacharycmontoya Yup! Can confirm this is the case; spoke with Munir as well and my thoughts were tags/service names should retain their case sensitivity for the customer's own consistency, while other vars would be "lowered" to for our own consistency

ddtrace/settings/_otel_remapper.py
mabdinur: Good catch. We only support the `none` configuration. All other configurations (ex: otlp,zipkin,console) should not influence ddtrace. Since `DD_TRACE_ENABLED` is `True` by default returning `True` here actually doesn't do anything. 

```suggestion
    return "False"
```

tests/opentelemetry/test_config.py
```diff
@@ -0,0 +1,298 @@
...
+    assert config._tracing_enabled is False, config._tracing_enabled
+    assert config._runtime_metrics_enabled is False, config._runtime_metrics_enabled
+    assert config.tags == {
+        "env": "prod",
+        "testtag1": "random1",
+        "testtag2": "random2",
+        "testtag3": "random3",
+        "testtag4": "random4",
+    }, config.tags
+    assert config.service == "Test"
```
mabdinur, line 77: This is checked twice in the same test. We need to make sure we are not doing this in multiple tests
```suggestion
```

tests/opentelemetry/test_config.py
```diff
@@ -0,0 +1,298 @@
...
+    assert config.version == "1.0"
+    assert config._otel_enabled is True, config._otel_enabled
```
mabdinur, line 79: we should move this up
```suggestion
```

tests/opentelemetry/test_config.py
```diff
@@ -0,0 +1,298 @@
...
+        "OTEL_SDK_DISABLED": "False",
+    },
+    err=b"Following style not supported by ddtrace: jaegar.\n"
+    b"An unrecognized trace exporter 'otlp' is being used; setting DD_TRACE_ENABLED to False.\n"
+    b"Unsupported OTEL logs exporter value detected: warning. Only the 'none' value is supported.\n",
+)
+def test_dd_otel_missing_dd_env_configuration():
+    from ddtrace import config
+
+    assert config.service == "Test", config.service
```
mabdinur, line 64:
```suggestion
    assert config.service == "Test", config.service
    assert config.version == "1.0", config.version
    assert config._otel_enabled is True, config._otel_enabled
```

tests/opentelemetry/test_config.py
```diff
@@ -0,0 +1,298 @@
...
+
+@pytest.mark.subprocess(
+    env={
+        "OTEL_RESOURCE_ATTRIBUTES": "service.version=1.0"
+    },
+)
+def test_otel_resource_attributes_version_tag():
+    from ddtrace import config
+
+    assert config.version == "1.0"
```
mabdinur, line 298:
```suggestion
    assert config.version == "1.0"

```

ddtrace/settings/_otel_remapper.py
ddmclav: Will do this in a separate PR!

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,185 @@
...
```
mabdinur, line 64: According the specification all unrecognized exporters should be logged as warnings. It does not say we should disable tracing if we see an unrecognized exporter. My understanding is that if we detected an unsupported exporter (ex: otlp,zipkin,console) we do nothing and let tracing be enabled by default.

```suggestion
        log.warning("An unrecognized trace exporter '%s' is being used; setting DD_TRACE_ENABLED to False.", otel_value)
        return ""
    else:
        return "False"
```

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,185 @@
...
+    elif otel_value == "always_off" or otel_value == "parentbased_always_off":
+        return "0.0"
+    elif otel_value == "traceidratio" or otel_value == "parentbased_traceidratio":
+        return os.environ.get("OTEL_TRACES_SAMPLER_ARG", "1")
+    else:
+        return otel_value
+
+
+def _remap_traces_exporter(otel_value):
+    """Remaps the otel trace exporter to ddtrace trace exporter"""
```
mabdinur, line 59:
```suggestion
    """Remaps the otel trace exporter to ddtrace enabled"""
```

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,189 @@
...
+    "service.version": VERSION_KEY,
+}
+
+
+def _remap_otel_log_level(otel_value):
+    """Remaps the otel log level to ddtrace log level"""
+    if otel_value == "debug":
+        return "True"
+    elif otel_value == "info":
+        return "False"
```
mabdinur, line 23: ddtrace-py only supports mapping `OTEL_LOG_LEVEL=debug` to `DD_TRACE_DEBUG=true`. All other log levels should not set `DD_TRACE_DEBUG`.

```suggestion
```

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,189 @@
...
+    accepted_styles = []
+    for style in otel_value.split(","):
+        style = style.strip().lower()
+        if style in ["b3", "b3multi", "b3single", "datadog", "tracecontext", "none"]:
+            if style == "b3single":
+                style = "b3"
+            if style not in accepted_styles:
+                accepted_styles.append(style)
+        else:
+            log.warning("Following style not supported by ddtrace: %s.", style)
```
mabdinur, line 43:
```suggestion
    try:
      for style in otel_value.split(","):
          style = style.strip().lower()
          if style in ["b3", "b3multi", "b3single", "datadog", "tracecontext", "none"]:
              if style == "b3single":
                  style = "b3"
              if style not in accepted_styles:
                  accepted_styles.append(style)
          else:
              log.warning("Following style not supported by ddtrace: %s.", style)
    except Exception:
         log.warning("OTEL Propagation styles do not match the expected format: %s. Datadog defaults will be used.", otel_value)
         accepted_styles = []
```

releasenotes/notes/otel-env-var-support-for-dd-trace-py-d4378445a1405bab.yaml
```diff
@@ -0,0 +1,13 @@
...
+    tracing: Ensures the following OpenTelemetry environment variables are mapped to an equivalent Datadog configuration (datadog environment variables taking precedence in cases where both are configured):
+        "OTEL_SERVICE_NAME": "DD_SERVICE",
+        "OTEL_LOG_LEVEL": "DD_TRACE_DEBUG",
+        "OTEL_PROPAGATORS": "DD_TRACE_PROPAGATION_STYLE",
+        "OTEL_TRACES_SAMPLER": "DD_TRACE_SAMPLE_RATE",
+        "OTEL_TRACES_EXPORTER": "DD_TRACE_ENABLED",
+        "OTEL_METRICS_EXPORTER": "DD_RUNTIME_METRICS_ENABLED",
+        "OTEL_LOGS_EXPORTER": "none",
+        "OTEL_RESOURCE_ATTRIBUTES": "DD_TAGS",
+        "OTEL_SDK_DISABLED": "DD_TRACE_OTEL_ENABLED"
```
brettlangdon, line 13: we might want to try a different formatting here, this is how it gets rendered:

- tracing: Ensures the following OpenTelemetry environment variables are mapped to an equivalent Datadog configuration (datadog environment variables taking precedence in cases where both are configured):
  "OTEL_SERVICE_NAME": "DD_SERVICE", "OTEL_LOG_LEVEL": "DD_TRACE_DEBUG", "OTEL_PROPAGATORS": "DD_TRACE_PROPAGATION_STYLE", "OTEL_TRACES_SAMPLER": "DD_TRACE_SAMPLE_RATE", "OTEL_TRACES_EXPORTER": "DD_TRACE_ENABLED", "OTEL_METRICS_EXPORTER": "DD_RUNTIME_METRICS_ENABLED", "OTEL_LOGS_EXPORTER": "none", "OTEL_RESOURCE_ATTRIBUTES": "DD_TAGS", "OTEL_SDK_DISABLED": "DD_TRACE_OTEL_ENABLED"



This alternative suggestion would get rendered like the following:

```rst
tracing: Ensures the following OpenTelemetry environment variables are mapped to an equivalent Datadog configuration (datadog environment variables taking precedence in cases where both are configured)::
      
        OTEL_SERVICE_NAME -> DD_SERVICE
        OTEL_LOG_LEVEL -> DD_TRACE_DEBUG
        OTEL_PROPAGATORS -> DD_TRACE_PROPAGATION_STYLE
        OTEL_TRACES_SAMPLER -> DD_TRACE_SAMPLE_RATE
        OTEL_TRACES_EXPORTER -> DD_TRACE_ENABLED
        OTEL_METRICS_EXPORTER -> DD_RUNTIME_METRICS_ENABLED
        OTEL_LOGS_EXPORTER -> none
        OTEL_RESOURCE_ATTRIBUTES -> DD_TAGS
        OTEL_SDK_DISABLED -> DD_TRACE_OTEL_ENABLED

```





- tracing: Ensures the following OpenTelemetry environment variables are mapped to an equivalent Datadog configuration (datadog environment variables taking precedence in cases where both are configured):

      OTEL_SERVICE_NAME -> DD_SERVICE
      OTEL_LOG_LEVEL -> DD_TRACE_DEBUG
      OTEL_PROPAGATORS -> DD_TRACE_PROPAGATION_STYLE
      OTEL_TRACES_SAMPLER -> DD_TRACE_SAMPLE_RATE
      OTEL_TRACES_EXPORTER -> DD_TRACE_ENABLED
      OTEL_METRICS_EXPORTER -> DD_RUNTIME_METRICS_ENABLED
      OTEL_LOGS_EXPORTER -> none
      OTEL_RESOURCE_ATTRIBUTES -> DD_TAGS
      OTEL_SDK_DISABLED -> DD_TRACE_OTEL_ENABLED

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,187 @@
...
+
+def otel_remapping():
+    """Checks for the existence of both OTEL and Datadog tracer environment variables and remaps them accordingly.
+    Datadog Environment variables take precedence over OTEL, but if there isn't a Datadog value present,
+    then OTEL values take their place.
+    """
+    user_envs = {key.upper(): value for key, value in os.environ.items()}
+
+    for otel_env, otel_value in user_envs.items():
+        if otel_env not in ENV_VAR_MAPPINGS.keys():
```
brettlangdon, line 162:
```suggestion
        if otel_env not in ENV_VAR_MAPPINGS:
```

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,187 @@
...
+    "OTEL_SDK_DISABLED": ("DD_TRACE_OTEL_ENABLED", _remap_otel_sdk_config),
+}
+
+
+def otel_remapping():
+    """Checks for the existence of both OTEL and Datadog tracer environment variables and remaps them accordingly.
+    Datadog Environment variables take precedence over OTEL, but if there isn't a Datadog value present,
+    then OTEL values take their place.
+    """
+    user_envs = {key.upper(): value for key, value in os.environ.items()}
```
brettlangdon, line 159: should they all be uppercase provided by the user anyways?

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,187 @@
...
+        if dd_env in user_envs:
+            log.debug(
+                "Datadog configuration %s is already set. OpenTelemetry configuration will be ignored: %s=%s",
+                dd_env,
+                otel_env,
+                otel_value,
+            )
+            continue
+
+        if otel_env not in ["OTEL_RESOURCE_ATTRIBUTES", "OTEL_SERVICE_NAME"]:
```
brettlangdon, line 175:
```suggestion
        if otel_env not in ("OTEL_RESOURCE_ATTRIBUTES", "OTEL_SERVICE_NAME"):
```

convention, for something immutable/lookup like this can use a tuple (or a set) instead.

this isn't a big deal, since this is a one time startup operation making micro optimizations on performance doesn't matter

if this were code that was run in a hot path at runtime, I'd suggest moving this list to the global scope and have them as a frozenset, since the "not in" lookup time would be the fastest in that case.


(this comment is just for fun/learning, no explicit action is needed to be taken. <- meaning, just read, then hit resolve without accepting/making any changes)

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+ENV_VAR_MAPPINGS = {
+    "OTEL_SERVICE_NAME": ("DD_SERVICE", _remap_default),
+    "OTEL_LOG_LEVEL": ("DD_TRACE_DEBUG", _remap_otel_log_level),
+    "OTEL_PROPAGATORS": ("DD_TRACE_PROPAGATION_STYLE", _remap_otel_propagators),
+    "OTEL_TRACES_SAMPLER": ("DD_TRACE_SAMPLE_RATE", _remap_traces_sampler),
+    "OTEL_TRACES_EXPORTER": ("DD_TRACE_ENABLED", _remap_traces_exporter),
+    "OTEL_METRICS_EXPORTER": ("DD_RUNTIME_METRICS_ENABLED", _remap_metrics_exporter),
+    "OTEL_LOGS_EXPORTER": ("", _remap_logs_exporter),  # Does not set a DDTRACE environment variable.
```
brettlangdon, line 148: does this need to be here if there is no mapping?

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,187 @@
...
+                otel_value,
+            )
+            continue
+
+        if otel_env not in ["OTEL_RESOURCE_ATTRIBUTES", "OTEL_SERVICE_NAME"]:
+            # Resource attributes and service name are case-insensitive
+            otel_value = otel_value.lower()
+
+        mapped_value = otel_config_remapper(otel_value)
+        if mapped_value != "":
```
brettlangdon, line 180: does this need to be explicitly `!= ""` ?

or can we do `if mapped_value:` which is better Python convention.

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,187 @@
...
+            os.environ[dd_env] = mapped_value
```
brettlangdon, line 181: should we be checking for `dd_env` being `""` since we have one mapping which maps to `""`, wouldn't that mean for `OTEL_LOGS_EXPORTER` we are setting `os.environ[""] = value` ?

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+OTEL_UNIFIED_TAG_MAPPINGS = {
+    "deployment.environment": ENV_KEY,
+    "service.name": "service",
+    "service.version": VERSION_KEY,
+}
+
+
+def _remap_otel_log_level(otel_value):
```
brettlangdon, line 18: can we add type hinting to all of these helpers?

it seems they are all `def _remap_*(otel_value: str) -> str:`, correct?

ddtrace/settings/_otel_remapper.py
brettlangdon: I guess this is handled by the remapper, but I guess I still wonder why we are trying to remap it if we aren't going to use it?

brettlangdon: alternative, instead of setting up a remapping for it, we can just add a `if user_envs.get("OTEL_LOGS_EXPORTER", "none") != "none": log.warning(...)` to the start of this remapping function.

it might make the code for this specific env variable more clear.

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,187 @@
...
+            dd_tags,
+            remaining_tags,
+        )
+    return ",".join(dd_tags)
+
+
+def _remap_otel_sdk_config(otel_value):
+    """Remaps the otel sdk config to ddtrace sdk config"""
+    if otel_value == "false":
+        return "True"
```
brettlangdon, line 128: can we get a comment here, I am not sure why false should mean true here?

brettlangdon (commented):
I didn't get a chance to go through all the remap functions yet, but wanted to share the current comments I had.

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,187 @@
...
+    "deployment.environment": ENV_KEY,
+    "service.name": "service",
+    "service.version": VERSION_KEY,
+}
+
+
+def _remap_otel_log_level(otel_value):
+    """Remaps the otel log level to ddtrace log level"""
+    if otel_value == "debug":
+        return "True"
```
brettlangdon, line 21: does the casing here matter?

not sure why `"True"` looks weird to me, maybe because I assume it is supposed to be `True` (boolean constant)

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,187 @@
...
+    else:
+        log.warning(
+            "ddtrace does not support otel log level '%s'. setting ddtrace to log level info.",
```
brettlangdon, line 24: should we state that we only support debug level?

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,187 @@
...
+        else:
+            log.warning("Following style not supported by ddtrace: %s.", style)
+    return ",".join(accepted_styles)
+
+
+def _remap_traces_sampler(otel_value):
+    """Remaps the otel trace sampler to ddtrace trace sampler"""
+    if otel_value in ["always_on", "always_off", "traceidratio"]:
+        log.warning("Trace sampler set to %s; only parent based sampling is supported.", otel_value)
+    if otel_value == "always_on" or otel_value == "parentbased_always_on":
```
brettlangdon, line 49: these could be `if otel_value in ("always_on", "parentbased_always_on"):`

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,187 @@
...
+                accepted_styles.append(style)
+        else:
+            log.warning("Following style not supported by ddtrace: %s.", style)
+    return ",".join(accepted_styles)
+
+
+def _remap_traces_sampler(otel_value):
+    """Remaps the otel trace sampler to ddtrace trace sampler"""
+    if otel_value in ["always_on", "always_off", "traceidratio"]:
+        log.warning("Trace sampler set to %s; only parent based sampling is supported.", otel_value)
```
brettlangdon, line 48: should we state that we are using the parentbased version more explicitly, e.g.

```suggestion
        log.warning("Trace sampler set from %s to parentbased_%s; only parent based sampling is supported.", otel_value, otel_value)
```


which if we do that, we could even add:

```python
otel_value = f"parentbased_{otel_value}"
```

to be more explicit about it, then your checks below only need to care about the `parentbased_*` variants.

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,187 @@
...
+    if otel_value == "always_on" or otel_value == "parentbased_always_on":
+        return "1.0"
+    elif otel_value == "always_off" or otel_value == "parentbased_always_off":
+        return "0.0"
+    elif otel_value == "traceidratio" or otel_value == "parentbased_traceidratio":
+        return os.environ.get("OTEL_TRACES_SAMPLER_ARG", "1")
+    else:
+        return otel_value
```
brettlangdon, line 56: do we need a warning here about unknown value type?

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,187 @@
...
+
+def _remap_metrics_exporter(otel_value):
+    """Remaps the otel metrics exporter to ddtrace metrics exporter"""
+    if otel_value != "none":
+        log.warning(
+            "An unrecognized runtime metrics exporter '%s' is being "
+            "used; setting DD_RUNTIME_METRICS_ENABLED to False.",
+            otel_value,
+        )
+    return "False"
```
brettlangdon, line 77: should this always be `False`?

do you want:

```python
if otel_value != "none":
    return "False"
return ""
```

not sure the log warning lines up with the static return value of `"False"`

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+def _remap_logs_exporter(otel_value):
+    """Logs warning when OTEL Logs exporter is configured. DDTRACE does not support this configuration."""
+    if otel_value != "none":
+        log.warning(
+            "Unsupported OTEL logs exporter value detected: %s. Only the 'none' value is supported.", otel_value
+        )
+        return ""
+    return ""
```
brettlangdon, line 87:
```suggestion
    return ""
```

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,187 @@
...
+            tag_pair = tag.split("=")
+            otel_user_tag_dict[tag_pair[0]] = tag_pair[1]
+
+        for otel_key, dd_key in OTEL_UNIFIED_TAG_MAPPINGS.items():
+            if otel_key in otel_user_tag_dict.keys():
+                dd_tags.append("{}:{}".format(dd_key, otel_user_tag_dict[otel_key]))
+
+        for key, value in otel_user_tag_dict.items():
+            if key.lower() not in OTEL_UNIFIED_TAG_MAPPINGS.keys():
+                if len(dd_tags) < 10:
```
brettlangdon, line 108: you can probably simplify this logic and handle the truncation later.

e.g. only generate the `dd_tags` list, then below:

```python
if len(dd_tags) > 10:
    dd_tags, remaining_tags = dd_tags[:10], dd_tags[10:]
    log.warning("too many tags")
```
it is a small cleanup, but would help reduce the places in the code where you need to know or care about "10".

ddtrace/settings/_otel_remapper.py
brettlangdon: oh, is this because this is "OTEL_SDK_DISABLED" and we are remapping to an "ENABLED" config?

ddtrace/settings/_otel_remapper.py
ddmclav: @brettlangdon Hey Brett! We included that just to be safe in case a user supplied a mis-cased env var spelling-- would it be better to remove?

ddtrace/settings/_otel_remapper.py
ddmclav: @brettlangdon Left in as the mapped function logs a warning if an improper value is passed-- is that alright?

ddtrace/settings/_otel_remapper.py
ddmclav: Hey! Small implementation detail, but: `_remap_logs_exporter` always returns an empty string. It's not a remapper per say, but it's used to detect invalid inputs and log a warning. The empty string env var will not be set. Let me know if that's alright

I renamed _remap_logs_exporter to _validate_logs_exporter to make this more clear

ddmclav: @brettlangdon We're also planning to add support for the logs exporter in the near future (so this will remap to a value later on)

ddtrace/settings/_otel_remapper.py
ddmclav: @brettlangdon This true value will be used to set os.environ[dd_env_variable] = ...

ddmclav: True needs to be a string in this case

ddtrace/settings/_otel_remapper.py
ddmclav: Changed! However we're planning on supporting non-parent based samplers in a future PR, which is why we've written it this way for now

ddtrace/settings/_otel_remapper.py
ddmclav: Done

ddtrace/settings/_otel_remapper.py
brettlangdon: ok, knowing that we will add support for it makes sense to keep a remapper setup (I am ok with the remap name btw)

ddtrace/settings/_otel_remapper.py
ddmclav: Added a warning, however an unsupported value will cause a value error to be raised downstream.

ddtrace/settings/_otel_remapper.py
brettlangdon: no major issue, as long as there is a good reason to do so then 👍🏻

ddtrace/settings/_otel_remapper.py
mabdinur: Yup, I gave suggestion where the unified tags (service, version, env) are always added to the front of the list and the other tags are appended to the end. That way we can always keep the first 10 without losing the important tags. We should also log a warning when we drop tags

ddtrace/settings/_otel_remapper.py
brettlangdon: all good, the reasoning makes sense.

ddtrace/settings/_otel_remapper.py
brettlangdon: 👍🏻 either way would have been fine

ddtrace/settings/_otel_remapper.py
```diff
@@ -66,53 +73,47 @@ def _remap_traces_exporter(otel_value):
...
+            key, value = tag.split("=")
+            otel_user_tag_dict[key] = value
 
         for key, value in otel_user_tag_dict.items():
-            if key.lower() not in OTEL_UNIFIED_TAG_MAPPINGS.keys():
-                if len(dd_tags) < 10:
-                    dd_tags.append("{}:{}".format(key, value))
-                else:
-                    remaining_tags.append("{}:{}".format(key, value))
+            if key.lower() in OTEL_UNIFIED_TAG_MAPPINGS.keys():
```
brettlangdon, line 107:
```suggestion
            if key.lower() in OTEL_UNIFIED_TAG_MAPPINGS:
```

ddtrace/settings/_otel_remapper.py
```diff
@@ -66,53 +73,47 @@ def _remap_traces_exporter(otel_value):
...
+                dd_key = OTEL_UNIFIED_TAG_MAPPINGS[key.lower()]
+                dd_tags = ["{}:{}".format(dd_key, value)] + dd_tags
```
brettlangdon, line 109:
```suggestion
                dd_tags.insert(0, f"{dd_key}:{value}")
```

ddtrace/settings/_otel_remapper.py
```diff
@@ -66,53 +73,47 @@ def _remap_traces_exporter(otel_value):
...
+            else:
+                dd_tags.append("{}:{}".format(key, value))
```
brettlangdon, line 111:
```suggestion
                dd_tags.append(f"{key}:{value}")
```

ddtrace/settings/_otel_remapper.py
brettlangdon: this will help with needlessly creating new Python lists just for the prepend.

brettlangdon (commented):
minor minor nits, looking good

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,188 @@
...
+    if otel_value != "none":
+        log.warning(
+            "Unsupported OTEL logs exporter value detected: %s. Only the 'none' value is supported.", otel_value
+        )
+    return ""
+
+
+def _remap_otel_tags(otel_value: str) -> str:
+    """Remaps the otel tags to ddtrace tags"""
+    dd_tags = []
```
brettlangdon, line 98:
```suggestion
    dd_tags: typing.List[str] = []
```

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,188 @@
...
+
+    try:
+        otel_user_tag_dict = dict()
```
brettlangdon, line 101:
```suggestion
        otel_user_tag_dict: typing.Dict[str, str] = dict()
```

brettlangdon (approved):

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,188 @@
+import os
+import typing
+
+from ..constants import ENV_KEY
+from ..constants import VERSION_KEY
+from ..internal.logger import get_logger
+
```
brettlangdon, line 7:
```suggestion


if sys.version_info >= (3, 8):
    from typing import Literal  # noqa:F401
else:
    from typing_extensions import Literal
```

ddtrace/settings/_otel_remapper.py
```diff
@@ -0,0 +1,188 @@
...
+    if otel_value == "none":
+        return "False"
+    log.warning(
+        "Metrics exporter value is set to unrecognized value: %s.",
+        otel_value,
+    )
+    return ""
+
+
+def _validate_logs_exporter(otel_value: str) -> typing.Literal[""]:
```
brettlangdon, line 87:
```suggestion
def _validate_logs_exporter(otel_value: str) -> Literal[""]:
```

PR #9321, opened 2024-05-20
fix(openai): wrap streamed response

### LINKED ISSUE #9077: Asynchronous context manager crash with langchain-openai >= 0.1.2
### Summary of problem
`ddtrace` breaks streaming with `langchain-openai >= 0.1.2`

2.8.2

23.3.1

The following code will crash:
```
async for message in ChatOpenAI(...).astream("My prompt"):
  print(message)
```

With the following exception:
```
  File "/.venv/lib/python3.12/site-packages/langchain_core/language_models/chat_models.py", line 293, in astream
    async for chunk in self._astream(
  File "/.venv/lib/python3.12/site-packages/langchain_openai/chat_models/base.py", line 560, in _astream
    async with response:
TypeError: 'async_generator' object does not support the asynchronous context manager protocol
```

The root cause appears to be in the way `ChatOpenAI` is patched. The langchain-openai code was recently updated to use context managers:
```
response = await self.async_client.create(messages=message_dicts, **params)
async with response:
  ...
```
(PR: https://github.com/langchain-ai/langchain/pull/18013/files#diff-197e34e7a297cef90049a95d035073870fb484896bab67f800a95fe190ab02bfR560)

However, the tracer returns a type that isn't compatible

Yun-Kim:
Fixes #9077.

This PR changes the way we patch streamed completion/chat responses in the OpenAI integration. Previously we had been returning generator functions that yielded the value of each chunk. However, this was incorrect as `OpenAIStream/OpenAIAsyncStream` objects can be used as context managers as of `openai>=1.6.0`, which means returning a generator function to replace `OpenAIStream` would break user applications that tried to access stream responses as context managers:

```python
with openai.OpenAI().completions.create(..., stream=True) as response:
    for chunk in response:
        print(chunk)
```
There is no introduced/changed functionality other than the fact that we now returned a wrapped `TracedOpenAIStream` object instead of the previous traced generator function.

Note: This PR may seem much larger than it really is, we also updated the tested versions of OpenAI so the riot requirement lockfiles have been changed.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


ddtrace/contrib/openai/utils.py
```diff
@@ -21,6 +22,87 @@
...
+        self._dd_integration = integration
+        self._is_completion = is_completion
+        self._kwargs = kwargs
+
+
+class TracedOpenAIStream(BaseTracedOpenAIStream):
+    def __enter__(self):
+        self.__wrapped__.__enter__()
+        return self
+
```
lievan, line 68: Think we also need
```    
def __next__(self):
        chunk = self.__wrapped__.__next__()
        _loop_handler(self._dd_span, chunk, self._streamed_chunks)
        return chunk
```

lievan (commented):
Looks good!

I think we also need to implement `__next__` in `TracedOpenAIStream/AsyncStream`

Currently getting:
```
    chunk1 = next(resp)
TypeError: 'TracedOpenAIStream' object is not an iterator
```

ddtrace/contrib/openai/utils.py
lievan: + something similar for the async version

lievan (approved):

erikayasuda (approved):

PR #9322, opened 2024-05-20
fix(tracer): ensure all metadata and configs are checked for flare + correct config file

erikayasuda:
## Overview
This PR fixes two bugs with the tracer flare.

First, the RC poller behavior for the tracer flare. Prior to this PR, it would assume that the first product when polling the RC backend would be the target config/metadata we need. However, this is not a safe assumption as multiple products could be updated and polled at once. This, in addition to the fact that currently the tracer flare shares the `AGENT_CONFIG` and `AGENT_TASK` product with the **agent** flare, means that we need to check all products to ensure we don't accidentally skip a tracer flare product.

Second, we were passing in the RC config, not the actual ddtrace config when generating the config file for the flare. This is fixed, and the correct config is being generated now.

## Checklist

## Reviewer Checklist

- [x] Title is accurate

### DIFF
--- a/ddtrace/internal/flare/_subscribers.py
+++ b/ddtrace/internal/flare/_subscribers.py
@@ -1,9 +1,10 @@
 from datetime import datetime
-import os
 from typing import Callable  # noqa:F401
 from typing import Optional  # noqa:F401
 
 from ddtrace.internal.flare.flare import Flare
+from ddtrace.internal.flare.handler import _generate_tracer_flare
+from ddtrace.internal.flare.handler import _prepare_tracer_flare
 from ddtrace.internal.logger import get_logger
 from ddtrace.internal.remoteconfig._connectors import PublisherSubscriberConnector  # noqa:F401
 from ddtrace.internal.remoteconfig._subscribers import RemoteConfigSubscriber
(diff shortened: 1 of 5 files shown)


ddtrace/internal/flare/_subscribers.py
```diff
@@ -61,18 +66,16 @@ def _get_data_from_connector_and_exec(self):
                         "There is already a tracer flare job started at %s. Skipping new request.",
                         str(self.current_request_start),
                     )
-                    return
-                self.current_request_start = datetime.now()
+                    continue
+                if _prepare_tracer_flare(self.flare, configs):
```
erikayasuda, line 70: Don't want to set a timestamp if we just got an agent flare request, not a tracer flare request

ddtrace/internal/flare/flare.py
```diff
@@ -53,8 +54,9 @@ def __init__(
         self.file_handler: Optional[RotatingFileHandler] = None
         self.url: str = trace_agent_url
         self._api_key: Optional[str] = api_key
+        self.ddconfig = ddconfig
```
erikayasuda, line 57: This is the solution I ended up going with, but it also means I won't be pulling the most recent config (just the config when ddtrace starts up). This was the simplest way to get the config so I went with it, but if we want to be pulling the most recent version, we'll need to do some more involved refactoring.

emmettbutler (approved):

PR #9326, opened 2024-05-21
chore(asm): use unpatched open for asm

christophe-papazian:
To improve performances and avoid unwanted interactions with exploit prevention, this PR ensures that we used unpatched open function for loading rules or blocking content for asm.
- add `_unpatched` module to keep track of unpatched references
- move threading to _unpatched
- ensure we use the unpatched version of open for blocking content and waf rule files

## Checklist

## Reviewer Checklist

- [x] Title is accurate

### DIFF
--- a/ddtrace/__init__.py
+++ b/ddtrace/__init__.py
@@ -7,13 +7,7 @@ from ddtrace.internal.module import ModuleWatchdog
 
 ModuleWatchdog.install()
 
-# Acquire a reference to the threading module. Some parts of the library (e.g.
-# the profiler) might be enabled programmatically and therefore might end up
-# getting a reference to the tracee's threading module. By storing a reference
-# to the threading module used by ddtrace here, we make it easy for those parts
-# to get a reference to the right threading module.
-import threading as _threading
-
+import ddtrace.internal._unpatched  # noqa
 from ._logger import configure_ddtrace_logger
 
 
--- a/ddtrace/__init__.py
+++ b/ddtrace/__init__.py
@@ -7,6 +7,7 @@ from ddtrace.internal.module import ModuleWatchdog
 
 ModuleWatchdog.install()
 
+# Ensure we capture references to unpatched modules as early as possible
 import ddtrace.internal._unpatched  # noqa
 from ._logger import configure_ddtrace_logger
 
--- a/ddtrace/internal/_unpatched.py
+++ b/ddtrace/internal/_unpatched.py
@@ -1,0 +1,10 @@
+# Acquire a reference to the open function from the builtins module. This is
+# necessary to ensure that the open function can be used unpatched when required.
+from builtins import open as unpatched_open  # noqa
+
+# Acquire a reference to the threading module. Some parts of the library (e.g.
+# the profiler) might be enabled programmatically and therefore might end up
+# getting a reference to the tracee's threading module. By storing a reference
+# to the threading module used by ddtrace here, we make it easy for those parts
+# to get a reference to the right threading module.
+import threading as _threading  # noqa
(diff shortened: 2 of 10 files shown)


avara1986 (approved):

ddtrace/__init__.py
```diff
@@ -7,13 +7,7 @@
...
 ModuleWatchdog.install()
 
-# Acquire a reference to the threading module. Some parts of the library (e.g.
-# the profiler) might be enabled programmatically and therefore might end up
-# getting a reference to the tracee's threading module. By storing a reference
-# to the threading module used by ddtrace here, we make it easy for those parts
-# to get a reference to the right threading module.
-import threading as _threading
-
+import ddtrace.internal._unpatched  # noqa
```
brettlangdon, line 10: pedantic:

we don't need to do the `_` prefix on the module name since `internal` is already a private namespace (explicitly not covered as part of the public API).

however you are correct that we don't want to import this as not `_` prefixed here.

we could do: `from ddtrace.internal import unpatched as _unpatched` if we don't want to do the `_` prefix for internal usage.

again, pedantic, I truly don't care if we keep it named this way or not, just sharing a different perspective.

ddtrace/internal/_unpatched.py
```diff
@@ -0,0 +1,10 @@
+# Acquire a reference to the open function from the builtins module. This is
+# necessary to ensure that the open function can be used unpatched when required.
+from builtins import open as unpatched_open  # noqa
```
brettlangdon, line 3: do we only need `from builtins import open` here? we don't have to rename, right?

ddtrace/internal/_unpatched.py
```diff
@@ -0,0 +1,10 @@
...
+
+# Acquire a reference to the threading module. Some parts of the library (e.g.
+# the profiler) might be enabled programmatically and therefore might end up
+# getting a reference to the tracee's threading module. By storing a reference
+# to the threading module used by ddtrace here, we make it easy for those parts
+# to get a reference to the right threading module.
+import threading as _threading  # noqa
```
brettlangdon, line 10: would `import threading` be enough here? we did `as _threading` before to make sure to not expose it as part of our public API

🤔

brettlangdon (commented):
some nits about naming, but otherwise looks good

ddtrace/internal/_unpatched.py
christophe-papazian: this one makes life easier because I need to update all `from ddtrace import _threading`

ddtrace/internal/_unpatched.py
christophe-papazian: no we don't, but it's kind of nice to be forced to import a different name as a reminder this is the unpatched version.

brettlangdon: I wonder if we should do the same for threading?

ddtrace/__init__.py
christophe-papazian: I already did that for `ddtrace/internal/_exceptions.py` so I was following the same pattern. 
Main advantage is that it's clear for everyone it's private.

brettlangdon: 🤷🏻 I think "internal" is clear, but like I said. I truly don't care, I'll resolve. Your reasoning and the consistency is good enough.

ddtrace/internal/_unpatched.py
christophe-papazian: This may be done in another refactor? As you want.

brettlangdon: yeah 👍🏻

ddtrace/__init__.py
P403n1x87, line 10: Perhaps we should add a comment here saying that this import acquires the references to the unpatched modules.

brettlangdon:
```suggestion
# Ensure we capture references to unpatched modules as early as possible
import ddtrace.internal._unpatched  # noqa
```

P403n1x87 (approved):

brettlangdon (approved):

gnufede (approved):

PR #9330, opened 2024-05-21
chore(asm): avoid terminate being called on native aspect

juanjux:
## Description

This makes the bytearray extend native aspect and the join aspect raise an exception instead of calling `terminate()` if there is an error when calling `PyObject_CallMethodObjArgs`. 

I checked all the aspects manually and only these ones called `terminate()` when there was an error, all the other ones errors were correctly intercepted by `pybind11` and converted to Python exceptions, so it must be something related to this specific Python C API call.

This also add Python front-ends for all the `os.path` aspects. While these would not call `terminate()` on exception, they would raise a `pybind11` exception which could confuse the users. The front-end aspects will call the original one on any exception, just like the other front-ends do, so if there is a real error the user will receive the right error.

## Checklist

- [X] Change(s) are motivated and described in the PR description
- [X] Testing strategy is described if automated tests are not included in the PR
- [X] Risks are described (performance impact, potential for breakage, maintainability)
- [X] Documentation is included (in-code, generated user docs, [public corp docs](https://github.com/DataDog/documentation/))
- [X] If this PR changes the public interface, I've notified `@DataDog/apm-tees`.

## Reviewer Checklist

- [x] Title is accurate


avara1986 (approved):

avara1986:
LGTM! nice work! :heart:

ddtrace/appsec/_iast/_taint_tracking/aspects.py
```diff
@@ -540,6 +540,86 @@ def format_value_aspect(
...
 
 
+def ospathjoin_aspect(*args, **kwargs):
+    try:
+        return _aspect_ospathjoin(*args, **kwargs)
+    except Exception as e:
+        iast_taint_log_error("IAST propagation error. ospathjoin_aspect. {}".format(e))
+        import os.path
+
+        return os.path.join(*args, **kwargs)
```
gnufede, line 550: Not sure about this approach, what if the `os.path.join` is a mock? I think we should pass the original function here

gnufede (changes requested):

PR #9349, opened 2024-05-22
chore(internal): log a warning if partial flushes has zero spans to send

romainkomorndatadog:
Adds a guard against `on_span_finish()` with partial flushing on running into an `IndexError` because there are no spans to flush (which may happen if `tracer.configure()` was called between the time a span was created and the time it was finished).

## Checklist

## Reviewer Checklist

- [ ] Title is accurate

### DIFF
--- a/ddtrace/_trace/processor/__init__.py
+++ b/ddtrace/_trace/processor/__init__.py
@@ -324,6 +324,14 @@ class SpanAggregator(SpanProcessor):
 
                 if should_partial_flush:
                     log.debug("Partially flushing %d spans for trace %d", num_finished, span.trace_id)
+                    if not finished:
+                        log_msg = "Partial flush triggered but no spans to flush (was tracer reconfigured?)"
+                        # Avoid potential crashes if, for some reason, we have no spans to flush even though
+                        # should_partial_flush is True
+                        if config._telemetry_enabled:
+                            telemetry.telemetry_writer.add_log("WARNING", log_msg)
+                        log.warning(log_msg)
+                        return
                     finished[0].set_metric("_dd.py.partial_flush", num_finished)
 
                 trace.num_finished -= num_finished


ddtrace/_trace/processor/__init__.py
```diff
@@ -324,6 +324,14 @@ def on_span_finish(self, span):
 
                 if should_partial_flush:
                     log.debug("Partially flushing %d spans for trace %d", num_finished, span.trace_id)
+                    if not finished:
+                        log_msg = "Partial flush triggered but no spans to flush (was tracer reconfigured?)"
+                        # Avoid potential crashes if, for some reason, we have no spans to flush even though
+                        # should_partial_flush is True
+                        if config._telemetry_enabled:
```
brettlangdon, line 331: I don't think we need the enabled check here? 🤔

brettlangdon: the telemetry writer will check if it is enabled, and only queue the log message if it is enabled:

https://github.com/DataDog/dd-trace-py/blob/8c2758ad21cbbf24ab5cc3db351cb2b7cad0ff44/ddtrace/internal/telemetry/writer.py#L630

brettlangdon (commented):
can we add a regression test?

the simple case we had should work well:

```python
with tracer.trace("test"):
    tracer.configure(partial_flush_enabled=True, partial_flush_min_spans=1)
```

ddtrace/_trace/processor/__init__.py
romainkomorndatadog: Huh. The import of `telemetry` is conditional over at https://github.com/DataDog/dd-trace-py/pull/9349/files#diff-b767582aad69d1a29223fbc6e78492dbba7096d7cc84e39dd7cde1f7797136e0R31 .

I thought that `if config._telemetry_enabled` is `False`, then `telemetry` wouldn't be imported and available and `telemetry.telemetry_writer.add_log()` would give me a `NameError: name 'telemetry' is not defined`, but.. that doesn't appear to be the case.

romainkomorndatadog: I'm on a roll today. Yes, it is the case. The reason my simple manual test didn't run into that error is precisely because if that `if` statement I added here. 😮‍💨

romainkomorndatadog: Without the `if`, you get:
```
% DD_INSTRUMENTATION_TELEMETRY_ENABLED=false python -c '\
import ddtrace
with ddtrace.tracer.trace("regression"):
    ddtrace.tracer.configure(partial_flush_min_spans=1)
'
error closing trace
  File "/Users/romain.komorn/go/src/github.com/DataDog/dd-trace-py/ddtrace/_trace/span.py", line 627, in __exit__
    self.finish()
  File "/Users/romain.komorn/go/src/github.com/DataDog/dd-trace-py/ddtrace/_trace/span.py", line 293, in finish
    self._finish_ns(time_ns())
  File "/Users/romain.komorn/go/src/github.com/DataDog/dd-trace-py/ddtrace/_trace/span.py", line 306, in _finish_ns
    cb(self)
  File "/Users/romain.komorn/go/src/github.com/DataDog/dd-trace-py/ddtrace/_trace/tracer.py", line 833, in _on_span_finish
    p.on_span_finish(span)
  File "/Users/romain.komorn/go/src/github.com/DataDog/dd-trace-py/ddtrace/_trace/processor/__init__.py", line 333, in on_span_finish
    telemetry.telemetry_writer.add_log("WARNING", log_msg)
NameError: name 'telemetry' is not defined
```

brettlangdon: > Huh. The import of `telemetry` is conditional over at https://github.com/DataDog/dd-trace-py/pull/9349/files#diff-b767582aad69d1a29223fbc6e78492dbba7096d7cc84e39dd7cde1f7797136e0R31 .
> 
> I thought that `if config._telemetry_enabled` is `False`, then `telemetry` wouldn't be imported and available and `telemetry.telemetry_writer.add_log()` would give me a `NameError: name 'telemetry' is not defined`, but.. that doesn't appear to be the case.

oh, I missed that 🤦🏻 

yeah... you are right, the if enabled will need to stay here :/

ddtrace/_trace/processor/__init__.py
```diff
@@ -324,6 +324,14 @@ def on_span_finish(self, span):
 
                 if should_partial_flush:
                     log.debug("Partially flushing %d spans for trace %d", num_finished, span.trace_id)
+                    if not finished:
+                        log_msg = "Partial flush triggered but no spans to flush (was tracer reconfigured?)"
```
brettlangdon, line 328: are we able to include the span repr in the message? that might help us narrow down when/where the bug is being triggered.

e.g. for your testing in CI it was the pytest test suite span.

brettlangdon (commented):
do we want to backport this? I would think no since it isn't a bug fix. if there isn't a release note we cannot release a patch with this anyways, so it would have to wait for another change to be backported.

we might want to change this to a `fix(tracing): do not raise exception if partial flush is triggered without any spans`, add a release note, and then keep the backports. wdyt?

releasenotes/notes/fix-tracing-dont_raise_exception_on_empty_partial_flush-131cd3268101f255.yaml
```diff
@@ -0,0 +1,29 @@
...
+#    * Be concise
+#    * Include actionable steps with the necessary code changes
+#    * Include relevant links (bug issues, upstream issues or release notes, documentation pages)
+#    * Use full sentences with sentence-casing and punctuation.
+#    * Before using Datadog specific acronyms/terminology, a release note must first introduce them with a definition.
+#
+#    Release notes should not:
+#    * Be vague. Example: ``fixes an issue in tracing``.
+#    * Use overly technical language
+#    * Use dynamic links (``stable/latest/1.x`` URLs). Instead, use static links (specific version, commit hash) whenever possible so that they don't break in the future.
```
brettlangdon, line 26:
```suggestion
```

releasenotes/notes/fix-tracing-dont_raise_exception_on_empty_partial_flush-131cd3268101f255.yaml
```diff
@@ -0,0 +1,29 @@
...
+fixes:
+  - |
+    tracing: fixes a potential crash where using partial flushes and tracer.configure() could result in an IndexError
```
brettlangdon, line 29:
```suggestion
    tracing: fixes a potential crash where using partial flushes and ``tracer.configure()`` could result in an IndexError
```

tests/internal/test_settings.py
```diff
@@ -605,3 +606,22 @@ def test_remoteconfig_header_tags(run_python_code_in_subprocess):
         env=env,
     )
     assert status == 0, f"err={err.decode('utf-8')} out={out.decode('utf-8')}"
+
+
+def test_tracer_reconfigure_does_not_crash_tracer(run_python_code_in_subprocess):
+    env = os.environ.copy()
+    env.update({"DD_TRACER_PARTIAL_FLUSH_ENABLED": "true", "DD_TRACER_PARTIAL_FLUSH_MIN_SPANS": "1"})
```
brettlangdon, line 613: _probably_ don't need the env vars?

doing `.configure(partial_flush_enabled=True, partial_flush_min_spans=1)` _should_` be enough?

but all sort of same/same, so this isn't a blocking comment, feel free to resolve.

romainkomorndatadog: 🤔 , yeah I guess not, it's the "incoming" tracer configuration that needs to have partial flush enabled.

releasenotes/notes/fix-tracing-dont_raise_exception_on_empty_partial_flush-131cd3268101f255.yaml
romainkomorndatadog: I literally do this with every single release note. 😮‍💨

ddtrace/_trace/processor/__init__.py
romainkomorndatadog: Test `stderr` says:
```
No spans to flush in partial flush for span: <Span(id=15838273028537712397,trace_id=135991983792686704567357058663853114310,parent_id=None,name=regression)> (was tracer reconfigured?)
```

ddtrace/_trace/processor/__init__.py
```diff
@@ -324,6 +324,14 @@ def on_span_finish(self, span):
...
                 if should_partial_flush:
                     log.debug("Partially flushing %d spans for trace %d", num_finished, span.trace_id)
+                    if not finished:
+                        log_msg = f"No spans to flush in partial flush for span: {span} (was tracer reconfigured?)"
+                        # Avoid potential crashes if, for some reason, we have no spans to flush even though
+                        # should_partial_flush is True.
+                        if config._telemetry_enabled:
+                            telemetry.telemetry_writer.add_log("WARNING", log_msg)
+                        log.warning(log_msg)
+                        return
```
brettlangdon, line 334: thinking about this, do we need to not increment `num_finished` in this case as well?

which makes me wonder if instead we need something like:

```python
if span not in trace:
    log.warning('wtf?')
    return
```

romainkomorndatadog: I think you're right. I didn't realize `trace.num_finished` was just an attribute we increment/decrement "manually".

romainkomorndatadog: I think that makes the original change under `if should_partial_flush:` more or less unreachable but "you never know".

Side note, some tests output:
```
tests/internal/test_settings.py::test_tracer_reconfigure_does_not_crash_tracer
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB set_trace >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /Users/romain.komorn/go/src/github.com/DataDog/dd-trace-py/tests/internal/test_settings.py(624)test_tracer_reconfigure_does_not_crash_tracer()
-> assert status == 0, f"err={err.decode('utf-8')} out={out.decode('utf-8')}"
(Pdb) err
b'No trace found for span <Span(id=3488788091398017934,trace_id=135992766012335207893023319893039562152,parent_id=None,name=regression)> (was tracer reconfigured?)\n'
```

```
tests/internal/test_settings.py::test_tracer_trace_removed_does_not_crash
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB set_trace >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /Users/romain.komorn/go/src/github.com/DataDog/dd-trace-py/tests/internal/test_settings.py(642)test_tracer_trace_removed_does_not_crash()
-> assert status == 0, f"err={err.decode('utf-8')} out={out.decode('utf-8')}"
(Pdb) err
b'No trace found for span <Span(id=9329823571518464311,trace_id=135992766250019695449858101370167084146,parent_id=4517369765126936596,name=regression)> (was tracer reconfigured?)\n'
```

```
tests/internal/test_settings.py::test_tracer_trace_span_removed_does_not_crash
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB set_trace >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /Users/romain.komorn/go/src/github.com/DataDog/dd-trace-py/tests/internal/test_settings.py(660)test_tracer_trace_span_removed_does_not_crash()
-> assert status == 0, f"err={err.decode('utf-8')} out={out.decode('utf-8')}"
(Pdb) err
b'Finished span <Span(id=3480700046215677949,trace_id=135992766170791532931548392515882328551,parent_id=11076758516579201008,name=regression)> not found in trace 135992766170791532931548392515882328551\n'
```

ddtrace/_trace/processor/__init__.py
```diff
@@ -304,7 +304,22 @@ def on_span_finish(self, span):
...
+
+            if span.trace_id not in self._traces:
+                log_msg = f"No trace found for span {span} (was tracer reconfigured?)"
+                if config._telemetry_enabled:
+                    telemetry.telemetry_writer.add_log("WARNING", log_msg)
+                log.warning(log_msg)
+                return
+
             trace = self._traces[span.trace_id]
+            if span not in trace.spans:
```
brettlangdon, line 316: I wonder if this is one of the contributors to the execution time increase on the tracer large benchmark?

ddtrace/_trace/processor/__init__.py
```diff
@@ -304,7 +304,22 @@ def on_span_finish(self, span):
         # type: (Span) -> None
         with self._lock:
             self._span_metrics["spans_finished"][span._span_api] += 1
+
+            if span.trace_id not in self._traces:
```
brettlangdon, line 308: I kept going back and forth myself... should we do `self._traces[span.trace_id].spans.append(span)`, which then would be a trace of one which is then finished, and would cause it to flush immediately.... but.... I can't tell if that is correct or not.

we'd maybe ensure that the span does get encoded and flushed, but should we be caring about spans which we didn't see a span start for?

romainkomorndatadog: I think I don't like that idea because my take is that `on_span_finish()` should be as read-only as possible but, at this point, we're already broken anyway. If the trace was supposed to have more than one span, than we're sending broken data instead of not sending any data. I suppose that's technically better?

brettlangdon: well, if we end up sending a single span, but the trace produces more later, it'll be fine since it all gets stitched together in the backend, we'll just send multiple payloads.

but is there a case where we can end up with a mismatch in spans being tracked and `num_finished` ? e.g. a situation where `num_finished` can never equal `len(trace.spans)` ?

that is sort of the inverse of the bug that is happening.

romainkomorndatadog: Assuming we don't have a race condition somewhere, if the trace isn't in `self._trace`, and we add it, then:
* `trace.num_finished += 1` should mean `trace.num_finished` is now 1
* `trace.num_finished == len(trace.spans)` because we're creating the trace and the span with a single span in it
* `len(finished) == 1`, so `trace.num_finished == len(finished)`
* `len(trace.spans) == 0` so we delete the trace

So by the end of `on_span_finish()` , we still don't have a trace with the span's trace ID, right? So it's like that trace never existed?

brettlangdon: > So by the end of on_span_finish() , we still don't have a trace with the span's trace ID, right? So it's like that trace never existed?

but that is fine, if we create another span for that trace id, then we'll create it in `on_span_start`

ddtrace/_trace/processor/__init__.py
```diff
@@ -324,6 +335,14 @@ def on_span_finish(self, span):
...
                 if should_partial_flush:
                     log.debug("Partially flushing %d spans for trace %d", num_finished, span.trace_id)
+                    if not finished:
+                        log_msg = f"No spans to flush in partial flush for span: {span}"
+                        # Avoid potential crashes if, for some reason, we have no spans to flush even though
+                        # should_partial_flush is True.
+                        if config._telemetry_enabled:
+                            telemetry.telemetry_writer.add_log("WARNING", log_msg)
+                        log.warning(log_msg)
+                        return
```
brettlangdon, line 345: I worry about the state of `num_finished` here, since we incremented it by one at the start.


```python
if num_finished != trace.num_finished:
    log_msg = f"Trying to flush fewer spans than expected {num_finished} != {trace.num_finished} when finishing spans {span}"
    ...

# We are holding a lock, and we removed all the finished spans from `trace`, so there shouldn't be any finished spans left
trace.num_finished = 0

# We checked and logged already
if not finished:
    return

if should_partial_flush:
    finished[0].set_metric("_dd.py.partial_flush", num_finished)
```

What do you think about this general approach/control flow change?

tests/internal/test_settings.py
```diff
@@ -605,3 +606,18 @@ def test_remoteconfig_header_tags(run_python_code_in_subprocess):
...
+            """
+        import ddtrace
+
+        span = ddtrace.tracer.trace("regression")
+        ddtrace.tracer.configure(partial_flush_enabled=1, partial_flush_min_spans=1)
+        span.finish()
+        """
+        ),
+    )
+    assert status == 0, f"err={err.decode('utf-8')} out={out.decode('utf-8')}"
```
brettlangdon, line 623:
```suggestion
@pytest.mark.subprocess()
def test_tracer_reconfigure_does_not_crash_tracer():
    import ddtrace

    span = ddtrace.tracer.trace("regression")
    ddtrace.tracer.configure(partial_flush_enabled=1, partial_flush_min_spans=1)
    span.finish()
```


basically the same thing.

brettlangdon (commented):
one concern, and a few nits/suggestions, but otherwise lgtm

tests/internal/test_settings.py
```diff
@@ -605,3 +605,24 @@ def test_remoteconfig_header_tags(run_python_code_in_subprocess):
         env=env,
     )
     assert status == 0, f"err={err.decode('utf-8')} out={out.decode('utf-8')}"
+
+
+def _stderr_contains_log(stderr: str) -> bool:
+    return stderr.startswith("Finished span not connected to a trace, adding to trace.")
```
romainkomorndatadog, line 611: I think tests on stderr are finicky and end up making life difficult if one of us does something that changes logging (from an unrelated part of the tracer) later on.

IMO, a test on the logger that checks that it was called with the expected args is less flaky and more maintainable.

ddtrace/_trace/processor/__init__.py
```diff
@@ -321,16 +335,29 @@ def on_span_finish(self, span):
                     finished = trace_spans
 
                 num_finished = len(finished)
+                if num_finished != trace.num_finished:
+                    log_msg = f"Finished span count of {num_finished} is not the expected {trace.num_finished}. {span}"
+                    if config._telemetry_enabled:
+                        telemetry.telemetry_writer.add_log("WARNING", log_msg)
+                    log.warning(log_msg)
```
romainkomorndatadog, line 342: Very, very nit-y (and very "6 vs half dozen"-y) : I think I like this better, because it explicitly shows that we **should** be at 0 after we remove the finished spans, and makes the forcing of `trace.num_finished=0` a result of that condition:  
```suggestion
                trace.num_finished -= num_finished           
                if trace.num_finished != 0:
                    log_msg = f"Finished span count of {num_finished} is not the expected {trace.num_finished}. {span}"
                    if config._telemetry_enabled:
                        telemetry.telemetry_writer.add_log("WARNING", log_msg)
                    log.warning(log_msg)
                    trace.num_finished = 0
```
Also because there are deleted lines It didn't let me suggest removing the next 3 lines.

romainkomorndatadog (commented):
Can't approve because it's originally my PR, but 👍

ddtrace/_trace/processor/__init__.py
romainkomorndatadog: (since this is a nit, feel free to resolve)

ddtrace/_trace/processor/__init__.py
```diff
@@ -321,16 +335,29 @@ def on_span_finish(self, span):
...
 
-                if should_partial_flush:
-                    log.debug("Partially flushing %d spans for trace %d", num_finished, span.trace_id)
-                    finished[0].set_metric("_dd.py.partial_flush", num_finished)
-
-                trace.num_finished -= num_finished
+                # We have removed all finished spans from the trace while holding the lock.
+                # There should be no more finished span for this trace.
+                trace.num_finished = 0
```
brettlangdon, line 346:
```suggestion
```

tests/internal/test_settings.py
brettlangdon: could always just make this a `"Finished span not connected ..." in stderr` too.

I'd love to keep this test simple if we can, otherwise the subprocess will need to import mock, mock logging, and then add more asserts.

short term, this seems elegant, but also if you feel strongly then happy to switch it around

romainkomorndatadog: I'd be content with only testing that the process didn't crash (eg: returncode 0).

Maybe do something like
```
"Finished span not connected to a trace, adding to trace." in stderr
```
 instead of `stderr.startswith()`?

That way it's at least less susceptible to something else logging before this hits.

romainkomorndatadog:
> do we want to backport this? I would think no since it isn't a bug fix

We do because #9350 depends on this PR merging, and that is a genuine fix that should be backported.

tests/internal/test_settings.py
brettlangdon, line 611:
```suggestion
    return "Finished span not connected to a trace, adding to trace." in stderr
```

tests/tracer/test_processors.py
```diff
@@ -692,3 +693,72 @@ def on_span_finish(self, span):
     with tracer.trace("test") as span:
         assert span.get_tag("on_start") is None
     assert span.get_tag("on_finish") is None
+
+
+def _stderr_contains_log(stderr: str) -> bool:
+    return stderr.startswith("Finished span not connected to a trace, adding to trace.")
```
brettlangdon, line 699:
```suggestion
    return "Finished span not connected to a trace, adding to trace." in stderr
```

ddtrace/_trace/processor/__init__.py
```diff
@@ -292,18 +292,32 @@ class _Trace(object):
         type=Dict[str, DefaultDict],
     )
 
-    def on_span_start(self, span):
-        # type: (Span) -> None
+    def _on_span_start(self, span: Span) -> None:
+        """This function MUST be called with self._lock held."""
```
emmettbutler, line 296: The strong wording of this note makes me wonder if there's a way to enforce the condition - something like `assert self._lock.held` (not sure if a suitable interface actually exists)

emmettbutler (approved):

ddtrace/_trace/processor/__init__.py
romainkomorndatadog: We could do something like
```suggestion
        """This function MUST be called with self._lock held."""
        if not self._lock.locked():
            log.warning("_on_span_start() called without holding self._lock")
            return
```
(since some people don't like to `assert` in production code)

romainkomorndatadog: Turned out to be a little uglier than I'd like because the `Lock` and `RLock` APIs differ.

brettlangdon (approved):

emmettbutler (approved):

ddtrace/_trace/processor/__init__.py
```diff
@@ -297,18 +297,43 @@ class _Trace(object):
...
         with self._lock:
             self._span_metrics["spans_finished"][span._span_api] += 1
+
+            # Calling finish on a span that we did not see the start for
+            # DEV: This can occur if the SpanAggregator is recreated while there is a span in progress
+            #      e.g. `tracer.configure()` is called after starting a span
+            if span.trace_id not in self._traces:
+                log_msg = f"Finished span not connected to a trace, adding to trace. {span}"
+                if config._telemetry_enabled:
+                    telemetry.telemetry_writer.add_log("WARNING", log_msg)
```
gnufede, line 333:
```suggestion
                    telemetry.telemetry_writer.add_log("WARN", log_msg)
```

ddtrace/_trace/processor/__init__.py
```diff
@@ -326,16 +351,27 @@ def on_span_finish(self, span):
                     finished = trace_spans
 
                 num_finished = len(finished)
+                trace.num_finished -= num_finished
+                if trace.num_finished != 0:
+                    log_msg = f"Finished span count of {num_finished} is not the expected {trace.num_finished}. {span}"
+                    if config._telemetry_enabled:
+                        telemetry.telemetry_writer.add_log("WARNING", log_msg)
```
romainkomorndatadog, line 358:
```suggestion
                        telemetry.telemetry_writer.add_log("WARN", log_msg)
```

emmettbutler:
@romainkomorndatadog the failure of the flask and django system-tests on this branch seems to be reliable

romainkomorndatadog:
@emmettbutler , I'm aware and investigating, yes. Unfortunately, debugging system-tests is a pain.

romainkomorndatadog:
@brettlangdon , creating a trace for the span if there's no existing trace for it is definitely what's causing the system-tests failures here.

I think it's coming from [our instrumentation of `os.fork()`](https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/contrib/subprocess/patch.py#L330). At first glance, it looks like the span's active in both processes (not sure whether it's the parent or child that's missing the trace, though).

romainkomorndatadog:
This repros very simply:
```
import os
import ddtrace

with ddtrace.tracer.trace("gonnafork") as span1:
    os.fork()
```
If you ignore the spam, and run `DD_TRACE_DEBUG=1 ddtrace-run python forktest.py` (so that `os.fork` is instrumented), you end up seeing the span finished twice:
```
finishing span name='gonnafork' id=18102054155643822375 trace_id=136143087665005747000608175049017086130 parent_id=None service=None resource='gonnafork' type=None start=1718367350.831949 end=1718367350.832698 duration=0.000749 error=0 tags={'_dd.p.dm': '-0', '_dd.p.tid': '666c347600000000', 'language': 'python', 'runtime-id': 'f1d810631514472cb6a959cd26bbc0fc'} metrics={'_dd.top_level': 1, '_dd.tracer_kr': 1.0, '_sampling_priority_v1': 1, 'process_id': 19910} (enabled:True)
...
Finished span not connected to a trace, adding to trace. <Span(id=18102054155643822375,trace_id=136143087665005747000608175049017086130,parent_id=None,name=gonnafork)>
finishing span name='gonnafork' id=18102054155643822375 trace_id=136143087665005747000608175049017086130 parent_id=None service=None resource='gonnafork' type=None start=1718367350.831949 end=1718367350.835386 duration=0.003437 error=0 tags={'_dd.p.dm': '-0', '_dd.p.tid': '666c347600000000', 'language': 'python', 'runtime-id': 'f1d810631514472cb6a959cd26bbc0fc'} metrics={'_dd.top_level': 1, '_dd.tracer_kr': 1.0, '_sampling_priority_v1': 1, 'process_id': 19910} (enabled:True)
logs request payload
```

Just to spam myself a little more and validate that it's actually the child process' span that is indeed being traced:
```
prefork os.getpid()=21196 os.getppid()=96204
postfork os.getpid()=21196 os.getppid()=96204
postfork os.getpid()=21219 os.getppid()=21196
Finished span not connected to a trace, adding to trace in os.getpid()=21219. <Span(id=1988567355488963484,trace_id=136143119435498915215208579446138639845,parent_id=None,name=gonnafork)>
```

In the original code, `self._traces[span.trace_id]` would've been empty, so the spans would've never been sent to the writer and we wouldn't duplicates.

@brettlangdon , what do you think of modifying [our os.fork instrumentation](https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/contrib/subprocess/patch.py#L327-L338) to avoid the context manager only start/stop the span in the forking process?

romainkomorndatadog:
Well, just to "mess around", I made some changes to our `os.fork()` wrapper and, at least, tests pass.

There are more issues, though. For one, this only addresses the `os.fork()` span itself, so any other span that wraps the call to `os.fork()` (like my original example) still fails.

I think there are tangible arguments for and against the idea of having the tracer clear all spans post-fork in the child (and we already do this with traces in the `SpanAggregator`, technically), but I think that introduces a bunch of risk (eg: any references to spans in the forked process are still liable to get finished or accessed).

So... maybe the approach of just submitting a new trace for any span that shows up without a `trace_id` is only a good idea in concept.

romainkomorndatadog:
@emmettbutler , @brettlangdon , @erikayasuda , asking for re-review since I'm updating the logic around the `os.fork()` integration. This also probably wants review from someone on the ASM side.

@juanjux , @christophe-papazian , git blame points to the two of you.

romainkomorndatadog:
So after yet another twist and turn... @brettlangdon and I decided that we don't want to alter the behavior of the span aggregator in this PR.

Any spans opened prior to forking could end up getting submitted if they're closed in the forked process (which is not currently the case). Deciding whether that's desirable behavior is outside of the scope of this change.

I'm leaving in the change to the `test_fork()` test just because it bugs me that we're forking and not having the child exit out as soon as it can, though.

emmettbutler (approved):

gnufede (approved):

juanjux (approved):

PR #9410, opened 2024-05-28
chore(ci_visibility): add multiprocessing support to internal coverage collector

romainkomorndatadog:
This adds the ability to collect coverage when using the `multiprocessing` module.

The `BaseProcess` class is patched with a custom `_bootstrap` method that ensures the `ModuleCodeCollector` is installed and enabled, and uses a pipe back to the parent process to send collected coverage at the end of the process' run, and the `close()`, `join()`, `kill()` and `terminate()` methods are all wrapped to ensure that the parent process absorbs the child's coverage data.

The `Process`'s `__init__()` method is also patched to inject a pipe between the parent and child process.

Additionally, support for the `spawn` and `forkserver` start methods relies on overriding the `get_preparation_data()` method and (ab)uses its pickling behavior to trigger the installation of the `ModuleCodeCollector` when the process starts.

No release not as this is still unreleased, and only used for internal development. 

## Checklist

## Reviewer Checklist

- [x] Title is accurate


anmarchenko (approved):

ddtrace/internal/coverage/instrumentation.py
```diff
@@ -2,6 +2,7 @@
 import typing as t
 
 from bytecode import Bytecode
+from bytecode import instr as bytecode_instr
```
P403n1x87, line 5:
```suggestion
from bytecode import Label
```

ddtrace/internal/coverage/instrumentation.py
```diff
@@ -14,6 +15,9 @@ def instrument_all_lines(code: CodeType, hook: HookType, path: str) -> t.Tuple[C
 
     last_lineno = None
     for i, instr in enumerate(abstract_code):
+        if isinstance(instr, bytecode_instr.Label):
```
P403n1x87, line 18:
```suggestion
        if isinstance(instr, Label):
```

P403n1x87: 🤔 Is this check really needed? I'd have thought that a label would cause an exception within the `try` block below.

romainkomorndatadog: It does. I just thought it was better, for code readability, to explicitly ignore types we know we want to ignore, rather than let the exception happen (although in terms of code readability, there was already the comment there 😓 ).

But maybe the `isinstance()` check's more costly, in this case, and since speed is of the essence here, I'm actually introducing a regression?

emmettbutler (approved):

ddtrace/internal/coverage/instrumentation.py
romainkomorndatadog: FWIW, it looks like `isinstance()` is faster:
```
% python -m timeit -n 10000 -s "from bytecode import Label; labels = [Label() for _ in range(1000)]" "
for label in labels:
    if isinstance(label, Label):
        continue
    try:
        if label.lineno is None:
            continue
    except AttributeError:
        pass
"
10000 loops, best of 5: 12.1 usec per loop
```
vs
```
% python -m timeit -n 10000 -s "from bytecode import Label; labels = [Label() for _ in range(1000)]" "
for label in labels:
    try:
        if label.lineno is None:
            continue
    except AttributeError:
        pass
"
10000 loops, best of 5: 243 usec per loop
```

PR #9418, opened 2024-05-28
chore: script that discovers minimum tested versions

emmettbutler:
This script discovers the minimum version of every package referenced in riotfile.py's Venv tree (excluding those envs that do not execute a `pytest` command) and writes that information to a CSV file. It also takes into account the installation dependencies in pyproject.toml.

For use in https://github.com/DataDog/dd-trace-py/pull/9323

## Checklist

## Reviewer Checklist

- [ ] Title is accurate


brettlangdon (commented):
do we need to do this without `riot`?

are we able to use `riot` to generate an artifact during build time? e.g. the `Build deploy` job could generate sdist, wheels, and some other artifacts like supported library versions

then that can be packaged with the other artifacts into the OCI artifact.


(we could even generate the result as Python and have it just embedded into the `sitecustomize.py` file if we don't want to read something from disk)

all_requirements.py
```diff
@@ -0,0 +1,95 @@
+import ast
+import csv
+
+from packaging.version import parse as parse_version
+
+
+def tree_pkgs(node):
```
brettlangdon, line 7: typing in this file could be nice to have.

all_requirements.py
```diff
@@ -0,0 +1,51 @@
...
+
+import riotfile
+
+
+def tree_pkgs_from_riot() -> Dict[str, Set[str]]:
+
+    return _tree_pkgs_from_riot(riotfile.venv)
+
+
+def _tree_pkgs_from_riot(node: riotfile.Venv) -> Dict[str, Set]:
```
brettlangdon, line 15:
```suggestion
def _tree_pkgs_from_riot(node: riotfile.Venv) -> Dict[str, Set[str]]:
```

all_requirements.py
```diff
@@ -0,0 +1,51 @@
...
+    result = {pkg: _format_version_specifiers(set(versions)) for pkg, versions in node.pkgs.items()}
+    for child_venv in node.venvs:
```
brettlangdon, line 17: I think you want `for venv in riotfile.venv.instances():` and then `venv.pkgs`

https://github.com/DataDog/riot/blob/aa599ad0b770d293dca90ae3bd27ed57ba59be6f/riot/riot.py#L265-L292


I _think_ ?

emmettbutler:
@brettlangdon I realized a reason using `import riotfile` isn't ideal: it would require the script to live in the same directory as the riotfile, because the riotfile does not exist inside of a python module with an `__init__` function.

I think this change is stable now.

brettlangdon:
> @brettlangdon I realized a reason using `import riotfile` isn't ideal: it would require the script to live in the same directory as the riotfile, because the riotfile does not exist inside of a python module with an `__init__` function.
> 
> I think this change is stable now.


it just means the script needs to be executed from the root directory no?

we could also just add the root directory to the python path either via shell script or directly in python.

🤷🏻 not trying to argue for one specific way vs the other

scripts/min_compatible_versions.py
```diff
@@ -0,0 +1,59 @@
...
+
+sys.path.append(str(pathlib.Path(__file__).parent.parent.resolve()))
+import riotfile  # noqa:E402
+
+
+OUT_FILENAME = "min_compatible_versions.csv"
+
+
+def _format_version_specifiers(spec: Set[str]) -> Set[str]:
+    return set([part.strip("~<>==") for v in [v.split(",") for v in spec if v] for part in v if "!=" not in part])
```
brettlangdon, line 18: how do we differentiate between `>3.0.0` and `>=3.0.0` ?

brettlangdon: we have some usage which is just `<2.0`, so that gives us an upper bound, not a lower bound.

min_compatible_versions.csv
```diff
@@ -0,0 +1,188 @@
...
+tortoise-orm,0
+typing-extensions,0
+typing_extensions,0
+urllib3,1.0
+uwsgi,0
+vcrpy,4.2.1
+vertica-python,0.6.0
+websockets,11.0
+webtest,0
+werkzeug,1.0
```
brettlangdon, line 184: the actual minimum here is `<1.0`, so "technically" would be `0.1`, but realistically proabbly means `0.16.1`?

brettlangdon (commented):
overall riot parsing looks good, other than I have a question about how to convert specifiers into actual minimum versions

emmettbutler:
@brettlangdon I've adjusted the script to avoid stripping out the specifier information (<, >=, etc) from the version string that it writes to the csv file. Because the determination of "minimum" is done based on naive gt/lt comparison there are probably some edge cases where the wrong thing gets written to the file, but I think that in the majority of cases this approach will be solid enough for use in the single-step guardrail logic.

brettlangdon:
> I think that in the majority of cases this approach will be solid enough for use in the single-step guardrail logic.

@emmettbutler what are you thinking, just parsing the specifiers in SSI and then comparing actual versions against them?

if yes, we need to keep in mind that `packaging` might not be present on the system.

emmettbutler:
@brettlangdon pretty much, yes. It seems to me that including the range markers in the minimum versions file is the best we can do without going to PyPI during that file's creation. I think we can compare version specifiers to the degree necessary without the `packaging` module.

emmettbutler:
@erikayasuda @ZStriker19 @brettlangdon let me know if you'd like any changes to this approach

min_compatible_versions.csv
```diff
@@ -0,0 +1,188 @@
...
+elasticsearch[async],0
+envier,==0.5.1
+exceptiongroup,0
+falcon,~=3.0
+fastapi,~=0.64.0
+flask,~=0.12.0
+flask-caching,~=1.10.0
+flask-login,~=0.6.2
+gevent,~=20.12.0
+git+https://github.com/gnufede/pytest-memray.git@24a3c0735db99eedf57fb36c573680f9bab7cd73,0
```
brettlangdon, line 72: we probably don't want this, right?

seems like we are also getting test dependencies in here, which we might not want because what we require for our tests might not indicate what we are compatible with.

brettlangdon (commented):
lgtm, but I am worried about including out test dependencies, and it isn't clear to me how to identify those from our `riotfile.py` 🤔

min_compatible_versions.csv
emmettbutler: Looking through the riotfile, I'm not even sure that I could accurately distinguish "test requirements" from "what the library is compatible with". Part of the thinking behind this approach is that what our tests install is the definition of what the library is compatible with.

min_compatible_versions.csv
```diff
@@ -0,0 +1,189 @@
...
+async_generator,~=1.10
+asyncpg,~=0.22.0
+asynctest,==0.13.0
+attrs,>=20
+austin-python,~=1.0
+blinker,0
+boto3,0
+botocore,~=1.13
+bottle,>=0.12
+bytecode,0
```
brettlangdon, line 31: here is one, in our tests we have `"bytecode": latest` for an iast test, but in `pyproject.toml` we have:

```
    "bytecode>=0.15.0; python_version>='3.12'",
    "bytecode>=0.14.0; python_version~='3.11.0'",
    "bytecode>=0.13.0; python_version<'3.11'",
```

min_compatible_versions.csv
```diff
@@ -0,0 +1,189 @@
...
+anthropic,~=0.28
+anyio,>=3.4.0
+aredis,0
+asgiref,~=3.0
+astunparse,0
+async_generator,~=1.10
+asyncpg,~=0.22.0
+asynctest,==0.13.0
+attrs,>=20
+austin-python,~=1.0
```
brettlangdon, line 26: this is not actually a dependency we should block on.

we use a version in our test suite for the `profile-diff` job, but it shouldn't matter what version of `austin-python` any of our customers are using.

also the test doesn't really tell us if we are compatible with `austin-python`, that is just the version we use in that test suite.

(same job, but same for `rich`)

PR #9420, opened 2024-05-29
support Anthropic chat models

### LINKED ISSUE #6055: Add Anthropic support similar to OpenAI
### Summary of problem
Thanks for adding OpenAI support, it is awesome!

I am curious if there any plans to add other providers of LLM such as [Anthropic](https://console.anthropic.com/docs/api/reference)?

1.14.0

Not using pip.

poetry==1.4.2

<details>
  
</details>

Anthropic LLM dedicated metrics similar to OpenAI. Ideally with its own dashboard.

wconti27:
Fixes #6055.

## Checklist

## Reviewer Checklist

- [ ] Title is accurate


ddtrace/contrib/anthropic/patch.py
```diff
@@ -0,0 +1,290 @@
...
+
+
+def _format_api_key(api_key: Union[str, SecretStr]) -> str:
+    """Obfuscate a given LLM provider API key by returning the last four characters."""
+    if hasattr(api_key, "get_secret_value"):
+        api_key = api_key.get_secret_value()
+
+    if not api_key or len(api_key) < 4:
+        return ""
+    return "...%s" % api_key[-4:]
```
RohitK89, line 60: Is it deliberate to not use an f-string here?

tests/contrib/anthropic/utils.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+# To (re)-generate the cassettes: pass a real Anthropic API key with
+# ANTHROPIC_API_KEY, delete the old cassettes and re-run the tests.
+# NOTE: be sure to check that the generated cassettes don't contain your
+#       API key. Keys should be redacted by the filter_headers option below.
+# NOTE: that different cassettes have to be used between sync and async
+#       due to this issue: https://github.com/kevin1024/vcrpy/issues/463
+#       between cassettes generated for requests and aiohttp.
+def get_anthropic_vcr(subdirectory_name=""):
```
Yun-Kim, line 22:
```suggestion
def get_request_vcr():
```
Not required to have subdirectory name as we don't have to cover different minor versions of anthropic (which required different sets of cassette files)

tests/snapshots/tests.contrib.anthropic.test_anthropic.test_anthropic_llm_sync.json
```diff
@@ -0,0 +1,40 @@
...
...). It is a foundational element of his philosophy of radical doubt.\\n\\nDescartes sought to find a solid foundation for knowledge by doubting everything that could be doubted. He reasoned that he could doubt the existence of his body, the external world, and even mathematical truths. However, he realized that he could not doubt the existence of his own mind and his own thinking. Even in the act of doubting, he was thinking.\\n\\nFrom this, Descartes concluded that the one thing he could be certain of was his own existence as a thinking being. Even if an evil demon were deceiving him about everything else, it could not deceive him about the fact that he was thinking, and therefore, he must exist.\\n\\nIn other words, \\\"I think, therefore I am\\\" means that the very act of thinking is proof of one's existence. It establishes self-awareness and consciousness as the bedrock of all other knowledge. This principle became a starting point for Descartes' further philosophical explorations and had a significant influence on the development of Western philosophy, particularly in the areas of epistemology (the study of knowledge) and metaphysics (the study of the nature of reality).\"}]",
```
Yun-Kim, line 17: Just an FYI, these hidden tags should not be present in these snapshot files (should disappear when you uncomment the `llmobs_is_pc_sampled(...)` check

Yun-Kim (commented):
Looking great! Added some comments to help clear some context / suggestions.

ddtrace/contrib/anthropic/patch.py
wconti27: Removed

ddtrace/contrib/anthropic/patch.py
```diff
@@ -0,0 +1,278 @@
...
+            if isinstance(message_set, dict):
+                if isinstance(message_set.get("content", None), str):
+                    span.set_tag_str(
+                        "anthropic.request.messages.%d.0.content" % (message_set_idx),
+                        integration.trunc(message_set.get("content", "")),
+                    )
+                    span.set_tag_str(
+                        "anthropic.request.messages.%d.0.message_type" % (message_set_idx),
+                        "text",
+                    )
```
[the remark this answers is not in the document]
wconti27: So the problem here is the `message.content` can be either a str, or a list of messages. See [here for content being an iterable of messages](https://docs.anthropic.com/en/api/messages-examples#:~:text=%5B%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%7B%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%22type%22%3A%20%22image%22%2C%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%22source%22%3A%20%7B%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%22type%22%3A%20%22base64%22%2C%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%22media_type%22%3A%20image_media_type%2C%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%22data%22%3A%20image_data%2C%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%7D%2C%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20%20%20%20%20%20%20%5D), and [here](https://docs.anthropic.com/en/api/messages-examples#:~:text=messages%3D%5B%0A%20%20%20%20%20%20%20%20%7B%22role%22%3A%20%22user%22%2C%20%22content%22%3A%20%22Hello%2C%20Claude%22%7D%0A%20%20%20%20%5D) for a simple string

ddtrace/contrib/anthropic/patch.py
```diff
@@ -0,0 +1,278 @@
...
+                elif isinstance(message_set.get("content", None), Iterable):
```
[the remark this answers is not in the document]
wconti27: I covered this above, links included. A message can include several content prompts as an iterable, see below:

```
    "messages": [
        {"role": "user", "content": [
            {"type": "image", "source": {
                "type": "base64",
                "media_type": "'$IMAGE_MEDIA_TYPE'",
                "data": "'$IMAGE_BASE64'"
            }},
            {"type": "text", "text": "What is in the above image?"}
        ]}
    ]
```

ddtrace/contrib/anthropic/patch.py
```diff
@@ -0,0 +1,278 @@
...
+                                span.set_tag_str(
+                                    "anthropic.request.messages.%d.%d.content" % (message_set_idx, message_idx),
+                                    integration.trunc(json.dumps(message.get("source", ""))),
+                                )
+
+                        span.set_tag_str(
+                            "anthropic.request.messages.%d.%d.message_type" % (message_set_idx, message_idx),
+                            message.get("type", "text"),
+                        )
+        for param, val in kwargs.items():
```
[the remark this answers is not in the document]
wconti27: Yeah there is supported tool functionality (non-beta). I'll get around to that if needed.

ddtrace/llmobs/_integrations/anthropic.py
```diff
@@ -0,0 +1,135 @@
...
+    def _extract_output_message(formatted_response):
+        """Extract output messages from the stored response."""
+        # TO-DO: do we want to truncate these outputs ?
+
+        output_messages = []
+        if isinstance(formatted_response.content, str):
+            return [{"content": formatted_response.content}]
+        if isinstance(formatted_response.content, list):
+            for response in formatted_response.content:
+                if isinstance(response.text, str):
```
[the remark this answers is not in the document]
wconti27: do we want to set the empty text string in this case or not set one at all. Your suggestion is returning a string

ddtrace/llmobs/_integrations/anthropic.py
```diff
@@ -0,0 +1,135 @@
...
+                    output_messages.append({"content": response.text})
+        return output_messages
+
+    # def record_usage(self, span: Span, usage: Dict[str, Any]) -> None:
```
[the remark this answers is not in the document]
wconti27: I added these metrics to spans on my last commit, do I need to set them as LLMObs tags or metrics as well?

ddtrace/contrib/anthropic/patch.py
```diff
@@ -0,0 +1,164 @@
...
+                span.set_tag_str(
+                    "anthropic.response.completions.%d.content" % (idx),
+                    integration.trunc(getattr(chat_completion, "text", "")),
+                )
+            span.set_tag_str(
+                "anthropic.response.completions.%d.type" % (idx),
+                chat_completion.type,
+            )
+        usage = getattr(chat_completions, "usage", {})
+        if usage:
```
Yun-Kim, line 122: To keep things consistent with the openai integration, can we move this code block to `AnthropicIntegration.record_usage(span, usage)`? Thanks!

ddtrace/contrib/anthropic/patch.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+@with_traced_module
+def traced_chat_model_generate(anthropic, pin, func, instance, args, kwargs):
+    chat_messages = get_argument_value(args, kwargs, 0, "messages")
+    integration = anthropic._datadog_integration
+
+    span = integration.trace(
+        pin,
+        "%s.%s" % (instance.__module__, instance.__class__.__name__),
```
Yun-Kim, line 52: If we're planning on tracing both `create()` and `stream()`, can we modify the name to also capture the method that was called?

ddtrace/contrib/anthropic/utils.py
```diff
@@ -0,0 +1,284 @@
...
+log = get_logger(__name__)
+
+
+def handle_stream_response(integration, resp, args, kwargs, span):
+    if _is_async_generator(resp):
+        return TracedOpenAIAsyncStream(resp, integration, span, args, kwargs)
+    elif _is_generator(resp):
+        return TracedOpenAIStream(resp, integration, span, args, kwargs)
+    elif _is_stream_manager(resp):
+        return TracedOpenAIStreamManager(resp, integration, span, args, kwargs)
```
Yun-Kim, line 21: The names of the Traced stream classes might need to be changed with `Anthropic` instead of `OpenAI`

ddtrace/contrib/anthropic/utils.py
```diff
@@ -0,0 +1,284 @@
...
+            _process_finished_stream(
+                self._dd_integration,
+                self._dd_span,
+                self._args,
+                self._kwargs,
+                self._streamed_chunks,
+                self._is_message_stream,
+            )
+            self._dd_span.finish()
+            self._dd_integration.metric(self._dd_span, "dist", "request.duration", self._dd_span.duration_ns)
```
Yun-Kim, line 85:
```suggestion
```

ddtrace/contrib/anthropic/utils.py
```diff
@@ -0,0 +1,284 @@
...
+            raise
+        except Exception:
+            self._dd_span.set_exc_info(*sys.exc_info())
+            self._dd_span.finish()
+            self._dd_integration.metric(self._dd_span, "dist", "request.duration", self._dd_span.duration_ns)
```
Yun-Kim, line 90:
```suggestion
```

ddtrace/contrib/anthropic/utils.py
```diff
@@ -0,0 +1,284 @@
...
+            raise
+
+    def __stream_text__(self):
```
Yun-Kim, line 93: Is this a helper method that doesn't exist in `Anthropic.Stream`? If so, we should rename it to `_stream_text`.

ddtrace/contrib/anthropic/utils.py
```diff
@@ -0,0 +1,284 @@
...
+            _process_finished_stream(
+                self._dd_integration,
+                self._dd_span,
+                self._args,
+                self._kwargs,
+                self._streamed_chunks,
+                self._is_message_stream,
+            )
+            self._dd_span.finish()
+            self._dd_integration.metric(self._dd_span, "dist", "request.duration", self._dd_span.duration_ns)
```
Yun-Kim, line 140:
```suggestion
```

ddtrace/contrib/anthropic/utils.py
```diff
@@ -0,0 +1,284 @@
...
+            raise
+        except Exception:
+            self._dd_span.set_exc_info(*sys.exc_info())
+            self._dd_span.finish()
+            self._dd_integration.metric(self._dd_span, "dist", "request.duration", self._dd_span.duration_ns)
```
Yun-Kim, line 145:
```suggestion
```

ddtrace/contrib/anthropic/utils.py
```diff
@@ -0,0 +1,284 @@
...
+        super().__init__(wrapped)
+        n = kwargs.get("n", 1) or 1
+        self._dd_span = span
+        self._streamed_chunks = [[] for _ in range(n)]
+        self._dd_integration = integration
+        self._kwargs = kwargs
+        self._args = args
+        self._is_message_stream = (
+            is_message_stream  # message stream helper will have accumulated the message for us already
+        )
```
Yun-Kim, line 56: I'm a bit confused on what this does. Could you elaborate what's happening here?

ddtrace/contrib/anthropic/utils.py
```diff
@@ -0,0 +1,284 @@
...
+            self._dd_integration.metric(self._dd_span, "dist", "request.duration", self._dd_span.duration_ns)
+            raise
+        except Exception:
+            self._dd_span.set_exc_info(*sys.exc_info())
+            self._dd_span.finish()
+            self._dd_integration.metric(self._dd_span, "dist", "request.duration", self._dd_span.duration_ns)
+            raise
+
+
+def _construct_unaccumulated_messages(streamed_chunks):
```
Yun-Kim, line 149: Will review this further once I get more clarification on my above comments, thanks

ddtrace/contrib/anthropic/utils.py
```diff
@@ -0,0 +1,284 @@
...
+
+def _is_async_generator(resp):
+    # type: (...) -> bool
+    import anthropic
+
+    if isinstance(resp, AsyncGenerator):
+        return True
+    if hasattr(anthropic, "AsyncStream") and isinstance(resp, anthropic.AsyncStream):
+        return True
+    return False
```
Yun-Kim, line 277: Curious, does Anthropic use AsyncGenerator/Generator objects? From what I've seen it looks like they only use AsyncStream/Stream. This corresponding code from the openai integration was checking for AsyncGenerator/Generators to provide backwards support for older versions, which we don't need to do for anthropic. (unless it does return custom Generator/AsyncGenerator objects)

ddtrace/llmobs/_integrations/anthropic.py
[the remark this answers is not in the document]
Yun-Kim: Ahh you're correct, we would only append here if the response.text is a string. Thanks for catching!

ddtrace/contrib/anthropic/utils.py
```diff
@@ -0,0 +1,284 @@
...
+                    message_usage = message.get("usage", {"output": 0, "input": 0})
+                    message_usage["output"] += getattr(chunk_usage, "output_tokens", 0)
+                    message_usage["input"] += getattr(chunk_usage, "input_tokens", 0)
+                    message["usage"] = message_usage
+                message_finished = True
+
+    return message, message_finished
+
+
+def _tag_streamed_chat_completion_response(integration, span, messages):
```
Yun-Kim, line 234: Do Anthropic streamed responses provide `usage` in the output response?

ddtrace/contrib/anthropic/utils.py
```diff
@@ -0,0 +1,284 @@
...
+    elif _is_generator(resp):
+        return TracedOpenAIStream(resp, integration, span, args, kwargs)
+    elif _is_stream_manager(resp):
+        return TracedOpenAIStreamManager(resp, integration, span, args, kwargs)
+
+
+def _process_finished_stream(integration, span, args, kwargs, streamed_chunks, message_accumulated=False):
+    messages = None
+    messages = kwargs.get("messages", None)
+    chat_messages = get_argument_value(args, kwargs, 0, "messages")
```
Yun-Kim, line 27: Sorry - just realized the openai integration has terrible variable naming. Can we change the names to `input_messages` and `output_messages` to be more clear what we're passing?

Yun-Kim: Honestly it would probably make this a bit cleaner to just pass in `kwargs` to `llmobs_set_tags()` similar to how we do it in the openai integration, since we're not actually doing anything with the input messages in this function.

ddtrace/contrib/anthropic/utils.py
```diff
@@ -0,0 +1,284 @@
...
+            span.set_metric("anthropic.response.completions.%d.usage.output" % idx, message["usage"].get("output", 0))
+
+
+def _loop_handler(span, chunk, streamed_chunks):
+    """Sets the anthropic model tag and appends the chunk to the correct index in the streamed_chunks list.
+
+    When handling a streamed chat/completion response, this function is called for each chunk in the streamed response.
+    """
+    if span.get_tag("anthropic.response.model") is None:
+        span.set_tag("anthropic.response.model", chunk.message.model)
```
Yun-Kim, line 254: Just curious, is this code block necessary / ever hit in Anthropic streamed responses? I remember for OpenAI this was only necessary because we wouldn't get the response model name in the base stream response object and had to infer it from the stream chunks.

ddtrace/llmobs/_integrations/anthropic.py
```diff
@@ -0,0 +1,121 @@
...
+
+class AnthropicIntegration(BaseLLMIntegration):
+    _integration_name = "anthropic"
+
+    def llmobs_set_tags(
+        self,
+        span: Span,
+        input_messages: Optional[Dict[str, Any]] = None,
+        formatted_response: Optional[Dict[str, Any]] = None,
+        err: bool = False,
```
Yun-Kim, line 35: I would advise being consistent with the OpenAIIntegration.llmobs_set_tags(...) function signature, because 1) it'll keep maintainability easier if everything is the same pattern, and 2) keeping both `resp` and `streamed_completions` may be easier to work with since we'll be able know to handle streamed vs non-streamed responses.

ddtrace/llmobs/_integrations/anthropic.py
```diff
@@ -0,0 +1,121 @@
...
+        #     return
+        parameters = {
+            "temperature": float(span.get_tag("anthropic.request.parameters.temperature") or 1.0),
+            "max_tokens": int(span.get_tag("anthropic.request.parameters.max_tokens") or 0),
+        }
+
+        input_messages = self._extract_input_message(input_messages)
+
+        span.set_tag_str(SPAN_KIND, "llm")
+        span.set_tag_str(MODEL_NAME, span.get_tag("anthropic.request.model") or "")
```
Yun-Kim, line 48: In the OpenAI integration we had to save `anthropic.response.model` instead of `anthropic.request.model` because users could technically specify a generic model name i.e. `gpt-3.5-turbo` in the request, while the response would return the specific model name we care more about, i.e. `gpt-3.5-turbo-0901`.
Is this the same case for Anthropic? If so, let's keep that pattern.

ddtrace/llmobs/_integrations/anthropic.py
```diff
@@ -0,0 +1,121 @@
...
+        span.set_tag_str(INPUT_MESSAGES, json.dumps(input_messages))
+        span.set_tag_str(METADATA, json.dumps(parameters))
+        if err or formatted_response is None:
+            span.set_tag_str(OUTPUT_MESSAGES, json.dumps([{"content": ""}]))
+        else:
+            output_messages = self._extract_output_message(formatted_response)
+            span.set_tag_str(OUTPUT_MESSAGES, json.dumps(output_messages))
+
+        # To-Do: record usage here as a metric of some sort?
```
Yun-Kim, line 57: Agreed, you can follow the same pattern as `OpenAIIntegration._set_llmobs_metrics_tags()`. Note though we are setting metric names `prompt_tokens` and `completion_tokens` for Anthropic, not `input_tokens` and `output_tokens` respectively

ddtrace/contrib/anthropic/patch.py
```diff
@@ -0,0 +1,164 @@
...
+                if isinstance(message_set.get("content", None), str):
+                    span.set_tag_str(
+                        "anthropic.request.messages.%d.0.content" % (message_set_idx),
+                        integration.trunc(message_set.get("content", "")),
+                    )
+                    span.set_tag_str(
+                        "anthropic.request.messages.%d.0.type" % (message_set_idx),
+                        "text",
+                    )
+                elif isinstance(message_set.get("content", None), list):
```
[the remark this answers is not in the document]
wconti27: The messages in the first level are previous messages from the conversation (real or artificial).  The messages in the second content level are separate new prompts, though it seems although the model acknowledges both prompts, it returns all responses in one content response.

```
        message = llm.messages.create(
            model="claude-3-opus-20240229",
            max_tokens=1024,
            messages=[
                {"role": "user", "content": "Hello, Claude, start each reply with your name."},
                {"role": "assistant", "content": "Claude: Sure!"},
                {
                    "role": "user",
                    "content": [
                        {"type": "text", "text": "Hello, Can you also address me as Bits in each reply?"},
                        {"type": "text", "text": "Can you explain what Descartes meant by 'I think, therefore I am'?"},
                    ],
                }
            ],
        )
```

reply:
```
Message(id='msg_01PAeWMeKf11fsyRnEDu2eFW', content=[TextBlock(text='Claude: Of course, Bits! Descartes\' famous phrase "I think, ...
```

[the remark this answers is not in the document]
wconti27: And to add, I do think this would be helpful information to include on spans.

ddtrace/contrib/anthropic/patch.py
```diff
@@ -0,0 +1,164 @@
...
+        params_to_tag = {k: v for k, v in kwargs.items() if k != "messages"}
+        span.set_tag_str("anthropic.request.parameters", json.dumps(params_to_tag))
+
+        chat_completions = func(*args, **kwargs)
+        if isinstance(chat_completions, anthropic.Stream) or isinstance(
+            chat_completions, anthropic.lib.streaming._messages.MessageStreamManager
+        ):
+            return handle_stream_response(integration, chat_completions, args, kwargs, span)
+        else:
+            return handle_non_stream_response(integration, chat_completions, chat_messages, span)
```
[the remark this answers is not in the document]
wconti27: For #2, I moved the finally block outside the  `traced_chat_model_generate()` func since we probably don't want to be closing the span for a stream (I guess I could put an if check for a stream, and leave it open if it is)

ddtrace/llmobs/_integrations/anthropic.py
wconti27: This is not the case. It will just throw an error if you try to do that with anthropic.

ddtrace/llmobs/_integrations/anthropic.py
wconti27: okay, makes sense 👍

ddtrace/llmobs/_integrations/anthropic.py
wconti27: So since the model name is consistent, should we only be setting `anthropic.request.model`?

ddtrace/contrib/anthropic/utils.py
wconti27: I don't think so

ddtrace/contrib/anthropic/patch.py
```diff
@@ -0,0 +1,164 @@
...
+            span.set_tag_str(
+                "anthropic.response.completions.%d.type" % (idx),
+                chat_completion.type,
+            )
+        usage = getattr(chat_completions, "usage", {})
+        if usage:
+            input_token = getattr(usage, "input_tokens", 0)
+            output_token = getattr(usage, "output_tokens", 0)
+            span.set_metric(
+                "anthropic.response.completions.usage.input",
```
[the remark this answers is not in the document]
wconti27: wait do we want `input_tokens` or `prompt_tokens`?

ddtrace/contrib/anthropic/patch.py
wconti27: Makes sense to me

ddtrace/contrib/anthropic/patch.py
```diff
@@ -0,0 +1,164 @@
...
+        "%s.%s" % (instance.__module__, instance.__class__.__name__),
+        submit_to_llmobs=True,
+        interface_type="chat_model",
+        provider="anthropic",
+        model=kwargs.get("model", ""),
+        api_key=_extract_api_key(instance),
+    )
+    chat_completions = None
+    try:
+        for message_set_idx, message_set in enumerate(chat_messages):
```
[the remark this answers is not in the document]
wconti27: Changed to message 👍

ddtrace/contrib/anthropic/patch.py
wconti27: Are we talking modify the span name? currently its `anthropic.request`

ddtrace/contrib/anthropic/utils.py
wconti27: Trying to find a better name. When calling `messages.stream()` as opposed to `messages.create(Stream=true)`, Anthropic uses a `messages.Stream` helper class that will have already accumulated the streamed message within the first stream content chunk. It means we don't have to iterate over every stream chunk to build up the full response message.

ddtrace/contrib/anthropic/utils.py
wconti27: Yes, `__stream_text__` is not used by `Anthropic.Stream` but is used when calling the `messages.stream` method and is from a helper class of `Anthropic.messages.Stream`. I needed to keep the same function name since we need to override the method. If we don't override it, then we will iterate though the unwrapped stream `Anthropic.messages.Stream` object, and not our `TracedAnthropicStream` object

ddtrace/contrib/anthropic/utils.py
wconti27: I will have better comments in my newest commit. This is basically for messages where the user did not use any streaming helper methods, and so a message will have to be built up by looking at each streamed chunk.

ddtrace/contrib/anthropic/utils.py
wconti27: I don't think so...? This is from the code I took from openai, so generator can be removed.

ddtrace/contrib/anthropic/utils.py
wconti27: yes

wconti27: do we want to tag usage in the integration or the LLMObs integration?

ddtrace/contrib/anthropic/utils.py
wconti27: Removed the generator checks

ddtrace/contrib/anthropic/patch.py
```diff
@@ -0,0 +1,164 @@
...
+        usage = getattr(chat_completions, "usage", {})
+        if usage:
+            input_token = getattr(usage, "input_tokens", 0)
+            output_token = getattr(usage, "output_tokens", 0)
+            span.set_metric(
+                "anthropic.response.completions.usage.input",
+                input_token,
+            )
+            span.set_metric(
+                "anthropic.response.completions.usage.output",
```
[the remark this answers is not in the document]
wconti27: output_tokens or completion_tokens?

ddtrace/contrib/anthropic/utils.py
wconti27: Disregard the above, I renamed it as requested 👍

ddtrace/contrib/anthropic/utils.py
wconti27: Removed this code

ddtrace/contrib/anthropic/utils.py
wconti27: I've updated this code to be easier to follow.

ddtrace/contrib/anthropic/patch.py
Yun-Kim: We can keep the name to `anthropic.request`, but I mean changing the resource name (to include the name of the method that was called)

ddtrace/contrib/anthropic/patch.py
```diff
@@ -0,0 +1,163 @@
...
+    )
+    chat_completions = None
+    try:
+        for message_idx, message in enumerate(chat_messages):
+            if isinstance(message, dict):
+                if isinstance(message.get("content", None), str):
+                    span.set_tag_str(
+                        "anthropic.request.messages.%d.0.content" % (message_idx),
+                        integration.trunc(message.get("content", "")),
+                    )
```
Yun-Kim, line 69: We should add a `is_pc_sampled_span()` before tagging the message content. We can always tag the `messages.%d.0.type` by default

ddtrace/contrib/anthropic/patch.py
```diff
@@ -0,0 +1,163 @@
...
+                    span.set_tag_str(
+                        "anthropic.request.messages.%d.0.type" % (message_idx),
+                        "text",
+                    )
+                elif isinstance(message.get("content", None), list):
+                    for prompt_idx, prompt in enumerate(message.get("content", [])):
```
Yun-Kim, line 75: According to Anthropic's API docs the individual content objects are referred to as `blocks` ([ref](https://docs.anthropic.com/en/api/messages)). Let's rename `prompt_idx/prompt` to `block_idx/block` to be consistent with their api docs

ddtrace/contrib/anthropic/patch.py
```diff
@@ -0,0 +1,163 @@
...
+                                )
+
+                        span.set_tag_str(
+                            "anthropic.request.messages.%d.%d.type" % (message_idx, prompt_idx),
+                            prompt.get("type", "text"),
+                        )
+                span.set_tag_str(
+                    "anthropic.request.messages.%d.role" % (message_idx),
+                    message.get("role", ""),
+                )
```
Yun-Kim, line 95: When you generate a span with two tags `request.messages.0.0.content` and `request.messages.0.role`, how does it show up on Datadog's span tags UI?

Usually specifying tags to have numeric values inside will make the UI render it as an array, but I'm concerned adding a `role` field will mess up rendering. Would you be able to attach a screenshot? Thanks!

ddtrace/contrib/anthropic/patch.py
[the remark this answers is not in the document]
Yun-Kim: 1. I'm good with leaving the `span.finish()` in the  stream response handler
2. Sorry about the back and forth, we can rename the handlers back to `handle_{}_...`. I didn't realize we were returning the traced streamed response, it makes more sense to go with what you originally named the methods
3. Can we make `handle_non_streamed_response()` not return anything and just perform the tagging? This way we can keep the `span.finish()` in the original `traced_chat_model_generate()`.

ddtrace/contrib/anthropic/utils.py
```diff
@@ -0,0 +1,268 @@
...
+        message["usage"] = message_usage
+
+    return message
+
+
+def _tag_streamed_chat_completion_response(integration, span, message):
+    """Tagging logic for streamed chat completions."""
+    if message is None:
+        return
+    for idx, content in enumerate(message["content"]):
```
Yun-Kim, line 238:
```suggestion
    for idx, block in enumerate(message["content"]):
```

ddtrace/contrib/anthropic/utils.py
```diff
@@ -0,0 +1,268 @@
...
+        span.set_tag_str("anthropic.response.completions.%d.type" % idx, str(integration.trunc(content["type"])))
+        span.set_tag_str("anthropic.response.completions.%d.content" % idx, str(integration.trunc(content["text"])))
```
Yun-Kim, line 240:
```suggestion
        span.set_tag_str("anthropic.response.completions.%d.text" % idx, str(integration.trunc(content["text"])))
```

ddtrace/contrib/anthropic/utils.py
```diff
@@ -0,0 +1,268 @@
...
+        span.set_tag_str("anthropic.response.completions.role", str(message["role"]))
+        if message.get("finish_reason", None) is not None:
```
Yun-Kim, line 242:
```suggestion
        if message.get("finish_reason") is not None:
```
`get()` by default returns None I believe

ddtrace/contrib/anthropic/utils.py
```diff
@@ -0,0 +1,268 @@
...
+    if hasattr(anthropic, "AsyncStream") and isinstance(resp, anthropic.AsyncStream):
+        return True
+    return False
+
+
+def _is_stream_manager(resp):
+    # type: (...) -> bool
+    import anthropic
+
+    return isinstance(resp, anthropic.lib.streaming._messages.MessageStreamManager)
```
Yun-Kim, line 268: https://github.com/anthropics/anthropic-sdk-python/blob/main/src/anthropic/__init__.py#L90 the main anthropic module has a reference to MessageStreamManager
```suggestion
    if hasattr(anthropic, "MessageStreamManager") and isinstance(resp, anthropic.MessageStreamManager):
        return True
    return False
```

ddtrace/llmobs/_integrations/anthropic.py
```diff
@@ -0,0 +1,147 @@
...
+                span.set_tag_str(API_KEY, f"...{str(api_key[-4:])}")
+            else:
+                span.set_tag_str(API_KEY, api_key)
+
+    def _extract_input_message(self, messages):
+        """Extract input messages from the stored prompt.
+        Anthropic allows for messages and multiple texts in a message, which requires some special casing.
+        """
+        if not isinstance(messages, Iterable):
+            log.warning("Anthropic input must be a list of messages.")
```
Yun-Kim, line 87: Do we need to return in this case?

ddtrace/llmobs/_integrations/anthropic.py
```diff
@@ -0,0 +1,147 @@
...
+        for message in messages:
+            if not isinstance(message, dict):
+                log.warning("Anthropic message input must be a list of message param dicts.")
+                continue
+
+            content = message.get("content", None)
+            role = message.get("role", None)
+
+            if role is None or content is None:
+                log.warning("Anthropic input message must have content and role.")
```
Yun-Kim, line 99: Doo we continue in this case?

ddtrace/llmobs/_integrations/anthropic.py
```diff
@@ -0,0 +1,147 @@
...
+
+            if isinstance(content, str):
+                input_messages.append({"content": self.trunc(content), "role": role})
```
Yun-Kim, line 102:
```suggestion
                input_messages.append({"content": content, "role": role})
```
We don't want truncation for llmobs payloads

ddtrace/llmobs/_integrations/anthropic.py
```diff
@@ -0,0 +1,147 @@
...
+
+            elif isinstance(content, list):
+                for entry in content:
```
Yun-Kim, line 105:
```suggestion
                for block in content:
```

ddtrace/llmobs/_integrations/anthropic.py
```diff
@@ -0,0 +1,147 @@
...
+                    if entry.get("type") == "text":
+                        input_messages.append({"content": self.trunc(entry.get("text", "")), "role": role})
```
Yun-Kim, line 107:
```suggestion
                        input_messages.append({"content": entry.get("text", ""), "role": role})
```

ddtrace/llmobs/_integrations/anthropic.py
```diff
@@ -0,0 +1,147 @@
...
+
+        elif isinstance(content, list):
+            for completion in content:
+                text = _get_attr(completion, "text", None)
+                if isinstance(text, str):
+                    output_messages.append({"content": self.trunc(text), "role": role})
+        return output_messages
+
+    def record_usage(self, span: Span, usage: Dict[str, Any]) -> None:
+        if not usage or not self.metrics_enabled:
```
Yun-Kim, line 133:
```suggestion
        if not usage:
```
We don't need to worry about dogstatsd metrics for anthropic

ddtrace/llmobs/_integrations/anthropic.py
```diff
@@ -0,0 +1,147 @@
...
+            return
+        for token_type in ("prompt", "completion"):
+            num_tokens = usage.get(token_type, None)
+            if num_tokens is None:
+                continue
+            span.set_metric("anthropic.response.usage.%s_tokens" % token_type, num_tokens)
```
Yun-Kim, line 139: If we have both prompt and completion token numbers, can we add a third `total_tokens` metric which is just the sum?

ddtrace/contrib/anthropic/patch.py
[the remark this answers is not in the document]
Yun-Kim: Resolved in a separate thread - we should continue to tag the individual content blocks if available.

ddtrace/llmobs/_integrations/anthropic.py
```diff
@@ -0,0 +1,147 @@
...
+            for completion in content:
+                text = _get_attr(completion, "text", None)
+                if isinstance(text, str):
+                    output_messages.append({"content": self.trunc(text), "role": role})
+        return output_messages
+
+    def record_usage(self, span: Span, usage: Dict[str, Any]) -> None:
+        if not usage or not self.metrics_enabled:
+            return
+        for token_type in ("prompt", "completion"):
```
Yun-Kim, line 135:
```suggestion
        for token_type in ("input", "output"):
```

ddtrace/llmobs/_integrations/anthropic.py
```diff
@@ -0,0 +1,147 @@
...
+            span.set_tag_str(OUTPUT_MESSAGES, json.dumps([{"content": ""}]))
+        else:
+            output_messages = self._extract_output_message(resp)
+            span.set_tag_str(OUTPUT_MESSAGES, json.dumps(output_messages))
+
+        usage = _get_attr(resp, "usage")
+        self.record_usage(
+            span,
+            {"prompt": _get_attr(usage, "input_tokens", 0), "completion": getattr(usage, "output_tokens", 0)},
+        )
```
Yun-Kim, line 64: `AnthropicIntegration.record_usage()` should be used to tag span metrics for the input/output/total token counts. We should have a separate `AnthropicIntegration._set_llmobs_metrics_tags()` to return the recorded span metric values, i.e.
```python
def _get_llmobs_metrics_tags(span):
    return {"input_tokens": span.get_metric("anthropic.response.usage.input_tokens"), "output_tokens": ..., "total_tokens": ...}
```
And set that on the span, i.e. `span.set_tag_str(METRICS, json.dumps(self._get_llmobs_metrics_tags(span))`

ddtrace/llmobs/_integrations/anthropic.py
```diff
@@ -0,0 +1,147 @@
...
+from .base import BaseLLMIntegration
+
+
+log = get_logger(__name__)
+
+
+API_KEY = "anthropic.request.api_key"
+MODEL = "anthropic.request.model"
+TOTAL_COST = "anthropic.tokens.total_cost"
+TYPE = "anthropic.request.type"
```
Yun-Kim, line 26: Are these constants used anywhere?

ddtrace/contrib/anthropic/patch.py
```diff
@@ -0,0 +1,163 @@
...
+    except Exception:
+        span.set_exc_info(*sys.exc_info())
+        if integration.is_pc_sampled_llmobs(span):
+            integration.llmobs_set_tags(span, response=None, args=args, kwargs=kwargs, err=bool(span.error))
+        span.finish()
+        raise
+
+
+def tag_non_streamed_response(integration, chat_completions, args, kwargs, span):
+    try:
```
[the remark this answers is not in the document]
wconti27: Removed

ddtrace/contrib/anthropic/patch.py
```diff
@@ -0,0 +1,163 @@
...
+        for idx, chat_completion in enumerate(chat_completions.content):
+            if integration.is_pc_sampled_span(span):
+                span.set_tag_str(
+                    "anthropic.response.completions.%d.content" % (idx),
+                    integration.trunc(str(getattr(chat_completion, "text", ""))),
+                )
```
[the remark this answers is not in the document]
wconti27: done

ddtrace/contrib/anthropic/patch.py
```diff
@@ -0,0 +1,163 @@
...
+            span.set_tag_str(
+                "anthropic.response.completions.%d.type" % (idx),
+                chat_completion.type,
+            )
+
+        # set message level tags
+        if getattr(chat_completions, "stop_reason", None) is not None:
+            span.set_tag_str("anthropic.response.completions.finish_reason", chat_completions.stop_reason)
+        if getattr(chat_completions, "role", None) is not None:
```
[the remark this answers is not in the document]
wconti27: I was just being extra safe, this can be removed

ddtrace/contrib/anthropic/patch.py
```diff
@@ -0,0 +1,163 @@
...
+            span.set_tag_str("anthropic.response.completions.role", chat_completions.role)
+
```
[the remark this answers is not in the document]
wconti27: Done

ddtrace/contrib/anthropic/utils.py
```diff
@@ -0,0 +1,280 @@
...
+                kwargs=kwargs,
+            )
+    except Exception:
+        log.warning("Error processing streamed completion/chat response.", exc_info=True)
+
+
+class BaseTracedAnthropicStream(wrapt.ObjectProxy):
+    def __init__(self, wrapped, integration, span, args, kwargs, _is_message_stream_manager=False):
+        super().__init__(wrapped)
+        n = kwargs.get("n", 1) or 1
```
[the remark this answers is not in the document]
wconti27: removed.

ddtrace/contrib/anthropic/utils.py
```diff
@@ -0,0 +1,280 @@
...
+                self._streamed_chunks,
+                self._is_message_stream_manager,
+            )
+            self._dd_span.finish()
+            raise
+        except Exception:
+            self._dd_span.set_exc_info(*sys.exc_info())
+            self._dd_span.finish()
+            raise
+
```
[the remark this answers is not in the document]
wconti27: yes, was planning to add that later with the async PR. I'll remove the async code from this PR and add it in with the later PR

ddtrace/contrib/anthropic/utils.py
```diff
@@ -0,0 +1,268 @@
...
+        content_text = ""
+        contents = getattr(chunk.message, "content", [])
+        for content in contents:
+            if content.type == "text":
+                content_text += content.text
+                content_type = "text"
+            elif content.type == "image":
+                content_text = "([IMAGE DETECTED])"
+                content_type = "image"
+            message["content"].append({"text": content_text, "type": content_type})
```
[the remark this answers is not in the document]
wconti27: I believe this is true when calling `messages.create(stream = True, ...)`. This is the most basic form of streaming, and doesn't build up a final payload. When calling `messages.stream(...)`, then using `resp.text_stream`, the message start contains a content block that iteratively gets built up by the library for each chunk.

ddtrace/contrib/anthropic/utils.py
```diff
@@ -0,0 +1,268 @@
...
+        message = _extract_from_chunk(chunk, message)
+
+        if "finish_reason" in message:
+            return message
+    return message
+
+
+def _extract_from_chunk(chunk, message={}) -> Tuple[Dict[str, str], bool]:
+    """Constructs a chat message dictionary from streamed chunks given chunk type"""
+    TRANSFORMATIONS_BY_BLOCK_TYPE = {
```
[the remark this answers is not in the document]
wconti27: I'm going to add a To-Do to look into this later, there's not a solid description of what encountering an error block means.

ddtrace/contrib/anthropic/utils.py
```diff
@@ -0,0 +1,268 @@
...
+def _on_message_delta_chunk(chunk, message):
+    # message delta events signal the end of the message
+    if getattr(chunk, "type", "") != "message_delta":
+        return message
+
+    delta_block = getattr(chunk, "delta", "")
+    chunk_finish_reason = getattr(delta_block, "stop_reason", "")
+    if chunk_finish_reason:
+        message["finish_reason"] = chunk_finish_reason
+        message["content"][-1]["text"] = message["content"][-1]["text"].strip()
```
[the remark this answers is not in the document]
wconti27: we were doing something similar in the openai integration once all the content was added.

ddtrace/contrib/anthropic/utils.py
wconti27: `.content` matches the other LLM integrations ... we don't use `.text` at all

ddtrace/contrib/anthropic/utils.py
wconti27: you're right

ddtrace/contrib/anthropic/utils.py
wconti27: Done.

ddtrace/llmobs/_integrations/anthropic.py
wconti27: In this case I would assume that anthropic will be returning an error, and just logging the warning should be fine.

ddtrace/llmobs/_integrations/anthropic.py
wconti27: Same as above, I think we should let whatever error comes from anthropic propagate down and only log a message.

ddtrace/llmobs/_integrations/anthropic.py
wconti27: why? I thought we were going with prompt and completion for the tag names?

ddtrace/llmobs/_integrations/anthropic.py
wconti27: and we set this tag in the LLMObs Integration correct?

ddtrace/llmobs/_integrations/anthropic.py
wconti27: Done

ddtrace/contrib/anthropic/utils.py
```diff
@@ -0,0 +1,268 @@
...
+
+def _construct_message(streamed_chunks):
+    """Iteratively build up a response message from streamed chunks.
+
+    The resulting message dictionary is of form:
+      {"content": [{"type": [TYPE], "text": "[TEXT]"}], "role": "...", "finish_reason": "...", "usage": ...}
+    """
+    message = {"content": []}
+    for chunk in streamed_chunks:
+        message = _extract_from_chunk(chunk, message)
```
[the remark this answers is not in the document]
wconti27: Gonna leave this for the streaming PR work

ddtrace/contrib/anthropic/patch.py
wconti27: oh wait, you are correct, the tags for role do not work. I think i will change the role and other tagging to be set on each content block

PR #9456, opened 2024-06-03
fix(tracing): only extract distributed headers if a trace is not already started

brettlangdon:
Today there are some integrations which always check for distributed tracing headers and if found will activate them as the current/parent context. This blind approach does not take into account if a parent trace has already been started.

This means if you have nested Flask apps (for example), then the child Flask app may make it's parent the upstream remote service, and not the first/initial Flask app in the process. This can result in a broken trace.

The scenario which "works" today which will be broken with this change is if there is "broken" instrumentation as the first span in a trace. This could either be custom instrumentation that did not check for distributed headers, or a built-in integration which is not properly parsing/activating distributed tracing headers.

This change would prioritize keeping the local trace accurate over ensuring a connection with the upstream calling service.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


erikayasuda (approved):
Are we going to backport this to 2.8 and 2.9 (after release)? Or are we introducing it in 2.10 since it would break some cases that currently work?

mabdinur (approved):

ddtrace/contrib/trace_utils.py
```diff
@@ -545,6 +545,17 @@ def activate_distributed_headers(tracer, int_config=None, request_headers=None,
...
+    if current_span:
+        log.debug(
+            "will not extract distributed headers, a Span(trace_id%d, span_id=%d) is already active",
+            current_span.trace_id,
+            current_span.span_id,
+        )
+        return
+
     if override or (int_config and distributed_tracing_enabled(int_config)):
         context = HTTPPropagator.extract(request_headers)
```
mabdinur, line 560:
```suggestion
        context = HTTPPropagator.extract(request_headers)
        current_span = tracer.current_span()
        if current_span and context:
            log.debug(
                "will not activate distributed headers, a Span(trace_id%d, span_id=%d) is already active",
                current_span.trace_id,
                current_span.span_id,
            )
            current_span.link_span(context, {"reason": "terminated_context"})
            return
```
```

File: ddtrace/contrib/trace_utils.py [hunk 2]
```diff
@@ -545,6 +545,17 @@ def activate_distributed_headers(tracer, int_config=None, request_headers=None,
...
+    # DEV: Only do this if there is an active Span, an active Context is fine to override
+    current_span = tracer.current_span()
+    if current_span:
+        log.debug(
+            "will not extract distributed headers, a Span(trace_id%d, span_id=%d) is already active",
+            current_span.trace_id,
+            current_span.span_id,
+        )
+        return
+
```
mabdinur, line 558:
```suggestion
```

mabdinur (approved):
Something like this would be ideal but we will likely need an RFC first. Also this doesn't resolve the performance issues in nested apps.

The current change looks good to me

ddtrace/contrib/trace_utils.py
brettlangdon: this is the main value we are trying to drive with this change, avoid extracting context at all in conditions where we don't need to.

brettlangdon (commented):
per discussion with @tabgok, I have added a "private" FF to control this behavior in case we need to roll back for any customers.

tabgok (approved):

avara1986 (approved):

PR #9457, opened 2024-06-03
feat(asm): standalone asm propagation

gnufede:
ASM: adds standalone asm propagation as described in "RFC: Standalone ASM billing V2"

## Checklist

## Reviewer Checklist

- [ ] Title is accurate

### DIFF
--- a/ddtrace/_trace/processor/__init__.py
+++ b/ddtrace/_trace/processor/__init__.py
@@ -14,7 +14,9 @@ from ddtrace import config
 from ddtrace._trace.span import Span  # noqa:F401
 from ddtrace._trace.span import _get_64_highest_order_bits_as_hex
 from ddtrace._trace.span import _is_top_level
+from ddtrace.appsec._constants import APPSEC
 from ddtrace.constants import _APM_ENABLED_METRIC_KEY as MK_APM_ENABLED
+from ddtrace.constants import MANUAL_DROP_KEY
 from ddtrace.constants import SAMPLING_PRIORITY_KEY
 from ddtrace.constants import USER_KEEP
 from ddtrace.internal import gitmetadata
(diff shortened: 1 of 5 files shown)


ddtrace/_trace/tracer.py
```diff
@@ -779,7 +779,9 @@ def _start_span(
             if span._local_root is None:
                 span._local_root = span
             for k, v in _get_metas_to_propagate(context):
-                if k != SAMPLING_DECISION_TRACE_TAG_KEY:
+                from ddtrace.appsec._constants import APPSEC
```
ZStriker19, line 782: probably don't need to import in the for loop?

ddtrace/_trace/processor/__init__.py
```diff
@@ -160,6 +162,10 @@ def process_trace(self, trace):
 
             if self.apm_opt_out:
                 chunk_root.set_metric(MK_APM_ENABLED, 0)
+                if APPSEC.PROPAGATION_HEADER not in chunk_root.get_tags():
```
ZStriker19, line 165: Would you want to move this code chunk below, after we sample the trace, right before we return it? Otherwise we may (outgoing request sent earlier causes sampling to run) more may not have a sampling_priority on the span. It seems like it could vary what `root_ctx._previous_sampling_priority` returns.

ddtrace/propagation/http.py
```diff
@@ -241,10 +243,21 @@ def _inject(span_context, headers):
...
         sampling_priority = span_context.sampling_priority
+
+        if (
+            asm_config._appsec_standalone_enabled
+            and APPSEC.PROPAGATION_HEADER not in span_context._meta
+            and span_context._previous_sampling_priority
+            and sampling_priority is not None
+            and sampling_priority < span_context._previous_sampling_priority
+        ):
+            headers[HTTP_HEADER_SAMPLING_PRIORITY] = str(span_context._previous_sampling_priority)
```
ZStriker19, line 256: Can we add a comment to explain what/why we're doing this? I'm a bit confused by the switching out of sampling_priority values.

gnufede: Yeah, sorry it's better explained in the RFC, but long story short, the goal would be to drop the current local span while keep propagating the original sampling priority.

Since setting the MANUAL_DROP tag mainly changes the sampling priority, we need to store the previous value somewhere and use it just when injecting the new headers.

ddtrace/_trace/tracer.py
gnufede: definitely, sorry I was just testing if this approach would work, moving the import to the top!

juanjux (approved):

gnufede:
Closing in favor of #9482

PR #9461, opened 2024-06-03
feat(tracing): add anthropic integration patching

### LINKED ISSUE #6055: Add Anthropic support similar to OpenAI

wconti27:
Fixes #6055.

Adds Datadog Integration support for Anthropic chat models. This PR adds support for basic Anthropic chats.

Note: Anthropic officially has support for Python 3.7, but their latest 0.28 release requires `jiter>0.4` which does not have support for Python 3.7. For this reason we will only officially support Anthropic with Python 3.8+.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


ddtrace/contrib/anthropic/__init__.py
```diff
@@ -0,0 +1,17 @@
+"""
+Do later.
```
Yun-Kim, line 2: We'll need a brief documentation about the Anthropic integration, including:
- Summary about what it traces, similar to the OpenAI integration docs
- Prompt and completion sampling (similar to OpenAI docs)
- streamed responses support section (once we add that in, not necessarily for this PR)
- Enabling (similar to openai docs)
- Global configuration (service, span char limit, pc sample rate)
- Instance configuration (similar to openai docs)

ddtrace/contrib/anthropic/patch.py
```diff
@@ -0,0 +1,158 @@
...
+config._add(
+    "anthropic",
+    {
+        "span_prompt_completion_sample_rate": float(os.getenv("DD_ANTHROPIC_SPAN_PROMPT_COMPLETION_SAMPLE_RATE", 1.0)),
+        "span_char_limit": int(os.getenv("DD_ANTHROPIC_SPAN_CHAR_LIMIT", 128)),
+    },
+)
+
+
+def _extract_api_key(instance: Any) -> str:
```
Yun-Kim, line 38:
```suggestion
def _extract_api_key(instance: Any) -> Optional[str]:
```

ddtrace/contrib/anthropic/patch.py
```diff
@@ -0,0 +1,158 @@
...
+@with_traced_module
+def traced_chat_model_generate(anthropic, pin, func, instance, args, kwargs):
+    chat_messages = get_argument_value(args, kwargs, 0, "messages")
+    integration = anthropic._datadog_integration
+
+    operation_name = func.__name__
+
+    span = integration.trace(
+        pin,
+        "%s.%s.%s" % (instance.__module__, instance.__class__.__name__, operation_name),
```
Yun-Kim, line 57: Do you think the complete path (module, class, operation) may be excessive for naming spans? I'm happy to go with just class name + operation name, but up to you.

ddtrace/contrib/anthropic/patch.py
```diff
@@ -0,0 +1,158 @@
...
+        interface_type="chat_model",
+        provider="anthropic",
+        model=kwargs.get("model", ""),
+        api_key=_extract_api_key(instance),
+    )
+
+    chat_completions = None
+    try:
+        for message_idx, message in enumerate(chat_messages):
+            if isinstance(message, dict):
```
Yun-Kim, line 68: Can a message object be a different type than `dict`? I'm wondering if we need this check. If we do, can we minimize indentation by doing a `if not isinstance(message, dict): continue` instead?

ddtrace/contrib/anthropic/patch.py
```diff
@@ -0,0 +1,158 @@
...
+                if isinstance(message.get("content", None), str):
+                    if integration.is_pc_sampled_span(span) and message.get("content", "") != "":
```
Yun-Kim, line 70: Why do we need to make the check that the content is an empty string?

ddtrace/contrib/anthropic/patch.py
```diff
@@ -0,0 +1,158 @@
...
+                            integration.trunc(message.get("content", "")),
+                        )
+                    span.set_tag_str(
+                        "anthropic.request.messages.%d.content.0.type" % (message_idx),
+                        "text",
+                    )
+                elif isinstance(message.get("content", None), list):
+                    for block_idx, block in enumerate(message.get("content", [])):
+                        if integration.is_pc_sampled_span(span):
+                            if block.get("type", None) == "text" and block.get("text", "") != "":
```
Yun-Kim, line 82: Same here, what if a user explicitly passes in an empty string? We should still tag that (unless this breaks something I wasn't aware of)

ddtrace/contrib/anthropic/patch.py
```diff
@@ -0,0 +1,158 @@
...
+                span.set_tag_str(
+                    "anthropic.request.messages.%d.role" % (message_idx),
+                    message.get("role", ""),
+                )
+        params_to_tag = {k: v for k, v in kwargs.items() if k != "messages"}
+        span.set_tag_str("anthropic.request.parameters", json.dumps(params_to_tag))
+
+        chat_completions = func(*args, **kwargs)
+
+        handle_non_streamed_response(integration, chat_completions, args, kwargs, span)
```
Yun-Kim, line 106: Just curious, how does a streamed response get handled in this case? I know we're adding support for that in another PR but wanted to be confident this won't break streamed responses

Yun-Kim (commented):
A couple minor comments but this looks great!

ddtrace/contrib/anthropic/patch.py
wconti27: Makes sense to me, it is a bit excessive rn

ddtrace/contrib/anthropic/patch.py
wconti27: added a test for this

ddtrace/contrib/anthropic/__init__.py
wconti27: done

ddtrace/contrib/anthropic/patch.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+@with_traced_module
+def traced_chat_model_generate(anthropic, pin, func, instance, args, kwargs):
+    chat_messages = get_argument_value(args, kwargs, 0, "messages")
+    integration = anthropic._datadog_integration
+
+    operation_name = func.__name__
+
+    span = integration.trace(
```
emmettbutler, line 55: Using `core.context_with_data` here would loosen the coupling between instrumentation and the LLMObs product. I think that's an important change to make in the near future.

tests/contrib/anthropic/test_anthropic.py
```diff
@@ -0,0 +1,139 @@
...
+                            "type": "text",
+                            "text": "Can you explain what Descartes meant by 'I think, therefore I am'?",
+                        }
+                    ],
+                },
+            ],
+            stream=True,
+        )
+        for chunk in stream:
+            print(chunk.type)
```
Yun-Kim, line 139:
```suggestion
            pass
```
We don't want to print the test output, just passing iterating over the stream is fine

emmettbutler (approved):

ddtrace/contrib/anthropic/__init__.py
```diff
@@ -0,0 +1,98 @@
...
+- ``anthropic.request.api_key``: Anthropic API key used to make the request (obfuscated to match the Anthropic UI representation ``sk-...XXXX`` where ``XXXX`` is the last 4 digits of the key).
+- ``anthropic.request.parameters``: Parameters used in anthropic package call.
+
+
+(beta) Prompt and Completion Sampling
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The following data is collected in span tags with a default sampling rate of ``1.0``:
+
+- Prompt inputs and completions for the ``Messages.create`` endpoint.
```
Yun-Kim, line 17:
```suggestion
Prompt texts and completion content for the ``Messages.create`` endpoint are collected in span tags with a default sampling rate of ``1.0``. These tags will have truncation applied if the text exceeds the configured character limit.
```

ddtrace/contrib/anthropic/__init__.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+Enabling
+~~~~~~~~
+
+The Anthropic integration is enabled automatically when you use
+:ref:`ddtrace-run<ddtracerun>` or :ref:`import ddtrace.auto<ddtraceauto>`.
+
+Note that these commands also enable the ``requests`` and ``aiohttp``
+integrations which trace HTTP requests from the Anthropic library.
```
Yun-Kim, line 27: Question, which request libraries are used by Anthropic? Just want to make sure this is an accurate statement

ddtrace/llmobs/_integrations/anthropic.py
```diff
@@ -0,0 +1,48 @@
...
+        model: Optional[str] = None,
+        api_key: Optional[str] = None,
+        **kwargs: Dict[str, Any],
+    ) -> None:
+        """Set base level tags that should be present on all Anthropic spans (if they are not None)."""
+        if model is not None:
+            span.set_tag_str(MODEL, model)
+        if api_key is not None:
+            if len(api_key) >= 4:
+                span.set_tag_str(API_KEY, f"...{str(api_key[-4:])}")
```
Yun-Kim, line 34:
```suggestion
                span.set_tag_str(API_KEY, f"sk-...{str(api_key[-4:])}")
```

ddtrace/contrib/anthropic/patch.py
```diff
@@ -0,0 +1,162 @@
...
+
+        chat_completions = func(*args, **kwargs)
+
+        if not isinstance(chat_completions, anthropic.Stream) and not isinstance(
+            chat_completions, anthropic.lib.streaming._messages.MessageStreamManager
+        ):
+            handle_non_streamed_response(integration, chat_completions, args, kwargs, span)
+    except Exception:
+        span.set_exc_info(*sys.exc_info())
+        span.finish()
```
Yun-Kim, line 113: We can actually remove this since we're finishing the span in the `finally` block. (This will change when we add streaming support i assume)

ddtrace/contrib/anthropic/async_message.py
```diff
@@ -0,0 +1,86 @@
...
+from ddtrace.internal.utils import get_argument_value
+
+from .utils import _extract_api_key
+from .utils import handle_non_streamed_response
+
+
+log = get_logger(__name__)
+
+
+@with_traced_module
```
Yun-Kim, line 15: I think we can just keep this method in `patch.py` to keep the traced methods in the same file (if we had to trace multiple methods I wouldn't mind but there's only two methods)

ddtrace/contrib/anthropic/__init__.py
wconti27: httpx, should I remove this?

releasenotes/notes/feat-anthropic-04a880a26ff44d9c.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    anthropic: This introduces tracing support for anthropic chat messages.
+      See `the docs <https://ddtrace.readthedocs.io/en/stable/integrations.html#anthropic>`_
+      for more information.
```
Yun-Kim, line 6:
```suggestion
      for more information.

```

ddtrace/contrib/anthropic/patch.py
Yun-Kim: We're aiming to do a refactor of the LLM integrations to use core context management instead in Q3/Q4. We'll keep this for now for the sake of deadlines, but just wanted to let you know we're definitely keeping that in mind @emmettbutler

ddtrace/contrib/anthropic/patch.py
```diff
@@ -0,0 +1,206 @@
...
+                    span.set_tag_str(
+                        "anthropic.request.messages.%d.content.%d.type" % (message_idx, block_idx),
+                        block.get("type", "text"),
+                    )
+            span.set_tag_str(
+                "anthropic.request.messages.%d.role" % (message_idx),
+                message.get("role", ""),
+            )
+        params_to_tag = {k: v for k, v in kwargs.items() if k != "messages"}
+        span.set_tag_str("anthropic.request.parameters", json.dumps(params_to_tag))
```
Yun-Kim, line 92: Just curious, how do tool calls get captured by this? I know tools can get very long so this may be excessive to tag and I would prefer avoiding tagging this parameter as well.

riotfile.py
```diff
@@ -2533,6 +2533,16 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
                 ),
             ],
         ),
+        Venv(
+            name="anthropic",
+            command="pytest {cmdargs} tests/contrib/anthropic",
+            pys=select_pys(min_version="3.7", max_version="3.11"),
```
Yun-Kim, line 2539: Just curious, is there a reason we don't support anthropic with python 3.12?

tests/contrib/anthropic/cassettes/anthropic_completion_error_async.yaml
```diff
@@ -0,0 +1,67 @@
+interactions:
```
Yun-Kim, line 1: Nit - can we use the same cassettes for both sync and async tests to minimize cassette files? Thanks

tests/contrib/anthropic/test_anthropic_async.py
```diff
@@ -0,0 +1,142 @@
+import pytest
```
Yun-Kim, line 1: We can also move the async tests to the same `test_anthropic.py` file

tests/contrib/anthropic/test_anthropic_async.py
```diff
@@ -0,0 +1,142 @@
...
+    assert span.resource == "AsyncMessages.create"
+    assert span.service == "test-svc"
+    assert span.get_tag("env") == "staging"
+    assert span.get_tag("version") == "1234"
+    assert span.get_tag("anthropic.request.model") == "claude-3-opus-20240229"
+    assert span.get_tag("anthropic.request.api_key") == "sk-...key>"
+
+
+@pytest.mark.asyncio
+# @pytest.mark.snapshot
```
Yun-Kim, line 34:
```suggestion
```

tests/contrib/anthropic/test_anthropic.py
```diff
@@ -0,0 +1,133 @@
...
+    span = mock_tracer.pop_traces()[0][0]
+    assert span.resource == "Messages.create"
+    assert span.service == "test-svc"
+    assert span.get_tag("env") == "staging"
+    assert span.get_tag("version") == "1234"
+    assert span.get_tag("anthropic.request.model") == "claude-3-opus-20240229"
+    assert span.get_tag("anthropic.request.api_key") == "sk-...key>"
+
+
+# @pytest.mark.snapshot(ignores=["metrics.anthropic.tokens.total_cost", "resource"])
```
Yun-Kim, line 32:
```suggestion
```

tests/contrib/anthropic/test_anthropic_patch.py
```diff
@@ -0,0 +1,21 @@
...
+
+class TestAnthropicPatch(PatchTestCase.Base):
+    __integration_name__ = "anthropic"
+    __module_name__ = "anthropic"
+    __patch_func__ = patch
+    __unpatch_func__ = unpatch
+    __get_version__ = get_version
+
+    def assert_module_patched(self, anthropic):
+        self.assert_wrapped(anthropic.resources.messages.Messages.create)
```
Yun-Kim, line 15: We want to assert that `AsyncMessages` is also wrapped/not/double wrapped

tests/contrib/anthropic/utils.py
```diff
@@ -0,0 +1,30 @@
+import os
+
+import vcr
+
+
+def iswrapped(obj):
```
Yun-Kim, line 6: Is this utility function used anywhere?

ddtrace/contrib/anthropic/patch.py
Yun-Kim: I would also prefer avoiding tagging `models` again since we already tag it as a base level tag, can we keep a small blacklist of kwarg names to not tag as parameters?

tests/contrib/anthropic/test_anthropic_async.py
Yun-Kim: Also, is it possible to reuse the same snapshot and cassette files for both sync and async tests? I would prefer to minimize the number of static test data files we keep.

Yun-Kim (commented):
Some last few nits

wconti27:
> Some last few nits

Httpx is used for both sync and async

ddtrace/contrib/anthropic/patch.py
wconti27: Ill add a test for tools, also yes ill add a blacklist

riotfile.py
wconti27: ill change this

tests/contrib/anthropic/test_anthropic_patch.py
wconti27: done

tests/contrib/anthropic/utils.py
wconti27: no, removing

tests/contrib/anthropic/test_anthropic_async.py
wconti27: done

ddtrace/contrib/anthropic/patch.py
wconti27: I've added a tools test, and added tools to the blacklist so it doesn't capture the param, same with model.

tests/contrib/anthropic/cassettes/anthropic_completion_error_async.yaml
wconti27: done

ddtrace/contrib/anthropic/__init__.py
wconti27: httpx does both async and sync

ddtrace/contrib/anthropic/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+These tags will have truncation applied if the text exceeds the configured character limit.
+
+
+Enabling
+~~~~~~~~
+
+The Anthropic integration is enabled automatically when you use
+:ref:`ddtrace-run<ddtracerun>` or :ref:`import ddtrace.auto<ddtraceauto>`.
+
+Note that these commands also enable the ``httx`` integration which traces HTTP requests from the Anthropic library.
```
Yun-Kim, line 25:
```suggestion
Note that these commands also enable the ``httpx`` integration which traces HTTP requests from the Anthropic library.
```
```suggestion
Note that these commands also enable the ``httx`` integration which traces HTTP requests from the Anthropic library.
```

ddtrace/contrib/anthropic/patch.py
```diff
@@ -0,0 +1,207 @@
...
+                    span.set_tag_str(
+                        "anthropic.request.messages.%d.content.%d.type" % (message_idx, block_idx),
+                        _get_attr(block, "type", "text"),
+                    )
+            span.set_tag_str(
+                "anthropic.request.messages.%d.role" % (message_idx),
+                message.get("role", ""),
+            )
+        params_to_tag = {k: v for k, v in kwargs.items() if k not in ["messages", "model", "tools"]}
+        span.set_tag_str("anthropic.request.parameters", json.dumps(params_to_tag))
```
Yun-Kim, line 93: I just saw on Anthropic's API [docs](https://docs.anthropic.com/en/docs/system-prompts) that you can specify a `system` prompt. For our purposes that is a prompt to the LLM, meaning we have to apply truncation and tag separately from other parameters.
I'm noticing that request parameter tagging is getting more complex (blacklist, prompt tagging). Can we extract this logic into a utility function?

ddtrace/contrib/anthropic/patch.py
```diff
@@ -0,0 +1,207 @@
...
```
Yun-Kim, line 166: Same here ^

riotfile.py
```diff
@@ -2533,6 +2533,16 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
...
             ],
         ),
+        Venv(
+            name="anthropic",
+            command="pytest {cmdargs} tests/contrib/anthropic",
+            pys=select_pys(min_version="3.7", max_version="3.12"),
+            pkgs={
+                "pytest-asyncio": latest,
+                "vcrpy": latest,
+                "anthropic": [latest, "~=0.28", "~=0.26"],
```
Yun-Kim, line 2543: Testing every other minor version is excessive considering their release cadence (~1 week per minor release), let's just test `0.26` and `latest`.
```suggestion
                "anthropic": ["~=0.26", latest],
```
```suggestion
                "anthropic": [latest, "~=0.28", "~=0.26"],
```

ddtrace/contrib/anthropic/patch.py
Yun-Kim: Let's also modify an existing test case to add a system prompt if possible, thanks

tests/contrib/anthropic/test_anthropic.py
```diff
@@ -0,0 +1,401 @@
...
+                tools=tools,
+            )
+        else:
+            response = message
+
+        final_response = next(
+            (block.text for block in response.content if hasattr(block, "text")),
+            None,
+        )
+        assert final_response is not None
```
Yun-Kim, line 196: I don't think we need to make the test this complicated (2 cassettes, 2 message calls, running an actual tool function, etc). We just need to test that we tag the correct values if we specify tool calling in a message call.

Can we just reproduce the first basic example in anthropic's tool call [docs](https://docs.anthropic.com/en/docs/tool-use-examples#single-tool)?

tests/contrib/anthropic/utils.py
```diff
@@ -0,0 +1,65 @@
...
+            },
+            "required": ["expression"],
+        },
+    }
+]
+
+
+def process_tool_call(tool_name, tool_input):
+    if tool_name == "calculator":
+        return calculate(tool_input["expression"])
```
Yun-Kim, line 65: We don't need this

tests/contrib/anthropic/utils.py
```diff
@@ -0,0 +1,65 @@
...
+def calculate(expression):
+    # Remove any non-digit or non-operator characters from the expression
+    expression = re.sub(r"[^0-9+\-*/().]", "", expression)
+
+    try:
+        # Evaluate the expression using the built-in eval() function
+        result = eval(expression)
+        return str(result)
+    except (SyntaxError, ZeroDivisionError, NameError, TypeError, OverflowError):
+        return "Error: Invalid expression"
```
Yun-Kim, line 42: We don't need to actually run anthropic tools, so we don't need these in our tests

Yun-Kim (commented):
Everything looks great, except for two main points:
- Request parameter tagging will need to be extracted into a separate helper since it's not so straightforward anymore (we'll need to apply blacklist and tag/truncate system prompts separately)
- Tool calling tests don't need to be so complicated, since we're just testing that we can tag the correct values if tools are specified in a message call. We don't need to make 2 cassettes or test running the actual tool.

tests/contrib/anthropic/test_anthropic.py
wconti27: This is the full example from their demo examples. Honestly I think we should keep it, since tools use will result in different context blocks. It doesn't really hurt to keep it

ddtrace/contrib/anthropic/patch.py
wconti27: done

ddtrace/contrib/anthropic/patch.py
wconti27: done

tests/contrib/anthropic/test_anthropic.py
wconti27: done

tests/contrib/anthropic/utils.py
wconti27: done

tests/contrib/anthropic/utils.py
wconti27: done

ddtrace/contrib/anthropic/utils.py
```diff
@@ -0,0 +1,57 @@
...
+
+    usage = _get_attr(chat_completions, "usage", {})
+    integration.record_usage(span, usage)
+
+
+def tag_params_on_span(span, kwargs, integration):
+    tagged_params = {}
+    for k, v in kwargs.items():
+        if k == "system":
+            tagged_params[k] = integration.trunc(v)
```
Yun-Kim, line 36:
```suggestion
        if k == "system" and integration.is_pc_sampled_span(span):
            tagged_params[k] = integration.trunc(v)
```

ddtrace/contrib/anthropic/utils.py
```diff
@@ -0,0 +1,57 @@
...
+        elif k not in ["messages", "model", "tools"]:
+            tagged_params[k] = v
```
Yun-Kim:
```suggestion
    for k, v in kwargs.items():
        if k == "system" and integration.is_pc_sampled_span(span):
            span.set_tag_str("anthropic.request.system", integration.trunc(v))
        elif k not in ("messages", "model", "tools"):
            tagged_params[k] = v
```
System should be its own tag

Yun-Kim (approved):
Just needs fix to separate system prompt but we're good to go 🚀

ddtrace/contrib/anthropic/utils.py
wconti27: done

ddtrace/_monkey.py
```diff
@@ -91,6 +91,7 @@
     "tornado": False,
     "openai": True,
     "langchain": True,
+    "anthropic": True,
```
mabdinur, line 94: This integration is marked as beta but it will be auto enabled. Can we make this integration opt in?

ddtrace/contrib/anthropic/patch.py
```diff
@@ -0,0 +1,205 @@
...
+        tag_params_on_span(span, kwargs, integration)
+
+        chat_completions = func(*args, **kwargs)
+
+        if isinstance(chat_completions, anthropic.Stream) or isinstance(
+            chat_completions, anthropic.lib.streaming._messages.MessageStreamManager
+        ):
+            pass
+        else:
+            handle_non_streamed_response(integration, chat_completions, args, kwargs, span)
```
mabdinur, line 101: Not a fan of this. Can we get rid of the `pass` and just make this an if statement without an else

PR #9501, opened 2024-06-10
fix(profiling): fix encoding-related issues in stack v2

sanchda:
This adjusts the stack V2 API in order to conform to the new proposed echion renderer.  This also fixes an encoding error in stack V2 which was caused by the old task renderer.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


sanchda:
Moving this back into draft.  There's some kind of issue in CI I didn't anticipate.

sanchda:
Looks like the issue was spurious, reopening.

ddtrace/internal/datadog/profiling/stack_v2/src/stack_renderer.cpp
```diff
@@ -29,18 +29,63 @@ StackRenderer::render_thread_begin(PyThreadState* tstate,
         return;
     }
 
+    ddup_push_threadinfo(sample, static_cast<int64_t>(thread_id), static_cast<int64_t>(native_id), name);
+    ddup_push_walltime(sample, 1000 * wall_time_us, 1);
```
danielsn, line 33: why not push cputime?

ddtrace/internal/datadog/profiling/stack_v2/src/stack_renderer.cpp
```diff
@@ -29,18 +29,63 @@ StackRenderer::render_thread_begin(PyThreadState* tstate,
...
     ddup_push_threadinfo(sample, static_cast<int64_t>(thread_id), static_cast<int64_t>(native_id), name);
     ddup_push_walltime(sample, 1000LL * wall_time_us, 1);
+
+    // We stash the current thread information in the StackRenderer instance, since a task may be begun
+    // after ending the sample.
+    stashed_thread_id = thread_id;
+    stashed_native_id = native_id;
+    stashed_thread_name = std::string(name);
+    stashed_wall_time_us = wall_time_us;
+    stashed_end_time_ns = now_ns;
```
taegyunkim, line 55:
```suggestion
    stashed_now_time_ns = now_ns;
```

ddtrace/internal/datadog/profiling/stack_v2/src/stack_renderer.cpp
```diff
@@ -29,18 +29,63 @@ StackRenderer::render_thread_begin(PyThreadState* tstate,
...
+        // Re-hydrate the thread context
+        ddup_push_threadinfo(sample,
+                             static_cast<int64_t>(stashed_thread_id),
+                             static_cast<int64_t>(stashed_native_id),
+                             stashed_thread_name);
+        ddup_push_walltime(sample, 1000 * stashed_wall_time_us, 1);
+        ddup_push_cputime(sample, 1000 * stashed_cpu_time_us, 1); // initialized to 0, so possibly a no-op
+
+        // If we have a nonzero timestamp, push that too
+        if (stashed_end_time_ns != 0) {
```
taegyunkim, line 83:
```suggestion
        if (stashed_now_time_ns != 0) {
```

ddtrace/internal/datadog/profiling/stack_v2/src/stack_renderer.cpp
```diff
@@ -29,18 +29,63 @@ StackRenderer::render_thread_begin(PyThreadState* tstate,
...
+            ddup_push_monotonic_ns(sample, stashed_end_time_ns);
```
taegyunkim, line 84:
```suggestion
            ddup_push_monotonic_ns(sample, stashed_now_time_ns);
```

taegyunkim (changes requested):
Looks like CI doesn't have DD_STACK_V2_REQUIRED set and build failures from these typos were not caught.

ddtrace/internal/datadog/profiling/stack_v2/CMakeLists.txt
```diff
@@ -26,7 +26,7 @@ if (NOT Python3_INCLUDE_DIRS)
 endif()
 
 # Add echion
-set(ECHION_COMMIT "d69048b63e7d49091659964915eb52ff0ced9fc8" CACHE STRING "Commit hash of echion to use")
+set(ECHION_COMMIT "38df39c45d192efc2e6f1be94a9d7136099564d0" CACHE STRING "Commit hash of echion to use")
```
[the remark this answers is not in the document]
sanchda: Agree, we'll get there

ddtrace/internal/datadog/profiling/stack_v2/src/stack_renderer.cpp
```diff
@@ -29,18 +29,63 @@ StackRenderer::render_thread_begin(PyThreadState* tstate,
...
+    stashed_native_id = native_id;
+    stashed_thread_name = std::string(name);
+    stashed_wall_time_us = wall_time_us;
+    stashed_end_time_ns = now_ns;
+}
+
+void
+StackRenderer::render_task_begin(std::string_view name)
+{
+    static bool failed = false;
```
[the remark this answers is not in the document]
sanchda: It should be global, since there's currently no way for us to deduce the scope of the problem.  If we fail to get a sample, we stop creating samples forever.

That said, I see there's actually a split here.  We should probably bind the error conditions for both the task and the thread begin sections.

ddtrace/internal/datadog/profiling/stack_v2/src/stack_renderer.cpp
```diff
@@ -29,18 +29,63 @@ StackRenderer::render_thread_begin(PyThreadState* tstate,
...
+    if (failed) {
+        return;
+    }
+    if (sample == nullptr) {
+        // This is possible and natural, we just re-hydrate the thread context from the stashed values.
+        sample = ddup_start_sample();
+        if (sample == nullptr) {
+            std::cerr << "Failed to create a sample.  Stack v2 sampler will be disabled." << std::endl;
```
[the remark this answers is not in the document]
sanchda: In some sense (e.g., temporary failures to allocate memory), I suppose so.  However, it's impossible to deduce that an error was spurious and I'm not confident that a retry would be safe (e.g., if the state of the runtime has degraded and we didn't fence it off within the libdatadog FFI).

I don't think these failures are spurious in the sense you mean, though (in that they happen, well, spuriously and thus a try-retry would be appropriate).

ddtrace/internal/datadog/profiling/stack_v2/src/stack_renderer.cpp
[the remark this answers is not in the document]
sanchda: Yup

ddtrace/internal/datadog/profiling/stack_v2/src/stack_renderer.cpp
sanchda: Whoops!  Thank you!

sanchda:
> Looks like CI doesn't have DD_STACK_V2_REQUIRED set and build failures from these typos were not caught.

This should be fixed now, since a recent PR removes that flag and forces these objects to be built on compatible platforms.

ddtrace/internal/datadog/profiling/stack_v2/include/stack_renderer.hpp
```diff
@@ -20,13 +20,23 @@ class StackRenderer : public RendererInterface
 {
     Sample* sample = nullptr;
 
+    // Stashed thread information
+    uintptr_t stashed_thread_id = 0;
+    unsigned long stashed_native_id = 0;
+    std::string stashed_thread_name;
+    microsecond_t stashed_wall_time_us = 0;
+    microsecond_t stashed_cpu_time_us = 0;
+    int64_t stashed_now_time_ns = 0;
```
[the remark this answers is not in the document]
sanchda: * Added comments in the header
* Refactored the so-called "stash" elements into their own "ThreadInfo" struct to make it clearer

ddtrace/internal/datadog/profiling/stack_v2/include/stack_renderer.hpp
```diff
@@ -27,6 +42,8 @@ class StackRenderer : public RendererInterface
                                      uintptr_t thread_id,
                                      unsigned long native_id) override;
 
+    virtual void render_task_begin(std::string_view name);
+    virtual void render_pid_tid(uintptr_t pid, uintptr_t tid) override;
```
taegyunkim, line 46:
```
/root/project/ddtrace/internal/datadog/profiling/stack_v2/include/stack_renderer.hpp:46:18: error: ‘virtual void Datadog::StackRenderer::render_pid_tid(uintptr_t, uintptr_t)’ marked ‘override’, but does not override
     virtual void render_pid_tid(uintptr_t pid, uintptr_t tid) override;
```

ddtrace/internal/datadog/profiling/stack_v2/src/stack_renderer.cpp
```diff
@@ -29,18 +29,63 @@ StackRenderer::render_thread_begin(PyThreadState* tstate,
...
+    thread_state.id = thread_id;
+    thread_state.native_id = native_id;
+    thread_state.name = std::string(name);
+    thread_state.now_time_ns = now_ns;
+    thread_state.wall_time_ns = 1000LL * wall_time_us;
+    thread_state.cpu_time_ns = 0; // Walltime samples are guaranteed, but CPU times are not. Initialize to 0
+                                  // since we don't know if we'll get a CPU time here.
+
+    // Finalize the thread information we have
     ddup_push_threadinfo(sample, static_cast<int64_t>(thread_id), static_cast<int64_t>(native_id), name);
```
taegyunkim, line 56: We're calling `ddup_push_threadinfo` twice in this function. Looking at the `push_label` definition [here](https://github.com/DataDog/dd-trace-py/blob/6e937a8521752179410e1e5e25a9b1e744815491/ddtrace/internal/datadog/profiling/dd_wrapper/src/sample.cpp#L89), there will be two labels with same keys. Do you know how this is handled in the backend and rest of the stack?

ddtrace/internal/datadog/profiling/stack_v2/include/stack_renderer.hpp
sanchda: Good catch--the `pid_tid` call was debris from a refactor.  I removed that function and propagated the changes.

ddtrace/internal/datadog/profiling/stack_v2/src/stack_renderer.cpp
sanchda: Yup, that was a mistake.

As for the question, you're correct that it'll push the duplicates all the way down to libdatadog.  I'm not 100% sure what libdatadog will do in that case.  Taking the first, taking the last, and pushing duplicate entries all feel like valid options at some level.  If duplicate tags are pushed all the way to the backend, I don't know what it will do!

ddtrace/internal/datadog/profiling/stack_v2/src/stack_renderer.cpp
```diff
@@ -33,14 +33,56 @@ StackRenderer::render_thread_begin(PyThreadState* tstate,
...
+            return;
+        }
+
+        // Add the thread context into the sample
+        ddup_push_threadinfo(sample,
+                             static_cast<int64_t>(thread_state.id),
+                             static_cast<int64_t>(thread_state.native_id),
+                             thread_state.name);
+        ddup_push_walltime(sample, thread_state.wall_time_ns, 1);
+        ddup_push_cputime(sample, thread_state.cpu_time_ns, 1); // initialized to 0, so possibly a no-op
```
taegyunkim, line 81: `thread_state.cpu_time_ns` will will be zero only when `render_cpu_time` has not yet been called, right?

taegyunkim (approved):
Overall LGTM, we just want to see whether this indeed fixes the problem with some minimal testing.

releasenotes/notes/fix-stack-v2-task-encoding-447f39478749027c.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    profiling: Fix an issue where task information coming from echion was
```
erikayasuda, line 4:
```suggestion
    profiling: Fixes an issue where task information coming from echion was
```

ddtrace/internal/datadog/profiling/stack_v2/src/stack_renderer.cpp
```diff
@@ -33,14 +33,56 @@ StackRenderer::render_thread_begin(PyThreadState* tstate,
...
+                                  // since we don't know if we'll get a CPU time here.
+
+    // Finalize the thread information we have
     ddup_push_threadinfo(sample, static_cast<int64_t>(thread_id), static_cast<int64_t>(native_id), name);
-    ddup_push_walltime(sample, 1000LL * wall_time_us, 1);
+    ddup_push_walltime(sample, thread_state.wall_time_ns, 1);
+}
+
+void
+StackRenderer::render_task_begin(std::string_view name)
```
erikayasuda, line 58: Dumb question, but how was this done before you added `render_task_begin`? Or is the concept of a `task` the new addition that was added in echion?

erikayasuda (approved):

releasenotes/notes/fix-stack-v2-task-encoding-447f39478749027c.yaml
sanchda: tyvm!

ddtrace/internal/datadog/profiling/stack_v2/src/stack_renderer.cpp
sanchda: > how was this done before

It was broken!

Task comprehension was done correctly within echion itself, but the rendering system I added (to echion) tried to conflate the concept of a task with the concept of a thread.  I had done it this way because the Profiling backend actually makes no distinction between the two concepts, except that they have different tags (and thus, different UI filters).

Unfortunately, that approach didn't do the state management correctly (tasks in Python live on a thread, and that thread is backed by a native thread, and we need to propagate that information), so I had to fix it upstream, and 

This PR mostly fixes dd-trace-py's consumption of the echion rendering API.

PR #9515, opened 2024-06-11
chore(internal): improve performance of telemetry dependency reporting

brettlangdon:
This change:

- Removes the usage of `TelemetryModuleWatchdog`
  - A strategy of diffing `sys.modules.keys()` is now used instead of watching for imports.
- Removes tracking of module origin
  - Since we are only interested in distribution name + version we don't need to keep track of all imported module origins, and then look up distributions from that. We can use import names to directly look up the distribution they are apart of.
- Add support for namespace packages
  - Added the use of `importlib_metdata.packages_distributions()` to ensure we are able to resolve namespace packages to their actual distribution name.
  - We only add `importlib_metadata` for Python <3.8, but `packages_distributions` was added in 3.10, the footprint is fairly small, so it is just vendored for the cases when we don't have it available (keeps `importlib_metadata` requirement fairly broad as well)

The main change here is to have the TelemetryWriter keep track of which `sys.modules.keys()` we have already resolved and sent information about. Then on each periodic run, compare current `sys.modules.keys()` against the cache, and send any new modules.

We also are no longer using origins to try and resolve the distributions for a given module, instead we use `importlib_metadata.distribution(name)`, if that fails to resolve, then we will check to see if that import name is mapped to any namespace packages, which we then try grab the distributions for that. (this means any given module name could resolve to multiple distributions to report about). 

I have really struggled to get a solid benchmark/performance measurement going, the only details I have are the internal app I was testing against went from 100s to about 2s of overhead added for dependency resolution, and the test suite itself still validates that the previous behavior is occurring.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


tests/telemetry/test_writer.py
```diff
@@ -380,13 +379,11 @@ def test_update_dependencies_event_when_disabled(telemetry_writer, test_agent_se
 
 @pytest.mark.skip(reason="FIXME: This test does not generate a dependencies event")
```
gnufede, line 380: Do we still have to skip this test?

tests/telemetry/test_writer.py
```diff
@@ -396,19 +393,16 @@ def test_update_dependencies_event_not_stdlib(telemetry_writer, test_agent_sessi
 
 @flaky(1717255857)
```
gnufede, line 394: How about this one?

ddtrace/internal/packages.py
```diff
@@ -77,6 +79,42 @@ def get_distributions():
...
+    """a mapping of importable package names to their distribution name(s)"""
+    try:
+        import importlib.metadata as importlib_metadata
+    except ImportError:
+        import importlib_metadata  # type: ignore[no-redef]
+
+    # Prefer the official API if available, otherwise fallback to the vendored version
+    if hasattr(importlib_metadata, "packages_distributions"):
+        return importlib_metadata.packages_distributions()
+    return _packages_distributions()
```
gnufede, line 93: Why not simply use this one and avoid all the above? Since it's going to be always available

gnufede (approved):
Just a few nits, great job @brettlangdon !! 🚀

juanjux (approved):

ddtrace/internal/packages.py
brettlangdon: I am worried about keeping our vendored version up to date, if we have an official upstream version I'd rather use that.

so anyone using Python 3.8 + 3.9 will use this vendored version, and anyone on 3.7 using an older version of `importlib_metdata` (we don't specify a min required version, just a max).

I am not sure if this was the right decision, I considered just using the vendored version for everyone. If someone has a strong opinion, I don't have one.

tests/telemetry/test_writer.py
brettlangdon:
```suggestion
```

tests/telemetry/test_writer.py
brettlangdon:
```suggestion
```

gnufede:
I guess this was some CI hang, but could you double check, @brettlangdon ?
> test_tracer_usage_multiprocess - test_rand.py 7m 1.49s (+6m 57.58s, +10696%) - [Details](https://app.datadoghq.com/ci/test/AgAAAZALWQA9x7iwSAAAAAAAAAAYAAAAAEFaQUxXUUhWQUFBQlM3eElSNDI3cUxTcgAAACQAAAAAMDE5MDBiNTktMmU0NS00NzJjLWIzYzYtNzMyOGZhMDk3Y2Nm?colorBy=service&spanViewType=metadata&env=none&spanID=11474191554179990409&eventStack=AgAAAZALWQA9x7iwSAAAAAAAAAAYAAAAAEFaQUxXUUhWQUFBQlM3eElSNDI3cUxTcgAAACQAAAAAMDE5MDBiNTktMmU0NS00NzJjLWIzYzYtNzMyOGZhMDk3Y2Nm&currentTab=performance)

ddtrace/internal/telemetry/writer.py
```diff
@@ -588,15 +546,18 @@ def _app_client_configuration_changed_event(self, configurations):
...
+    def _update_dependencies_event(self):
         """Adds events to report imports done since the last periodic run"""
 
         if not config._telemetry_dependency_collection or not self._enabled:
             return
 
         with self._lock:
-            packages = update_imported_dependencies(self._imported_dependencies, newly_imported_deps)
+            latest_modules = set(sys.modules.keys())
+            new_modules = latest_modules - self._imported_dependencies
```
P403n1x87, line 557: Is this faster than the module watchdog solution when `sys.modules` is a large collection? Do we have some numbers?

erikayasuda (commented):
The change makes sense, but I'm also curious about @P403n1x87 's question regarding the performance improvement in the case that there are a lot of dependencies. I know you said your test app went from 100s to 2s. How many dependencies were you working with for that test app? I think naively, it seems like the `ModuleWatchDog` method of adding the new module with `after_import()` seems like it would be faster than iterating through the entire imported dependency list every time `periodic()` is called, but that's probably my lack of knowledge on how `ModuleWatchDog` works?

ddtrace/internal/telemetry/writer.py
brettlangdon: I don't have any numbers for this, but easy enough to measure the set creating and diffing overhead with a large dataset.

What do you think is a realistic size of sys.modules to test with? 1000 entries? 10,000 entries? I would think anything higher than that would be crazy.


Also, the part of the module watchdog approach which was slow was going from module import -> file -> file origin -> distribution. This was the main source of overhead the current implementation. As long as performance overhead (when I shared numbers) is acceptable, I'd greatly prefer this approach to having the additional moving pieces of a module watchdog.

brettlangdon:
```python
import random
import string
import pyperf


def random_string():
    return "".join(random.choice(string.ascii_letters) for _ in range(10))


def bench_module_diff(seen_modules, sys_modules):
    latest_modules = set(sys_modules.keys())
    _ = latest_modules - seen_modules


runner = pyperf.Runner()

names = [random_string() for _ in range(10_000)]

for num_existing in [0, 500, 2_000, 5_000, 10_000]:
    for num_modules in [500, 2_000, 5_000, 10_000]:
        if num_modules < num_existing:
            continue

        runner.bench_func(
            f"module_diff[{num_existing},{num_modules}]",
            bench_module_diff,
            set(names[:num_existing]),
            {names[i]: True for i in range(num_modules)},
        )
```

The results are `module_diff[{num modules names already "seen"}, {num modules in sys.modules}]`


```
.....................
module_diff[0,500]: Mean +- std dev: 9.22 us +- 0.59 us
.....................
module_diff[0,2000]: Mean +- std dev: 65.7 us +- 3.9 us
.....................
module_diff[0,5000]: Mean +- std dev: 233 us +- 6 us
.....................
module_diff[0,10000]: Mean +- std dev: 378 us +- 6 us
.....................
module_diff[500,500]: Mean +- std dev: 9.67 us +- 0.64 us
.....................
module_diff[500,2000]: Mean +- std dev: 94.8 us +- 3.4 us
.....................
module_diff[500,5000]: Mean +- std dev: 232 us +- 5 us
.....................
module_diff[500,10000]: Mean +- std dev: 378 us +- 6 us
.....................
module_diff[2000,2000]: Mean +- std dev: 63.5 us +- 3.1 us
.....................
module_diff[2000,5000]: Mean +- std dev: 277 us +- 5 us
.....................
module_diff[2000,10000]: Mean +- std dev: 388 us +- 7 us
.....................
module_diff[5000,5000]: Mean +- std dev: 207 us +- 4 us
.....................
module_diff[5000,10000]: Mean +- std dev: 521 us +- 17 us
.....................
module_diff[10000,10000]: Mean +- std dev: 367 us +- 9 us
```


I think something between the "500" and "2000" cases are the most realistic for most applications? So somewhere between 0.01ms and 0.1ms per-operation to get the list of newly imported modules.

brettlangdon (commented):
I updated the implementation to be more inline with the previous of tracking module names when imported, rather than always diffing `sys.modules` every time.

For Python >= 3.8 we use `sys.audit` "import" event, and Python 3.7 we use the ModuleWatchdog.


The big improvement here is still to just resolve distributions from module names instead of trying to resolve file origin -> distribution.

ddtrace/internal/telemetry/modules.py
```diff
@@ -0,0 +1,92 @@
...
+        return new_modules
+
+    def install_import_hook():
+        global MODULE_HOOK_INSTALLED, NEW_MODULES, ALL_MODULES
+
+        # If we have not called get_newly_imported_modules yet, we can initialize to all imported modules
+        if not NEW_MODULES:
+            NEW_MODULES = set(sys.modules.keys())
+            ALL_MODULES = NEW_MODULES.copy()
+        sys.addaudithook(audit_hook)
```
erikayasuda, line 47: Double checking my understanding here... The 3 approaches we have are:
1. Periodically go through all imported dependencies (what you originally had)
2. Use ModuleWatchDog which triggers the `after_import` event after each import
3. This approach, where we trigger on all sys audit events, and then check if it's an `import` event type before proceeding.

Naively it feels like the ModuleWatchdog approach is still the most "efficient" in the sense that it only triggers the hook after imports specifically, whereas the sys audit hooks will check for not only `import` events, but also things like `open`, `exec`, etc. Is the perf improvement here based on the assumption that those methods probably won't be called very much, so it's not that much additional performance overhead? Or is there something pretty performance heavy on the ModuleWatchdog side that I'm missing here?

erikayasuda (approved):

ddtrace/internal/telemetry/modules.py
erikayasuda: Oh wait I think I see my answer:
> The big improvement here is still to just resolve distributions from module names instead of trying to resolve file origin -> distribution.

brettlangdon: So there are a few pieces.

The main performance improvement is to not try to resolve the module origin. We don't need it for this telemetry data. That part of the code is slow, and it isn't what we actually want for telemetry data. Instead we want the "distribution" not the module (e.g. the `packaging` module comes from the `setuptools` distribution).


The other sys.audit change is because, if we don't need to resolve origin, we don't need all the data provided by ModuleWatchdog. We really only need the module name that was imported to then try to resolve a distribution for it. This way we don't need to worry about installing a loader into the `sys.meta_path`. `sys.audit` will be less invasive and more performant in this case when only collecting the imported module name. So I want to prefer `sys.audit` if it is available.


If we start to use `sys.audit` for more than just this, we may want to centralize a single handler... but... future us problem.

PR #9520, opened 2024-06-12
chore(telemetry): use enum for telemetry log levels

romainkomorndatadog:
Makes type-checking more accurate the telemetry writer's `add_log()` method by specifying an enum.

For example, improperly specifying `TELEMETRY_LOG_LEVEL.POTATO` instead of `.ERROR` would result in:
```
ddtrace/appsec/_iast/_metrics.py:77: error: "Type[TELEMETRY_LOG_LEVEL]" has no attribute "POTATO"  [attr-defined]
```

This was prompted by #9349 having failures in `system-tests` due to the use of `WARNING` instead of `WARN`. 

The telemetry API docs currently specify:

Log level|Remote collection policy
---|--- 
ERROR|Enabled by default, configured by environment variable `DD_TELEMETRY_LOG_COLLECTION_ENABLED`
WARN|Enabled by default, configured by same environment variable as ERROR
DEBUG|Can be enabled locally by the end user when asked by the support team. 

## Checklist

## Reviewer Checklist

- [x] Title is accurate


gnufede (approved):
:shipit:

erikayasuda (approved):

romainkomorndatadog:
This and #9349 are going to conflict with one another depending who merges first (#9349 will have to get updated since it uses `add_log()` as well).

tests/telemetry/test_telemetry_metrics.py
```diff
@@ -427,11 +428,11 @@ def test_send_multiple_log_metric_no_duplicates_for_each_interval_check_time(tel
     with override_global_config(dict(_telemetry_dependency_collection=False)):
         for _ in range(3):
             sleep(0.1)
-            telemetry_writer.add_log("WARNING", "test error 1")
+            telemetry_writer.add_log(TELEMETRY_LOG_LEVEL.WARN, "test error 1")
```
mabdinur, line 431: nit: We should use warning instead of warn. I think this is the convention used in pythons built in logging library. It would be nice to make it consistent.

ddtrace/internal/telemetry/writer.py
```diff
@@ -55,6 +55,7 @@
 from .constants import TELEMETRY_INJECT_WAS_ATTEMPTED
 from .constants import TELEMETRY_LIB_INJECTION_FORCED
 from .constants import TELEMETRY_LIB_WAS_INJECTED
+from .constants import TELEMETRY_LOG_LEVEL  # noqa:F401
```
mabdinur, line 58: Do we need no noqa here? Can we format this file using 'hatch run fmt:lint'

romainkomorndatadog: I had to `noqa` here because it's only used in the type hint comment and would show up as an unused import.

We don't have a `fmt` hatch env, we do have a `lint` env with a `fmt` script, but it doesn't touch this.

tests/telemetry/test_telemetry_metrics.py
romainkomorndatadog: Fair point. I'll make that change. I thought it made sense to match what the backend expects but it makes more sense to match what's pythonic.

mabdinur (approved):

PR #9527, opened 2024-06-12
ci: add datadog static analyzer

juli1:
## What does this PR do?

This PR adds supports for Datadog static analysis. It only checks rules that validates the GitHub actions are safe and secure.

## Motivation

We want to ensure our GitHub actions are safe and secure.

## Related

[PR](https://github.com/DataDog/datadog-agent/pull/26627) in `datadog-agent`


.github/workflows/datadog-static-analysis.yml
```diff
@@ -0,0 +1,21 @@
...
+jobs:
+  static-analysis:
+    runs-on: ubuntu-latest
+    name: Datadog Static Analyzer
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+    - name: Check code meets quality and security standards
+      id: datadog-static-analysis
+      uses: DataDog/datadog-static-analyzer-github-action@v1
```
erikayasuda, line 14: Sorry, I'm unfamiliar with this, so possibly dumb question. I was reading the docs for DD static analysis and saw:
> Static Analysis takes the guesswork out of adhering to an organization’s code standards, enabling your development team to ship compliant code without significant impacts to developer velocity.

Do we define these rules somewhere? Just want to know what rules we'll be applying to the repo so we can fix any issues before merging this (otherwise I'm worried it'll block our PRs or become an un-required check that gets ignored)

juli1: The rules used are defined in the `static-analysis.datadog.yml` file, defined at the root of the repository. For this PR, we only add rules requested by security but more rules are available per [our documentation](https://docs.datadoghq.com/code_analysis/static_analysis_rules/).

erikayasuda: Sorry to clarify, I meant where are rulesets like `sit-ci-best-practices` defined? I did a quick search in GitHub and the docs but I just see them being called in other repos `static-analysis.datadog.yml` files

juli1: For visibility: clarified on Slack since this ruleset is not documented now.

.github/workflows/datadog-static-analysis.yml
```diff
@@ -0,0 +1,21 @@
...
+      with:
+        dd_api_key: ${{ secrets.DD_API_KEY }}
```
romainkomorndatadog, line 16: Since we already have an API key for CI purposes, I'd like to suggest that we create keys specifically for this purpose and give them specific names.

eg: `secrets.STATIC_ANALYZER_DD_API_KEY` (or a better name).

That way, if a key leaks, we can know which one to change.

juli1: Yes, this is what we have done on the datagod-agent. See https://github.com/DataDog/datadog-agent/pull/26627/files

PR #9555, opened 2024-06-14
fix(flask): fix crashes with flask-like frameworks

sabrenner:
## What does this PR do
Adds `provide_automatic_options` as an extra named argument to `patched_add_url_rule`. Previously, when using a framework like [`flask-openapi3`](https://pypi.org/project/flask-openapi3/), and running the Hello World example linked using `ddtrace-run`, users would see the following:

```
File "/Users/sam.brenner/Development/ml-obs/sandboxes/support/app.py", line 18, in <module>
    def get_book(query: BookQuery):
  File "/Users/sam.brenner/Development/ml-obs/sandboxes/support/.venv/lib/python3.10/site-packages/flask_openapi3/scaffold.py", line 182, in decorator
    self._add_url_rule(rule, view_func=view_func, **options)
  File "/Users/sam.brenner/Development/ml-obs/sandboxes/support/.venv/lib/python3.10/site-packages/flask_openapi3/openapi.py", line 411, in _add_url_rule
    self.add_url_rule(rule, endpoint, view_func, provide_automatic_options, **options)
  File "/Users/sam.brenner/Development/ml-obs/sandboxes/support/.venv/lib/python3.10/site-packages/ddtrace/contrib/flask/patch.py", line 433, in patched_add_url_rule
    return _wrap(*args, **kwargs)
TypeError: patched_add_url_rule.<locals>._wrap() takes from 1 to 3 positional arguments but 4 were given
```

After this change, this error is no longer present.

This is because `flask-openapi3` passes in `provide_automatic_options` as an unnamed argument as the fourth positional argument, but is read on Flask as a keyword arg.

## Testing

Added a regression test that instruments a barebones `flask-openapi3` app, and runs Datadog instrumentation on it, asserting no errors. Additionally, this change was verified manually in a running app, for both instrumentation and execution.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


ddtrace/contrib/flask/patch.py
```diff
@@ -426,7 +426,9 @@ def _wrap(rule, endpoint=None, view_func=None, **kwargs):
 def patched_add_url_rule(wrapped, instance, args, kwargs):
     """Wrapper for flask.app.Flask.add_url_rule to wrap all views attached to this app"""
 
-    def _wrap(rule, endpoint=None, view_func=None, **kwargs):
+    # certain flask-like frameworks, like flaskopenapi-3, may call add_url_rule with extra arguments
+    # adding a catch-all *args avoids number of arguments mismatch
+    def _wrap(rule, endpoint=None, view_func=None, *args, **kwargs):
```
emmettbutler, line 431:
```suggestion
    def _wrap(rule, *args, endpoint=None, view_func=None, **kwargs):
```
I think this will be necessary to adhere to valid Python syntax

emmettbutler (changes requested):

ddtrace/contrib/flask/patch.py
sabrenner: ah yes fair, but I am getting new errors with that. I will work through those to see what's up!

sabrenner: so I did some more digging (apologies, my Flask knowledge in general isn't super), but it looks like `flask-openapi3` passes the `provide_automatic_options` argument, which is read as a `kwarg` in base Flask, as an `arg`, so it's not picked up properly in our integration (since we only account for `endpoint` and `view_func` as keyword args, and assume everything else is passed in as an `kwarg`, not `arg`. I added it in as a variable in [c1d1b4b](https://github.com/DataDog/dd-trace-py/pull/9555/commits/c1d1b4b807503d0f5ed1bd8ed7e508c279700f88). AFAICT this is fine, and Flask tests are passing, and my local dummy app using `flask-openapi3` doesn't crash. lmk what you think, and i'll update the PR description and release note to be a bit more specific.

emmettbutler (changes requested):
I'm surprised that this code change makes a difference in behavior. Could you add a regression test that proves that it has the intended effect?

sabrenner:
Hey @emmettbutler ! Sorry for delays, I just got back around to this PR today.

> I'm surprised that this code change makes a difference in behavior. Could you add a regression test that proves that it has the intended effect?

I added a test. I also had to re-run `scripts/ddtest scripts/compile-and-prune-test-requirements`, which generated a BUNCH of LOC for riot requirements files. Let me know if I did this incorrectly or inefficiently, happy to rerun a command if it's doable with a smaller amount of changes 😅 

As for the test itself, ran it locally without the changes to `patch.py`, and had the error show up:

```
ddtrace_run_python_code_in_subprocess = <function ddtrace_run_python_code_in_subprocess.<locals>._run at 0x108e6df30>

    def test_flask_openapi3_instrumentation(ddtrace_run_python_code_in_subprocess):
        code = """
    from flask_openapi3 import Info, Tag
    from flask_openapi3 import OpenAPI
    
    info = Info(title="my API", version="1.0.0")
    app = OpenAPI(__name__, info=info)
    
    @app.get("/")
    def hello_world():
        return 'Hello, World!'
    """
        env = os.environ.copy()
        out, err, status, pid = ddtrace_run_python_code_in_subprocess(code, env=env)
>       assert status == 0, (out, err)
E       AssertionError: (b'', b'Traceback (most recent call last):
E           File "/private/var/folders/94/lnn2crkx7914c2835zmwrwvh0000gq/T/pytest-of...**kwargs)
E         TypeError: patched_add_url_rule.<locals>._wrap() takes from 1 to 3 positional arguments but 4 were given
E         ')
E       assert 1 == 0

tests/contrib/flask/test_flask_openapi3.py:17: AssertionError
```

As shown in the CI runs on this PR, with the fix in place, those errors don't show up. Happy to expand on this test if this isn't what you had in mind!

emmettbutler (approved):

PR #9588, opened 2024-06-18
fix(futures): fix incorrect context propgation with ThreadPoolExecutor

brettlangdon:
There is a bug when scheduling work onto a `ThreadPoolExecutor` and not waiting for the response (e.g. `pool.submit(work)`, and ignoring the future) we not properly associate the spans created in the task with the trace that was active when submitting the task.

The reason for this bug is because we propagate the currently active span (parent) to the child task, however, if the parent span finishes before the child task can create it's first span, we no longer consider the parent span active/available to inherit from. This is because our context management code does not work if passing spans between thread or process boundaries.

The solution is to instead pass the active span's Context to the child task. This is a similar process as passing context between two services/processes via HTTP headers (for example). This change will allow the child task's spans to be properly associated with the parent span regardless of the execution order.

This issue can be highlighted by the following example:

```python

pool = ThreadPoolExecutor(max_workers=1)

def task():
    parent_span = tracer.current_span()
    assert parent_span is not None
    time.sleep(1)


with tracer.trace("parent"):
    for _ in range(10):
        pool.submit(task)
```

The first execution of `task` will (probably) succeed without any issues because the parent span is likely still active at that time. However, when each additional task executes the assertion will fail because the parent span is no longer an active span so `tracer.current_span()` will return `None`.

This example shows that only the first execution of `task` will be properly associated with the parent span/trace, the other calls to `task` will be disconnected traces.

This fix will resolve this inconsistent and unexpected behavior to ensure that the spans created in `task` will always be properly associated with the parent span/trace.

This change may impact people who were expecting to access the current span in the child task, but before creating any spans in the child task (the code sample above), as the span will no longer be available via `tracer.current_span()`.

## Checklist

## Reviewer Checklist

- [x] Title is accurate

### DIFF
--- a/ddtrace/contrib/futures/threading.py
+++ b/ddtrace/contrib/futures/threading.py
@@ -1,4 +1,7 @@
+from typing import Optional
+
 import ddtrace
+from ddtrace._trace.context import Context
 
 
 def _wrap_submit(func, args, kwargs):
@@ -17,9 +20,10 @@ def _wrap_submit(func, args, kwargs):
     #
     #      The resolution is to not create/propagate a new context if one does not exist, but let the
     #      future's thread create the context instead.
-    current_ctx = None
+    # DEV: Be sure to propagate a Context and not a Span since we are crossing thread boundaries
+    current_ctx: Optional[Context] = None
     if ddtrace.tracer.context_provider._has_active_context():
-        current_ctx = ddtrace.tracer.context_provider.active()
+        current_ctx = ddtrace.tracer.current_trace_context()
 
     # The target function can be provided as a kwarg argument "fn" or the first positional argument
     self = args[0]
@@ -31,7 +35,7 @@ def _wrap_submit(func, args, kwargs):
     return func(self, _wrap_execution, current_ctx, fn, fn_args, kwargs)
 
 
-def _wrap_execution(ctx, fn, args, kwargs):
+def _wrap_execution(ctx: Optional[Context], fn, args, kwargs):
     """
     Intermediate target function that is executed in a new thread;
     it receives the original function with arguments and keyword
--- a/ddtrace/contrib/futures/threading.py
+++ b/ddtrace/contrib/futures/threading.py
@@ -10,20 +10,8 @@ def _wrap_submit(func, args, kwargs):
     thread. This wrapper ensures that a new `Context` is created and
     properly propagated using an intermediate function.
     """
-    # If there isn't a currently active context, then do not create one
-    # DEV: Calling `.active()` when there isn't an active context will create a new context
-    # DEV: We need to do this in case they are either:
-    #        - Starting nested futures
-    #        - Starting futures from outside of an existing context
-    #
-    #      In either of these cases we essentially will propagate the wrong context between futures
-    #
-    #      The resolution is to not create/propagate a new context if one does not exist, but let the
-    #      future's thread create the context instead.
     # DEV: Be sure to propagate a Context and not a Span since we are crossing thread boundaries
-    current_ctx: Optional[Context] = None
-    if ddtrace.tracer.context_provider._has_active_context():
-        current_ctx = ddtrace.tracer.current_trace_context()
+    current_ctx: Optional[Context] = ddtrace.tracer.current_trace_context()
 
     # The target function can be provided as a kwarg argument "fn" or the first positional argument
     self = args[0]
--- a/tests/contrib/futures/test_propagation.py
+++ b/tests/contrib/futures/test_propagation.py
@@ -1,4 +1,4 @@
-import concurrent
+import concurrent.futures
 import time
 
 import pytest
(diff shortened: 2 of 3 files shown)


juanjux (approved):

ddtrace/contrib/futures/threading.py
```diff
@@ -17,9 +20,10 @@ def _wrap_submit(func, args, kwargs):
     #
     #      The resolution is to not create/propagate a new context if one does not exist, but let the
     #      future's thread create the context instead.
-    current_ctx = None
+    # DEV: Be sure to propagate a Context and not a Span since we are crossing thread boundaries
+    current_ctx: Optional[Context] = None
```
mabdinur, line 24:
```suggestion
    current_ctx: Optional[Context] = ddtrace.tracer.current_trace_context()
```

ddtrace/contrib/futures/threading.py
```diff
@@ -17,9 +20,10 @@ def _wrap_submit(func, args, kwargs):
...
     if ddtrace.tracer.context_provider._has_active_context():
-        current_ctx = ddtrace.tracer.context_provider.active()
+        current_ctx = ddtrace.tracer.current_trace_context()
```
mabdinur, line 26:
```suggestion
```

ddtrace/contrib/futures/threading.py
```diff
@@ -17,9 +20,10 @@ def _wrap_submit(func, args, kwargs):
     #
     #      The resolution is to not create/propagate a new context if one does not exist, but let the
     #      future's thread create the context instead.
-    current_ctx = None
+    # DEV: Be sure to propagate a Context and not a Span since we are crossing thread boundaries
+    current_ctx: Optional[Context] = None
     if ddtrace.tracer.context_provider._has_active_context():
```
mabdinur, line 25: We can remove this line. We don't need to check this internal api. If the tracer does not have an active context then the suggestion above will return None.

tests/contrib/futures/test_propagation.py
```diff
@@ -1,14 +1,24 @@
-import concurrent
+import concurrent.futures
```
mabdinur, line 1: Do we need to update this import?

tests/contrib/futures/test_propagation.py
```diff
@@ -384,3 +437,58 @@ def test_concurrent_futures_with_gevent():
...
+
+    futures = []
+
+    def work():
+        # This is our expected scenario
+        assert tracer.current_trace_context() is not None
+        assert tracer.current_span() is None
+
+        # DEV: This is the regression case that was raising
+        # tracer.current_span().set_tag("work", "done")
```
mabdinur, line 453: Should we remove this comment?
```suggestion
        tracer.current_span().set_tag("work", "done")
```

tests/contrib/futures/test_propagation.py
```diff
@@ -154,7 +186,10 @@ def fn():
                 self.assertEqual(result, 42)
 
         # the trace must be completed
-        self.assert_structure(dict(name="executor.thread"))
```
mabdinur, line 157: Why do are we removing self.assert_structure from tests? Is this method broken? Do we need to remove it in this PR?

tests/contrib/futures/test_propagation.py
brettlangdon: my editor wasn't happy, and this is technically more correct, since `futures` might not be imported by `concurrent`.

"no", but 🤷🏻

tests/contrib/futures/test_propagation.py
brettlangdon: yes, because we are not propagating a context there isn't a strong link between the two.

this `assert_structure` stuff isn't very good/smart, because the `executor.thread` span doesn't have a `_parent` which points to the `main.thread` span, all of this stuff breaks.

we could do it in a separate PR, but the reason these assertions need to be refactored is because of the change being made in this PR.

tests/contrib/futures/test_propagation.py
brettlangdon: no, the comment is here to showcase the example that would cause the exception.

uncommenting will cause these tests to fail.

ddtrace/contrib/futures/threading.py
brettlangdon: right! I was staring at the code too long, I thought that the dev comment above was still true.

tests/contrib/futures/test_propagation.py
mabdinur: Ahh I misread this test but user's should never use  the result of ` tracer.current_span()` without a None check. `tracer.current_span()` has the type Optional[Span] so `tracer.current_span().set_tag(...)` should log a typing error/warning.

mabdinur (approved):

tests/contrib/futures/test_propagation.py
```diff
@@ -224,15 +262,14 @@ def fn():
...
-        self.assert_structure(
-            dict(name="main.thread"),
-            (
-                (
-                    dict(name="executor.thread"),
-                    (dict(name="nested.thread"),),
-                ),
-            ),
-        )
+        spans = self.get_spans()
```
erikayasuda, line 265: NB: Why are we calling `self.get_spans()` instead of `self.get_root_spans()` here (to access the `"main.thread"` span)?

erikayasuda (approved):

tests/contrib/futures/test_propagation.py
brettlangdon: 🤷🏻 I think just a decision in the moment, since there are only 3 spans I figured it would be simple enough to code this way

PR #9604, opened 2024-06-20
feat(llmobs): capture input args and resp in function decorators

Yun-Kim:
This PR introduces automatic input/output annotation for task/tool/workflow/agent/retrieval spans created using function decorators. Specifically, the input args/kwargs provided to a traced function will be captured as a dictionary of key-value pairs, and the return value(s) of the function will also be captured as a JSON-serialized object (or a tuple of JSON-serialized objects if multiple return values).

### Example
```python
@workflow
def traced_workflow(prompt, arg_2, kwarg_1=None):
    formatted_output = ...
    return formatted_output
```
Screenshot 2024-06-25 at 3 30 05 PM

### Limitations / Future Steps

There are 2 limitations/special cases introduced by this PR:
1. If a user manually annotates the span I/O, auto-annotation will be overwritten. This is to avoid complications with data merging auto/manual annotations due to the different types of I/O data that can be annotated manually. Future steps include resolving this behavior by auto-merging auto/manual annotations when possible.

2. This PR does not include auto annotation for LLM/embedding spans, or output annotation for retrieval spans, due to the specialized I/O that those span kinds have which raises a whole new can of worms on how to automatically store/format I/O for those span kinds based on the traced function signature. Future steps include figuring out how we can automate annotation for more specialized I/O cases such as LLM/embedding/retrieval spans.

### Notes

Additionally, this PR adds a private option to each decorator to disable automatic annotation (not public) in case users don't want automatic annotation to happen.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


sabrenner (approved):
LGTM from the standpoint of what was outlined in the PR description, but one opinion:

IMHO i do feel that it would make more sense to _not_ override direct usage of `LLMObs.annotate()` calls by users in their methods as a default behavior.

from a users perspective, i would want the use of `LLMObs.annotate` to be intentional. If they already just basically dump I/O from `args`, `kwargs` and return value directly, then obv it's not doing anything different from this PR. but if they do some specific formatting for tagging I/O on these annotations, I would think they wouldn't expect that to change (but, with this PR, it would get overwritten), and shouldn't be expected to necessarily use a private option to disable it in that case. wdyt?

ddtrace/llmobs/decorators.py
```diff
@@ -66,8 +67,15 @@ def wrapper(*args, **kwargs):
...
-                with traced_operation(name=span_name, session_id=session_id, ml_app=ml_app):
-                    return func(*args, **kwargs)
+                with traced_operation(name=span_name, session_id=session_id, ml_app=ml_app) as span:
+                    func_signature = signature(func)
+                    bound_args = func_signature.bind_partial(*args, **kwargs)
+                    if _automatic_io_annotation and bound_args.arguments:
+                        LLMObs.annotate(span=span, input_data=bound_args.arguments)
+                    resp = func(*args, **kwargs)
+                    if _automatic_io_annotation and resp and operation_kind != "retrieval":
+                        LLMObs.annotate(span=span, output_data=resp)
```
lievan, line 77: if a user manually annotated `output_data` I think it will be over-written here? Or would a user need to specify `_automatic_io_annotation=False` every time they want to manually annotate a span? 

I'm guessing we need to add logic in annotate to make sure automatic annotations do not override manual annotations

Yun-Kim: Discussed offline - we will only do auto-annotations if manual annotations are not present

ddtrace/llmobs/decorators.py
```diff
@@ -66,8 +69,20 @@ def wrapper(*args, **kwargs):
                 if span_name is None:
                     span_name = func.__name__
                 traced_operation = getattr(LLMObs, operation_kind, "workflow")
-                with traced_operation(name=span_name, session_id=session_id, ml_app=ml_app):
-                    return func(*args, **kwargs)
+                with traced_operation(name=span_name, session_id=session_id, ml_app=ml_app) as span:
+                    func_signature = signature(func)
+                    bound_args = func_signature.bind_partial(*args, **kwargs)
+                    if _automatic_io_annotation and bound_args.arguments and span.get_tag(INPUT_VALUE) is None:
```
sabrenner, line 75: nit: we might not need the `span.get_tag(INPUT_VALUE)` check here, right? i think the only case this could be set is if somehow `LLMObs.annotate()` was called before the function was invoked. but correct me if I'm wrong! i think how it was before already allowed users to override automatic annotations for inputs, it was just the outputs that needed the change.

sabrenner (approved):
lgtm, thanks for this change! just left one nit

lievan (approved):

releasenotes/notes/feat-llmobs-function-decorator-automatic-io-annotation-acf41b686e39d20c.yaml
```diff
@@ -0,0 +1,5 @@
+---
+features:
+  - |
+    LLM Observability: This introduces automatic input/output annotation for task/tool/workflow/agent/retrieval spans traced by function decorators.
```
lievan, line 4: Thoughts on:
"task/tool/workflow/agent/retrieval spans created using decorators will have inputs and outputs be automatically annotated.
Manual annotations for input/output values override automatic annotations."

erikayasuda (approved):

releasenotes/notes/feat-llmobs-function-decorator-automatic-io-annotation-acf41b686e39d20c.yaml
erikayasuda: nit: I personally like @Yun-Kim 's first sentence, but I like @lievan 's second sentence.

PR #9605, opened 2024-06-21
chore(telemetry): decouple tests from implementation internals and fix flaky tests

mabdinur:
Changes to the Telemetry Writer
- Ensures `app-closing` event is queued last (after dependencies, configurations, metrics, and log events)
- Ensures all list are reset when `TelemetryWriter.reset_queues()` is called

- Refactors
  - 0cc7cf8ec58981f519032b337abf599bda8d711c Removes `filter_heartbeats` fixture. This fixture removes events from telemetry sessions by default. This can make tests hard to debug/understand. Instead of removing heartbeats via a fixture we now pass the `filter_heartbeats` argument to `TelemetryTestSession.get_event(...)`. So the filter can be enabled/disabled when querying events. This should make the tests easier to read and maintain.
  - 324f1eb0d98bf8d641e43c9bf1cee7488c083b21 Removes `_assert_dependencies_sort_and_remove` and `get_default_telemetry_env`. These helpers make the code harder to read and are unnecessary. I believe these helpers were introduced just to "make tests pass". I refactored impacted tests so they no longer require these helpers.
  - 61de29a0de9f3cae31a45e74e479884a4814d0cf Refactors _assert_metric and _assert_log helpers to avoid asserting all fields in a telemetry payloads. Payloads contain sequence ids and timestamps that are highly dependent on implementation details and test helpers. This change limits the assertion to only match the expected metric/log.
  - 4f045e04b1f80231d5348443ca9b5aa7e30cad68 Removes `initial_event_count = len(test_agent_session.get_events())` and assertions around the initial count of events from telemetry tests. Previously TelemetryTestSession captured all payloads sent to the agent. This made request counts unreliable and these counters were added to improve the reliability of tests. This PR attempts to resolve the underlying issue so this hack is no longer needed.
  - fe45da9264f54b34e6dc2be79362710ba042956a Renames `_update_dependencies_event` method to `_app_dependencies_loaded_event` to align with other telemetry  writer events.
  - 629c5819057d87b7186a2909d7c37cd70a0d7852 4f045e04b1f80231d5348443ca9b5aa7e30cad68 Ensure `TelemetryAgentSession.get_events(request_type=.....)` is used to retrieve events.  

- Fixes
  -  370a3c979c418db31bf64e83881ffb8e6bc3aaee In `tests/utitils:spanshot_context` `RemoteDisconnected` is imported  from `ddtrace.ext.http`. This is incorrect `RemoteDisconnected` should be imported from `http.client`. This results in a syntax error when snapshot tests to connect to the agent: [failure](https://app.circleci.com/pipelines/github/DataDog/dd-trace-py/64068/workflows/51694040-d0a5-4416-ba41-37f25e3b431a/jobs/3968306) 
   - 633f097ee966847a08d784c01a97a2ffcfbe01e7 Fix for decoding telemetry payloads: Ensures only telemetry payloads are decoded in `TelemetryTestSession.get_requests(...)`. Previously this method would capture and attempt to decode ALL payloads sent to the agent. The decoder expected utf8 encoding and would fail to unpack msgpack trace payloads cauing tests to fail. Moving forward the telemetry test agent session will only capture and decode telemetry requests.
  - 89d3a313520c60eb811f2bcc17f2ce368ed71378 DD_PROFILING_EXPORT_LIBDD_ENABLED can not be set to true on macos. This platform does not support ddup. However this configuration can be enabled in ci 
  -  a9e2d7ea34227a6e77b397c275c3e99d7a6bc0de Ensures telemetry events use `periodic(force_flish=True)` instead of `periodic()`. Tests that require events to be flushed should override the default batching behavior of periodic and ensure events are sent to the agent. This will improve the reliability of tests (by default peridoic batches events and only sends events every minute). 
  - 0cc7cf8ec58981f519032b337abf599bda8d711c Ensure app-closing is queued as the final event. This aligns the python implementation with other client libraries (node, go, java, etc). 
  - 98e02ec6a39867b43f7e42f9e3e2eb0347498e52 DD_PROFILING_EXPORT_LIBDD_ENABLED (ddtrace.settings.profiling.export.libdd_enabled) configuration is platform dependent. This configuration can not be enabled on platforms that do not support libdatadog. For now we should avoid enabling this configuration in telemetry configurations tests.  

## Checklist

## Reviewer Checklist

- [x] Title is accurate


tests/telemetry/test_telemetry.py
```diff
@@ -65,23 +33,24 @@ def test_telemetry_enabled_on_first_tracer_flush(test_agent_session, ddtrace_run
...
-    )
+    events = test_agent_session.get_events()
 
-    assert len(events) == 4
-    assert events[0]["request_type"] == "app-closing"
-    assert events[1]["request_type"] == "app-integrations-change"
-    assert events[2]["request_type"] == "app-started"
-    assert events[3]["request_type"] == "generate-metrics"
+    assert len(events) == 5
+    # Generate metrics is sent after app-closed event. This is because the span aggregator is shutdown after the
```
erikayasuda, line 44: Should this be flipped to reflect the ordering you have in the assertions?
```suggestion
    # app-closed is sent after the generate-metrics event. This is because the span aggregator is shutdown after the
```

tests/telemetry/test_writer.py
```diff
@@ -45,37 +44,28 @@ def test_add_event(telemetry_writer, test_agent_session, mock_time):
...
 
 def test_app_started_event(telemetry_writer, test_agent_session, mock_time):
     """asserts that _app_started_event() queues a valid telemetry request which is then sent by periodic()"""
     with override_global_config(dict(_telemetry_dependency_collection=False)):
-        initial_event_count = len(test_agent_session.get_events())
         # queue an app started event
         telemetry_writer._app_started_event()
         # force a flush
-        telemetry_writer.periodic()
+        telemetry_writer.periodic(force_flush=True)
```
erikayasuda, line 63: Is adding `force_flush=True` meant to be the refresh state between tests?

erikayasuda (approved):
Just a few clarifying questions

tests/utils.py
```diff
@@ -1026,7 +1027,7 @@ def snapshot_context(
             while r is None and time.time() - attempt_start < 60:
                 try:
                     r = conn.getresponse()
-                except http.client.RemoteDisconnected:
```
mabdinur, line 1029: ddtrace.ext.http is shadowing the built http module. This results in this error: https://app.circleci.com/pipelines/github/DataDog/dd-trace-py/64068/workflows/51694040-d0a5-4416-ba41-37f25e3b431a/jobs/3968306

tests/telemetry/test_writer.py
mabdinur: We use telemetry_writer.reset_queues() to reset the state between tests. This is done in the telemetryagentsession fixture. 


By default the periodic writer thread is configured to run every 10 seconds. On every call periodic aggregates metrics and logs and on the six invocation (on 1 minute intervals) it submits telemetry events to the agent (these intervals can be overridden via envars).

Here we call periodic with forced_flushed=True to ensure periodic always sends all queued telemetry events.

PR #9609, opened 2024-06-21
chore(ci_visibility): collect module-level line execution and imports

romainkomorndatadog:
This introduces changing to the internal coverage collector to capture import-time dependencies (necessary for proper representation of test coverage with ITR).

This is achieved in two ways:
- by collecting coverage during the module execution part of the import process
- by instrumenting import statements (both at module level and below) to capture executed imports

The data is stored in three separate structures on the `ModuleCodeCollector` instance:
- `_import_time_covered`: contains lines covered in imports, by path string
- `_import_time_name_to_path`: containing a mapping of module names to path strings
- `_import_names_by_path`: containing a mapping of which paths executed which imports, with the package and import name

The dependency collection is optional (passed in as a flag to the coverage installer), and the collected data is also used optionally when reporting coverage (since a BFS is executed to traverse all dependencies for the files covered).

Also added is a set of tests that try and cover both module-level imports, and in-function imports, using both regular and namespace packages, with relative imports.

Some other changes are piggybacked into this PR, including:
- Removing the instrumentation for Python 3.9 (since it is identical to 3.8)
- Introducing instrumentation for Python 3.7 (which is nearly identical to 3.8, except that `CodeType.replace()` is not available in 3.7)
- Removal of the native C-based bytecode implementation since it is no longer in use for 3.7 as a fallback
- A fix to threading coverage to ensure that we don't block on an empty coverage queue (and still exit the coverage context even in the event of an exception)

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/internal/coverage/code.py
```diff
@@ -191,54 +291,96 @@ def report_seen_lines(cls):
...
                 file_segments.append([file_segment[0], 0, file_segment[1], 0, -1])
-            files.append({"filename": path, "segments": file_segments})
+            files.append({"filename": path_str, "segments": file_segments})
 
         return files
 
     def transform(self, code: CodeType, _module: ModuleType) -> CodeType:
         code_path = Path(code.co_filename)
+        if _module is None:
+            return code
```
P403n1x87, line 315:
```suggestion
        if _module is None:
            return code

        code_path = Path(code.co_filename)
```

ddtrace/internal/coverage/code.py
```diff
@@ -191,54 +291,96 @@ def report_seen_lines(cls):
...
 
         if not any(code_path.is_relative_to(include_path) for include_path in self._include_paths):
             # Not a code object we want to instrument
             return code
 
-        return self.instrument_code(code)
+        if any(code_path.is_relative_to(_) for _ in (stdlib_path, purelib_path)):
```
P403n1x87, line 321: I think the logic for detecting stdlib might be more complicated because of how things are installed https://github.com/P403n1x87/dd-trace-py/blob/c81984899843c776254ed30d35de232d87d3e915/ddtrace/internal/packages.py#L226

ddtrace/internal/coverage/instrumentation_py3_12.py
```diff
@@ -9,44 +9,129 @@
...
+            _, nested_lines = instrument_all_lines(nested_code, hook, path, package)
             lines.update(nested_lines)
 
         # Register the hook and argument for the code object
-        _CODE_HOOKS[code] = (hook, path)
+        _CODE_HOOKS[code] = (hook, path, import_names)
+
+        # Special case for empty modules (eg: __init__.py ):
+        # Make sure line 0 is marked as executable, and add package dependency
+        if not lines and code.co_name == "<module>" and code.co_code == bytes([151, 0, 121, 0]):
```
P403n1x87, line 132: Can we replace the literal bytes with constants from `dis` to make this more readable?

ddtrace/internal/module.py
```diff
@@ -265,12 +284,25 @@ def get_code(_loader, fullname):
...
+                for _ in sys.meta_path:
+                    if isinstance(_, ModuleWatchdog):
+                        for (
+                            cond,
+                            hook,
+                        ) in _._import_exception_hooks:
+                            if (isinstance(cond, str) and cond == module.__name__) or (
+                                callable(cond) and cond(module.__name__)
+                            ):
+                                hook(self, module)
```
P403n1x87, line 304: Would it make sense to move this logic into a dedicated function/method?

tests/coverage/test_coverage_relative_regular_packages.py
```diff
@@ -0,0 +1,394 @@
...
+    imports_rp_dot_normal()
+    ModuleCodeCollector.stop_coverage()
+
+    executable = _get_relpath_dict(cwd_path, ModuleCodeCollector._instance.lines)
+    covered = _get_relpath_dict(cwd_path, ModuleCodeCollector._instance._get_covered_lines(include_imported=False))
+    covered_with_imports = _get_relpath_dict(
+        cwd_path, ModuleCodeCollector._instance._get_covered_lines(include_imported=True)
+    )
+
+    # In Python 3.11+, empty __init__.py modules have line 0:
```
erikayasuda, line 33: Oh, I think this answers my question?

erikayasuda (approved):
Sorry this took so long. This review is mostly questions for clarification, so mostly NBs.

ddtrace/internal/coverage/code.py
```diff
@@ -10,28 +12,41 @@
...
 _original_exec = exec
 
 ctx_covered = ContextVar("ctx_covered", default=None)
+ctx_is_import_coverage = ContextVar("ctx_is_import_coverage", default=False)
 ctx_coverage_enabled = ContextVar("ctx_coverage_enabled", default=False)
 
 
-class ModuleCodeCollector(BaseModuleWatchdog):
+def _get_ctx_covered_lines() -> t.DefaultDict[str, t.Set]:
+    return ctx_covered.get()[-1] if ctx_coverage_enabled.get() else defaultdict(set)
```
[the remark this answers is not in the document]
romainkomorndatadog: The data in the context coverage is a stack (so that nested usage of the context-based collector works properly).

If we've entered the context (which is what sets `ctx_coverage_enabled` to `True` ( at https://github.com/DataDog/dd-trace-py/pull/9609/files#diff-3985b6c4c52757d7d726c3002c85935f582641d687acd0f6ab50f93e1294205dR232-R233 )), then we've already created the stack and added an entry to it.

The only way `ctx_coverage_enabled` becomes `False` is if we `__exit__` the context collector and the stack is empty (at https://github.com/DataDog/dd-trace-py/pull/9609/files#diff-3985b6c4c52757d7d726c3002c85935f582641d687acd0f6ab50f93e1294205dR245-R246 ).

romainkomorndatadog (commented):
Thanks for the questions/nits, @erikayasuda . Will address/fix.

anmarchenko (approved):
Very interesting idea! Did you by any chance run test-environment with these changes to estimate how much slow down does this approach bring?

romainkomorndatadog:
> Very interesting idea! Did you by any chance run test-environment with these changes to estimate how much slow down does this approach bring?

TL;DR: we're slightly faster than the original approach, with better functionality.

It's a bit fuzzy because the current strategy using coverage.py does not capture import-time dependencies. I mucked around with some other attempts to capture coverage at import time but `pytest`'s hooks weren't being called at the right time.

That said, our test env results are a tiny bit faster with this approach than with the original coverage.py-based approach. Anecdotally, the gathering of dependencies does come at a cost of about 10% more overhead than without, but overall has ~5-10% less overhead than the current approach.

There's one other aspect that is fuzzy here, and that is that this approach basically makes all tests in a suite have the same dependencies, which makes the behavior much more like suite-level skipping than test-level skipping, and I'll almost certainly end up updating the plugins to do suite-level skipping as a result, and the overhead with suite-level skipping is a little lower than with test-level skipping because we're sending less data over the wire and doing the BFS for dependencies coverage less often.

P403n1x87 (approved):
LGTM

PR #9615, opened 2024-06-21
fix(profiling): export lock acquire/release call location and variable name

taegyunkim:
We used to export "<filename>:<line_number>" for the name of a lock. Whenever lock is initialized, we traverse the stack frame to populate these. However, we can't get the variable name of the lock as we haven't yet fully initialized and assigned it to a variable.  

Now we traverse the frame whenever `_acquire/_release` is called to get the caller information. Also, when `_acquire` is called for the first time, we inspect local and global variables in the caller frame to find the name of this lock variable. With `DD_PROFILING_LOCK_NAME_INSPECT_DIR` enabled, the lock profiler also inspects attributes of local and global variables to find the name of the lock. This behavior can be turned off with the flag. 

## Checklist

## Reviewer Checklist

- [x] Title is accurate

### DIFF
--- a/ddtrace/profiling/collector/_lock.py
+++ b/ddtrace/profiling/collector/_lock.py
@@ -88,7 +88,6 @@ class _ProfiledLock(wrapt.ObjectProxy):
         self._self_capture_sampler = capture_sampler
         self._self_endpoint_collection_enabled = endpoint_collection_enabled
         self._self_export_libdd_enabled = export_libdd_enabled
-
         frame = sys._getframe(2 if WRAPT_C_EXT else 3)
         code = frame.f_code
 
@@ -113,17 +112,6 @@ class _ProfiledLock(wrapt.ObjectProxy):
         return self.__wrapped__.__aexit__(*args, **kwargs)
 
     def _acquire(self, inner_func, *args, **kwargs):
-        print("In ProfiledLock.acquire")
-
-        for i in range(7):
-            try:
-                frame = sys._getframe(i)
-                if frame is not None:
-                    code = frame.f_code
-                    print("Frame %d: %s:%d:%s" % (i, os.path.basename(code.co_filename), frame.f_lineno, code.co_name))
-            except ValueError:
-                break
-
         if not self._self_capture_sampler.capture():
             return inner_func(*args, **kwargs)
 
@@ -144,7 +132,6 @@ class _ProfiledLock(wrapt.ObjectProxy):
                 frames, nframes = _traceback.pyframe_to_frames(frame, self._self_max_nframes)
 
                 if self._self_export_libdd_enabled:
-                    print("Exporting using libdd")
                     thread_native_id = _threading.get_thread_native_id(thread_id)
 
                     handle = ddup.SampleHandle()
@@ -154,9 +141,6 @@ class _ProfiledLock(wrapt.ObjectProxy):
                     handle.push_threadinfo(thread_id, thread_native_id, thread_name)
                     handle.push_task_id(task_id)
                     handle.push_task_name(task_name)
-                    print(
-                        end, self._self_name, end - start, thread_id, thread_native_id, thread_name, task_id, task_name
-                    )
 
                     if self._self_tracer is not None:
                         handle.push_span(self._self_tracer.current_span(), self._self_endpoint_collection_enabled)
--- a/ddtrace/profiling/collector/_lock.py
+++ b/ddtrace/profiling/collector/_lock.py
@@ -110,6 +110,7 @@ class _ProfiledLock(wrapt.ObjectProxy):
                 end = self._self_acquired_at = compat.monotonic_ns()
                 thread_id, thread_name = _current_thread()
                 task_id, task_name, task_frame = _task.get_task(thread_id)
+                lock_name = self._get_lock_call_loc_with_name() or self._self_name
 
                 if task_frame is None:
                     frame = sys._getframe(1)
@@ -123,7 +124,7 @@ class _ProfiledLock(wrapt.ObjectProxy):
 
                     handle = ddup.SampleHandle()
                     handle.push_monotonic_ns(end)
-                    handle.push_lock_name(self._self_name)
+                    handle.push_lock_name(lock_name)
                     handle.push_acquire(end - start, 1)  # AFAICT, capture_pct does not adjust anything here
                     handle.push_threadinfo(thread_id, thread_native_id, thread_name)
                     handle.push_task_id(task_id)
@@ -136,7 +137,7 @@ class _ProfiledLock(wrapt.ObjectProxy):
                     handle.flush_sample()
                 else:
                     event = self.ACQUIRE_EVENT_CLASS(
-                        lock_name=self._self_name,
+                        lock_name=lock_name,
                         frames=frames,
                         nframes=nframes,
                         thread_id=thread_id,
@@ -169,6 +170,7 @@ class _ProfiledLock(wrapt.ObjectProxy):
                         end = compat.monotonic_ns()
                         thread_id, thread_name = _current_thread()
                         task_id, task_name, task_frame = _task.get_task(thread_id)
+                        lock_name = self._get_lock_call_loc_with_name() or self._self_name
 
                         if task_frame is None:
                             frame = sys._getframe(1)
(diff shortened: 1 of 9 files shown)


ddtrace/profiling/collector/_lock.py
```diff
@@ -233,6 +235,33 @@ def __enter__(self, *args, **kwargs):
     def __exit__(self, *args, **kwargs):
         self._release(self.__wrapped__.__exit__, *args, **kwargs)
 
+    # Get lock acquire/release call location and variable name the lock is assigned to
+    def _get_lock_call_loc_with_name(self) -> typing.Optional[str]:
+        try:
+            # We expect the call stack to be like this:
```
danielsn, line 241: Can we add a debug_assert here so we'll catch violations in test?

taegyunkim: I'm not sure whether Python or this repo has support for something similar to Rust's `std::debug_assert!`.

ddtrace/profiling/collector/_lock.py
```diff
@@ -233,6 +235,47 @@ def __enter__(self, *args, **kwargs):
...
+
+            var_name = None
+
+            # Search for global variables
+            for name, value in frame.f_globals.items():
+                if value == self:
+                    var_name = name
+                    break
+                for attribute in dir(value):
+                    if not attribute.startswith("__") and getattr(value, attribute) == self:
```
sanchda, line 259: Yeah, I think omitting `__`-prefixed attributes is pragmatic.  For a given object, in the typical case, _most_ of those will be inherited junk that won't be a lock (... until a class has a private lock we won't know about now), so this saves time.

One feature of this implementation is that we'll miss attributes of attributes.  In general, we'd have to do a recursive scan through the attribute tree in order to exhaust all possibilities.  I don't think that's worth doing, just working through the dynamics of the solution for my own benefit.

ddtrace/profiling/collector/_lock.py
```diff
@@ -233,6 +235,47 @@ def __enter__(self, *args, **kwargs):
...
         self._release(self.__wrapped__.__exit__, *args, **kwargs)
 
+    # Get lock acquire/release call location and variable name the lock is assigned to
+    def _get_lock_call_loc_with_name(self) -> typing.Optional[str]:
+        try:
+            # We expect the call stack to be like this:
+            # 0: this
+            # 1: _acquire/_release
+            # 2: acquire/release (or __enter__/__exit__)
+            # 3: caller frame
```
sanchda, line 245: This matches my mental model too--big improvement!

ddtrace/profiling/collector/_lock.py
```diff
@@ -233,6 +235,47 @@ def __enter__(self, *args, **kwargs):
...
+            for name, value in frame.f_globals.items():
+                if value == self:
+                    var_name = name
+                    break
+                for attribute in dir(value):
+                    if not attribute.startswith("__") and getattr(value, attribute) == self:
+                        var_name = attribute
+                        break
+            # Search for local variables
+            for name, value in frame.f_locals.items():
```
sanchda, line 263: All of these lookups--the list of locals, globals, and their attributes--have highly variable search bounds.

We know from some of our other reports that whenever we have instrumentation with this form, there's going to be some user who runs headfirst into the worst case scenario and then has a bad time.

Since this code will run on-GIL, assuming control from the user whenever a lock is created (or even acquired/released if my other comment isn't appropriate), I think there's value in setting a maximum search complexity.  I'd propose just counting the number of comparisons and stopping once a user-configurable maximum has been reached (it can be a profiling-related configurable setting in `ddtrace/settings/profiling.py`)

tests/profiling/collector/test_asyncio.py
```diff
@@ -70,16 +69,23 @@ async def test_lock_events_tracer(tracer):
             pass
     events = r.reset()
     # The tracer might use locks, so we need to look into every event to assert we got ours
+    lock1_acquire, lock1_release, lock2_acquire, lock2_release = (
+        "test_asyncio.py:59:lock",
```
sanchda, line 73: Beautiful!

ddtrace/profiling/collector/_lock.py
```diff
@@ -169,6 +170,7 @@ def _release(self, inner_func, *args, **kwargs):
                         end = compat.monotonic_ns()
                         thread_id, thread_name = _current_thread()
                         task_id, task_name, task_frame = _task.get_task(thread_id)
+                        lock_name = self._get_lock_call_loc_with_name() or self._self_name
```
sanchda, line 173: I think I'm confused.

If I'm reading the implementation correctly, this function will produce the context around the lock's identity.  The calling "user code" (file) in which it was created, the line number, and then the name of the variable or attribute to which it was assigned.  These are generally creation-time aspects.

However, `lock_name` is populated whenever the lock is released or acquired, despite not really producing any novel information related to the location of release or acquisition.  (the exception that comes to mind is if the lock is transferred to a different symbol, but I suspect this operation is rare)

If these are mostly true statements, would it make more sense to assign the lock name to an object attribute and populate the location upon our first acquisition of the lock?  The symbol lookup seems like it might be expensive.

ddtrace/profiling/collector/_lock.py
```diff
@@ -233,6 +235,47 @@ def __enter__(self, *args, **kwargs):
...
+                if value == self:
+                    var_name = name
+                    break
+                for attribute in dir(value):
+                    if not attribute.startswith("__") and getattr(value, attribute) == self:
+                        var_name = attribute
+                        break
+
+            if var_name:
+                call_loc += ":%s" % var_name
```
sanchda, line 273: If we're here, then we tried to find a variable name and we weren't able to.  We know that users are going to hit this condition and they're going to wonder why their lock is unnamed.

Users are generally not going to read the documentation for every feature.  They're going to interact with these objects in the UX first and foremost, so the names of things tend to reveal not only the context of objects within their application, but also the state of the _instrumentation_ when that context was gathered.  I propose adding some kind of default name indicating that the lookup for the symbol failed.

I wouldn't be totally heartbroken if you decided to keep it blank, since I personally interpret what you have to mean "There's no more info, go read the docs if this is a bummer for you."

sanchda (approved):
I added some comments for discussion, but nothing I wrote is set in stone or critical to the implementation.

I think there may be some value in comprehending the overhead of the symbol lookup operation, but I have literally never read a single comment from any user that our lock profiler has high overhead (the other collectors, unfortunately, are the opposite), so I think such a study can wait.

Ultimately, good stuff--a massive step in the right direction for the product!

ddtrace/profiling/collector/_lock.py
taegyunkim: Added a test case for a private lock, and it just works! 

re: recursive scan
Yeah, also added a test case to show that we can't find the name in this case. I also agree that the more worrying part is that we iterate through too many globals and some customers might be hit by that.

ddtrace/profiling/collector/_lock.py
```diff
@@ -233,6 +235,47 @@ def __enter__(self, *args, **kwargs):
...
+            # And we expect addtioinal frame if WRAPT_C_EXT is False
+            frame = sys._getframe(3 if WRAPT_C_EXT else 4)
+            code = frame.f_code
+            call_loc = "%s:%d" % (os.path.basename(code.co_filename), frame.f_lineno)
+
+            var_name = None
+
+            # Search for global variables
+            for name, value in frame.f_globals.items():
+                if value == self:
```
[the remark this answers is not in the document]
taegyunkim: https://docs.python.org/3/reference/expressions.html#is

Changed to use `is`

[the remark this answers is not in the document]
sanchda: Good call

ddtrace/profiling/collector/_lock.py
taegyunkim: Updated to populate lock name at the first invocation to the `acquire`.

ddtrace/profiling/collector/_lock.py
taegyunkim: Maybe that's a TODO in UI? In UI, we could check whether we have variable name by checking that the lock name is of the form `%s:%d:%s`or what not. And then show a button or something like "Can't find variable name for the lock?" and link to the docs?

brettlangdon (approved):
change to ddtrace/settings/profiling.py lgtm

ddtrace/profiling/collector/_lock.py
danielsn: We could mock it using a config variable?

taegyunkim: Oh that would work. I'll follow up with another PR.

PR #9627, opened 2024-06-24
feat(llmobs): use default placeholder instead of rejecting annotation

Yun-Kim:
This PR changes the behavior of `LLMObs.annotate()` to use default placeholder values of `[Unserializable object: <repr(obj)>]` if a specified object is not JSON serializable, whereas previously it would not be annotated at all. Additionally, given a list or dictionary of values to annotate, previously none of the values would be annotated onto the span, but this PR now annotates all serializable values normally and defaults to the placeholder for the non-serializable values.

Note that if an object is not JSON serializable and it has no `repr()` implementation (which is rare), this will still result in previous behavior of logging a warning and rejecting the annotation entirely.

### Example
```python
# Before
LLMObs.annotate(input_data={"abc": "def", "unserializable": UNSERIALIZABLE_OBJECT})
# logs warning, does not annotate input_data at all

# Now
LLMObs.annotate(input_data={"abc": "def", "unserializable": UNSERIALIZABLE_OBJECT})
# annotates input_data with {"abc": "def", "unserializable": "[Unserializable object: <Unserializable object at 0x103f4de70>]"}
```

## Checklist

## Reviewer Checklist

- [x] Title is accurate


sabrenner (approved):
lgtm

ddtrace/llmobs/_llmobs.py
```diff
@@ -560,7 +561,7 @@ def _tag_embedding_io(cls, span, input_documents=None, output_text=None):
                 span.set_tag_str(OUTPUT_VALUE, output_text)
             else:
                 try:
-                    span.set_tag_str(OUTPUT_VALUE, json.dumps(output_text))
+                    span.set_tag_str(OUTPUT_VALUE, json.dumps(output_text, default=_unserializable_default_repr))
                 except TypeError:
                     log.warning("Failed to parse output text. Output text must be JSON serializable.")
```
lievan, line 566: Should we update this error log to something more generic like "Failed to set input/output value for X span kind"?

>"Output text must be JSON serializable."

Also I think the wording of "must" here may be confusing, since this codepath is not reached even when the I/O is not JSON serializable

lievan, line 566: How about moving this warning log inside the `_unserializable_default_repr` function?

Something like:
```
def _unserializable_default_repr(obj):
    log.warning("Failed to parse I/O. I/O must be JSON serializable. Defaulting to X".)
    return "[Unserializable object: {}]".format(repr(obj))
```

ddtrace/llmobs/_llmobs.py
```diff
@@ -560,7 +561,7 @@ def _tag_embedding_io(cls, span, input_documents=None, output_text=None):
                 span.set_tag_str(OUTPUT_VALUE, output_text)
             else:
                 try:
-                    span.set_tag_str(OUTPUT_VALUE, json.dumps(output_text))
+                    span.set_tag_str(OUTPUT_VALUE, json.dumps(output_text, default=_unserializable_default_repr))
                 except TypeError:
```
lievan, line 565: should we expand this try/catch to be more generic (maybe just to Exception)? I don't think we should be expecting TypeError's anymore because of the `default` function, and `_unserializable_default_repr` doesn't throw a `TypeError`.

Yun-Kim: `_unserializable_default_repr()` does throw a `TypeError` if `repr(obj)` is not a string object. The exceptions should still stay there, as we need to catch those cases.

ddtrace/llmobs/_llmobs.py
Yun-Kim: I like the idea of adding a warning log to inside the unserializable_default_repr() method, but this warning log still needs to be there in cases the object is actually unserializable (and its default repr() method is also not serializable())

Yun-Kim: I'll update the wording, but I'd like to keep insistence on the I/O object being JSON serializable (the placeholder is just a default and should not be relied on to be normal behavior)

ZStriker19 (approved):

lievan (approved):

PR #9668, opened 2024-06-28
chore(ci_visibility): performance improvements for internal coverage collector

romainkomorndatadog:
This brings performance improvements of the internal coverage collector for Python 3.8 to 3.12 .

For 3.8 to 3.11, the improvements come from custom bytecode modification by limiting the impact to:
- inserting instructions for calling the collection hook
- updating absolute/relative jump values based on new instruction offsets and introduction of extended arguments (due to increased jump sizes or increased number of constants)
- recomputing `code.co_lnotab` or `code.co_linetable` to reflect updated offsets

For 3.12, the improvement comes from using `sys.monitoring` on line events.

For 3.7, further improvements are needed as the `code` object format is markedly different.

The instrumentations are broken down into per-version file, and while there is some repeated code between all versions, older versions should not need to be modified again and newer (eg: `3.13`) versions should be able to rely on the `3.12` implementation.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


ddtrace/contrib/pytest/plugin.py
```diff
@@ -821,6 +821,22 @@ def pytest_runtest_protocol(item, nextitem):
             _stop_coverage(pytest)
 
 
+def pytest_load_initial_conftests(early_config, parser, args):
+    from envier import En
+
+    from ddtrace.internal.coverage.code import ModuleCodeCollector
+
+    class CoverageConfig(En):
+        __prefix__ = "dd_coverage"
```
P403n1x87, line 830: nit
```suggestion
        __prefix__ = "dd.coverage"
```

ddtrace/internal/coverage/_native.c
P403n1x87, line 1: this can be removed

ddtrace/internal/coverage/_native.pyi
P403n1x87, line 1: this can be removed

ddtrace/internal/coverage/instrumentation_py3_10.py
```diff
@@ -0,0 +1,285 @@
...
+            return cls.BACKWARD if "BACKWARD" in dis.opname[opcode] else cls.FORWARD
+
+    class Jump(ABC):
+        def __init__(self, start: int, argbytes: list[int]) -> None:
+            self.start = start
+            self.end: t.Optional[int] = None
+            self.arg = int.from_bytes(argbytes, "big", signed=False)
+            self.argsize = len(argbytes)
+
+    # TODO: There are no absolute jumps in CPython >= 3.10
```
P403n1x87, line 28: IT'S A LIE!

ddtrace/internal/coverage/instrumentation_py3_10.py
```diff
@@ -0,0 +1,285 @@
+import sys
+
+
+if sys.version_info >= (3, 10) and sys.version_info < (3, 11):
```
P403n1x87, line 4: can we remove these checks?

setup.py
```diff
@@ -423,6 +423,13 @@ def get_exts_for(name):
             ],
             extra_compile_args=debug_compile_args,
         ),
+        Extension(
+            "ddtrace.internal.coverage._native",
+            sources=[
+                "ddtrace/internal/coverage/_native.c",
+            ],
+            extra_compile_args=debug_compile_args,
+        ),
```
P403n1x87, line 432:
```suggestion
```

ddtrace/internal/coverage/instrumentation_py3_10.py
romainkomorndatadog: Swapped them for asserts (I don't love it, but it's the only other option I've found).

romainkomorndatadog:
Note: There are further changes coming in #9609 that I don't really feel like backporting into this one just for the sake of it because I'll end up creating merge conflicts with myself.

ddtrace/internal/coverage/instrumentation_py3_10.py
```diff
@@ -0,0 +1,312 @@
...
+LOAD_CONST = dis.opmap["LOAD_CONST"]
+CALL = dis.opmap["CALL_FUNCTION"]
+POP_TOP = dis.opmap["POP_TOP"]
+
+
+def trap_call(trap_index: int, arg_index: int) -> t.Tuple[Instruction, ...]:
+    return (
+        *instr_with_arg(LOAD_CONST, trap_index),
+        *instr_with_arg(LOAD_CONST, arg_index),
+        Instruction(-1, CALL, 1),
```
P403n1x87, line 149: Should we define `NO_OFFSET = -1` and then do
```suggestion
        Instruction(NO_OFFSET, CALL, 1),
```
to make this more readable?

P403n1x87 (commented):
Great job getting 3.8 and 3.9 implemented, and fixing the bugs in the other implementations! 💪

ddtrace/internal/coverage/instrumentation_py3_10.py
romainkomorndatadog: Can I offer to throw that in as a freebie in #9609 ?

anmarchenko (approved):

ddtrace/internal/coverage/instrumentation.py
```diff
@@ -1,42 +1,16 @@
...
-            lines.add(last_lineno)
-        except AttributeError:
-            # pseudo-instruction (e.g. label)
-            pass
-
-    return abstract_code.to_code(), lines
+import sys
+
+
+# Import are noqa'd otherwise some formatters will helpfully
```
romainkomorndatadog, line 4:
```suggestion
# Import are noqa'd otherwise some formatters will helpfully remove them
```

P403n1x87 (approved):
👍 👍

ddtrace/internal/coverage/instrumentation_py3_10.py
```diff
@@ -0,0 +1,314 @@
...
+    # need to add EXTENDED_ARGs if the argument is too large.
+    process_branches = True
+    exts: t.List[t.Tuple[Instruction, int]] = []
+    while process_branches:
+        process_branches = False
+        for branch in branches:
+            jump_instr = branch.start
+            new_arg = branch.arg
+            jump_instr.arg = new_arg & 0xFF
+            new_arg >>= 8
```
erikayasuda, line 253: Dumb question, where did the 8 and 0xFF come from? Could these be consts maybe?

ddtrace/internal/coverage/instrumentation_py3_10.py
```diff
@@ -0,0 +1,314 @@
...
+            new_consts[original_offset], nested_lines = instrument_all_lines(nested_code, trap_func, trap_arg)
+            seen_lines.update(nested_lines)
+
+    new_linetable = update_location_data(code, traps, [(instr.offset, s) for instr, s in exts])
+
+    return (
+        code.replace(
+            co_code=bytes(new_code),
+            co_consts=tuple(new_consts),
+            co_stacksize=code.co_stacksize + 4,  # TODO: Compute the value!
```
erikayasuda, line 310: Same question here. Where did the 4 come from? 😅 Is this just a safe buffer?

ddtrace/internal/coverage/instrumentation_py3_12.py
```diff
@@ -0,0 +1,52 @@
...
+    def _line_event_handler(code: CodeType, line: int) -> t.Any:
+        hook, path = _CODE_HOOKS[code]
+        return hook((line, path))
+
+    # Register the line callback
+    sys.monitoring.register_callback(
+        sys.monitoring.COVERAGE_ID, sys.monitoring.events.LINE, _line_event_handler
+    )  # noqa
+
+    def instrument_all_lines(code: CodeType, hook: HookType, path: str) -> t.Tuple[CodeType, t.Set[int]]:
```
erikayasuda, line 37: Why do we keep the recursive approach just for 3.12?

erikayasuda: And the fallback (3.7)

erikayasuda (approved):
Just some NB questions

ddtrace/internal/coverage/instrumentation_py3_12.py
erikayasuda: Never mind, I realized `sys.monitoring` was added in 3.12

ddtrace/internal/coverage/instrumentation_py3_10.py
romainkomorndatadog: So the short answer is "to be safe". The long answer is "I don't actually know, @P403n1x87 did that and I'm trying to reverse-engineer my way to the answer and can't." 😓

ddtrace/internal/coverage/instrumentation_py3_10.py
erikayasuda: No the explanation makes sense. I just haven't done this stuff in a while so... This was a clarification I just personally needed 😅

ddtrace/internal/coverage/instrumentation_py3_10.py
erikayasuda: OK, no need to block on this. Just one of the magic numbers I couldn't reverse engineer either. Would be curious to know where it came from if someone knows though.

P403n1x87: The magic number 4 was chosen for the following reasons:
- when we add our instrumentation, we push additional values to the frame stack (the callable hook object plus its argument); because in theory we might be adding the instrumentation at a point where the stack is at peak usage, it is just safe to add 2 to it (this is an upper bound, but not necessarily the optimal value)
- when I was testing with 2 I was getting segmentation faults that went away by simply increasing the stack size to 4

PR #9677, opened 2024-06-28
feat(llmobs): capture more model parameters in openai chat/completion spans

Yun-Kim:
This PR captures more model/input parameters in OpenAI chat/completion spans. Currently we only capture `temperature` and `max_tokens`, but there are plenty of other [parameters](https://platform.openai.com/docs/api-reference/completions/create) that we can capture as part of the request metadata to LLM Observability.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


ddtrace/llmobs/_integrations/openai.py
```diff
@@ -161,10 +161,11 @@ def _llmobs_set_meta_tags_from_completion(
         if isinstance(prompt, str):
             prompt = [prompt]
         span.set_tag_str(INPUT_MESSAGES, json.dumps([{"content": str(p)} for p in prompt]))
-        parameters = {"temperature": kwargs.get("temperature", 0)}
-        if kwargs.get("max_tokens"):
-            parameters["max_tokens"] = kwargs.get("max_tokens")
+
+        parameters = {k: v for k, v in kwargs.items() if k not in ("model", "prompt")}
+        parameters["temperature"] = kwargs.get("temperature", 0)
```
lievan, line 166: Is this line here ensuring _temperature_ always has a default value? I think the default value of temperature for completions is `1` ((source)[https://platform.openai.com/docs/api-reference/chat/create?lang=python])

IMO we should avoid hard coding the logic that temperature defaults to 0. Similar to the other arguments, we should only set temperature if it's able to be extracted out of kwargs.

mabdinur (approved):

ddtrace/llmobs/_integrations/openai.py
```diff
@@ -187,10 +187,10 @@ def _llmobs_set_meta_tags_from_chat(
                 continue
             input_messages.append({"content": str(getattr(m, "content", "")), "role": str(getattr(m, "role", ""))})
         span.set_tag_str(INPUT_MESSAGES, json.dumps(input_messages))
-        parameters = {"temperature": kwargs.get("temperature", 0)}
-        if kwargs.get("max_tokens"):
-            parameters["max_tokens"] = kwargs.get("max_tokens")
+
+        parameters = {k: v for k, v in kwargs.items() if k not in ("model", "messages", "tools", "functions")}
```
lievan, line 191: is there a guarantee that these parameters are json serializable? If not might be worth to add a try/catch here, unless this method call is being wrapped in another try/catch

Yun-Kim: Good catch, will add a default repr if the object is not json serializable.

lievan (approved):

PR #9680, opened 2024-07-01
chore(telemetry): improve the maintainability of the telemetry writer

mabdinur:
Bundled 10 telemetry improvements and refactors into this PR. Each change is capture by a standalone commit. Merging 10 different PRs to make telemetry collection slightly better seems cumbersome and not the best use of CI and everyone's time.

This PR does not introduce any new functionality.

- dc5f75fdd5dd8d0307acf52a71e2871882fa17b5 Refactor global functions and constants in `../telemetry/writer.py` 
   - Moves `AGENT_ENDPOINT`, `AGENTLESS_ENDPOINT_V2`, `_get_endpoint_v2`, and `_agentless_telemetry_url` from a global scope to fields/methods in the `_TelemetryClient` class. Also updates tests to align with this change. This allows us to better encapsulate telemetry logic.
- 2490e1d9438ce270248f7b813bfac139deafd501, baeebfe7a03281593ba29cab7f9886d86362117e Ensures the telemetry writer is started before all products. 
    - Also ensures telemetry atexit hooks are executed last (ex: before tracer, agent writer, profiling). This will allow us to capture more start up errors and ensure app-started event is sent first and app-closed event is sent last.
- baeebfe7a03281593ba29cab7f9886d86362117e Ensures telemetry collections results in little to no performance impact when `DD_INSTRUMENTATION_TELEMETRY_ENABLED` is `False`. 
    - With this change telemetry worker thread, atexit hooks and forksafe hooks will not be enabled/started if telemetry is disabled via envar. If telemetry is not manually disabled these components will be configured when `ddtrace` is imported (and not just on ddtrace-run or when the first telemetry event/log/metric is queued).
    - This change also moves all exception hook logic from `ddtrace.internal.telemetry.__init__` to `ddtrace.internal.telemrtry.writer.py`. This provides better encapsulation and resolves some circular import errors.
- df5235087a812abec05a4cccb80b6ddf290ebadd Remove ``ddtrace.internal.telemetry.disable_and_flush(...)`. 
    - This function is not required. Telemetry will always be disabled and flushed by `TelemetryWriter.app_shutdown` (this method is registered as an atexit hook). 
- a983f081b0106a4a47f777368bd51d8fcba32d18 Remove most usages of `TelemetryWriter.enable()`. 
    - Telemetry is always enabled when ddtrace is imported or a telemetry event is queued.  There is no need to call this method in other places in the codebase or in tests.
- ba572c514fa42b226a78ff8f80f4206b71248e77 Avoid inlining imports from `ddtrace.internal.telemetry`. 
    - This was initially done as a performance optimization and to resolve circular imports. Telemetry is now started before all products, this resolves most of the issues with circular imports. Since all expensive operations are now hidden behind the `TelemetryWriter._enabled` flag importing telemetry no longer has a performance impact if `DD_INSTRUMENTATION_TELEMETRY_ENABLED=False`. This will reduce instances where disabling telemetry results in sytanx/import errors.
-  99a305dbf3b8f2111b1b4ed277fac0d313b50b4b Moves telemetry test from tracer testsuite to the telemetry testsuite.
    -  Updates tests to align with the refactor of install_exception hook and uninstall_exception hooks  
    - Renames `_excepthook` to `_telemetry_excepthook`. This will help with testing. 
- a4b2a40894c06e8348f5b9707ad317662c7dc6f5 Update test_setting_origin_mode to mock starting the telemetry writer.
   - With this PR telemetry events are only sent after app_started is queued. This event is queued when the first span is sent to the datadog agent. Since this test does not use a datadog agent telemetry is not enabled.
- b54e869084ab28418243cd56e8d460e66e725289 Decouple telemetry and LLMobs. There is no need to disable teleemtry in llmobs. This will reduce the likely hood of deadlocks. Stopping a thread in another thread can be messy.
- 1f3ac18398a9f2e6a54ea76a0bf5be75785f83de Avoids calling TelemetryWriter.stop() when TelemetryWriter._worker is None
  - When a process forks the telemetry writer attempts to stop the telemetry worker thread. If the worker thread has already been halted this will raise an Exception and crash the the telemetry writer. We should always check if telemetry is running before calling stop. 

## Checklist

## Reviewer Checklist

- [x] Title is accurate


emmettbutler (approved):

mabdinur:
@gnufede This is only commit that requires ASM approval: https://github.com/DataDog/dd-trace-py/commit/ba572c514fa42b226a78ff8f80f4206b71248e77#diff-937b920fe5e1ac6fa83809e5846f61c67cf31cb76d23f1b0fabd48e6d1f37589. In this commit I removed `if config._telemetry_enabled` from all the `ddtrace.internal.telemetry` imports. There is no need to inline imports any more. Since telemetry is started before all products (ASM included) we will not run into any circular imports.

romainkomorndatadog:
Ran a quick smoke test against https://github.com/romainkomorndatadog/gevent-test-repo and it looks like it works. 👍

romainkomorndatadog:
Side note, this PR description is hella chef's kiss. 😍

gnufede (approved):
All good from AppSec standpoint. I'm only curious about the performance hit of the telemetry import if it's not needed (when telemetry is disabled), especially in environments like Serverless.

romainkomorndatadog (approved):

ddtrace/internal/telemetry/writer.py
```diff
@@ -821,3 +828,47 @@ def _stop_service(self, join=True, *args, **kwargs):
...
+                    )
+                    error_msg = "{}:{} {}".format(filename, lineno, str(value))
+                    self.add_integration(integration_name, True, error_msg=error_msg)
+
+            if self._enabled and not self.started:
+                self._app_started_event(False)
+
+            self.app_shutdown()
+
+        return self._ORIGINAL_EXCEPTHOOK(tp, value, root_traceback)
```
romainkomorndatadog, line 866: This causes exceptions:
```
  File "/Users/romain.komorn/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/telemetry/writer.py", line 866, in _telemetry_excepthook
    return self._ORIGINAL_EXCEPTHOOK(tp, value, root_traceback)
TypeError: exceptiongroup_excepthook() takes 3 positional arguments but 4 were given
```

romainkomorndatadog: You might have wanted something like
```
_ORIGINAL_EXCEPTHOOK = staticmethod(sys.excepthook)
```
in the class, but maybe it makes more sense to just stash `_ORIGINAL_EXCEPTHOOK` at the module level instead of as a class attribute.

PR #9685, opened 2024-07-01
chore(auto-inject): add package instrumentation skipping and denylist

ZStriker19:
Currently the python tracer will abort instrumentation entirely if a package that's outside of the supported range is detected. Instead it should follow the behavior of other tracers and simply skip instrumenting that particular package, and then instrument the rest of the supported packages.

In addition, the Python tracer should skip instrumentation if uwsgi is detected since it's only compatible with uwsgi under specific conditions: https://ddtrace.readthedocs.io/en/stable/advanced_usage.html?highlight=uwsgi#uwsgi

## Checklist

## Reviewer Checklist

- [ ] Title is accurate


lib-injection/sources/sitecustomize.py
```diff
@@ -333,21 +366,21 @@ def _inject():
...
                 )
+                event = gen_telemetry_payload(telemetry_data)
                 send_telemetry(event)
             except Exception as e:
-                event = gen_telemetry_payload(
-                    [create_count_metric("library_entrypoint.error", ["error_type:" + type(e).__name__.lower()])]
+                telemetry_data.append(
+                    create_count_metric("library_entrypoint.error", [f"error_type:{type(e).__name__.lower()}"])
                 )
+                event = gen_telemetry_payload(telemetry_data)
```
erikayasuda, line 383: Probably just my lack of context/knowledge about telemetry, but doesn't this mean we keep adding more and more events to `telemetry_data` and then generate the telemetry payload with the expanding data? Should we be "resetting" the `telemetry_data` every time we send it?

lib-injection/sources/sitecustomize.py
```diff
@@ -282,17 +296,18 @@ def _inject():
                     "DD_INJECT_FORCE set to True, allowing unsupported runtimes and continuing.",
                     level="debug",
                 )
-        if telemetry_data:
+        if runtime_incomp or denylist_package:
```
erikayasuda, line 299: nit: Could we just do this instead? And remove the `denylist_package` flag?
```suggestion
        if runtime_incomp or not FORCE_INJECT:
```

PR #9700, opened 2024-07-02
ci: bump cibuildwheel to 2.19.2

gnufede:
Updates CI Build Wheel to the latest release, as our CI is blocked by this issue related to the EOL of Centos 7: 
https://github.com/pypa/cibuildwheel/issues/1915

## Checklist

- [ ] The PR description includes an overview of the change
- [ ] The PR description articulates the motivation for the change
- [ ] The change includes tests OR the PR description describes a testing strategy
- [ ] The PR description notes risks associated with the change, if any
- [ ] Newly-added code is easy to change
- [ ] The change follows the [library release note guidelines](https://ddtrace.readthedocs.io/en/stable/releasenotes.html)
- [ ] The change includes or references documentation updates if necessary

## Reviewer Checklist

- [ ] Title is accurate
- [ ] Newly-added code is easy to change
- [ ] If necessary, author has acknowledged and discussed the performance implications of this PR as reported in the benchmarks PR comment

### DIFF
--- a/.github/workflows/build_python_3.yml
+++ b/.github/workflows/build_python_3.yml
@@ -60,7 +60,7 @@ jobs:
 
       - name: Build wheels arm64
         if: matrix.os == 'arm-4core-linux'
-        run: /home/runner/.local/bin/pipx run cibuildwheel==2.19.1 --platform linux
+        run: /home/runner/.local/bin/pipx run cibuildwheel==2.16.5 --platform linux
         env:
           # configure cibuildwheel to build native archs ('auto'), and some
           # emulated ones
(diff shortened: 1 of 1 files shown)


.github/workflows/build_python_3.yml
```diff
@@ -87,6 +87,7 @@ jobs:
               mv $w {dest_dir}
             done &&
             rm -rf ./tempwheelhouse
+          CIBW_ENVIRONMENT_MACOS: "MACOSX_DEPLOYMENT_TARGET=12"
```
sanchda, line 90: Do you know whether setting this closer to setup.py would help native tools pick up minimum required dependencies?  As far as I can tell, this parameter is inferred from the environment in a lot of cases.  If we were more strict, we may not have to set it here.

... it's also possible that we've just hit the limit of what we can accomplish with these legacy OSX targets and our only option is to limit what we support.  Any clue how old "12" is versus the old target (10.9)?

gnufede: Don't mind this PR, I didn't see yours merged 🙈

gnufede:
Already fixed by #9697

.github/workflows/build_python_3.yml
sanchda: Oh.  Well, I think we'll have to do a PR like this sometime anyway, so getting runtime versions sorted out is a good thing.  :)

PR #9701, opened 2024-07-02
chore(tracer): use simple queue file for extra service names instead of multiprocessing

christophe-papazian:
To prevent further problems with multiprocessing in config:
- remove multiprocessing queue from main config file
- add a new File_Queue using low level locking to write into temporary file
- Use the new queue to exchange messages between multiple processes for extra service names
- update tests
- added a new unit-tests workflow on github to launch any tests on linux(x86), macos(arm64) and windows(x86). (For now, only the test relevant to that PR is used).

This is an alternative solution to https://github.com/DataDog/dd-trace-py/pull/9626 ensuring that we still report extra service names.

[APPSEC-53927]

## Checklist

## Reviewer Checklist

- [x] Title is accurate

[APPSEC-53927]: https://datadoghq.atlassian.net/browse/APPSEC-53927?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ddtrace/internal/_file_queue.py
```diff
@@ -0,0 +1,63 @@
...
+                f.write(data + "\x00")
+                unlock(f)
+        except Exception:  # nosec
+            pass
+
+    def get_all(self) -> typing.Set[str]:
+        """Pop all unique strings from the queue."""
+        try:
+            with unpatched_open(self.filename, "r+") as f:
+                lock(f)
```
gnufede, line 55: Do we need to lock to read? I wonder what happens if there are many services and a lot of contention here

ddtrace/internal/_file_queue.py
```diff
@@ -0,0 +1,63 @@
...
+    def __init__(self) -> None:
+        self.directory = tempfile.gettempdir()
+        self.filename = os.path.join(self.directory, "file_queue")
+
+    def put(self, data: str) -> None:
+        """Push a string to the queue."""
+        try:
+            with unpatched_open(self.filename, "a") as f:
+                lock(f)
+                f.write(data + "\x00")
```
gnufede, line 46: It would be nice to write only unique data, to avoid the file from getting too big? Alternatively, limit the size it can get, no?

gnufede (approved):

ddtrace/internal/_file_queue.py
christophe-papazian: this is the only way to not miss a service name, as the read also delete the file to clear the queue

ddtrace/internal/_file_queue.py
christophe-papazian: the main thread should delete the file every 5 seconds. I could add a failsafe that would check the file size and not write if too large, but it would cost a little time. not sure right now if it's better or worse.

emmettbutler (approved):

ddtrace/internal/_file_queue.py
christophe-papazian: Please note that the read should not happen more than every 5sec, as it's only done by the main thread when sending data through RC.

sanchda (approved):
I think the implementation has a few potential issues on Linux, but I don't know whether those issues matter here.

I can't speak to the behavior of the MSVC-controlled file locks--Windows has a reputation for having far superior IO primitives in general than Unices, so perhaps this works even better for MS than it does for Linux.

ddtrace/internal/_file_queue.py
```diff
@@ -0,0 +1,63 @@
...
+    def unlock(f):
+        msvcrt.locking(f.fileno(), msvcrt.LK_UNLCK, size(f))
+
+
+class File_Queue:
+    """A simple file-based queue implementation for multiprocess communication."""
+
+    def __init__(self) -> None:
+        self.directory = tempfile.gettempdir()
+        self.filename = os.path.join(self.directory, "file_queue")
```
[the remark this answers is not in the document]
christophe-papazian: each invocation would create a different non existing temporary directory, there is no possible collision.

ddtrace/internal/_file_queue.py
```diff
@@ -0,0 +1,63 @@
...
+            pass
+
+    def get_all(self) -> typing.Set[str]:
+        """Pop all unique strings from the queue."""
+        try:
+            with unpatched_open(self.filename, "r+") as f:
+                lock(f)
+                data = f.read()
+                os.unlink(self.filename)
+                unlock(f)
```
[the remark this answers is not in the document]
christophe-papazian: if we unlink before locking, we don't need to lock at all when reading. 
There is no perfect solution here I think, but the reader is single and won't read often, while the writers may be many and will write often. It's acceptable to have a few data losses for this feature, as it will be sent regularly to the backend.

ddtrace/internal/_file_queue.py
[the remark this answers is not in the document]
sanchda: If I understand the docs correctly, `gettempdir()` only deduces the location of the "first" canonical temporary directory.  It's supposed to replace fiddly checks against `/tmp`, `/dev/shm`, `$TEMP`, and a few other weird and random places that everybody--especially me--get wrong almost every time.

For instance, on my local Linux machine:

```
>>> import tempfile as t
>>> t.gettempdir()
'/tmp'
```

ddtrace/internal/_file_queue.py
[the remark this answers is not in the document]
christophe-papazian: While looking for a solution for windows, I improved the whole thing. It should be perfect now. We don't delete the file, just truncate it to 0 length. after locking, we ensuring to seek at the right position for the operation.

harm-nedap:
If anyone else comes across this, my Python project runs in a container with read_only filesystem and this PR broke it because `FileNotFoundError: [Errno 2] No usable temporary directory found in ['/tmp', '/var/tmp', '/usr/tmp','/app']`. I've had to set the environment variable `DD_REMOTE_CONFIGURATION_ENABLED: 'false'` to make it work again.

sanchda:
#10004 fixes behavior which was erroneously introduced here.  CC @harm-nedap; disabling the configuration manually should be unnecessary after the linked PR is merged (thank you for calling out this behavior and directing others to the fix, though!)

harm-nedap:
Thanks for linking that PR, I had not seen it yet :)!

PR #9709, opened 2024-07-03
chore(setup): early check for valid Rust toolchain

sanchda:
Adds an early check for a valid Rust toolchain.  Since Rust isn't installed everywhere we might get built, this provides a clearer and more immediate signal that system dependencies must be resolved.

Uh, if there's some fancy, new-fangled fixture for this in buildtools or whatever, I'm ignorant.  I tried to read the docs, but didn't see anything that did what I thought needed to be done here.

Here's a local failure
Screenshot 2024-07-03 at 9 28 51 AM

And you can see it dumps out faster in CI if there's a problem.
Screenshot 2024-07-03 at 9 36 02 AM

Screenshot 2024-07-03 at 9 36 07 AM

## Checklist

## Reviewer Checklist

- [x] Title is accurate


sanchda:
* **#9709** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-py/9709?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @sanchda and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

sanchda:
> Might be nice if we could optionally build without extensions (because I like incidents).

I think we need to do a few things before we can achieve that goal.  Notably, we'd need to distinguish between extensions which are critical for core functionality (unfortunately, this includes the Rust extension which is the focus of this PR) and those that represent "optional" features (like profiling and appsec).  At the same time, we'd need some controls in place to ensure that everything is built and shipped in our final wheel (we've missed this before, when certain extensions were optional).

We just need a lot more maturity around extensions before we can realistically get there.  Right now, the default posture is to fail if "almost" anything that isn't necessary for ddtrace-as-a-product-platform-plus-all-the-products fails to build.

Maybe this is something we can talk about internally?  We have more and more upcoming work dealing with moving functionality into native extensions, so it's definitely meaningful to get ahead of these issues.

romainkomorndatadog (approved):

PR #9714, opened 2024-07-04
chore(er): capture exception API

P403n1x87:
We add a new programmatic API for Exception Replay to allow users to capture caught exceptions explicitly.

## Checklist

## Reviewer Checklist

- [ ] Title is accurate
- [ ] Newly-added code is easy to change
- [ ] If necessary, author has acknowledged and discussed the performance implications of this PR as reported in the benchmarks PR comment


EWachnowezki:
just checked the API method name and looks good to me :)

EWachnowezki (approved):

ddtrace/debugging/_exception/auto_instrument.py
```diff
@@ -210,3 +212,16 @@ def on_span_finish(self, span: Span) -> None:
...
+
+
+def capture_exception() -> None:
+    import sys
+
+    from ddtrace import tracer
+
+    try:
+        span = tracer.current_span()
+        if span is not None:
```
shatzi, line 224: how this works when
1. a user call capture_exception() multiple times? will the first one/last one win? do we capture all snapshots multiple times?
2. how this works if the exception also propagate to the span it self? is that same as call "capture_exception" twice?

ddtrace/debugging/_exception/auto_instrument.py
```diff
@@ -210,3 +212,16 @@ def on_span_finish(self, span: Span) -> None:
...
             span.set_tag_str(EXCEPTION_ID_TAG, str(exc_id))
+
+
+def capture_exception() -> None:
+    import sys
+
+    from ddtrace import tracer
+
+    try:
+        span = tracer.current_span()
```
shatzi, line 223: if trace._current_span() is not the service entry, will this still work? 

I heard there is a tag ET support to index that span error if it not the service entry. maybe this should also set that tag. However, I can't seem to find it on their docs. so maybe this should add tags to service entry span (or both entry and current) or simple abort for non-entry?

shatzi (commented):
This is awesome. I think there are two things missing:
1. we need documentation both on change log here and on exception replay docs to expose this feature
2. figure out if `capture_exception()` works on non service entry spans with Error Tracking. if that not the case - make sure we document this as known limitation or have a follow up PR to find a solution

ddtrace/debugging/_exception/auto_instrument.py
P403n1x87: > if trace._current_span() is not the service entry, will this still work?

This will put debug info on the current span, regardless of what the span is. So any limitations are probably not due to "us" I'd expect

ddtrace/debugging/_exception/auto_instrument.py
P403n1x87: Multiple calls would cause the data to be emitted multiple times. We could probably guard against this by checking if the span has `error` already set, and skip.

> how this works if the exception also propagate to the span

I'm not quite sure I understand this. With this call we are setting an error on the span. However, we don't expect this to be used when the exceptions are not handled. This will only work in an exception handler, as anywhere else in the code there won't be an exception to attach to a span. If the exception is re-raised within an exception block however, then the tracer will trigger the capturing of debug data a second time.

ddtrace/debugging/_exception/auto_instrument.py
shatzi: yup. this is broken as non entry spans are not get indexed and there is no way for the user to query on that and it would not appear in Error tracking. so not sure why we allow the user to do it in the first place.

ddtrace/debugging/_exception/auto_instrument.py
shatzi: You got it. Therefore, I think this logic is little broken and should emit debug info only the last/first error that was assigned to the span.

P403n1x87:
Closing in favour of #10430.

PR #9715, opened 2024-07-04
chore(tracer): remove attr and cattr from remoteconfig

christophe-papazian:
Remove dependencies to `attrs` and `cattrs` in remoteconfig.

- Use dataclasses
- use post_init to simulate validators and parsers

After that PR, `cattrs` is **no longer used** in ddtrace and removed from install dependencies

## Checklist

## Reviewer Checklist

- [x] Title is accurate

### DIFF
--- a/ddtrace/internal/remoteconfig/client.py
+++ b/ddtrace/internal/remoteconfig/client.py
@@ -130,7 +130,8 @@ class Root:
     version: int
 
     def __post_init__(self):
-        assert self._type == "root"
+        if self._type != "root":
+            raise ValueError("Root: invalid root type")
         if isinstance(self.expires, str):
             self.expires = parse_isoformat(self.expires)
         for k, v in self.keys.items():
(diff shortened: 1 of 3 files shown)


juanjux (approved):

gnufede (approved):

pyproject.toml
```diff
@@ -29,7 +29,6 @@ dependencies = [
     "bytecode>=0.15.0; python_version>='3.12'",
     "bytecode>=0.14.0; python_version~='3.11.0'",
     "bytecode>=0.13.0; python_version<'3.11'",
-    "cattrs",
```
erikayasuda, line 32: Generally speaking, what triggered this change? Did we recently run into issues with customer code, or has this just always been an issue so we're fixing it now?

ddtrace/internal/remoteconfig/client.py
```diff
@@ -85,100 +84,148 @@ class RemoteConfigError(Exception):
...
+        if self._type != "root":
+            raise ValueError("Root: invalid root type")
+        if isinstance(self.expires, str):
+            self.expires = parse_isoformat(self.expires)
+        for k, v in self.keys.items():
+            if isinstance(v, dict):
+                self.keys[k] = Key(**v)
+        for k, v in self.roles.items():
+            if isinstance(v, dict):
+                self.roles[k] = Role(**v)
```
erikayasuda, line 142: nit: In the `__post_init__()`s, are we OK with leaving the values that don't conform to the specified type as is?

erikayasuda (approved):

ddtrace/internal/remoteconfig/client.py
christophe-papazian: We have definitely less dynamic checks than with attr, but those classes are only used in that file, so we can be sure it will be under control. Here failures to conform can only come from an unexpected RC payload that will be rejected.

PR #9743, opened 2024-07-08
chore(internal): update cache max size to for module/package/file

villainb-dg:
## Summary

Current implementation uses the default size of the `cached` decorator which is 256. This is often (always?) lower than the number of files of a Python program loaded in memory creating cache churn. This is problematic for functions like `filename_to_package` since it performs an expensive operation by calling `_root_module` that walks the filesystem.

Simple benchmark show a 7800% improvement on cache acces for `filename_to_package` function:

Current implementation:

```python
def fill_cache(i=10000):
    for j in range(i):
        filename_to_package(f"/home/benjamin/workspace/dd-trace-py/lib/python3.10/site-packages/ddtrace/internal/test{j}.py")

timeit.timeit("fill_cache(257)", setup="from __main__ import fill_cache", number=10000)
```

```
115.0306228880072
```

With patch:

```python
def fill_cache(i=10000):
    for j in range(i):
        filename_to_package(f"/home/benjamin/workspace/dd-trace-py/lib/python3.10/site-packages/ddtrace/internal/test{j}.py")

timeit.timeit("fill_cache(257)", setup="from __main__ import fill_cache", number=10000)
```

```
1.4706893929978833
```

## Risks

The main risk of this change would be an increase of memory usage from the agent. At most we are adding 7 x 64kB of memory (448kB) as opposed to 1.75kB today. This remains minimal so shouldn't be a problem.

## Checklist

## Reviewer Checklist

- [x] Title is accurate

### DIFF
--- a/ddtrace/internal/packages.py
+++ b/ddtrace/internal/packages.py
@@ -2,13 +2,15 @@ import logging
 import os
 import sys
 import sysconfig
-import typing as t
 from types import ModuleType
+import typing as t
 
 from ddtrace.internal.compat import Path
 from ddtrace.internal.module import origin
-from ddtrace.internal.utils.cache import cached, callonce
+from ddtrace.internal.utils.cache import cached
+from ddtrace.internal.utils.cache import callonce
 from ddtrace.settings.third_party import config as tp_config
+
 
 LOG = logging.getLogger(__name__)
 
--- a/ddtrace/internal/packages.py
+++ b/ddtrace/internal/packages.py
@@ -4,7 +4,6 @@ import sys
 import sysconfig
 import typing as t
 from types import ModuleType
-from functools import lru_cache as cached
 
 from ddtrace.internal.compat import Path
 from ddtrace.internal.module import origin
--- a/ddtrace/internal/packages.py
+++ b/ddtrace/internal/packages.py
@@ -77,7 +77,7 @@ def get_distributions():
     return pkgs
 
 
-@cached(maxsize=65536)
+@cached(maxsize=256)
 def get_version_for_package(name):
     # type: (str) -> str
     """returns the version of a package"""
@@ -194,7 +194,7 @@ def _third_party_packages() -> set:
     ) - tp_config.excludes
 
 
-@cached(maxsize=65536)
+@cached(maxsize=16384)
 def filename_to_package(filename: t.Union[str, Path]) -> t.Optional[Distribution]:
     mapping = _package_for_root_module_mapping()
     if mapping is None:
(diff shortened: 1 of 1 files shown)


P403n1x87:
@villainb-dg thanks for the performance analysis! The cache size is in terms of entries, so actual memory usage might be order of 10/100 times higher, depending on whether some of the data loaded by the expensive operations would be released or not. Whilst increasing the cache size is definitely something that we should do here, we should also try to be as conservative as possible. Do you reckon you can re-run your analysis with smaller cache sizes? 🙏

ddtrace/internal/packages.py
```diff
@@ -194,7 +194,7 @@ def _third_party_packages() -> set:
     ) - tp_config.excludes
 
 
-@cached()
+@cached(maxsize=65536)
```
P403n1x87, line 197: Because many file paths will map to the same package, we would expect this to be the larger cache needed. So as a rule of thumb, I'd say we can make the cache of functions that don't take a path as argument much smaller. And if we can also lower the size of path-based caches by a bit it would be great too, as paths could be quite long, and storing them when not really needed would cost some memory.

villainb-dg:
> @villainb-dg thanks for the performance analysis! The cache size is in terms of entries, so actual memory usage might be order of 10/100 times higher, depending on whether some of the data loaded by the expensive operations would be released or not. Whilst increasing the cache size is definitely something that we should do here, we should also try to be as conservative as possible. Do you reckon you can re-run your analysis with smaller cache sizes? 🙏

Ah yes good point. I can re-run the analysis with smaller cache sizes. :+1:

ddtrace/internal/packages.py
villainb-dg: Yes this is really a trade-off between CPU utilization and memory here. I can definitely lower the size of the various caches although the ones storing smaller objects use less memory so it might be okay for them to be bigger :)

P403n1x87:
> Do you know if we perhaps have metrics on the number of file names we typically see?

Indeed that's a challenge. AFAIK we don't have this kind of data, but perhaps @mabdinur who has worked on telemetry and packages might know better than me. One way we could estimate this could be to grab a few `requirements.txt` files from some popular projects and do a similar analysis as above on them.

villainb-dg:
What do you propose we do with a `requirements.txt` file? I'm not sure how to go from that to the number of files loaded by a program?

P403n1x87:
> What do you propose we do with a `requirements.txt` file? I'm not sure how to go from that to the number of files loaded by a program?

We could create virtual environments and `pip install -r requirements.txt` in them, then call `filename_to_package` for any files that are in the venv site-packages folder (recursively)

brettlangdon:
This change LGTM, are there more tests/benchmarks/analysis we want to be doing here before merging the change?

villainb-dg:
Taking a relatively big internal repository, I can count more than 4.5k files in the cache. I think 16k is a reasonable size and strikes the right balance for now in terms of max memory usage vs. CPU.

brettlangdon (approved):

villainb-dg:
Some tasks are failing and I'm not sure why. Can someone help me understand if this is related to my change?

PR #9767, opened 2024-07-09
chore(telemetry): queue app-started on first sent payload

mabdinur:
## Background

The instrumentation telemetry writer is enabled when `ddtrace` is imported, this is done so all products can begin queuing telemetry metrics, logs, and configurations immediately. However the `TelemetryWriter.app_started` is called when the first trace is sent to the agent (when the AgentWriter is started), this delay allows us to capture start up errors and configurations.

## Description

- Ensuring `telemetry_writer.app_started()` is called when the first CIVisibility, Profiling, and LLMOBS payloads are sent to the agent. Previously `app-started` (and `app-closing`) events were only sent when the first trace was submitted. With this change `app-started` and `app-closed` events are sent when tracing is disabled or products are run with a datadog agent. 

- Renames `TelemetryWriter._app_started_event` to `TelemetryWriter.app_started`. The `_app_started_event` method is not internal to the telemetry writer. This name change should hopefully make this method easier to understand.

## Checklist

## Reviewer Checklist

- [x] Title is accurate


ddtrace/profiling/exporter/http.py
```diff
@@ -234,6 +235,8 @@ def export(
         return profile, libs
 
     def _upload(self, client, path, body, headers):
+        # Enable telemetry before the first Trace is sent
```
mabdinur, line 238:
```suggestion
        # Enable telemetry before the first profile is sent
```

emmettbutler (approved):

sanchda (changes requested):
Looks good, very logical.  I made one proposal, but otherwise LGTM

ddtrace/internal/writer/writer.py
```diff
@@ -321,6 +320,7 @@ def _send_payload(self, payload, count, client):
         return response
 
     def write(self, spans=None):
+        telemetry_writer.app_started()
```
mabdinur, line 323:
```suggestion
        # Queues an app-started event before the first civisibility/llmobs/trace payload is sent
        telemetry_writer.app_started()
```

mabdinur:
> Looks good, very logical.  I made one proposal, but otherwise LGTM

Awesome😄. I addressed that concern. Thanks for calling that out.

ddtrace/profiling/scheduler.py
```diff
@@ -66,6 +67,8 @@ def flush(self):
                 self.before_flush()
             except Exception:
                 LOG.error("Scheduler before_flush hook failed", exc_info=True)
+        # Enable telemetry before the first profile is sent
+        telemetry_writer.app_started()
```
sanchda, line 71: IMHO, I recommend moving this to the first line of `flush()`, since that way it'll guarantee to run.  In some configurations (autoinject/SSI !!!), the current line may not execute because some flush configurations can't/don't/won't check for recorders.

sanchda (approved):
Left one comment, which would improve this for SSI/autoinjected configurations, but otherwise looks great!  Thank you!!!

sanchda (approved):
tyvm

PR #9773, opened 2024-07-09
fix(langchain): tag chat model content appropriately with structured output

sabrenner:
[MLOB-1063]
## What does this PR do?

When using a `ChatAnthropic` or `ChatOpenAI` chat model with `with_structured_output` ([ref](https://python.langchain.com/v0.1/docs/modules/model_io/chat/structured_output/)), the output content tag for that span would be empty. This PR adds some parsing logic and fallback handling to tag it properly.

**Before**
Screenshot 2024-07-10 at 1 04 52 PM

**After**
Screenshot 2024-07-10 at 1 05 26 PM

## Notes
1. Most LOC due to 2 new cassettes, snapshot files, and requirements. I had to specify a specific version of `anthropic` for the `langchain-anthropic` package to be compatible.
2. I'll have another PR up for the relevant LLMObs fix + test once this one is merged.

## Checklist

## Reviewer Checklist

- [x] Title is accurate

[MLOB-1063]: https://datadoghq.atlassian.net/browse/MLOB-1063?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ddtrace/contrib/langchain/patch.py
```diff
@@ -421,16 +429,19 @@ def traced_chat_model_generate(langchain, pin, func, instance, args, kwargs):
...
-                log_chat_completions = []
-            else:
-                log_chat_completions = [
-                    [
-                        {"content": message.text, "message_type": message.message.__class__.__name__}
-                        for message in messages
-                    ]
-                    for messages in chat_completions.generations
-                ]
+            log_chat_completions = []
```
Yun-Kim, line 432: Logs are unofficially deprecated, I don't think we need to make corresponding changes to the logs code as we will be removing them soon.

ddtrace/llmobs/_integrations/langchain.py
```diff
@@ -250,6 +250,39 @@ def _metrics_tags(cls, span: Span) -> List[str]:
...
+
+            message = chat_completion.message
+            content = message.content
+
+            tool_calls = message.tool_calls
+
+            # for anthropic, langchain puts it on the content
+            if content:
+                if isinstance(content, str):
+                    return content
```
Yun-Kim, line 273: What happens if we have multiple content blocks in the `message.content` field? Looks like anthropic has the tool call is also exposed in `tool_calls`, but the content blocks in a `message.content` can contain an actual message (`content[0]["text"]`) as well as tool call arguments (i.e. `content[1]["input"]`) 
[ref](https://python.langchain.com/v0.2/docs/integrations/chat/anthropic/#content-blocks)

releasenotes/notes/langchain-structured-output-content-c84ec40c4a087d88.yaml
```diff
@@ -0,0 +1,4 @@
+---
+fixes:
+  - |
+    langchain: Fix chat model output not being tagged appropriately when using `with_structured_output`.
```
Yun-Kim, line 4:
```suggestion
    langchain: Fixes chat model output not being tagged appropriately when using `with_structured_output`.
```

Yun-Kim (commented):
I think this PR can actually be split up into two, as it's trying to do a couple things here:
- Support tagging tool calls from responses (AFAIK we do not support this yet in the langchain integration)
- Do our best to capture the completion content without always looking at `completion.text`. This might include tool calls, but I think the bigger priority is the structured output.
Can we move out the support for tagging tool calls into a separate PR and only handle structured output in this PR?

sabrenner:
> Support tagging tool calls from responses (AFAIK we do not support this yet in the langchain integration)

I don't think we're really doing this here. LangChain is just using tool calls under the hood to facilitate expanding LLM output parsing, and that's what we're picking up on in this PR to tag the chat completion, we're not officially supporting the tracing of that tool itself. It's just that we want to tag the actual content of what the output of the chat model is. I do agree we should support actually tracing/tagging the parsing tool call (what was the input/output, etc.), but I don't think this PR touches that.

Lmk if I'm not understanding this, happy to chat offline!

ddtrace/llmobs/_integrations/langchain.py
sabrenner: good point! let me try this out.

sabrenner:
closing in favor of #9893 and an additional follow-up PR to trace output parsers

PR #9776, opened 2024-07-10
chore(asm): improve accept header management for threats

christophe-papazian:
- use the standard `application/json` instead of `text/json` for json response body content type
- consider the quality factors of the accept headers in auto mode to chose between html and json response.
- factorize the logic
- update relevant tests and add more test cases for quality factors on custom blocking

This will enable the python tracer to pass new system tests concerning quality factors in Accept header.
https://github.com/DataDog/system-tests/pull/2718

APPSEC-54007

## Checklist

## Reviewer Checklist

- [x] Title is accurate


juanjux (approved):

ddtrace/_trace/trace_handlers.py
```diff
@@ -153,6 +154,43 @@ def _maybe_start_http_response_span(ctx: core.ExecutionContext) -> None:
...
+    json_score = 0.0
+    ctypes = ctype.split(",")
+    for ct in ctypes:
+        if len(ct) > 128:
+            # ignore long (and probably malicious) headers to avoid performances issues
+            continue
+        m = re.match(r"([^/;]+/[^/;]+)(?:;q=([01](?:\.\d*)?))?", ct.strip())
+        if m:
+            if m.group(1) == "text/html":
+                html_score = max(html_score, min(1.0, float(1.0 if m.group(2) is None else m.group(2))))
```
erikayasuda, line 175: What does the value of `m.group(2)` actually represent?

ddtrace/_trace/trace_handlers.py
```diff
@@ -153,6 +154,43 @@ def _maybe_start_http_response_span(ctx: core.ExecutionContext) -> None:
...
+            elif m.group(1) == "text/*":
+                html_score = max(html_score, min(1.0, float(0.2 if m.group(2) is None else m.group(2))))
+            elif m.group(1) == "application/json":
+                json_score = max(json_score, min(1.0, float(1.0 if m.group(2) is None else m.group(2))))
+            elif m.group(1) == "application/*":
+                json_score = max(json_score, min(1.0, float(0.2 if m.group(2) is None else m.group(2))))
+    return html_score > json_score
```
erikayasuda, line 182: nit: In the case of a tie (not sure how likely/unlikely this is), we're defaulting to using `application/json`?

tests/appsec/contrib_appsec/utils.py
```diff
@@ -802,11 +811,30 @@ def test_request_suspicious_request_block_match_request_body(
...
+        get_tag,
+        asm_enabled,
+        metastruct,
+        root_span,
+        query,
+        status,
+        rule_id,
+        action,
+        headers,
+        use_html,
```
erikayasuda, line 837: nit: Should we be passing this in hardcoded? Looks like we could also validate the behavior of `_use_html()` and/or `_ctype_from_headers()` here, but if we already have enough coverage for that then feel free to ignore this.

erikayasuda (approved):

ddtrace/_trace/trace_handlers.py
christophe-papazian: group(0) is the entire match, group(1) is the first capturing subgroup `([^/;]+/[^/;]+)` that will be the mime type and group(2) is the second capturing subgroup `([01](?:\.\d*)?))` which should be a float between 0 and 1. As the second group is optional (inside a (...)?), it can be None even if a match happens.

ddtrace/_trace/trace_handlers.py
christophe-papazian: Yes, that was already the case before if there was nothing in the accept header, we were defaulting to json. 
Now, it's the same, in case of a tie.

tests/appsec/contrib_appsec/utils.py
christophe-papazian: we are covering the usage of `_ctype_from_headers` here with the new `Accept` headers value that could only lead to the expected answer with the appropriate computation from `_ctype_from_headers`
There will also be tested on system tests

PR #9777, opened 2024-07-10
refactor(debugger): remove attr usage

P403n1x87:
We refactor the debugger code to remove any usages of the attr package in favour of the built-in dataclasses library.

## Checklist

## Reviewer Checklist

- [x] Title is accurate

### DIFF
--- a/tests/appsec/iast_packages/inside_env_runner.py
+++ b/tests/appsec/iast_packages/inside_env_runner.py
@@ -61,12 +61,13 @@ def try_patched(module_name, expect_no_change=False):
 if __name__ == "__main__":
     mode = sys.argv[1]
     import_module = sys.argv[2]
-    expect_no_change = sys.argv[3]
+    if len(sys.argv) >= 4:
+        expect_no_change = sys.argv[3]
 
     if mode == "unpatched":
         sys.exit(try_unpatched(import_module))
     elif mode == "patched":
         sys.exit(try_patched(import_module, expect_no_change=expect_no_change == "True"))
 
-    print("Use: [python from pyenv] inside_env_runner.py patched|unpatched module_name")
+    print("Use: [python from pyenv] inside_env_runner.py patched|unpatched module_name True|False")
     sys.exit(1)


ddtrace/debugging/_probe/model.py
```diff
@@ -79,68 +79,77 @@ def update(self, other: "Probe") -> None:
         if self.version == other.version:
             return
 
-        for attrib in (_.name for _ in self.__attrs_attrs__ if _.eq):
+        for attrib in (f.name for f in self.__dataclass_fields__.values() if f.compare):
```
gnufede, line 82: I think `dataclasses.fields(self)` would be equivalent to `self.__dataclass_fields__.values()`

ddtrace/debugging/_probe/model.py
```diff
@@ -79,68 +80,77 @@ def update(self, other: "Probe") -> None:
...
+@dataclass
 class LineLocationMixin(ProbeLocationMixin):
-    source_file = attr.ib(type=Path, converter=_resolve_source_file, eq=False)  # type: ignore[misc]
-    line = attr.ib(type=int, eq=False)
+    source_file: str = field(compare=False)
+    line: int = field(compare=False)
+
+    def __post_init__(self):
+        super().__post_init__()
+        self.resolved_source_file = _resolve_source_file(self.source_file)
```
shatzi, line 150: can we add this logic into source_file init ? 
maybe we also like to keep the original value (aka, when we report errors or when _resolve_source_file fails)

ddtrace/debugging/_probe/model.py
```diff
@@ -79,68 +80,77 @@ def update(self, other: "Probe") -> None:
...
 
     def location(self):
-        return (str(self.source_file) if self.source_file is not None else None, self.line)
+        return (str(self.resolved_source_file) if self.source_file is not None else None, self.line)
```
shatzi, line 153: if self.source_file or this need to check resolved_source_file instead?

shatzi (approved):
awesome stuff

ddtrace/debugging/_probe/model.py
P403n1x87: unfortunately `dataclasses` don't have the ability of using a custom converter so we have to use an extra field here

ddtrace/debugging/_probe/model.py
P403n1x87: I think we can probably remove the `if` now 🤔  as `self.source_file` should never be `None`

PR #9788, opened 2024-07-11
fix(wsgi): don't report stopiteration errors

tabgok:
There are several integrations which send StopIteration errors as part of normal operation (say, when we hit the last element of an interable). We are currently flagging such spans as errors, when its expected behavior.

After this change, we still generate the spans, but they are no longer reported as errors.

Some salient points:

* "Errors" on spans are directly tied to UI/Datadog app behaviors and are desired to be actionable by, or at least significant to, the customer.
* The wsgi method in this issue is expected to raise StopIteration. Whether or not it is caught, raising an exception is normal behavior for this piece of code.
* The StopIteration raised by wsgi may be caught (customer doesn't care and it should not be an error) or not caught (customer does care and it should be an error). We don't know at the time of this span close if the issue is caught or not, so we can't make a 100% informed decision.
* Given those points, the decision is to only ignore StopIteration in wsgi explicitly for the method which is expected to raise StopIteration. While the exception may not be caught elsewhere, we are relying on other mechanisms (such as crash detection) to report the issue.

Related issue: https://github.com/miguelgrinberg/flask-sock/issues/64#issuecomment-1664709789

## Checklist

## Reviewer Checklist

- [x] Title is accurate


releasenotes/notes/ignore_stopiteration_exception-43806201b7fc8005.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    tracing: Ensures the status of a Span is not set to error when a``StopIteration`` exception is raised
```
brettlangdon, line 4:
```suggestion
    tracing: Ensures the status of a Span is not set to error when a ``StopIteration`` exception is raised
```

romainkomorndatadog (changes requested):
Requesting changes because this seems incorrect (or maybe just too broad) to me. My apologies in advance if I'm misunderstanding what this change is doing.

I don't know that I agree with the premise that "`StopIteration` is not an error". For example: I can do something in code that raises an uncaught `StopIteration` error that will cause the tracer to crash:
```
% ddtrace-run python
  File "/Users/romain.komorn/.pyenv/versions/bototest/bin/ddtrace-run", line 5, in <module>
    from ddtrace.commands.ddtrace_run import main
  File "/Users/romain.komorn/go/src/github.com/DataDog/dd-trace-py/ddtrace/__init__.py", line 11, in <module>
    import ddtrace.internal._unpatched  # noqa
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "/Users/romain.komorn/go/src/github.com/DataDog/dd-trace-py/ddtrace/internal/module.py", line 236, in _exec_module
    next(iter([]))
StopIteration
```

Another example is that I could make the same mistake in a Flask route, which would cause it to not execute/render properly. That would be an error, and the associated span should reflect that, and have that exception added.

Wouldn't this change prevent that?

romainkomorndatadog:
> Should we only set `StopIteraton` as an ignored exception for specific spans within the WSGI/web framework integrations ?

I think this would be a more correct approach.

tabgok:
> Wouldn't this change prevent that?

Counterpoint: 

```
try:
    next(iter([]))
except Exception as e:
   pass
```

is reported as an error with our current logic

tabgok:
Romain, Brett, and I met to discuss.

Some salient points:

1. "Errors" on spans are directly tied to UI/Datadog app behaviors and are desired to be actionable by, or at least significant to, the customer.
2. The `wsgi` method in this issue is _expected_ to raise `StopIteration`.  Whether or not it is caught, raising an exception is normal behavior for this piece of code.
3. The `StopIteration` raised by `wsgi` may be caught (customer doesn't care and it should not be an error) or not caught (customer does care and it should be an error).  We don't know at the time of this span close if the issue is caught or not, so we can't make a 100% informed decision.

Given those points, the decision is to only ignore `StopIteration` in `wsgi` explicitly for the method which is expected to raise `StopIteration`.  While the exception may not be caught elsewhere, we are relying on other mechanisms (such as crash detection) to report the issue.

tests/contrib/wsgi/test_wsgi.py
```diff
@@ -30,11 +30,20 @@ def chunked_response_generator_error(start_response):
             raise GeneratorExit()
 
 
+def raise_stop_iteration_exception(start_response):
+    status = "200 OK"
+    headers = [("Content-type", "text/plain")]
+    start_response(status, headers)
+    next(iter([]))
```
brettlangdon, line 37: is this the actual use case?

I would think it is `return iter([])`, no?

ddtrace/contrib/wsgi/wsgi.py
```diff
@@ -151,6 +163,12 @@ def blocked_view():
             result = core.dispatch_with_results(
                 "wsgi.request.complete", (ctx, closing_iterable, self.app_is_iterator)
             ).traced_iterable
+
+            if stop_iteration_exception and result.value:
+                # Close the request and app spans
+                result.value._finish_spans()
+                core.dispatch("wsgi.app.success", (ctx, closing_iterable))
```
brettlangdon, line 170:
```suggestion
            if stop_iteration_exception:
                if result.value:
                    # Close the request and app spans
                    result.value._finish_spans()
                    core.dispatch("wsgi.app.success", (ctx, closing_iterable))
```


Are there cases when we have a stop iteration exception but not a result value?


Should we always be raising the stop iteration exception even if we don't have a result value?

tabgok: Whoops! Your code is more correct, the intent was to be defensive and I messed it up - accepting your fix

tests/contrib/wsgi/test_wsgi.py
tabgok: Updated with a comment

releasenotes/notes/ignore_stopiteration_exception-43806201b7fc8005.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    wsgi: Ensures the status of wsgi Spans are not set to error when a``StopIteration`` exception is raised
```
brettlangdon, line 4:
```suggestion
    wsgi: Ensures the status of wsgi Spans are not set to error when a ``StopIteration`` exception is raised
```

brettlangdon (approved):

romainkomorndatadog (approved):
Thanks for the updated logic.

PR #9806, opened 2024-07-12
chore(internal): use `lru_cache` instead of custom cache implementation in packages module

villainb-dg:
## Summary

The custom cache implementation evicts half of its stored elements whenever it overflows. This impacts the performance of the filename to package lookup since half of the files need to be re-looked up even if they are frequently accessed. This PR proposes to instead use the `lru_cache` implementation available from the standard library.

## Motivation

Random access test shows that the custom cache implementation performs way worst for expensive functions.

Current implementation:

```python
def fill_cache():
   for j in range(1000000):
        filename_to_package(f"/home/benjamin/workspace/dd-trace-py/lib/python3.10/site-packages/ddtrace/internal/test{randint(0,257)}.py")
```

```
Timings:
0:00:04.913800
0:00:04.441695
0:00:04.586526
```

With `lru_cache`:

```
Timings:
0:00:01.193794
0:00:00.915702
0:00:00.894196
```

## Checklist

## Reviewer Checklist

- [x] Title is accurate


ddtrace/internal/packages.py
```diff
@@ -1,3 +1,4 @@
+from functools import lru_cache as cached
```
villainb-dg, line 1: Note that the default size of `lru_cache` is 128 so this change depends on #9743 otherwise the creating of various caches needs to set the `maxsize` argument

P403n1x87:
🙌  Thanks! I'd be curious to see how these number changes if we cythonised our caching implementation. That's something we never got around to doing.

brettlangdon (approved):

PR #9809, opened 2024-07-12
chore(internal): avoid re-wrapping get_code() in calls of module exec

romainkomorndatadog:
This avoids scenarios where `reload()` deepens the number of calls to `get_code()`, leading to `RecursionError` in cases where a module is reloaded many times.

The internal `get_code()` method is only bound if the module loader's `get_code()` method does not have the `_dd_get_code` attribute. 

There is no release note because this isn't fixing an already-released, non-RC version. 

## Checklist

## Reviewer Checklist
- Title is accurate


P403n1x87 (approved):

gnufede (approved):

romainkomorndatadog:
This looked like infinite recursion but I couldn't reproduce it locally, because what was actually happening was that a version of the `moto` package was reloading `moto.s3.urls` thousands of times in test scenarios.

juanjux (approved):

ddtrace/internal/module.py
```diff
@@ -241,6 +242,8 @@ def get_code(_loader, fullname):
 
                 return code
 
+            get_code.__setattr__("_dd_get_code", True)
```
brettlangdon, line 245:
```suggestion
            setattr(get_code, "_dd_get_code", True)
```

is there a difference? we use `getattr` above 🤷🏻

tests/internal/test_module.py
```diff
@@ -502,3 +502,14 @@ def test_module_watchdog_importlib_resources_files():
...
+
+
+@pytest.mark.subprocess
+def test_module_watchdog_does_not_rewrap():
+    """Ensures that self.loader.get_code() does not get re-wrapped (leading to a deeper stack)"""
+    from importlib import reload
+
+    from ddtrace import internal
+
+    for _ in range(10000):
```
brettlangdon, line 514: the idea here is to force a stack depth limit reached exception?

are there any other ways we can make sure we aren't double wrapping? like check that the loader's get_code isn't a wrapper of a wrapper?

romainkomorndatadog: This actually mimics the issue we ran into, and I'm paranoid about `reload()` causing us to run into real-world problem, so it's kind of a twofer for me. It's a bit more integration-testy then unit-testy, I guess.

ddtrace/internal/module.py
```diff
@@ -231,7 +231,8 @@ def _exec_module(self, module: ModuleType) -> None:
         pre_exec_hook = None
 
         _get_code = getattr(self.loader, "get_code", None)
-        if _get_code is not None:
+        # DEV: avoid recursively wrapping the loader's get_code method (eg: in case of nested imporlib.reload() calls)
```
erikayasuda, line 234: nit:
```suggestion
        # DEV: avoid recursively wrapping the loader's get_code method (eg: in case of nested importlib.reload() calls)
```

tests/internal/test_module.py
erikayasuda, line 514: How long does this test take? I think it's valuable to have the repro condition as a test, but just curious what the duration is for this loop

erikayasuda (approved):

tests/internal/test_module.py
romainkomorndatadog: In its new iteration:
```
0.54s call     tests/internal/test_module.py::test_module_watchdog_reloads_dont_cause_errors
```
Prior to that it was on the order of 3s or so.

romainkomorndatadog: @brettlangdon , added a separate test specifically to check we weren't changing the wrapper.

brettlangdon (approved):

PR #9818, opened 2024-07-12
feat(llmobs): improved support for LLM tool calls

hollyyzhuang:
What does this PR do?

- Previously, LLM tool calling responses (provided by OpenAI/Anthropic) are stored in the `output.messages.%.content` field and sent to EVP. However, the LLM tool calling responses are formatted in string, which is not so readable when displaying `output.messages.%.content`. 

- Example of bad display:
Screenshot 2024-07-17 at 10 18 21 AM

- To support better UI for tool call messages, a new subfield is added in the message data structure in the span EVP intake to support storing tool call information rather than relying on the same `output.messages.%.content` field. 
- To reflect this backend change, the code for ddtrace llmobs is updated: 1). parse the tool call information from LLMs and store in a single JSON dictionary, which was previously a string 2). instead of saving this info in `content`, it will be saved in a new field `tool_calls`, which will then be sent to EVP. 

- The new field `tool_calls` stores `tool_calls_info`, which is a list of JSON dictionary (`tool_call_info`) because sometimes the LLM tool calling responses can contain more than 1 tool calls. The structure of `tool_call_info` is standardized across providers: name (string), argument (dictionary), tool_id (string), and type (string). 

- Screenshot 2024-07-17 at 10 29 05 AM

## Checklist

## Reviewer Checklist
- Title is accurate


brettlangdon (changes requested):
Can you update this PR to remove the `.DS_Store` files?

.DS_Store
Yun-Kim, line 1: Let's remove this (local cache file), thanks

ddtrace/.DS_Store
Yun-Kim, line 1: Same here

ddtrace/llmobs/.DS_Store
Yun-Kim, line 1: Same

ddtrace/llmobs/_integrations/anthropic.py
```diff
@@ -150,16 +150,20 @@ def _extract_output_message(self, response):
 
         elif isinstance(content, list):
             for completion in content:
+                tool_calls_info = []
```
Yun-Kim, line 153: is there a need to instantiate the list here if we're only going to handle it in the tool call block? (i.e. line 158)

ddtrace/llmobs/_integrations/anthropic.py
```diff
@@ -150,16 +150,20 @@ def _extract_output_message(self, response):
...
-                        output_messages.append(
-                            {"content": "[tool: {}]\n\n{}".format(name, json.dumps(inputs)), "role": role}
-                        )
+                        tool_call_info = {
+                            "name": _get_attr(completion, "name", ""),
+                            "arguments": _get_attr(completion, "input", ""),
+                            "tool_id": _get_attr(completion, "id", ""),
+                            "type": _get_attr(completion, "type", ""),
+                        }
+                        tool_calls_info.append(tool_call_info)
```
Yun-Kim, line 165: Why are we appending tool call info to a list if we're just directly adding the tool call info to the output messages?

ddtrace/llmobs/_integrations/openai.py
```diff
@@ -199,20 +199,24 @@ def _llmobs_set_meta_tags_from_chat(
...
-                    )
-            output_messages.append({"content": str(content).strip(), "role": choice.message.role})
+                    tool_call_info = {
+                        "name": getattr(tool_call.function, "name", ""),
+                        "arguments": json.loads(getattr(tool_call.function, "arguments", {})),
+                        "tool_id": getattr(tool_call, "id", ""),
+                        "type": getattr(tool_call, "type", ""),
+                    }
+                    tool_calls_info.append(tool_call_info)
+            output_messages.append({"content": "", "role": choice.message.role, "tool_calls": tool_calls_info})
```
Yun-Kim, line 219: In case an openai response contains both content and tool calls, can we ensure that we capture both in the same message? Right now it looks like we set content to be an empty string.

Yun-Kim (commented):
Could we add tests for anthropic and openai to make sure we are sending the expected payloads to LLMObs?

ddtrace/llmobs/_integrations/openai.py
hollyyzhuang: yeah sorry about that - i meant to set it to just content which will default to an empty string if it's empty and still store content if both content and tool calls are present

tests/contrib/anthropic/test_anthropic_llmobs.py
```diff
@@ -15,6 +15,14 @@
 the necessary information to make the API call.\n\nNo other tools are needed to answer this question. \
 We can proceed with calling the get_weather tool with the provided location.\n</thinking>'
 WEATHER_OUTPUT_MESSAGE_2 = '[tool: get_weather]\n\n{"location": "San Francisco, CA"}'
+WEATHER_OUTPUT_MESSAGE_2_Modified = [
```
Yun-Kim, line 18:
```suggestion
WEATHER_OUTPUT_MESSAGE_2 = [
```
Do we still need the old weather output message 2?

ddtrace/llmobs/_integrations/anthropic.py
```diff
@@ -143,24 +143,28 @@ def _extract_input_message(self, messages, system_prompt=None):
...
-            return [{"content": content, "role": role}]
+            return [{"content": content, "role": role, "tool_calls": []}]
 
         elif isinstance(content, list):
             for completion in content:
+                tool_calls_info = []
                 text = _get_attr(completion, "text", None)
                 if isinstance(text, str):
-                    output_messages.append({"content": text, "role": role})
+                    output_messages.append({"content": text, "role": role, "tool_calls": tool_calls_info})
```
Yun-Kim, line 157: Do we need to specify `"tool_calls"` as a field if there is no tool calls to report?

tests/contrib/anthropic/test_anthropic_llmobs.py
hollyyzhuang: just testing for now - will delete later

ddtrace/llmobs/_integrations/anthropic.py
hollyyzhuang: for the anthropic case, we can get rid of this. Do we want to update openai to also not return "tool_calls" as a field if there is no tool calls to report? Right now the logic for openai is returning an empty list of "tool_calls" if only content is present.

ddtrace/llmobs/_integrations/anthropic.py
Yun-Kim: Following up on offline discussion - anthropic currently supports one tool use dictionary per content block.

releasenotes/notes/change-tool-call-output-openai-anthropic-f59f6ef816589cf8.yaml
```diff
@@ -0,0 +1,7 @@
+---
+fixes:
```
Yun-Kim, line 2:
```suggestion
features:
```
Since this is a new feature, not a fix

releasenotes/notes/change-tool-call-output-openai-anthropic-f59f6ef816589cf8.yaml
```diff
@@ -0,0 +1,7 @@
...
+  - |
+    tool_calls: This fix resolves an issue where the previous tool call response format 
+    caused bad UI display. The response format has been updated better formatting of
+    tool calls responses from LLM in JSON dictionary format stored in tool_calls for
+    both providers (Anthropic and OpenAI).
```
Yun-Kim, line 7:
```suggestion
    LLM Observability: This introduces improved support for capturing tool call responses from the OpenAI and Anthropic integrations.
```

tests/contrib/anthropic/test_anthropic_llmobs.py
```diff
@@ -346,7 +361,7 @@ def test_tools_sync(self, anthropic, ddtrace_global_config, mock_llmobs_writer,
                         "content": WEATHER_OUTPUT_MESSAGE_1,
                         "role": "assistant",
                     },
-                    {"content": WEATHER_OUTPUT_MESSAGE_2, "role": "assistant"},
+                    {"content": WEATHER_INPUT_MESSAGE_2, "role": "assistant"},
```
Yun-Kim, line 364: Why are we still expecting the old placeholder message value in the input? Are we not expecting the new tool call format in the input as well?

tests/contrib/anthropic/test_anthropic_llmobs.py
```diff
@@ -438,7 +457,7 @@ async def test_tools_async(self, anthropic, ddtrace_global_config, mock_llmobs_w
...
```
Yun-Kim, line 460: Same here

tests/contrib/anthropic/test_anthropic_llmobs.py
```diff
@@ -494,7 +513,18 @@ def test_tools_sync_stream(self, anthropic, ddtrace_global_config, mock_llmobs_w
...
                     {"content": message[0]["text"], "role": "assistant"},
-                    {"content": message[1]["text"], "role": "assistant"},
+                    {
+                        "content": "",
+                        "role": "assistant",
+                        "tool_calls": [
+                            {
+                                "name": "get_weather",
+                                "arguments": {"location": "San Francisco, CA"},
+                                "tool_id": "",
```
Yun-Kim, line 523: No tool ID here?

tests/contrib/anthropic/test_anthropic_llmobs.py
```diff
@@ -588,7 +618,18 @@ async def test_tools_async_stream_helper(
...
-                    {"content": WEATHER_OUTPUT_MESSAGE_2, "role": "assistant"},
+                    {
+                        "content": "",
+                        "role": "assistant",
+                        "tool_calls": [
+                            {
+                                "name": "get_weather",
+                                "arguments": {"location": "San Francisco, CA"},
+                                "tool_id": "",
+                                "type": "tool_use",
```
Yun-Kim, line 629: Same

ddtrace/llmobs/_integrations/anthropic.py
```diff
@@ -117,11 +117,19 @@ def _extract_input_message(self, messages, system_prompt=None):
...
+                        if isinstance(input_data, str):
+                            input_data = json.loads(input_data)
+                        tool_call_info = {
+                            "name": _get_attr(block, "name", ""),
+                            "arguments": input_data,
+                            "tool_id": _get_attr(block, "id", ""),
+                            "type": _get_attr(block, "type", ""),
+                        }
+                        if text is None:
+                            text = ""
```
Yun-Kim, line 131:
```suggestion
```

ddtrace/llmobs/_integrations/anthropic.py
```diff
@@ -117,11 +117,19 @@ def _extract_input_message(self, messages, system_prompt=None):
                         input_messages.append({"content": "([IMAGE DETECTED])", "role": role})
 
                     elif _get_attr(block, "type", None) == "tool_use":
-                        name = _get_attr(block, "name", "")
-                        inputs = _get_attr(block, "input", "")
-                        input_messages.append(
-                            {"content": "[tool: {}]\n\n{}".format(name, json.dumps(inputs)), "role": role}
-                        )
+                        text = _get_attr(block, "text", None)
```
Yun-Kim, line 120:
```suggestion
                        text = _get_attr(block, "text", "")
```

ddtrace/llmobs/_integrations/anthropic.py
```diff
@@ -156,11 +164,18 @@ def _extract_output_message(self, response):
...
+                        if isinstance(input_data, str):
+                            input_data = json.loads(input_data)
+                        tool_call_info = {
+                            "name": _get_attr(completion, "name", ""),
+                            "arguments": input_data,
+                            "tool_id": _get_attr(completion, "id", ""),
+                            "type": _get_attr(completion, "type", ""),
+                        }
+                        if text is None:
+                            text = ""
```
Yun-Kim, line 177:
```suggestion
```
Can we change line 162 to default to an empty string, not None, and remove this check?

tests/contrib/anthropic/test_anthropic_llmobs.py
Yun-Kim: Offline discussion - looks like we are not correctly extracting tool iD from streamed tool call responses. Will address this in a separate PR

tests/contrib/anthropic/test_anthropic_llmobs.py
```diff
@@ -14,7 +14,14 @@
 parameter. The user has provided the location of "San Francisco, CA" in their question, so we have \
 the necessary information to make the API call.\n\nNo other tools are needed to answer this question. \
 We can proceed with calling the get_weather tool with the provided location.\n</thinking>'
-WEATHER_OUTPUT_MESSAGE_2 = '[tool: get_weather]\n\n{"location": "San Francisco, CA"}'
+WEATHER_OUTPUT_MESSAGE_2 = [
```
Yun-Kim, line 17: Could we rename this to `WEATHER_OUTPUT_TOOL_CALL` instead to be more accurate? Thanks!

Yun-Kim (commented):
Very small comments but otherwise LGTM, ready to approve after the comments are addressed!

ddtrace/llmobs/_integrations/anthropic.py
hollyyzhuang: I think i saw a case where text is literally just None ("text": None) from openai output. But let me double check!

Yun-Kim: Ahh in that case, feel free to revert, happy either way.

Yun-Kim (approved):
Great work @hollyyzhuang! 🥳

brettlangdon (approved):
thanks for removing the extra files, and the release note lgtm

PR #9830, opened 2024-07-15
fix: telemetry exception hooks

mabdinur:
## Description

When wrapping pre-configured exception hooks the telemetry writer converts exception hooks from type `function` to type `method `. This adds the `self` argument to pre-configured exception which raised the following error: `TypeError: xxxsome_funcxxxx() takes 3 positional arguments but 4 were given`.

To resolve this issue this PR casts `TelemetryWriter._ORIGINAL_EXCEPTHOOK` is cast to a static method. This ensures the telemetry writer does not modify the argument of preconfigured hooks.

## Impact 

When an application raises an unhandled exception and ddtrace instrumentation telemetry is enabled:
 - the ddtrace telemetry exception hook will successfully run and provide crash reporting to the Datadog Agent
 - the ddtrace telemtry exception hook will then attempt to run other configured exception hooks, if they exist (hooks registered by user's application or other third party libraries)
     - the ddtrace telemetry writer will fail to run these exception hooks
     - applications will log `TypeError: xxxsome_funcxxxx() takes 3 positional arguments but 4 were given`

## Motivation 

Resolves the following errors: https://github.com/DataDog/dd-trace-py/actions/runs/9891512830/job/27322136218#step:5:141

This bug was introduced in: https://github.com/DataDog/dd-trace-py/commit/0cdc0831a62ee2575833298c49ff468cb14d51fb

## Checklist

## Reviewer Checklist
- Title is accurate


brettlangdon:
I am not sure I understand why the change is needed (or works). We grab a reference to `sys.excepthook` and then call it with the expected parameters.

are we installing the telemetry writer multiple times and overwriting `sys.excepthook` multiple times?

mabdinur:
> I am not sure I understand why the change is needed (or works). We grab a reference to `sys.excepthook` and then call it with the expected parameters.
> 
> are we installing the telemetry writer multiple times and overwriting `sys.excepthook` multiple times?

I am also struggling to write a regression test. This behavior is difficult to reproduce: https://github.com/DataDog/dd-trace-py/actions/runs/9891512830/job/27322136218#step:5:141. `sys.excepthook` should only be set once and this occurs when the telemetry module is loaded. Maybe the `telemetry/__init__.py` module is being reloaded and this is causing the failures?

@romainkomorndatadog do you know the root cause?

tests/telemetry/test_telemetry.py
```diff
@@ -217,7 +217,6 @@ def process_trace(self, trace):
     ]["message"]
 
 
-@pytest.mark.skip(reason="We don't have a way to capture unhandled errors in bootstrap before telemetry is loaded")
```
mabdinur, line 220: Not sure why this test was skipped. It is currently passing in ci

brettlangdon (commented):
I wonder if we should take a different approach, rather than repalcing `sys.excepthook` with an instance method, should we just have a single global excepthook?

```python
ORIGINAL_EXCEPTHOOK = sys.excepthook

def excepthook(type, value, traceback):
    core.dispatch("sys.excepthook", (type, value, traceback))


sys.excepthook = excepthook
```

there is also a `sys.audit` event for `sys.excepthook`, so we could use that instead maybe?

romainkomorndatadog:
The problem is that stashing `sys.excepthook` as an instance attribute and then invoking it with `self._ORIGINAL_EXCEPTHOOK` will lead to it being invoked like an instance method, with `self` as the first argument, which isn't expected and leads to
```
TypeError: exceptiongroup_excepthook() takes 3 positional arguments but 4 were given
```
It is actually not being called with the expected parameters.

romainkomorndatadog:
As far as testing this, we should only need to make sure exceptions get sent, right, so wouldn't a test that does the following be sufficient?
```
import ddtrace
raise Exception
```
Then we should be able to see that an exception was sent via telemetry?

mabdinur:
> As far as testing this, we should only need to make sure exceptions get sent, right, so wouldn't a test that does the following be sufficient?
> 
> ```
> import ddtrace
> raise Exception
> ```
> 
> Then we should be able to see that an exception was sent via telemetry?

In this commit I reverted the staticmethod fix and added a new regression test: https://github.com/DataDog/dd-trace-py/pull/9830/commits/810a695fa7532701551117b2e5f7c00bf3370e84. The exception hook is called and it successfully sends an app-started event. To reproduce this error we may need a more complex scenario.

romainkomorndatadog:
I don't really understand how the code an fail in one case and not another. Are you sure that code path's being exercised when the test is passing?

mabdinur:
I added regression test and updated the PR description.

romainkomorndatadog (approved):

tests/telemetry/test_telemetry.py
```diff
@@ -221,8 +221,7 @@ def process_trace(self, trace):
...
```
erikayasuda, line 224: So this is now false because we moved the import right? https://github.com/DataDog/dd-trace-py/blob/5db374c1a63aab3be408b53568e4297461a03f86/ddtrace/__init__.py#L20

erikayasuda (approved):

tests/telemetry/test_telemetry.py
mabdinur: Yup. Before we only enabled telemetry on ddtrace-run (or when the first event was queued)

erikayasuda:
Should be unblocked after this merges: https://github.com/DataDog/system-tests/pull/2781/files

tests/telemetry/test_telemetry.py
```diff
@@ -243,19 +242,37 @@ def test_app_started_error_unhandled_exception(test_agent_session, run_python_co
...
+raise Exception('bad_code')
+"""
     )
+    assert b"pre_ddtrace_exc_hook called" in out
     assert status == 1, stderr
     assert b"bad_code" in stderr
     # Regression test for python3.12 support
     assert b"RuntimeError: can't create new thread at interpreter shutdown" not in stderr
+    # Regression test for invalid number of arguments in wrapped exception hook
+    assert b"TypeError: pre_ddtrace_exc_hook() takes 3 positional arguments but 4 were given" not in stderr
```
brettlangdon, line 267:
```suggestion
    assert b"3 positional arguments but 4 were given" not in stderr
```

non-blocking nit, just makes it a bit more robust in case the function name or something else changes.

brettlangdon (approved):
non-blocking nit, but lgtm

PR #9850, opened 2024-07-16
feat(llmobs): submit langchain embedding spans

yahya-mouman:
## Changes

This PR adds instrumentation to submit langchain embedding spans to LLM Observability. The embedding spans sent to LLM Observability will contain the following I/O data:

input.documents: when part of a workflow captures the embeddings as a json list. when it is not a workflow but a single embedding operation captures each embedding input as a Document dictionary containing a text field
output.value: instead of unnecessarily storing very long vector values, adds a placeholder text [X embeddings returned with size Y] 

Ticket : https://datadoghq.atlassian.net/browse/MLOB-937

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/contrib/langchain/patch.py
```diff
@@ -603,6 +604,14 @@ def traced_embedding(langchain, pin, func, instance, args, kwargs):
         integration.metric(span, "incr", "request.error", 1)
         raise
     finally:
+        if integration.is_pc_sampled_llmobs(span):
+            integration.llmobs_set_tags(
+                "embedding",
+                span,
+                input_texts=input_texts,
+                output_embedding=func(*args, **kwargs),
```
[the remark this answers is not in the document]
Yun-Kim: In line 596, we define `embeddings` as the result of the wrapped function. However we should define `embeddings = None` before we try calling the wrapped function, because otherwise it may not exist (in case of error) like we do with `completions` ([ex](https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/contrib/langchain/patch.py#L211))

ddtrace/llmobs/_integrations/langchain.py
```diff
@@ -10,7 +10,7 @@
 from ddtrace.constants import ERROR_TYPE
 from ddtrace.internal.logger import get_logger
 from ddtrace.llmobs import LLMObs
-from ddtrace.llmobs._constants import INPUT_MESSAGES
+from ddtrace.llmobs._constants import INPUT_MESSAGES, INPUT_DOCUMENTS
```
Yun-Kim, line 13: Our styling used in the python tracer involves single line imports, ordered alphabetically
```suggestion
from ddtrace.llmobs._constants import INPUT_DOCUMENTS
from ddtrace.llmobs._constants import INPUT_MESSAGES  
```

ddtrace/llmobs/_integrations/langchain.py
```diff
@@ -42,21 +42,30 @@
...
+            span: Span,
+            inputs: Any,
+            response: Any = None,
+            error: bool = False,
     ) -> None:
         """Sets meta tags and metrics for span events to be sent to LLMObs."""
         if not self.llmobs_enabled:
             return
+        if operation not in SUPPORTED_OPERATIONS:
+            log.warning("Unsupported operation : %s" % operation)
```
Yun-Kim, line 67: Thanks for adding the unsupported case check! If the operation is not supported, can we return immediately instead?

ddtrace/llmobs/_integrations/langchain.py
```diff
@@ -194,6 +205,40 @@ def _llmobs_set_meta_tags_from_chain(
...
+            is_workflow: bool = False,
+    ) -> None:
+        span.set_tag_str(SPAN_KIND, "workflow" if is_workflow else "embedding")
+        span.set_tag_str(MODEL_NAME, span.get_tag(MODEL) or "")
+        span.set_tag_str(MODEL_PROVIDER, span.get_tag(PROVIDER) or "")
+
+        input_tag_key = INPUT_VALUE if is_workflow else INPUT_DOCUMENTS
+        output_tag_key = OUTPUT_VALUE
+
+        if input_texts is not None:
```
Yun-Kim, line 223: is `input_texts` guaranteed to be a list of strings? let's try to be more defensive (we've been bitten before by assuming argument types) and ensure that we work with a list of strings

ddtrace/llmobs/_integrations/langchain.py
```diff
@@ -194,6 +205,40 @@ def _llmobs_set_meta_tags_from_chain(
...
+            try:
+                formatted_inputs = self.format_io(input_texts)
+                if isinstance(input_texts, str):
+                    span.set_tag_str(input_tag_key, formatted_inputs)
+                else:
+                    span.set_tag_str(input_tag_key, json.dumps(self.format_io(input_texts)))
```
Yun-Kim, line 229: We need to do slightly different formatting of the input documents/value depending on if the span is a workflow or embedding type span (i.e. embedding spans --> input_documents containing Document() objects, workflow spans --> input_value containing json serialized input texts). I think this is currently formatting all input_value/documents to json text

Yun-Kim (commented):
Implementation looks great, just a couple questions 😄 can we add tests in `test_langchain_llmobs.py` to `class TestLLMObsLangchainCommunity` and `class TestLangchainTraceStructureWithLlmIntegrations`?

ddtrace/contrib/langchain/patch.py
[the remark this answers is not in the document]
yahya-mouman: Gotcha !
I tried to get it directly but the issue was that it needed to be initialized before. I was trying to be as less intrusive to the tracing code as possible 😅 But if I can just initialize embeddings before that would be simpler.

ddtrace/llmobs/_integrations/langchain.py
yahya-mouman: Yes sure ! I was thinking that we still want to see the spans with general tagging if it was an operation that wasn't dealt with from our side yet.

yahya-mouman:
> Implementation looks great, just a couple questions 😄 can we add tests in `test_langchain_llmobs.py` to `class TestLLMObsLangchainCommunity` and `class TestLangchainTraceStructureWithLlmIntegrations`?

I updated and addressed all the previous comments. The only thing left now is writing the test case but before doing that i wanted to confirm the implementation details especially the parsing difference between the workflow vs non-workflow cases 🙏

ddtrace/llmobs/_integrations/langchain.py
Yun-Kim: This should technically never get run, but still great error handling. If it's an operation we don't support but somehow end up in `llmobs_set_tags(...)`, we may actually result in errors trying to format/extract data that isn't there outside of the chat/chain/llm/embedding scenario, so I think we should just return instead.

ddtrace/llmobs/_integrations/langchain.py
```diff
@@ -194,6 +207,50 @@ def _llmobs_set_meta_tags_from_chain(
...
+
+        input_tag_key = INPUT_VALUE if is_workflow else INPUT_DOCUMENTS
+        output_tag_key = OUTPUT_VALUE
+
+        if (input_texts is not None
+                and (isinstance(input_texts, str) or all(isinstance(text, str) for text in input_texts))):
+            if is_workflow:
+                try:
+                    formatted_inputs = self.format_io(input_texts)
+                    if isinstance(input_texts, str):
```
Yun-Kim, line 230:
```suggestion
                    if isinstance(formatted_inputs, str):
```

ddtrace/llmobs/_integrations/langchain.py
```diff
@@ -194,6 +207,50 @@ def _llmobs_set_meta_tags_from_chain(
...
+                        span.set_tag_str(input_tag_key, json.dumps(self.format_io(input_texts)))
+                except TypeError:
+                    log.warning("Failed to serialize embedding input data to JSON")
+            else:
+                if isinstance(input_texts, str):
+                    span.set_tag_str(input_tag_key, input_texts)
+                elif (isinstance(input_texts, list)
+                      and all(isinstance(text, str) for text in input_texts) and not is_workflow):
+                    json_documents_list = json.dumps(input_texts)
+                    span.set_tag_str(input_tag_key, '{"documents":%s}' % json_documents_list)
```
Yun-Kim, line 242: Two suggestions here:
1. Let's avoid double nesting conditionals if possible 
```suggestion
            elif isinstance(input_texts, str):
                    span.set_tag_str(input_tag_key, input_texts)
            elif (isinstance(input_texts, list)
                      and all(isinstance(text, str) for text in input_texts) and not is_workflow):
                    json_documents_list = json.dumps(input_texts)
                    span.set_tag_str(input_tag_key, '{"documents":%s}' % json_documents_list)
```
2. Can we construct the `Document` objects similar to how we do it in the OpenAI integration? That might be a simpler/more readable approach (might be worth to add a try/catch similar to how you did in line 234)

Yun-Kim (commented):
Looks mostly great, I added a comment/suggestion on the input documents formatting.
Just needs tests, and release note similar to this [PR](https://github.com/DataDog/dd-trace-py/pull/9663).

tests/contrib/langchain/test_langchain_llmobs.py
```diff
@@ -122,6 +122,20 @@ def _invoke_chain(cls, chain, prompt, mock_tracer, cassette_name, batch=False):
         LLMObs.disable()
         return mock_tracer.pop_traces()[0]
 
+    def _embed_query(cls, embedding_model, query, mock_tracer, cassette_name):
+        LLMObs.enable(ml_app=cls.ml_app, integrations_enabled=False, _tracer=mock_tracer)
+        with get_request_vcr(subdirectory_name=cls.cassette_subdirectory_name).use_cassette(cassette_name):
+            embedding_model.embed_query(query)
```
yahya-mouman, line 127: @Yun-Kim Not sure if there's a different method call for langchain v0. I couldn't find a difference in the public documentation though.

tests/contrib/langchain/test_langchain_llmobs.py
```diff
@@ -337,6 +337,54 @@ def test_llmobs_chain_schema_io(self, langchain, mock_llmobs_span_writer, mock_t
         )
         _assert_expected_llmobs_llm_span(trace[1], mock_llmobs_span_writer, mock_io=True)
 
+    @pytest.mark.skipif(sys.version_info < (3, 10, 0), reason="Requires unnecessary cassette file for Python 3.9")
```
yahya-mouman, line 339: @Yun-Kim I noticed that there's a cassette file ending by 39 in the cassettes directory for langchain. hopefully this is the correct annotation to add.

ddtrace/contrib/langchain/patch.py
```diff
@@ -590,6 +593,7 @@ def traced_embedding(langchain, pin, func, instance, args, kwargs):
                 for idx, text in enumerate(input_texts):
                     span.set_tag_str("langchain.request.inputs.%d.text" % idx, integration.trunc(text))
             span.set_metric("langchain.request.input_count", len(input_texts))
+            span.set_tag_str("langchain.request.inputs.chunk_size", chunk_size)
```
Yun-Kim, line 596: For simplicity, we don't actually need to enhance the APM integration. If you'd like to, I would suggest moving the APM integration change into a separate PR.

ddtrace/contrib/langchain/patch.py
```diff
@@ -567,6 +567,7 @@ def traced_embedding(langchain, pin, func, instance, args, kwargs):
     """
     try:
         input_texts = get_argument_value(args, kwargs, 0, "texts")
+        chunk_size = get_argument_value(args, kwargs, 1, "chunk_Size")
```
Yun-Kim, line 570: Just curious, which langchain embedding models is `chunk_size` used as an argument for? I see that it is present for `OpenAIEmbeddings.embed_documents()` but if it is not present for all embedding models signatures I would be careful adding this.

ddtrace/llmobs/_integrations/langchain.py
```diff
@@ -194,6 +203,50 @@ def _llmobs_set_meta_tags_from_chain(
...
+                ):
+                    formatted_inputs = self.format_io(input_texts)
+                    formatted_str = (
+                        formatted_inputs
+                        if isinstance(formatted_inputs, str)
+                        else json.dumps(self.format_io(input_texts))
+                    )
+                    span.set_tag_str(input_tag_key, formatted_str)
+            elif isinstance(input_texts, str):
+                span.set_tag_str(INPUT_DOCUMENTS, json.dumps(Document(text=str(input_texts))))
```
Yun-Kim, line 234:
```suggestion
                span.set_tag_str(INPUT_DOCUMENTS, json.dumps([Document(text=str(input_texts))]))
```
I believe input documents should be always a list of documents

releasenotes/notes/feat-llmobs-submit-langchain-embedding-spans-89c8704ef41cfee3.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    LLM Observability: The Langchain integration now submits embedding spans to LLM Observability.
```
Yun-Kim, line 4:
```suggestion
    LLM Observability: The Langchain integration now submits embedding spans to LLM Observability.

```

tests/contrib/langchain/test_langchain_llmobs.py
Yun-Kim: I believe the methods are the same in both langchain v0 and v1, `embed_query()` and `embed_documents()`. We can keep this as is here.

tests/contrib/langchain/test_langchain_llmobs.py
Yun-Kim: Yes for some reason Python 3.9 requires different cassettes, so we can safely skip this test in that case.

tests/contrib/langchain/test_langchain_llmobs.py
```diff
@@ -323,6 +336,54 @@ def test_llmobs_chain_schema_io(self, langchain, mock_llmobs_span_writer, mock_t
...
+            cassette_name="openai_embedding_query.yaml",
+        )
+        assert mock_llmobs_span_writer.enqueue.call_count == 1
+        mock_llmobs_span_writer.enqueue.assert_called_with(
+            _expected_llmobs_llm_span_event(
+                span,
+                span_kind="embedding",
+                model_name=embedding_model.model,
+                model_provider="openai",
+                input_documents="hello world",
```
Yun-Kim, line 355: `input_documents` should be a list of dictionaries containing the `text` field, hopefully [this](https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/llmobs/utils.py#L47) may give you more context (we only use text field for embeddings documents, ignore the other fields)

tests/contrib/langchain/test_langchain_llmobs.py
```diff
@@ -323,6 +336,54 @@ def test_llmobs_chain_schema_io(self, langchain, mock_llmobs_span_writer, mock_t
...
+            cassette_name="openai_embedding_document.yaml",
+        )
+        assert mock_llmobs_span_writer.enqueue.call_count == 1
+        mock_llmobs_span_writer.enqueue.assert_called_with(
+            _expected_llmobs_llm_span_event(
+                span,
+                span_kind="embedding",
+                model_name=embedding_model.model,
+                model_provider="openai",
+                input_documents=["hello world", "goodbye world"],
```
Yun-Kim, line 379: Same comment as the above!

tests/contrib/langchain/test_langchain_llmobs.py
```diff
@@ -323,6 +336,54 @@ def test_llmobs_chain_schema_io(self, langchain, mock_llmobs_span_writer, mock_t
...
+                output_value="[2 embedding(s) returned with size 1536]",
+                token_metrics={"input_tokens": 4, "output_tokens": 0, "total_tokens": 4},
+                tags={"ml_app": "<ml-app-name>", "langchain.request.inputs.chunk_size": 4},
+                integration="langchain",
+            )
+        )
+
 
 @pytest.mark.skipif(PATCH_LANGCHAIN_V0, reason="These tests are for langchain >= 0.1.0")
 class TestLLMObsLangchainCommunity(BaseTestLLMObsLangchain):
```
Yun-Kim, line 389: Can we add the same test cases for langchain community? Thanks!

Yun-Kim (commented):
Just a small question about the `chunk_size` argument and testing suggestions, but otherwise looks great for final review once those are addressed!

ddtrace/contrib/langchain/patch.py
yahya-mouman: I believe it's only the openAIembeddings but i thought that was the only type we supported anyways 🤔

tests/contrib/langchain/test_langchain_llmobs.py
```diff
@@ -517,6 +578,52 @@ def test_llmobs_anthropic_chat_model(self, langchain_anthropic, mock_llmobs_span
...
+                model_provider="openai",
+                input_documents=[{"text": "hello world"}],
+                output_value="[1 embedding(s) returned with size 1536]",
+                tags={"ml_app": "langchain_test"},
+                integration="langchain",
+            )
+        )
+
+    @pytest.mark.skipif(sys.version_info < (3, 10, 0), reason="Requires unnecessary cassette file for Python 3.9")
+    def test_llmobs_embedding_documents(self, langchain_core, langchain_openai, mock_llmobs_span_writer, mock_tracer):
```
Yun-Kim, line 605: Unfortunately the cassette generated for the OpenAI embed_documents() call is over 200k lines of code long. For the sake of avoiding maintaining such a large file, can we use different providers for this embed_documents() test? For example, [FakeEmbeddings](https://python.langchain.com/v0.1/docs/integrations/text_embedding/fake/).

You might need to add this skip case:
```python
try:
    from langchain_community.embeddings import FakeEmbeddings
except (ModuleNotFoundError, ImportError):
    self.skipTest("langchain-community not installed which is required for this test.")
```

tests/contrib/langchain/test_langchain_llmobs.py
```diff
@@ -517,6 +578,52 @@ def test_llmobs_anthropic_chat_model(self, langchain_anthropic, mock_llmobs_span
...
+        embedding_model = langchain_openai.OpenAIEmbeddings()
+        trace = self._embed_documents(
+            embedding_model=embedding_model,
+            documents=["hello world", "goodbye world"],
+            mock_tracer=mock_tracer,
+            cassette_name="openai_embedding_document.yaml",  # cassette doesn't exist on langchain directory
```
Yun-Kim, line 611: If you are using FakeEmbeddings as a test case, you will most likely not need a cassette as that is local

tests/contrib/langchain/test_langchain_llmobs.py
```diff
@@ -323,6 +336,54 @@ def test_llmobs_chain_schema_io(self, langchain, mock_llmobs_span_writer, mock_t
...
+                input_documents=[{"text": "hello world"}],
+                output_value="[1 embedding(s) returned with size 1536]",
+                token_metrics={"input_tokens": 2, "output_tokens": 0, "total_tokens": 2},
+                tags={"ml_app": "langchain_test"},
+                integration="langchain",
+            )
+        )
+
+    @pytest.mark.skipif(sys.version_info < (3, 10, 0), reason="Requires unnecessary cassette file for Python 3.9")
+    def test_llmobs_embedding_documents(self, langchain, mock_llmobs_span_writer, mock_tracer):
```
Yun-Kim, line 364: Same here for the embedding model provider, let's use `FakeEmbeddings` instead to minimize maintaining such a large file

Yun-Kim (commented):
Just one test suggestion, but looks great to me!

tests/contrib/langchain/test_langchain_llmobs.py
yahya-mouman: Done 👍

tests/contrib/langchain/test_langchain_llmobs.py
```diff
@@ -517,6 +584,60 @@ def test_llmobs_anthropic_chat_model(self, langchain_anthropic, mock_llmobs_span
...
+    @pytest.mark.skipif(sys.version_info < (3, 10, 0), reason="Requires unnecessary cassette file for Python 3.9")
+    def test_llmobs_embedding_query(self, langchain_community, langchain_openai, mock_llmobs_span_writer, mock_tracer):
+        if langchain_community is None:
+            pytest.skip("langchain-community not installed which is required for this test.")
+        embedding_model = langchain_community.embeddings.FakeEmbeddings(size=1536)
+        trace = self._embed_query(
+            embedding_model=embedding_model,
+            query="hello world",
+            mock_tracer=mock_tracer,
+            cassette_name=None,  # FakeEmbeddings does not need a cassette
```
Yun-Kim, line 596: Since we already have a `openai_embedding_query.yaml` cassette, can we use OpenAIEmbeddings for this test?

tests/contrib/langchain/test_langchain_llmobs.py
```diff
@@ -517,6 +585,60 @@ def test_llmobs_anthropic_chat_model(self, langchain_anthropic, mock_llmobs_span
         assert mock_llmobs_span_writer.enqueue.call_count == 1
         _assert_expected_llmobs_llm_span(span, mock_llmobs_span_writer, input_role="user")
 
+    @pytest.mark.skipif(sys.version_info < (3, 10, 0), reason="Requires unnecessary cassette file for Python 3.9")
+    def test_llmobs_embedding_query(self, langchain_community, mock_llmobs_span_writer, mock_tracer):
+        if langchain_community is None:
+            pytest.skip("langchain_community not installed which is required for this test.")
+        embedding_model = langchain_community.embeddings.OpenAIEmbeddings()
+        trace = self._embed_query(
```
Yun-Kim, line 593:
```suggestion
        with mock.patch("langchain_openai.OpenAIEmbeddings._get_len_safe_embeddings", return_value=[0.0] * 1536):
            trace = self._embed_query(
```
There is a langchain internal method that relies on an external dependency that we need to mock.

tests/contrib/langchain/test_langchain_llmobs.py
```diff
@@ -517,6 +585,60 @@ def test_llmobs_anthropic_chat_model(self, langchain_anthropic, mock_llmobs_span
         assert mock_llmobs_span_writer.enqueue.call_count == 1
         _assert_expected_llmobs_llm_span(span, mock_llmobs_span_writer, input_role="user")
 
+    @pytest.mark.skipif(sys.version_info < (3, 10, 0), reason="Requires unnecessary cassette file for Python 3.9")
+    def test_llmobs_embedding_query(self, langchain_community, mock_llmobs_span_writer, mock_tracer):
+        if langchain_community is None:
+            pytest.skip("langchain_community not installed which is required for this test.")
+        embedding_model = langchain_community.embeddings.OpenAIEmbeddings()
```
Yun-Kim, line 592:
```suggestion
    def test_llmobs_embedding_query(self, langchain_openai, mock_llmobs_span_writer, mock_tracer):
        if langchain_openai is None:
            pytest.skip("langchain_openai not installed which is required for this test.")
        embedding_model = langchain_openai.OpenAIEmbeddings()
```
Can we use the `langchain_openai` package instead? This is more up-to-date than `langchain_community`.

Yun-Kim (commented):
Suggestions for failing tests

tests/contrib/langchain/test_langchain_llmobs.py
yahya-mouman: Question : How would I be able to know about this myself ? 😅 Wouldn't the test surface the dependency ?

tests/contrib/langchain/test_langchain_llmobs.py
```diff
@@ -598,6 +726,12 @@ def _call_openai_llm(OpenAI):
         with get_request_vcr(subdirectory_name="langchain_community").use_cassette("openai_completion_sync.yaml"):
             llm.invoke("Can you explain what Descartes meant by 'I think, therefore I am'?")
 
+    @staticmethod
+    def _call_openai_embedding(OpenAIEmbeddings):
+        embedding = OpenAIEmbeddings()
+        with get_request_vcr(subdirectory_name="langchain_community").use_cassette("openai_embedding_query.yaml"):
```
Yun-Kim, line 732:
```suggestion
        with mock.patch("langchain_openai.OpenAIEmbeddings._get_len_safe_embeddings", return_value=[0.0] * 1536):
            with get_request_vcr(subdirectory_name="langchain_community").use_cassette("openai_embedding_query.yaml"):
```
Since we're using OpenAIEmbeddings we need to mock out that internal tokenizer method that I mentioned earlier in the other tests as well 😄

tests/contrib/langchain/test_langchain_llmobs.py
```diff
@@ -323,6 +343,56 @@ def test_llmobs_chain_schema_io(self, langchain, mock_llmobs_span_writer, mock_t
...
+                output_value="[1 embedding(s) returned with size 1536]",
+                tags={"ml_app": "langchain_test"},
+                integration="langchain",
+            )
+        )
+
+    @pytest.mark.skipif(sys.version_info < (3, 10, 0), reason="Requires unnecessary cassette file for Python 3.9")
+    def test_llmobs_embedding_documents_langchain_base(self, langchain, mock_llmobs_span_writer, mock_tracer):
+        embedding_model = langchain.embeddings.OpenAIEmbeddings()
+        with mock.patch("langchain.embeddings.OpenAIEmbeddings._get_len_safe_embeddings", return_value=[0.0] * 1536):
```
Yun-Kim, line 374: For langchain v0 tests, you don't actually need to mock out `_get_len_safe_embeddings()`, this seems to only be required for langchain community tests.

tests/contrib/langchain/test_langchain_llmobs.py
```diff
@@ -323,6 +343,56 @@ def test_llmobs_chain_schema_io(self, langchain, mock_llmobs_span_writer, mock_t
         )
         _assert_expected_llmobs_llm_span(trace[1], mock_llmobs_span_writer, mock_io=True)
 
+    @pytest.mark.skipif(sys.version_info < (3, 10, 0), reason="Requires unnecessary cassette file for Python 3.9")
+    def test_llmobs_embedding_query_langchain_base(self, langchain, mock_llmobs_span_writer, mock_tracer):
+        embedding_model = langchain.embeddings.OpenAIEmbeddings()
+        with mock.patch("langchain.embeddings.OpenAIEmbeddings._get_len_safe_embeddings", return_value=[0.0] * 1536):
```
Yun-Kim, line 349: For langchain v0 tests, you don't actually need to mock out _get_len_safe_embeddings(), this seems to only be required for langchain community tests.

tests/contrib/langchain/test_langchain_llmobs.py
```diff
@@ -324,6 +344,58 @@ def test_llmobs_chain_schema_io(self, langchain, mock_llmobs_span_writer, mock_t
         )
         _assert_expected_llmobs_llm_span(trace[1], mock_llmobs_span_writer, mock_io=True)
 
+    @pytest.mark.skipif(sys.version_info < (3, 10, 0), reason="Requires unnecessary cassette file for Python 3.9")
+    def test_llmobs_embedding_query_langchain_base(self, langchain, mock_llmobs_span_writer, mock_tracer):
```
Yun-Kim, line 348:
```suggestion
    def test_llmobs_openai_embedding_query(self, langchain, mock_llmobs_span_writer, mock_tracer):
```

tests/contrib/langchain/test_langchain_llmobs.py
```diff
@@ -324,6 +344,58 @@ def test_llmobs_chain_schema_io(self, langchain, mock_llmobs_span_writer, mock_t
...
+                model_provider="openai",
+                input_documents=[{"text": "hello world"}],
+                output_value="[1 embedding(s) returned with size 1536]",
+                tags={"ml_app": "langchain_test"},
+                integration="langchain",
+            )
+        )
+
+    @pytest.mark.skipif(sys.version_info < (3, 10, 0), reason="Requires unnecessary cassette file for Python 3.9")
+    def test_llmobs_embedding_documents_langchain_base(self, langchain, mock_llmobs_span_writer, mock_tracer):
```
Yun-Kim, line 373:
```suggestion
    def test_llmobs_openai_embedding_documents(self, langchain, mock_llmobs_span_writer, mock_tracer):
```

tests/contrib/langchain/test_langchain_llmobs.py
```diff
@@ -519,6 +591,63 @@ def test_llmobs_anthropic_chat_model(self, langchain_anthropic, mock_llmobs_span
         assert mock_llmobs_span_writer.enqueue.call_count == 1
         _assert_expected_llmobs_llm_span(span, mock_llmobs_span_writer, input_role="user")
 
+    @pytest.mark.skipif(sys.version_info < (3, 10, 0), reason="Requires unnecessary cassette file for Python 3.9")
+    def test_llmobs_embedding_query_langchain_community(
```
Yun-Kim, line 595:
```suggestion
    def test_llmobs_openai_embedding_query_(
```

tests/contrib/langchain/test_langchain_llmobs.py
```diff
@@ -519,6 +591,63 @@ def test_llmobs_anthropic_chat_model(self, langchain_anthropic, mock_llmobs_span
...
+                model_provider="openai",
+                input_documents=[{"text": "hello world"}],
+                output_value="[1 embedding(s) returned with size 1536]",
+                tags={"ml_app": "langchain_test"},
+                integration="langchain",
+            )
+        )
+
+    @pytest.mark.skipif(sys.version_info < (3, 10, 0), reason="Requires unnecessary cassette file for Python 3.9")
+    def test_llmobs_embedding_documents_langchain_community(
```
Yun-Kim, line 624:
```suggestion
    def test_llmobs_fake_embedding_documents(
```

Yun-Kim (approved):
Small nits for test name changes (the base class already reflect the langchain version being tested) but otherwise LGTM! Great work @yahya-mouman 🥳

tests/contrib/langchain/test_langchain_llmobs.py
```diff
@@ -682,6 +821,24 @@ def test_llmobs_langchain_with_openai_disabled(self):
...
         self._assert_trace_structure_from_writer_call_args(["llm"])
 
+    @run_in_subprocess(env_overrides=openai_env_config)
+    def test_llmobs_langchain_with_embedding_model_openai_enabled(self):
+        from langchain_openai import OpenAIEmbeddings
+
+        patch(langchain=True, openai=True)
+        LLMObs.enable(ml_app="<ml-app-name>", integrations_enabled=False, agentless_enabled=True)
+        self._call_openai_embedding(OpenAIEmbeddings)
+        self._assert_trace_structure_from_writer_call_args(["workflow"])
```
Yun-Kim, line 831:
```suggestion
        self._assert_trace_structure_from_writer_call_args(["workflow", "embedding"])
```
Just noticed this, is this not being traced correctly?

tests/contrib/langchain/test_langchain_llmobs.py
```diff
@@ -601,6 +733,13 @@ def _call_openai_llm(OpenAI):
         with get_request_vcr(subdirectory_name="langchain_community").use_cassette("openai_completion_sync.yaml"):
             llm.invoke("Can you explain what Descartes meant by 'I think, therefore I am'?")
 
+    @staticmethod
+    def _call_openai_embedding(OpenAIEmbeddings):
+        embedding = OpenAIEmbeddings()
+        with mock.patch("langchain_openai.OpenAIEmbeddings._get_len_safe_embeddings", return_value=[0.0] * 1536):
```
Yun-Kim, line 739:
```suggestion
```
When running with OpenAI, we do not want to mock out `_get_len_safe_embeddings()` because that involves the call to `OpenAI.embeddings.create(...)`.

Yun-Kim:
@yahya-mouman there are significant flakiness/skipping issues in the Langchain test suite. Let's merge this PR once #9768 is merged.

tests/contrib/langchain/test_langchain_llmobs.py
```diff
@@ -315,6 +335,58 @@ def test_llmobs_chain_schema_io(self, langchain, mock_llmobs_span_writer, mock_t
         )
         _assert_expected_llmobs_llm_span(trace[1], mock_llmobs_span_writer, mock_io=True)
 
+    @pytest.mark.skipif(sys.version_info < (3, 10, 0), reason="Requires unnecessary cassette file for Python 3.9")
```
Yun-Kim, line 338: instead of skipping this for Python 3.9, we can just use both sets of cassette files we have, i.e.
```python
cassette_name="openai_embedding_query_39.yaml" if PY39 else "openai_embedding_query.yaml",
```

tests/contrib/langchain/test_langchain_llmobs.py
```diff
@@ -315,6 +335,58 @@ def test_llmobs_chain_schema_io(self, langchain, mock_llmobs_span_writer, mock_t
...
+                model_name=embedding_model.model,
+                model_provider="openai",
+                input_documents=[{"text": "hello world"}],
+                output_value="[1 embedding(s) returned with size 1536]",
+                tags={"ml_app": "langchain_test"},
+                integration="langchain",
+            )
+        )
+
+    @pytest.mark.skipif(sys.version_info < (3, 10, 0), reason="Requires unnecessary cassette file for Python 3.9")
```
Yun-Kim, line 363:
```suggestion
```
Same here, we have python 3.9 test cassettes for `openai.embed_documents` in langchain v0.

tests/contrib/langchain/test_langchain_llmobs.py
```diff
@@ -315,6 +335,58 @@ def test_llmobs_chain_schema_io(self, langchain, mock_llmobs_span_writer, mock_t
...
+    def test_llmobs_embedding_documents(self, langchain, mock_llmobs_span_writer, mock_tracer):
+        embedding_model = langchain.embeddings.OpenAIEmbeddings()
+        with mock.patch(
+            "langchain.embeddings.OpenAIEmbeddings._get_len_safe_embeddings", return_value=[[0.0] * 1536] * 2
+        ):
+            trace = self._embed_documents(
+                embedding_model=embedding_model,
+                documents=["hello world", "goodbye world"],
+                mock_tracer=mock_tracer,
+                cassette_name="openai_embedding_document.yaml",
```
Yun-Kim, line 373:
```suggestion
                cassette_name="openai_embedding_document_39.yaml" if PY39 else "openai_embedding_document.yaml",
```

tests/contrib/langchain/test_langchain_llmobs.py
```diff
@@ -499,6 +571,61 @@ def test_llmobs_anthropic_chat_model(self, langchain_anthropic, mock_llmobs_span
         assert mock_llmobs_span_writer.enqueue.call_count == 1
         _assert_expected_llmobs_llm_span(span, mock_llmobs_span_writer, input_role="user")
 
+    @pytest.mark.skipif(sys.version_info < (3, 10, 0), reason="Requires unnecessary cassette file for Python 3.9")
```
Yun-Kim, line 574:
```suggestion
```
I think we can actually get rid of this skip as it seems to be fine running in Python 3.9 without any extra cassette. Try this out and let me know if that works :)

tests/contrib/langchain/test_langchain_llmobs.py
```diff
@@ -499,6 +571,61 @@ def test_llmobs_anthropic_chat_model(self, langchain_anthropic, mock_llmobs_span
...
+                model_name=embedding_model.model,
+                model_provider="openai",
+                input_documents=[{"text": "hello world"}],
+                output_value="[1 embedding(s) returned with size 1536]",
+                tags={"ml_app": "langchain_test"},
+                integration="langchain",
+            )
+        )
+
+    @pytest.mark.skipif(sys.version_info < (3, 10, 0), reason="Requires unnecessary cassette file for Python 3.9")
```
Yun-Kim, line 601: Since this doesn't require cassettes, we can remove this skip as well.
```suggestion
```

Yun-Kim (commented):
Should be good to go after the testing comments are addressed now that the flakiness should be addressed!

tests/contrib/langchain/test_langchain_llmobs.py
yahya-mouman: Yes i believe it still works !

yahya-mouman:
> Should be good to go after the testing comments are addressed now that the flakiness should be addressed!

Great job on fixing the flakiness of the tests 🥇 I believe everything is done and good to go with this PR

PR #9872, opened 2024-07-19
chore: iast aspects microbenchmarks

juanjux:
## Description

Micro benchmarks for each aspect, comparing the patched and unpatched performance. 

Note: this moves some IAST fixtures from `tests.*` to `benchamarks.*` because the environment under which the benchmarks are run doesn't have access to `tests.*` (but tests have access to benchmarks dirs).

## Checklist

## Reviewer Checklist
- Title is accurate


benchmarks/appsec_iast_aspects/config.yaml
```diff
@@ -0,0 +1,5 @@
+no_iast: &base_variant
+iast_enabled: 0
+
+iast_enabled: &iast_enabled
+iast_enabled: 1
```
erikayasuda, line 5: How does this work with the scenario.py file?

erikayasuda (approved):
Approving with one NB question (because I'm not super familiar with benchmark configs), but I think this should not require core review. Will probably make a PR to specify only ASM approval for these.

benchmarks/appsec_iast_aspects/config.yaml
juanjux: It fills the `iast_enabled` member from the `IAST_Aspect` class (which inherits from the benchmark one). The PR is not yet finished because while it works it doesn't give a report with the granularity I would like and need to investigate how to do it (I moved it back to draft).

gnufede (approved):

juanjux:
Waiting for some way to make the individual scenarios faster so we don't timeout.

PR #9898, opened 2024-07-22
chore: privatizing public API of integrations in flask/patch.py

rachelyangdog:
The API should just be what is necessary. A lot of integrations expose implementation details through the patch.py We can’t just remove these functions right away as it will break our API so for each we will need to deprecate first and then remove in 2.x.

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/contrib/flask/patch.py
```diff
@@ -70,11 +72,21 @@
 )
 
 
-def get_version():
+def _get_version():
     # type: () -> str
     return get_version_for_package("flask")
 
 
+def get_version():
```
mabdinur, line 80: `get_version` is defined in every integration. If we plan to rename this function in the flask integration we should do it for our other integrations. Here's the PR that introduced this function: https://github.com/DataDog/dd-trace-py/commit/ab6321c9dd2beccecbc8e8028cd4144689db98fe.

ZStriker19: I think the plan is to go one integration at a time. @rachelyangdog lmk if that sounds right.

ZStriker19, line 80: We need to make sure that everywhere any of the now deprecated methods are called, we switch them to the non-deprecated private version. E.g. we need to change line 111. This'll fix a lot of the failing tests.

rachelyangdog: yes! i was going to make a PR for each integration

ddtrace/contrib/flask/__init__.py
```diff
@@ -108,6 +108,6 @@ def index():
 
         patch = _patch.patch
         unpatch = _patch.unpatch
-        get_version = _patch.get_version
+        get_version = _patch._get_version
```
ZStriker19, line 111: We shouldn't change this. The get_version we expose here should raise a deprecation warning.

erikayasuda:
@DataDog/apm-idm-python Could we have a reviewer on this? Sorry the automation didn't automatically tag someone

ZStriker19 (approved):

ddtrace/contrib/flask/patch.py
```diff
@@ -71,11 +73,21 @@
 )
 
 
-def get_version():
```
mabdinur, line 74: Can we revert this change? get_version is defined every integration and is used in ddtrace._monkey. Renaming this method would require a significant change that will updated tens of files.

PR #9912, opened 2024-07-23
feat(llmobs): make agentless mode a global config setting

Yun-Kim:
This PR introduces a new global config setting `_llmobs_agentless_enabled`, which is based off an environment variable `DD_LLMOBS_AGENTLESS_ENABLED` that we already use to configure the LLMObs span writing setting, just previously via `os.getenv()` directly. Now we use the global config object to keep track of this setting.

## Checklist

## Reviewer Checklist
- Title is accurate


lievan (approved):

romainkomorndatadog:
I don't know how I feel about this, given that we already have `DD_CIVISIBILITY_AGENTLESS_ENABLED`. And we already made a change to expand the agentless behavior to the telemetry writer in #8948 .

Are we going to end up in a scenario where multiple products have some sort of agenltess mode of their own?

I guess it would've been worth raising that concern when the original change to introduce agentless mode for LLMObs was added, though.

Yun-Kim:
@romainkomorndatadog so `DD_LLMOBS_AGENTLESS_ENABLED` has been a configuration (just not a global config setting) for some time now, in hindsight we should've coordinated a better way to allow products to submit agentless payloads instead of developing redundant code separately.

However since we already have these two env vars `DD_LLMOBS/CIVISIBILITY_AGENTLESS_ENABLED`, I think we should combine these into a new `DD_AGENTLESS_ENABLED` env var so that we keep a central configuration to handle agentless cases. (We would keep but deprecate our separate env vars until the next major version I assume). 

Do you think there might be a case where a user might want to configure agentless mode for one product but not the other?

romainkomorndatadog:
> Do you think there might be a case where a user might want to configure agentless mode for one product but not the other?

I can't imagine why, unless someone wanted to be able to have attached infrastructure metrics for some product and not others (or... if one product doesn't work like it's supposed to in either mode).

I guess there are broader product discussions to have about which products can behave in an agentless mode and what that implies for tracers. I assume LLM Obs has agentless behavior across all the tracers, like CI Visibility does?

Yun-Kim:
@romainkomorndatadog we can hold off on a central `DD_AGENTLESS_ENABLED` setting until we get a more organized cross-product discussion about agentless mode. Regardless, `DD_LLMOBS_AGENTLESS_ENABLED` has already been a setting for months now and is used by customers, so this shouldn't be blocked on that.

taegyunkim (approved):

PR #9928, opened 2024-07-24
chore: privatizing public API of integrations in tornado patch.py

rachelyangdog:
The API should just be what is necessary. A lot of integrations expose implementation details through the patch.py We can’t just remove these functions right away as it will break our API so for each we will need to deprecate first and then remove in 2.x.

## Checklist

## Reviewer Checklist
- Title is accurate

### DIFF
--- a/releasenotes/notes/deprecate-patch-py-methods-tornado-unittest-urllib-vertica-7e5e7bd8f5613323.yaml
+++ b/releasenotes/notes/deprecate-patch-py-methods-tornado-unittest-urllib-vertica-7e5e7bd8f5613323.yaml
@@ -1,0 +1,12 @@
+---
+deprecations:
+  - |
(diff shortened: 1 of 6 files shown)


releasenotes/notes/deprecate-patch-py-methods-tornado-unittest-urllib-vertica-7e5e7bd8f5613323.yaml
```diff
@@ -0,0 +1,12 @@
+---
+deprecations:
+  - |
```
ZStriker19, line 3: Can we add tick marks around patch and unpatch for each?

ZStriker19 (approved):
Looks good, could we just update the PR name to reflect the various integrations modified here?

Issue #9917, opened 2024-07-24
OpenTelemetry API context propagation is broken starting with opentelemetry-api==1.25.0


andreaimprovised:
### Summary of problem

Starting in opentelemetry-api==1.25.0, context propagation between ddtrace and opentelemetry does not work.

There is an exception logged that suggests that "ddcontextvars_context" is not used.

```
ERROR:opentelemetry.context:Failed to load context: ddcontextvars_context, fallback to contextvars_context
Traceback (most recent call last):
  File "/path/to/opentelemetry/context/__init__.py", line 52, in _load_runtime_context
    ).load()()
  File "/path/to/importlib_metadata/__init__.py", line 208, in load
    module = import_module(match.group('module'))
  File "/path/to/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "/path/to/ddtrace/internal/module.py", line 220, in _exec_module
    self.loader.exec_module(module)
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/path/to/ddtrace/opentelemetry/_context.py", line 6, in <module>
    from opentelemetry.trace import NonRecordingSpan as OtelNonRecordingSpan
  File "/path/to/ddtrace/internal/module.py", line 220, in _exec_module
    self.loader.exec_module(module)
  File "/path/to/opentelemetry/trace/__init__.py", line 89, in <module>
    from opentelemetry.trace.propagation import (
  File "/path/to/ddtrace/internal/module.py", line 220, in _exec_module
    self.loader.exec_module(module)
  File "/path/to/opentelemetry/trace/propagation/__init__.py", line 16, in <module>
    from opentelemetry.context import create_key, get_value, set_value
ImportError: cannot import name 'create_key' from partially initialized module 'opentelemetry.context' (most likely due to a circular import) (/path/to/opentelemetry/context/__init__.py)
```

[...]

PR #9929, opened 2024-07-24
fix(opentelemetry): resolve circular imports

mabdinur:
Resolves: https://github.com/DataDog/dd-trace-py/issues/9917

## Background

ddtrace replaces the default opentelemetry RuntimeContext with a datadog-aware runtime context. This allows the ddtrace library to correlate datadog and opentelemetry spans in the same trace.

The OpenTelemetry API exposes an [entrypoint](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-api/src/opentelemetry/context/__init__.py#L48) for overriding the default RuntimeContext ([ContextVarsRuntimeContext](https://github.com/open-telemetry/opentelemetry-python/blob/v1.25.0/opentelemetry-api/src/opentelemetry/context/contextvars_context.py#L19)). The ddtrace library implements this [entrypoint](https://github.com/DataDog/dd-trace-py/blob/v2.10.0rc3/pyproject.toml#L56) and overrides `ContextVarsRuntimeContext` with [DDRuntimeContext](https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/opentelemetry/_context.py#L26). 

### When `DD_TRACE_OTEL_ENABLED=True` and `opentelemetry-api==1.25.0`  is used the following occurs:

-> ddtrace  imports [set_tracer_provider](https://github.com/DataDog/dd-trace-py/blob/9cce458b5ffeabe1dfee22687c89b8635856b9b6/ddtrace/bootstrap/preload.py#L107) from `openelemetry.trace`
-> On import [openelemetry.trace](https://github.com/open-telemetry/opentelemetry-python/blob/3f95781a64cc6b897e6e34df907cd187c1e4c2f4/opentelemetry-api/src/opentelemetry/trace/__init__.py#L86) the `opentelemetry.context` package is imported.
-> On [import opentelemetry.context](https://github.com/open-telemetry/opentelemetry-python/blob/3f95781a64cc6b897e6e34df907cd187c1e4c2f4/opentelemetry-api/src/opentelemetry/context/__init__.py#L69) the `_RUNTIME_CONTEXT` object is initialized this calls the `_load_runtime_context` method.
-> On  [_load_runtime_context](https://github.com/open-telemetry/opentelemetry-python/blob/3f95781a64cc6b897e6e34df907cd187c1e4c2f4/opentelemetry-api/src/opentelemetry/context/__init__.py#L47) the DDRuntimeContext class is imported from `ddtrace.opentelemetry._context`.
->  On [import ddtrace.opentelemetry._context](https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/opentelemetry/_context.py#L6) the `opentelemetry.trace` package is loaded. This packaged is used in DDRuntimeContext to access [types and functions](https://github.com/DataDog/dd-trace-py/blob/9cce458b5ffeabe1dfee22687c89b8635856b9b6/ddtrace/opentelemetry/_context.py#L9) used to generate OpenTelemetry compatible spans.
-> On import `opentelemetry.trace` the `opentelemetry.context.__init__ `module is indirectly imported. 
-> BOOM circular import
-> `_load_runtime_context` handles the raised exception,`DDRuntimeContext` is not set, and the default `ContextVarsRuntimeContext` is used. This breaks span parenting and ddtrace opentelemetry support.

### What changed in opentelemetry-api==1.25.0?

Previously, `_load_runtime_context` was lazily loaded and `_RuntimeContext` was initialized when the first span was created. This changed in this commit: https://github.com/open-telemetry/opentelemetry-python/commit/52abb610dbdf1751cc745720055119ca1656ec16. With this change the _RuntimeContext is ALWAYS initialized as a side effect of importing the `opentelemetry.trace` package.

## Description

Since `_load_runtime_context` is no longer lazy loaded we must ensure `opentelemetry.context` is NOT imported as a side effect of importing `DDRuntimeContext`. We achieve this behavior by moving the `ddtrace.opentelemetry._context.py` module and inlining problematic imports.

## Regression Test

Run the ddtrace opentelemetry testsuite with `opentelemetry-api==1.25.0` (currently we use 1.22.0). This should be sufficient to reproduce this bug.

## Checklist

## Reviewer Checklist
- Title is accurate


brettlangdon:
Do we need to get rid of `ddtrace/opentelemetry/_context.py` ?

mabdinur:
> Do we need to get rid of `ddtrace/opentelemetry/_context.py` ?

I moved ddtrace/opentelemetry/_context.py to ddtrace/internal/opentelemetry/context.py. 

ddtrace/opentelemetry/_init_.py imports the ddtrace TracerProvider. Users are expected to use this class to enable opentelemetry support in ddtrace. The ddtrace TracerProvider imports classes and functions from the opentelemetry.trace package and this triggers this circular import. The easiest way to avoid breaking our public api and resolving the circular import is to move the DDRuntimeContext object from the ddtrace.opentelemetry package.

mabdinur:
> > I renamed ddtrace/opentelemetry/_context.py to ddtrace/internal/opentelemetry/context.py.
> 
> I still see the original file there.
> 
> https://github.com/DataDog/dd-trace-py/blob/05d8b274ec7b687bc0ee6221412b8997941c3aa4/ddtrace/opentelemetry/_context.py

Ahh I must've reverted a change during a rebase. The file is removed: https://github.com/DataDog/dd-trace-py/pull/9929/files#diff-62a0b204575d0586747747152ea77d0466cf05d05ca2bd5f4327478ea9583682.

For consistency we can also move `ddtrace.opentelemetry._span` and `ddtrace.opentelemetry._trace` to `ddtrace/opentelemetry`. I can include this refactor in a future PR.

ZStriker19 (approved):
Thanks for updating codeowners and suitespec too

riotfile.py
```diff
@@ -2351,12 +2351,11 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
         Venv(
             name="opentelemetry",
             command="pytest {cmdargs} tests/opentelemetry",
-            # FIXME: this test suite breaks on 3.7
             pys=select_pys(min_version="3.8"),
             pkgs={
                 "pytest-randomly": latest,
                 "pytest-asyncio": "==0.21.1",
-                "opentelemetry-api": ["~=1.0.0", "~=1.3.0", "~=1.4.0", "~=1.8.0", "~=1.11.0", "~=1.15.0", latest],
+                "opentelemetry-api": ["~=1.0.0", "~=1.15.0", latest],
```
erikayasuda, line 2358: nit: Should we pin on 1.25.0 explicitly and latest? Just so we know the "latest" that should have worked in case the next latest also is incompatible? 
```suggestion
                "opentelemetry-api": ["~=1.0.0", "~=1.15.0", "~=1.25.0", latest],
```

ddtrace/internal/opentelemetry/context.py
```diff
@@ -79,7 +83,6 @@ def detach(self, token):
 
     @property
     def _ddcontext_provider(self):
-        # type: () -> DDBaseContextProvider
```
erikayasuda, line 82: nit: Are we removing these other type hints because it's all or none? Or consistency?

erikayasuda (approved):
Just some NB nits. Thanks for the great description!

ddtrace/internal/opentelemetry/context.py
mabdinur: I went for the all or none approach. I don't have a strong opionon here. I guess something is better than nothing. Also these are python2 compatible type hints. We need to remove them and replace them with python3 type hints anyway.

zacharycmontoya (approved):

riotfile.py
mabdinur: I made this change in the next PR: https://github.com/DataDog/dd-trace-py/pull/9941/files#diff-8c2dff2cbcc0f4fff63050381d4c7546ae7195383ec8f9cc993f8494b77a34c3. We actually need to test 1.26.0 and not just 1.25.0. Testing both is a bit overkill.

PR #9940, opened 2024-07-25
fix(psycopg): support async iteration

quinna-h:
Motivation:

Previously, using async iteration with `psycopg` (`async for`) would throw an exception. 

Example:

```
def connect_to_db():
    # db connection logic...

async def testfunc():
    connectstr = get_conn_str()
    async with AsyncConnectionPool(
        conninfo=connectstr, min_size=4, max_size=30
    ) as async_pool:
        async with async_pool.connection() as conn:
            async with conn.cursor() as cur:
                await cur.execute(
                    sql.SQL(
                        """
                        SELECT
                            *
                        FROM table1
                        LIMIT 5
                    """
                    )
                )
                async for row in cur:
                    print(row)
```

Running `ddtrace-run python test.py` would return `TypeError: 'async for' requires an object with __aiter__ method, got Psycopg3TracedAsyncCursor`.

This fix implements the `__aiter__` method for the `TracedAsyncCursor` parent class, and has `Psycopg3TracedAsyncCursor` inherit from it. 

## Checklist

## Reviewer Checklist
- Title is accurate


riotfile.py
```diff
@@ -1195,6 +1195,7 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
             command="pytest {cmdargs} tests/contrib/psycopg",
             pkgs={
                 "pytest-randomly": latest,
+                "psycopg-pool": latest,
```
ZStriker19, line 1198: We don't use this in the test you wrote do we? Can remove if so.

ddtrace/contrib/psycopg/async_cursor.py
```diff
@@ -21,6 +26,33 @@ async def __aexit__(self, exc_type, exc_val, exc_tb) -> None:
         # messages will be the same.
         return await self.__wrapped__.__aexit__(exc_type, exc_val, exc_tb)
 
+    def __aiter__(self):
```
ZStriker19, line 29: Does this follow the behavior of the original Pyscopg3AsyncCursor? I think that this probably avoids the error, but would lead to unexpected behavior since it seems that this just returns self, and the expected effect of aiter is: https://github.com/psycopg/psycopg/blob/9c8664117ff20fff7eb3e72c31868764439eddcd/psycopg/psycopg/cursor_async.py#L239-L251

ZStriker19:
Good start, I left a couple of comments. We should also add a release note. Reach out to @rachelyangdog if you're not sure how to do that part.

ddtrace/contrib/psycopg/async_cursor.py
quinna-h: I was referencing other places in the repo where we implement asynchronous iteration via the `_aiter_` and `_anext_` methods: https://github.com/DataDog/dd-trace-py/blob/3237351b4a35dfd52e3d3d05c81bfe5b51279372/ddtrace/contrib/openai/utils.py#L106
so in the case of processing rows, it would just call the `_anext_` method until `row is None` which raises `StopAsyncIteration`.

However, reading [here](https://peps.python.org/pep-0525/) it seems like async generators are more performant than async iterators anyway, which would especially matter in the case of handling large datasets. So I can switch back to the original implementation.

ddtrace/contrib/psycopg/async_cursor.py
```diff
@@ -21,6 +21,25 @@ async def __aexit__(self, exc_type, exc_val, exc_tb) -> None:
...
+            row = load(self._pos)
+            if row is None:
+                break
+            self._pos += 1
+            yield row
+
+    async def _fetch_pipeline(self) -> None:
+        if self._execmany_returning is not False and not self.pgresult and self._conn._pipeline:
+            async with self._conn.lock:
+                await self._conn.wait(self._conn._pipeline._fetch_gen(flush=True))
```
wconti27, line 41: Although we need to implement `__aiter__`, we don't want to re implement the wrapped cursor's functionality. Instead, we want to ensure we are correctly calling the wrapped method from our new `__aiter__` method. 
```suggestion
    def __aiter__(self):
        return self.__wrapped__.__aiter__()
```

wconti27: this change should also probably be moved up into the dbapi async cursor parent class of this

tests/contrib/psycopg/test_psycopg_async.py
```diff
@@ -337,3 +338,26 @@ async def test_cursor_from_connection_shortcut(self):
...
+                    )
+                )
+                await cur.execute("INSERT INTO {} (a, b) VALUES (1, 'aa');".format(TEST_TABLE))
+                await cur.execute("INSERT INTO {} (a, b) VALUES (2, 'bb');".format(TEST_TABLE))
+                await cur.execute("INSERT INTO {} (a, b) VALUES (3, 'cc');".format(TEST_TABLE))
+
+                count = 1
+                async for row in cur:
+                    assert row[0] == count
+                    count += 1
```
mabdinur, line 363: Can we validate the spans generate by this test case?
Something like 
```suggestion
                    count += 1
                    spans = self.get_spans()
                    self.assertEqual(len(spans), X)
                    # maybe some assertions on span names or resource names
```

tests/contrib/psycopg/test_psycopg_async.py
```diff
@@ -337,3 +338,26 @@ async def test_cursor_from_connection_shortcut(self):
         rows = await cur.fetchall()
         assert len(rows) == 1, rows
         assert rows[0][0] == "one"
+
+    async def test_cursor_async_connect_execute(self):
```
mabdinur, line 342: Great regression test for postgres! Can we also add a test case that directly test the TracedAyncCursor class: https://github.com/DataDog/dd-trace-py/blob/ebf3fa99ba5cd58a57fa4c5f1950fb6068754645/tests/contrib/dbapi_async/test_dbapi_async.py#L19.

releasenotes/notes/fix-psycopg-async-iteration-d48a62f8aa597457.yaml
```diff
@@ -0,0 +1,4 @@
+---
+fixes:
+  - |
+    Fixes issue of psycopg async iterator previously not returning objects, by implementing async generator.
```
mabdinur, line 4: nit: We usually put the scope of the change at the front of the release note (so name of integration / feature / product / etc). That way user's can quickly scan our release notes and know what applies to them.

```suggestion
    psycopg: Fixes issue of psycopg async iterator previously not returning objects, by implementing async generator.
```

emmettbutler (commented):
I'll wait until existing conversations have been resolved before doing a thorough review.

tests/contrib/dbapi_async/test_dbapi_async.py
```diff
@@ -95,6 +98,36 @@ async def test_fetchone_wrapped_is_called_and_returned(self):
         assert "__result__" == await traced_cursor.fetchone("arg_1", kwarg1="kwarg1")
         cursor.fetchone.assert_called_once_with("arg_1", kwarg1="kwarg1")
 
+    @mark_asyncio
+    async def test_cursor_async_connection(self):
+        """Checks whether connection can execute operations with async iteration."""
+
+        cursor = self.cursor
+        traced_cursor = TracedAsyncCursor(cursor, Pin("dbapi_service", tracer=self.tracer), {})
+        async with traced_cursor as cursor:
```
mabdinur, line 107: Can we simplify this line to:
```suggestion
        async with  TracedAsyncCursor(self.cursor, Pin("dbapi_service", tracer=self.tracer), {}) as cursor:
```

tests/contrib/dbapi_async/test_dbapi_async.py
```diff
@@ -95,6 +98,36 @@ async def test_fetchone_wrapped_is_called_and_returned(self):
...
+            await cursor.execute("DROP TABLE IF EXISTS {}".format(TEST_TABLE))
+            await cursor.execute(
+                """CREATE TABLE {} (
+                        a INT,
+                        b VARCHAR(32)
+                        )
+                        """.format(
+                    TEST_TABLE
+                )
+            )
```
mabdinur, line 117: Do we need to create a table for this test? Can we just execute a `select` query. It would be nice if this test did not require clean up.

mabdinur, line 117: We can put the query on one line and allow dbapi to format the query. 
```suggestion
            await cursor.execute(f"CREATE TABLE  %s ( a INT, b VARCHAR(32))", TEST_TABLE)
```

tests/contrib/dbapi_async/test_dbapi_async.py
```diff
@@ -95,6 +98,36 @@ async def test_fetchone_wrapped_is_called_and_returned(self):
...
+            await cursor.execute("INSERT INTO {} (a, b) VALUES (1, 'aa');".format(TEST_TABLE))
+            await cursor.execute("INSERT INTO {} (a, b) VALUES (2, 'bb');".format(TEST_TABLE))
+            await cursor.execute("INSERT INTO {} (a, b) VALUES (3, 'cc');".format(TEST_TABLE))
+
+            count = 1
+            async for row in cursor:
+                assert row[0] == count
```
mabdinur, line 124: Do we need assertions on the row count. Can we just validate the `span.resource` field.

tests/contrib/dbapi_async/test_dbapi_async.py
```diff
@@ -95,6 +98,36 @@ async def test_fetchone_wrapped_is_called_and_returned(self):
...
+                count += 1
+                spans = self.get_spans()
+                assert spans
```
mabdinur, line 127: We should validate the number of number of spans. Ex:
```suggestion
                assert len(spans) == X
```

tests/contrib/dbapi_async/test_dbapi_async.py
```diff
@@ -95,6 +98,36 @@ async def test_fetchone_wrapped_is_called_and_returned(self):
...
+                span = spans[0]
+                assert span.service == "dbapi_service", "Service from pin"
```
mabdinur, line 129:
```suggestion
                assert span.service == "dbapi_service", "Service from pin"
                assert span.resource == XXXX
                assert span.name == YYYY
```

tests/contrib/psycopg/test_psycopg_async.py
```diff
@@ -337,3 +338,30 @@ async def test_cursor_from_connection_shortcut(self):
         rows = await cur.fetchall()
         assert len(rows) == 1, rows
         assert rows[0][0] == "one"
+
+    @AsyncioTestCase.run_in_subprocess(env_overrides=dict(DD_TRACE_SPAN_ATTRIBUTE_SCHEMA="v0"))
+    async def test_cursor_async_connect_execute(self):
```
mabdinur, line 343: Same nits as the test above

tests/contrib/psycopg/test_psycopg_async.py
```diff
@@ -337,3 +338,30 @@ async def test_cursor_from_connection_shortcut(self):
         rows = await cur.fetchall()
         assert len(rows) == 1, rows
         assert rows[0][0] == "one"
+
+    @AsyncioTestCase.run_in_subprocess(env_overrides=dict(DD_TRACE_SPAN_ATTRIBUTE_SCHEMA="v0"))
```
mabdinur, line 342: I don't think we need to run this test in a subprocess
```suggestion
```

releasenotes/notes/fix-psycopg-async-iteration-d48a62f8aa597457.yaml
```diff
@@ -0,0 +1,4 @@
+---
+fixes:
+  - |
+    psycopg: Fixes issue of psycopg async iterator previously not returning objects, by implementing async generator.
```
mabdinur, line 4: nit: We can simplify this release note
```suggestion
    psycopg: Ensures traced async cursors return an asynchronous iterator object.
```

tests/contrib/psycopg/test_psycopg_async.py
```diff
@@ -337,3 +337,17 @@ async def test_cursor_from_connection_shortcut(self):
...
+        async with psycopg.AsyncConnection.connect(**POSTGRES_CONFIG) as conn:
+            async with conn.cursor() as cur:
+                await cur.execute("""select 'one' as x""")
+                await cur.execute("""select 'blah'""")
+
+                async for row in cur:
+                    spans = self.get_spans()
+                    assert len(spans) == 2
+                    span = spans[0]
+                    assert span.name == "postgres.query"
```
mabdinur, line 353: Can we add an assertion for the second span.
Also can we assert the resource name. This will allow us to make sure a span was generated for the `select 'one' as x` and `select 'blah'` operations. 
```suggestion
                    assert spans[0].name == "postgres.query"
                    assert spans[0].resource == ...
                    assert spans[1].name == "postgres.query"
                    assert spans[1].resource == ...
```

tests/contrib/dbapi_async/test_dbapi_async.py
```diff
@@ -95,6 +95,27 @@ async def test_fetchone_wrapped_is_called_and_returned(self):
...
 
+    @mark_asyncio
+    async def test_cursor_async_connection(self):
+        """Checks whether connection can execute operations with async iteration."""
+
+        def method():
+            pass
+
+        async with TracedAsyncCursor(self.cursor, Pin("dbapi_service", tracer=self.tracer), {}) as cursor:
+            cursor._trace_method(method, "my_name", "my_resource", {"extra1": "value_extra1"}, False)
```
mabdinur, line 106: Can we add a comment describing why this line is required. I am not sure if we need this in this test. This test should validate that `cursor.execute("""select 'one' as x""")` and `cursor.execute("""select 'blah'""")` generate spans and `async for row in cursor` uses a valid async iterator

mabdinur (approved):

wconti27 (approved):

PR #9942, opened 2024-07-25
chore(hatch): adds lint:checks script

mabdinur:
There is no one hatch script that runs all checks in `pre-check` ([ex](https://app.circleci.com/pipelines/github/DataDog/dd-trace-py/66755/workflows/193957fc-1523-468f-ba87-0acfeb8817f2/jobs/4099402)).

This PR adds the `hatch run lint:check` command which validate changes against all pre_check rules.

## Potential next steps

Update [gen_pre_checks](https://github.com/DataDog/dd-trace-py/blob/v2.10.0rc4/scripts/gen_circleci_config.py#L41) to use `hatch run lint:check` validate all linting rules. This will help prevent `lint:checks` from becoming outdated.

## Checklist

## Reviewer Checklist
- Title is accurate


hatch.toml
```diff
@@ -44,6 +44,8 @@ fmt = [
     "black {args:.}",
     "ruff --fix {args:.}",
     "style",
+    "typing",
+    "spelling",
```
mabdinur, line 48: We could also add the following hatch scripts to the list:`riot`, `security`, `test`, and `suitespec-check`

mabdinur, line 48: Should we include these scripts as well?
```suggestion
    "spelling",
    "riot",
    "security",
    "test",
    "suitespec-check",
```

romainkomorndatadog:
I don't think this makes sense. `fmt` isn't for linting, it's for auto-formatting.

romainkomorndatadog:
eg: we don't run the `fmt` script in CI. It's there as a convenience, but the real flow show be:

1. edit code
2. optionally `hatch lint:fmt`
3. `git commit` which invokes the pre-commit hook that lints code
4. `git push`

romainkomorndatadog:
Having something that matches the pre checks does make sense to me (and I think that should be getting called by pre commit hooks, or maybe we should introduce a pre-push hook).

I have a pet peeve against things getting autoformatted in hooks (because that has bitten me multiple times in the past where files end up modified but left out of the eventual commit). I think "things that can modify files" and "things that run checks" should be kept separate.

hatch.toml
```diff
@@ -44,6 +44,12 @@ fmt = [
...
+    "riot",
+    "security",
+    "test",
+    "suitespec-check",
```
romainkomorndatadog, line 52: What do you think of this, then, which we would invoke with `hatch run lint:checks` ?

```suggestion
]
checks = [
    "style",
    "typing",
    "spelling",
    "riot",
    "security",
    "test",
    "suitespec-check",
```

lib-injection/sitecustomize.py
```diff
@@ -21,7 +21,8 @@
 
 
 def parse_version(version: str) -> Tuple:
-    constraint_idx = re.search(r"\d", version).start()
+    v = re.search(r"\d", version)
```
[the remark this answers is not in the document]
mabdinur: This was required to fix a linting error. `re.search(r"\d", version)` can return `None`.  We need to do a none check before calling `.start()`.

Yesterday I was experimenting with expanding the scope of this PR to address @romainkomorndatadog's concerns. It would be nice if one hatch script encapsulated all the checks run in pre-check. This will allow us to easily run pre-check locally. I will revert this change and avoid updating `gen_circleci_config.py`. I'll limit the scope of this PR to adding `lint:checks` to `hatch.toml`. This should be a good first step.

hatch.toml
```diff
@@ -45,6 +45,15 @@ fmt = [
     "ruff --fix {args:.}",
     "style",
 ]
+checks = [
```
mabdinur, line 48:
```suggestion
# checks should encapsulate all rules used in pre-check
checks = [
```

emmettbutler (approved):

PR #9999, opened 2024-07-30
feat(llmobs): shared HTTPWriter-based span writer

tomshen:
Uses a shared HTTPWriter-based LLM Obs writer for both agent and agentless.

## Checklist

## Reviewer Checklist
- Title is accurate


tests/llmobs/test_llmobs_span_writer.py
```diff
@@ -1,93 +1,76 @@
...
+from ddtrace import config
 from ddtrace.llmobs._writer import LLMObsSpanWriter
 from tests.llmobs._utils import _chat_completion_event
 from tests.llmobs._utils import _completion_event
 
 
-INTAKE_ENDPOINT = "https://llmobs-intake.datad0g.com/api/v2/llmobs"
-DD_SITE = "datad0g.com"
-dd_api_key = os.getenv("DD_API_KEY", default="<not-a-real-api-key>")
+INTAKE_URL = "https://llmobs-intake.%s" % config._dd_site
```
Yun-Kim, line 12: Can we continue to use staging `datad0g.com` for testing?
```suggestion
DD_SITE = "datad0g.com"
INTAKE_ENDPOINT = "https://llmobs-intake.{}/api/v2/llmobs".format(DD_SITE)
```

tests/llmobs/test_llmobs_span_writer.py
```diff
@@ -1,93 +1,76 @@
...
-    llmobs_span_writer = LLMObsSpanWriter(site="datad0g.com", api_key=dd_api_key, interval=1, timeout=1)
+    llmobs_span_writer = LLMObsSpanWriter(is_agentless=True, interval=1, timeout=1)
     llmobs_span_writer.start()
     llmobs_span_writer.enqueue(_chat_completion_event())
     llmobs_span_writer.periodic()
-    mock_writer_logs.debug.assert_has_calls([mock.call("sent %d LLMObs %s events to %s", 1, "span", INTAKE_ENDPOINT)])
-
-
-@pytest.mark.vcr_logs
-def test_send_completion_bad_api_key(mock_writer_logs):
```
Yun-Kim, line 51: Why did we get rid of this test case?

tests/llmobs/test_llmobs_span_writer.py
tomshen: this change is because we no longer pass in datadog_site as a parameter to the constructor, since we use the same class for both agent and agentless, and pull datadog site from config directly.

I could support optionally passing in datadog site via a constuctor parameter? that would let us pass in `datad0g.com` for testing.

alternatively, if there's a way to set the config values for unit tests, we could do that instead

tests/llmobs/test_llmobs_span_writer.py
tomshen: similar to the datadog site change, I also have the span writer pull in an API key from config. since we don't pass in an API key, we don't need to test a case where we pass in an API key.

we could test this if we had a way to set the API key config (or add back passing an API key to test this case)

tests/llmobs/test_llmobs_span_writer.py
Yun-Kim: No problem, we have the `override_global_config()` test utility to set config values for tests: 
```python
from tests.utils import override_global_config

with override_global_config(dict(_dd_site="datad0g.com")):
    # rest of test
```

tests/llmobs/test_llmobs_span_writer.py
Yun-Kim: If we can use the `override_global_config()` helper, then I think this should be addressed - I'd like to ensure we have testing for both cases (agent proxy and agentless) if possible 😄

tests/llmobs/test_llmobs_span_writer.py
```diff
@@ -7,92 +6,123 @@
...
 
 def test_send_on_exit(mock_writer_logs, run_python_code_in_subprocess):
-    out, err, status, pid = run_python_code_in_subprocess(
-        """
+    with override_global_config(
+        dict(
+            _dd_site=DATADOG_SITE,
+            _dd_api_key="foobar.baz",
+        )
+    ):
```
Yun-Kim, line 123: We can't override the global config here as that doesn't reach the subprocess, here we should modify the env used in the subprocess test like [here](https://github.com/DataDog/dd-trace-py/blob/main/tests/contrib/openai/test_openai_v1.py#L1203-L1229).
Note that the env just needs to include the `"PYTHONPATH"`, you can replace `OPENAI_API_KEY/DD_OPENAI_METRICS_ENABLED` in that example with `DD_API_KEY/DD_SITE`.

tests/llmobs/test_llmobs_span_writer.py
Yun-Kim: I think the test failure here is because the previous test relied on the direct request/response from the naive `LLMObsSpanWriter` with the LLMObs public intake. I think `HTTPWriter` handles 403 responses differently (different logs I think)? Might be worth figuring out what the actual logs/error handling are from 403 responses in the HTTPWriter.

ddtrace/llmobs/_llmobs.py
```diff
@@ -80,18 +79,11 @@ def __init__(self, tracer=None):
...
-        else:
-            self._llmobs_span_writer = LLMObsSpanAgentWriter(
-                interval=float(os.getenv("_DD_LLMOBS_WRITER_INTERVAL", 1.0)),
-                timeout=float(os.getenv("_DD_LLMOBS_WRITER_TIMEOUT", 5.0)),
-            )
+        self._llmobs_span_writer = LLMObsSpanWriter(
+            is_agentless=config._llmobs_agentless_enabled,
+            interval=float(os.getenv("_DD_LLMOBS_WRITER_INTERVAL", 1.0)),
+            timeout=float(os.getenv("_DD_LLMOBS_WRITER_TIMEOUT", 5.0)),
+        )
```
Yun-Kim, line 86: I feel like this doesn't need its own function anymore, can we just directly start the `LLMObsSpanWriter` in `LLMObs.__init__()`?

ddtrace/llmobs/_writer.py
```diff
@@ -21,9 +22,12 @@
...
 from ddtrace.internal.writer import WriterClientBase
-from ddtrace.llmobs._constants import EVP_PROXY_AGENT_ENDPOINT
-from ddtrace.llmobs._constants import EVP_SUBDOMAIN_HEADER_NAME
-from ddtrace.llmobs._constants import EVP_SUBDOMAIN_HEADER_VALUE
+
+from ._constants import AGENTLESS_BASE_URL
+from ._constants import AGENTLESS_ENDPOINT
+from ._constants import EVP_PROXY_AGENT_ENDPOINT
+from ._constants import EVP_SUBDOMAIN_HEADER_NAME
+from ._constants import EVP_SUBDOMAIN_HEADER_VALUE
```
Yun-Kim, line 30: For consistency, let's keep imports absolute rather than relative.

ddtrace/llmobs/_writer.py
```diff
@@ -217,11 +204,15 @@ def __init__(self):
         super(LLMObsEventClient, self).__init__(encoder)
 
 
+class LLMObsEventAgentlessEventClient(LLMObsEventClient):
```
Yun-Kim, line 207: Can we rename this to `LLMObsAgentlessEventClient`? The extra `event` seems unnecessary

ddtrace/llmobs/_writer.py
```diff
@@ -217,11 +204,15 @@ def __init__(self):
...
+    ENDPOINT = AGENTLESS_ENDPOINT
+
+
 class LLMObsEventProxiedEventClient(LLMObsEventClient):
```
Yun-Kim, line 211: Can we rename this to `LLMObsAgentProxyEventClient`? The extra `event` seems unnecessary

ddtrace/llmobs/_writer.py
```diff
@@ -232,35 +223,41 @@ def __init__(
...
+            clients.append(LLMObsEventProxiedEventClient())
+            intake_url = agent.get_trace_url()
+            headers[EVP_SUBDOMAIN_HEADER_NAME] = EVP_SUBDOMAIN_HEADER_VALUE
+
+        super(LLMObsSpanWriter, self).__init__(
             intake_url=intake_url,
             clients=clients,
             processing_interval=interval,
             timeout=timeout,
             dogstatsd=dogstatsd,
```
Yun-Kim, line 247: Quick question - were the `Error submitting packet: %s, dropping the packet and closing the socket` errors resolved, or were the unrelated to this change?

releasenotes/notes/llmobs-agentless-span-retry-300e4a035ed3942f.yaml
```diff
@@ -0,0 +1,3 @@
+features:
+  - |
+    LLM Observability: Updates the agentless span writer to retry sending spans if the initial attempt fails.
```
Yun-Kim, line 3:
```suggestion
    LLM Observability: Adds retry logic to the agentless span writer to mitigate potential networking issues, like timeouts or dropped connections.
```

tests/llmobs/test_llmobs_span_writer.py
```diff
@@ -7,107 +7,149 @@
 from ddtrace.llmobs._writer import LLMObsSpanWriter
```
Yun-Kim, line 7: Could we rename this file to `test_llmobs_span_agentless_writer.py` for clarity?

Yun-Kim (commented):
Some small suggestions but otherwise looks great! Thanks for the cleanup 🙇

ddtrace/llmobs/_writer.py
tomshen: Those errors are still happening.

tomshen: The errors should be unrelated to this change. If I check out the main branch of this repo and run the test script, I still get the error. It's coming from the OpenAI integration.

Yun-Kim: Great, thanks for clarifying. I think I've found the crux of this issue (unrelated), will address this once we merge this refactor.

Yun-Kim (approved):

PR #10030, opened 2024-08-01
fix(asm): switch back to default if remote config stop sending the security rule file

christophe-papazian:
Security rule file is provided by ASM_DD through remote config. 
This PR fix an issue where the WAF could be disabled if remote config was deleting or deprecating the ASM_DD rule file. 

After that PR, the WAF will stay enabled with the default recommended rule file.

Also
- rename `rules` attribute to the more precise `rule_filename` in `AppSecSpanProcessor`
- add a unit test to ensure that we never send an empty rule list to the waf
- this will also be tested with the new system tests : https://github.com/DataDog/system-tests/pull/2839

APPSEC-54105

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/appsec/_remoteconfiguration.py
```diff
@@ -86,8 +86,9 @@ def _add_rules_to_list(features: Mapping[str, Any], feature: str, message: str,
     if rules is not None:
         try:
             if ruleset.get(feature) is None:
-                ruleset[feature] = []
-            ruleset[feature] += rules
+                ruleset[feature] = rules
+            else:
+                ruleset[feature] = ruleset[feature] + rules
```
gnufede, line 91: This would be equivalent I think, although maybe too condensed?
```suggestion
            ruleset.setdefault(feature, []).extend(rules)
```

gnufede (approved):

ddtrace/appsec/_remoteconfiguration.py
christophe-papazian: not equivalent, because the initial list would be modified, and I don't want that if feature is initialized from the default rule set. So this code was made to ensure immutability. We rebuild a list if we have to append something.

PR #9931, opened 2024-07-24
chore: remove setuptools dependency

brettlangdon:
This PR removes the need to have `setuptools` listed as a dependency for our package.

Our code does not actually need `setuptools` or `pkg_resources` for anything. The only places where it is used it is checking if the user is using setuptools (e.g. `ddtrace/sourcecode/setuptools_auto`), or in our test suite.

An added benefit here is for our lib-injection/OCI images this removed 7+mb per-python/arch combination by removing `setuptools` dependency (saves ~75mb or 20% savings).

[](https://datadoghq.atlassian.net/browse/APMON-1292)

## Checklist
- [x] PR author has checked that all the criteria below are met
- The PR description includes an overview of the change
- The PR description articulates the motivation for the change
- The change includes tests OR the PR description describes a testing strategy
- The PR description notes risks associated with the change, if any
- Newly-added code is easy to change
- The change follows the [library release note guidelines](https://ddtrace.readthedocs.io/en/stable/releasenotes.html)
- The change includes or references documentation updates if necessary
- Backport labels are set (if [applicable](https://ddtrace.readthedocs.io/en/latest/contributing.html#backporting))

## Reviewer Checklist
- [x] Reviewer has checked that all the criteria below are met 
- Title is accurate
- All changes are related to the pull request's stated goal
- Avoids breaking [API](https://ddtrace.readthedocs.io/en/stable/versioning.html#interfaces) changes
- Testing strategy adequately addresses listed risks
- Newly-added code is easy to change
- Release note makes sense to a user of the library
- If necessary, author has acknowledged and discussed the performance implications of this PR as reported in the benchmarks PR comment
- Backport labels are set in a manner that is consistent with the [release branch maintenance policy](https://ddtrace.readthedocs.io/en/latest/contributing.html#backporting)

PR #10035, opened 2024-08-01
chore: remove setuptools dependency (#9931) [backport 2.10]

emmettbutler:
This PR removes the need to have `setuptools` listed as a dependency for our package.

Our code does not actually need `setuptools` or `pkg_resources` for anything. The only places where it is used it is checking if the user is using setuptools (e.g. `ddtrace/sourcecode/setuptools_auto`), or in our test suite.

An added benefit here is for our lib-injection/OCI images this removed 7+mb per-python/arch combination by removing `setuptools` dependency (saves ~75mb or 20% savings).

[](https://datadoghq.atlassian.net/browse/APMON-1292)

- [x] checklist


brettlangdon (commented):
we'll need to add a release note.

emmettbutler:
@brettlangdon this is a backport change - maybe we should add the release note in a PR against `main`?

PR #10057, opened 2024-08-02
chore(ci_visibility): introduce pytest test visibility plugin v2 using external API

romainkomorndatadog:
This is the first step of the new "v2" version of the pytest ddtrace plugin, bringing a partial implementation that covers suite-level test visibility.

The new version of the plugin is unsupported, and is enabled via the `_DD_CIVISIBILITY_USE_PYTEST_V2` environment variable.

A notable change from the original implementation (that will become a breaking change upon release) is that module, suite, and test names will be derived from the pytest item's nodeid attribute, without inspection of the object's parent items, packages, etc. In most cases, the result should be the same as in the original implementation. As an example, the nodeid `tests/tracer/test_trace_utils.py::test_bad_http_code[234-False]` would become: `module=test.tracer`, `suite=test_trace_utils.py`, `test=test_bad_http_code[234-False]`.

Another breaking change is that class names will now automatically be included in the test name (whereas in the past the `--ddtrace-include-class-name` flag needed to be specified).

The testing strategy is to wait until the new plugin is feature complete before introducing a new job matching the original `pytest` suite, but with the plugin enabled via the environment variable.

There is no release note as this will remain unannounced until feature parity is reached with the initial version and substantial testing has been completed.  

## Checklist

## Reviewer Checklist
- Title is accurate


juan-fernandez (approved):

ddtrace/internal/ci_visibility/recorder.py
```diff
@@ -977,9 +977,6 @@ def _register_module_handlers():
 def _on_discover_suite(discover_args: CISuite.DiscoverArgs):
     log.debug("Handling discovery for suite args %s", discover_args)
     module = CIVisibility.get_module_by_id(discover_args.suite_id.parent_id)
-    if discover_args.suite_id in module.children:
```
vitor-de-araujo, line 980: @romainkomorndatadog Just a question: this affects both the current and the new version, right?

romainkomorndatadog: It does not, because the current version does not use this API.

I'm removing this because there's already another check in [CIVisibilityParentItem.add_child()](https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/internal/ci_visibility/api/ci_base.py#L492-L498) that does a similar thing (but only if `self._session_settings.reject_duplicates` ).

This change does mean we unnecessarily pay the cost penalty of `CIVisibilitySuite.__init__()`, but that's an optimization I'm willing to leave to later.

vitor-de-araujo (approved):

PR #10066, opened 2024-08-02
fix(profiling): stack_v2 properly tracks Threading

sanchda:
Looks like we dropped this part in the first cut for stack v2.  Adding it back in.

Basically, stack v2 needs to know when threads are created in order to be able to track them off-GIL.  In the future, we should probably harmonize on a common backplane for thread tracking, but for now... let's not.

## Checklist

## Reviewer Checklist
- Title is accurate


sanchda:
* **#10066** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-py/10066?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @sanchda and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

ddtrace/internal/datadog/profiling/stack_v2/src/sampler.cpp
```diff
@@ -56,17 +58,63 @@ Sampler::get()
...
+    pthread_atfork(nullptr, nullptr, _stack_v2_atfork_child);
 
     // Register our rendering callbacks with echion's Renderer singleton
     Renderer::get().set_renderer(renderer_ptr);
 }
 
+void
+Sampler::register_thread(uintptr_t id, uint64_t native_id, const char* name)
+{
+    const std::lock_guard<std::mutex> thread_info_guard{ thread_info_map_lock };
```
taegyunkim, line 93: Could you add a comment saying that `thread_info_map_lock` and `thread_info_map` are defined in 'echion/threads.h'

ddtrace/internal/datadog/profiling/stack_v2/src/sampler.cpp
```diff
@@ -56,17 +58,63 @@ Sampler::get()
...
+
+    auto it = thread_info_map.find(id);
+    if (it == thread_info_map.end()) {
+        try {
+            thread_info_map.emplace(id, std::make_unique<ThreadInfo>(id, native_id, name));
+        } catch (const ThreadInfo::Error& e) {
+            // ignore for now
```
taegyunkim, line 100: Would it benefit us if we print to stderr or log?

ddtrace/internal/datadog/profiling/stack_v2/src/sampler.cpp
```diff
@@ -56,17 +58,63 @@ Sampler::get()
...
+        }
+    } else {
+        try {
+            it->second = std::make_unique<ThreadInfo>(id, native_id, name);
+        } catch (const ThreadInfo::Error& e) {
+            // ignore
```
taegyunkim, line 106: Same here

ddtrace/profiling/collector/threading.py
```diff
@@ -37,3 +41,25 @@ def _set_original(
     ):
         # type: (...) -> None
         threading.Lock = value  # type: ignore[misc]
+
+
+# Also patch threading.Thread so echion can track thread lifetimes
```
taegyunkim, line 46: Basically we need to do what echion's `monkey/threading.py` does, right?
https://github.com/sanchda/echion/blob/main/echion/monkey/threading.py

And the reason we're not able to invoke echion's Python code directly is that we're using its C++ APIs and we don't depend on its Python APIs.

ddtrace/internal/datadog/profiling/stack_v2/src/sampler.cpp
```diff
@@ -56,17 +58,63 @@ Sampler::get()
...
+    // Run the atfork handler to ensure that we're tracking the correct process
+    _stack_v2_atfork_child();
+    pthread_atfork(nullptr, nullptr, _stack_v2_atfork_child);
 
     // Register our rendering callbacks with echion's Renderer singleton
     Renderer::get().set_renderer(renderer_ptr);
 }
 
+void
+Sampler::register_thread(uintptr_t id, uint64_t native_id, const char* name)
```
taegyunkim, line 91: When I look at echion's `track_thread`
https://github.com/sanchda/echion/blob/b2f2d49f2f5d5c4dd78d1d9b83280db8ac2949c0/echion/coremodule.cc#L278-L288

It updates thread info if it's already tracked. Don't we need to do the same?

taegyunkim (changes requested):

ddtrace/internal/datadog/profiling/stack_v2/src/sampler.cpp
sanchda: Sure thing

ddtrace/internal/datadog/profiling/stack_v2/src/sampler.cpp
sanchda: Honestly, I don't know.  There are a few moving parts.

1. I don't fully understand the failure surface here.  How might we fail to receive a thread ID which is also a valid pthread pointer?  I don't think this is likely--or really, possible--but if I'm wrong then these exceptions can be incredibly chatty.  The tracer goes to extreme lengths to avoid flooding stdio, and since we don't have a clear integration with the logger (so far!), we'd need to construct a new strategy for dealing with this.
2.  We can integrate with the logger more naturally downstream, but then we have a conditional in the context of a wrapper (for threading, no less), which is an uncomfortable log point for me.

Let me propose the following.
1. In the near term (this PR), we'll do what we do elsewhere in the repo, which is log to stderr once.  This will at least show that instrumentation _can_ fail and we need to be aware of it.
2. In the mid-term, I have a separate PR where I'm adding some counters to the system.  We should count the failure rate.
3. In the long-term, this is a good place to integrate with the logger probably.  I mean, it'd still be in the code path which is called by the stdlib wrapper (which I said before makes me uncomfortable), but at least it'd be in the called code instead of the top-level wrapper itself.

ddtrace/internal/datadog/profiling/stack_v2/src/sampler.cpp
sanchda: Unless I'm mistaken, I believe we do.
1. Check that the iterator is the end iterator.  If so, then we didn't find the thing, so add it.
2. Otherwise, the iterator found a matching element, so just update the element (`it->second = std::make_unique<...>(...)...`)

ddtrace/internal/datadog/profiling/stack_v2/src/sampler.cpp
sanchda: Fixed in c83e1e35f

ddtrace/internal/datadog/profiling/stack_v2/src/sampler.cpp
sanchda: I did proposal 1 in c83e1e35f

ddtrace/internal/datadog/profiling/stack_v2/src/sampler.cpp
sanchda: See the other comment; added one-time logging in c83e1e35f

ddtrace/internal/datadog/profiling/stack_v2/src/sampler.cpp
taegyunkim: Oh, I have neglected that part. Thanks for pointing it out.

taegyunkim:
Thank you for fixing this!

ddtrace/internal/datadog/profiling/stack_v2/src/sampler.cpp
taegyunkim: I like your proposal for using counters.

PR #10068, opened 2024-08-04
feat: add aiokafka support

ls-pieter-vanisacker:
This adds support for Kafka distributed tracing using the aiokafka library.
It also adds supports for DSM when using aiokafka.

There's a bunch of tests included which cover the basic usage.

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/contrib/aiokafka/__init__.py
```diff
@@ -0,0 +1,39 @@
+"""
```
emmettbutler, line 1: This docstring is default and should be replaced with one that applies to aiokafka.

ddtrace/contrib/aiokafka/patch.py
```diff
@@ -0,0 +1,248 @@
...
+    "kafka",
+    dict(
+        _default_service=schematize_service_name("kafka"),
+        distributed_tracing_enabled=asbool(os.getenv("DD_KAFKA_PROPAGATION_ENABLED", default=False)),
+    ),
+)
+
+
+def get_version():
+    # type: () -> str
```
emmettbutler, line 41:
```suggestion
def get_version() -> str:
```

ddtrace/contrib/aiokafka/patch.py
```diff
@@ -0,0 +1,248 @@
...
+    return getattr(aiokafka, "__version__", "")
+
+
+class TracedAIOKafkaProducerMixin:
+    def __init__(self, *args, **kwargs):
+        super(TracedAIOKafkaProducerMixin, self).__init__(*args, **kwargs)
+        self._dd_bootstrap_servers = kwargs.get("bootstrap_servers")
+
+
+class TracedAIOKafkaConsumerMixin:
```
emmettbutler, line 51: This pattern looks like it was copied from the existing Kafka integration. It may work, but it's more complex than necessary. The confluent-kafka integration is written this way to work around the fact that native code can't be monkeypatched. Because aiokafka is pure Python, it can be monkeypatched with wrapt. I'd recommend that simpler approach here.

ddtrace/contrib/aiokafka/patch.py
```diff
@@ -0,0 +1,248 @@
...
+    if not pin or not pin.enabled():
+        return await func(*args, **kwargs)
+
+    topic = get_argument_value(args, kwargs, 0, "topic")
+    value = get_argument_value(args, kwargs, 1, "value", True) or None
+    key = get_argument_value(args, kwargs, 2, "key", True) or ""
+    partition = get_argument_value(args, kwargs, 3, "partition", True) or None
+    headers = get_argument_value(args, kwargs, 5, "headers", True) or []
+
+    with pin.tracer.trace(
```
emmettbutler, line 112: Code in contrib/ shouldn't directly call `trace`, instead using `core.context_with_data` paired with an event handler implemented in `_trace/trace_handlers`.

ddtrace/contrib/aiokafka/patch.py
```diff
@@ -0,0 +1,248 @@
...
+        schematize_messaging_operation(kafkax.PRODUCE, provider="kafka", direction=SpanDirection.OUTBOUND),
+        service=trace_utils.ext_service(pin, config.kafka),
+        span_type=SpanTypes.WORKER,
+    ) as span:
+        core.dispatch("aiokafka.produce.start", (instance, topic, value, key, headers, span))
+        span.set_tag_str(MESSAGING_SYSTEM, kafkax.SERVICE)
```
emmettbutler, line 118: The logic in this block is similar enough to the existing logic for tagging Kafka spans that I recommend sharing existing code via a new utils function rather than copypasting.

ddtrace/contrib/aiokafka/patch.py
```diff
@@ -0,0 +1,248 @@
...
+            for record in records:
+                _instrument_message(record, pin, start_ns, instance, err)
+    return result
+
+
+def _instrument_message(message, pin, start_ns, instance, err):
+    ctx = None
+    if message is not None and config.kafka.distributed_tracing_enabled and message.headers:
+        ctx = Propagator.extract(dict(message.headers))
+    with pin.tracer.start_span(
```
emmettbutler, line 200: Replacing this with `core.context_with_data` would clarify the separation of concerns between the Tracing product and the Kafka integration.

emmettbutler (changes requested):
Great work! I suggested a few structural changes. You're certainly on the right track. Let me know if you'd like more detailed assistance.

ddtrace/contrib/aiokafka/patch.py
ls-pieter-vanisacker: Fixed

ddtrace/contrib/aiokafka/patch.py
ls-pieter-vanisacker: Ok, so there's one event being fired by the wrapped functions. And that event gets dealt with by:
- `_trace/tracehandlers` for normal tracing functionality
- `internal/datastreams/aiokafka.py` for DSM
Is that the correct approach to go about?

ls-pieter-vanisacker: Is there any existing simple integration that implements that pattern which I could use as a base?

ddtrace/contrib/aiokafka/patch.py
ls-pieter-vanisacker: I would first like to make sure the logic fully works (and I also understand it) with the `core.context_with_data`. After that I'll have more knowledge of what to extract.

ddtrace/contrib/aiokafka/patch.py
ls-pieter-vanisacker: @emmettbutler I've reworked a bit of the code to make use of the `core.context_with_data`, see the last commit.
Please let me know if this aligning more with how it should be done.

I've also disabled tracing on some functions for now until the basics are sorted out.
And I disabled the test for distributed tracing and DSM as well for now.

ddtrace/contrib/aiokafka/patch.py
```diff
@@ -0,0 +1,175 @@
...
+        parent=None,
+        span_name=schematize_messaging_operation(PRODUCE, provider="kafka", direction=SpanDirection.OUTBOUND),
+        span_type=SpanTypes.WORKER,
+        service=trace_utils.ext_service(pin, config.aiokafka),
+        tags=create_send_span_tags(instance, args, kwargs),
+        pin=pin,
+    ) as ctx:
+        result = await func(*args, **kwargs)
+        core.dispatch("aiokafka.send_and_wait.post", (ctx,))
+        return result
```
emmettbutler, line 116: This pattern closes the span when the `with` block exits, without the need for a `dispatch` call.

```suggestion
    ) as ctx, ctx[ctx["call_key"]]:
        result = await func(*args, **kwargs)
        return result
```

ddtrace/_trace/trace_handlers.py
```diff
@@ -798,6 +803,7 @@ def listen():
     core.on("botocore.kinesis.GetRecords.post", _on_botocore_kinesis_getrecords_post)
     core.on("redis.async_command.post", _on_redis_command_post)
     core.on("redis.command.post", _on_redis_command_post)
+    core.on("aiokafka.send_and_wait.post", _on_aiokafka_send_and_wait_post)
```
emmettbutler, line 806:
```suggestion
```

ddtrace/_trace/trace_handlers.py
```diff
@@ -746,6 +746,11 @@ def _on_redis_command_post(ctx: core.ExecutionContext, rowcount):
         ctx[ctx["call_key"]].set_metric(db.ROWCOUNT, rowcount)
 
 
+def _on_aiokafka_send_and_wait_post(ctx: core.ExecutionContext):
+    span = ctx["call"]
+    span.finish()
+
+
```
emmettbutler, line 753:
```suggestion
```

ddtrace/contrib/aiokafka/patch.py
```diff
@@ -0,0 +1,175 @@
...
+
+    with core.context_with_data(
+        "aiokafka.getone",
+        parent=None,
+        span_name=schematize_messaging_operation(CONSUME, provider="kafka", direction=SpanDirection.INBOUND),
+        span_type=SpanTypes.WORKER,
+        service=trace_utils.ext_service(pin, config.aiokafka),
+        tags=create_get_span_tags(instance, args, kwargs),
+        pin=pin,
+    ) as ctx:
```
emmettbutler, line 143:
```suggestion
    ) as ctx, ctx[ctx["call_key"]] as call:
```

ddtrace/contrib/aiokafka/patch.py
```diff
@@ -0,0 +1,175 @@
...
+        err = None
+        result = None
+        try:
+            result = await func(*args, **kwargs)
+        except Exception as e:
+            err = e
+            raise err
+        finally:
+            _instrument_message(ctx["call"], result, err)
+            ctx["call"].finish()
```
emmettbutler, line 153:
```suggestion
            _instrument_message(call, result, err)
```

ddtrace/contrib/aiokafka/patch.py
```diff
@@ -0,0 +1,175 @@
...
+        return result
+
+
+def _instrument_message(span, message, err):
```
emmettbutler, line 157: Because it has directly to do with spans, this logic would be more at home in `ddtrace/_trace`, probably as part of an event handler listened for with `core.on`.

emmettbutler (changes requested):

ddtrace/contrib/aiokafka/patch.py
ls-pieter-vanisacker: @emmettbutler I've reworked the code to make use of the event handler listeners.

It does feel like there's no clear split between the integration and tracing code though.

Still to be done is supporting DSM and wrapping other methods for sending/receiving data.

ls-pieter-vanisacker: DSM is part of https://github.com/DataDog/dd-trace-py/pull/10068/commits/d8774d8a9e5cc1fa9ef1e905c392dfd9d2cd8af9

ddtrace/contrib/aiokafka/patch.py
ls-pieter-vanisacker: Fixed

emmettbutler:
@ls-pieter-vanisacker both of those tags have to do with "analyzed spans", a deprecated sampling feature. I think it's ok if you leave them out of your implementation.

ls-pieter-vanisacker:
@emmettbutler Would you mind having another look at the PR, are you seeing any things missing to be able to merge it?

emmettbutler (approved):

ls-pieter-vanisacker:
@emmettbutler can you share a bit more info on what is needed to get this merged? Is there any DD internal approval/test processes that this still needs to run through?

emmettbutler:
@ls-pieter-vanisacker there's an approval still pending from another datadog team (cc @tabgok). There are also some tests that need to be manually triggered. I'll see if I can do that.

ls-pieter-vanisacker:
@emmettbutler any news on this?

ddtrace/contrib/aiokafka/patch.py
```diff
@@ -0,0 +1,258 @@
...
+async def traced_getone(func, instance, args, kwargs):
+    pin = Pin.get_from(instance)
+    if not pin or not pin.enabled():
+        return await func(*args, **kwargs)
+
+    err = None
+    result = None
+    child_of = None
+    try:
+        result = await func(*args, **kwargs)
```
wconti27, line 197: Since our tracing setup does not allow us to change a span's parent, we have traditionally called the wrapped consume function before starting the consume span. However, we also need to record the start time of the original wrapped call, then reset the span's start time to be before the original consume. Please see how this is done for our confluent-kafka integration. [Here](https://github.com/DataDog/dd-trace-py/blob/e7786202b22cc7bfea2694dc7fb31967f03c0d29/ddtrace/contrib/internal/kafka/patch.py#L212) we capture the start time, then overwrite the spans start time [here](https://github.com/DataDog/dd-trace-py/blob/e7786202b22cc7bfea2694dc7fb31967f03c0d29/ddtrace/contrib/internal/kafka/patch.py#L248).

ddtrace/contrib/aiokafka/patch.py
```diff
@@ -0,0 +1,258 @@
...
+                    span_name=schematize_messaging_operation(
+                        CONSUME, provider="kafka", direction=SpanDirection.INBOUND
+                    ),
+                    span_type=SpanTypes.WORKER,
+                    service=trace_utils.ext_service(pin, config.aiokafka),
+                    call_key="instrumented_getmany",
+                    distributed_context=child_of,
+                    tags=create_get_span_tags(instance),
+                    pin=pin,
+                ) as ctx, ctx[ctx["call_key"]] as span:
```
wconti27, line 250: Unfortunately, our current trace model does not support many-to-many parent-child propagation. Generally, for batch message operations we have taken the following approach:
- If the batch operation calls `aiokafka.getone` n times for n records, then we will want to create a span for each of these calls. Many times this is not the case, and the batch operation will actually be 1 operation, returning N records. For each of these n spans, we set a span link to the original producer.  
- In the case of 1 operation returning N records, we should create 1 span for the batch operation, similar to `aiokafka.getone`. Again, we will not support parent-child propagation. Instead, we will use span links and set n links on the single span, each to their respective producer. 

Again, thank you for the contribution, and please don't hesitate to reach out for help. Also, if you would like, I could make direct changes to your branch to speed up the process. 

(FYI, span links are very new to ddtrace, and there may be limited examples to follow. We were planning to begin using them for auto-instrumentation this quarter)

wconti27 (changes requested):
See comments above.

ddtrace/contrib/aiokafka/patch.py
ls-pieter-vanisacker: @wconti27 feel free to many any changes as you see fit. Especially for the span link part.

sapirm-eleos:
Hey @wconti27 / @emmettbutler , any plans on getting this or some other aiokafka tracing support merged?

dubloom:
Taking over the work. A new PR based on this one will be opened soon.

menishak:
Thank you, @dubloom!
Do you have any estimate on when this PR might be merged?
It would really help us a lot! 🙏

dubloom:
Hello @menishak, I'm actively working on it. I can give you an estimate by early next week.

dubloom:
Hello @menishak.
I was able to make progress on the PR. I still miss some DSM support but I think I should be able to open the PR next week !

bm1549:
Closing this as it's been superseded by #15111 and #15112

PR #10084, opened 2024-08-05
fix(llmobs): do not emit integration metrics in agentless mode

Yun-Kim:
In #9912, we changed `DD_LLMOBS_AGENTLESS_ENABLED` to be stored in the global config setting `config._llmobs_agentless_enabled`, but forgot to also reference the config setting in the `OpenAIIntegration.metrics_enabled()` check. Instead, currently the openai integration only checks the env var meaning metrics will still be emitted in agentless mode if a user configures using `config._llmobs_agentless_enabled` or `LLMObs.enable(agentless_enabled=True)`.

This PR fixes that to make metrics emitting check the central global config setting. 

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/llmobs/_integrations/base.py
```diff
@@ -61,7 +61,7 @@ def __init__(self, integration_config: IntegrationConfig) -> None:
     def metrics_enabled(self) -> bool:
         """Return whether submitting metrics is enabled for this integration, or global config if not set."""
         env_metrics_enabled = asbool(os.getenv("DD_{}_METRICS_ENABLED".format(self._integration_name.upper())))
-        if not env_metrics_enabled and asbool(os.getenv("DD_LLMOBS_AGENTLESS_ENABLED")):
+        if not env_metrics_enabled and config._llmobs_agentless_enabled:
```
mabdinur, line 64: Is checking `config._llmobs_agentless_enabled` sufficient? Is is possible to call `metrics_enabled()` before `LLMObs.enable(agentless=True)`

mabdinur (approved):

ddtrace/llmobs/_integrations/base.py
Yun-Kim: Yes this should be sufficient, `LLMObs.enable()` is called at app startup if a user specifies `ddtrace-run`, otherwise our official guidance is to call `LLMObs.enable()` at the top of their entrypoint file. We do not officially support use cases where users may call `LLMObs.enable()` in the middle of their app execution.

lievan (approved):

PR #10085, opened 2024-08-05
chore(w3c): do not set _dd.parent_id to 16 zeros

mabdinur:
## Motivation

- Simplify the extraction logic for w3c tracecontext headers. 
- Avoids unnecessary tags to the Datadog agent.

## Description

Currently when w3c tracestate headers are extracted by the ddtrace HttpPropagator:
1) The `p` value in the tracestate is stored in the `_dd.parent_id` propagation tag
2) If the `p` value is not found in the tracestate then `_dd.parent_id` is set to `0000000000000000`.

The second condition is no longer required. Datadog internal services no longer handle the case where `Span.meta["_dd.parent_id"] == 0000000000000000`. `0000000000000000` is treated as invalid value and ignored. We should avoid setting this value in client libraries.

## Checklist

## Reviewer Checklist
- Title is accurate


erikayasuda (approved):

ddtrace/propagation/http.py
```diff
@@ -945,7 +944,7 @@ def _resolve_contexts(contexts, styles_w_ctx, normalized_headers):
                     dd_context = None
                     if PROPAGATION_STYLE_DATADOG in styles_w_ctx:
                         dd_context = contexts[styles_w_ctx.index(PROPAGATION_STYLE_DATADOG)]
-                    if context._meta.get(LAST_DD_PARENT_ID_KEY, DEFAULT_LAST_PARENT_ID) != DEFAULT_LAST_PARENT_ID:
+                    if LAST_DD_PARENT_ID_KEY in context._meta:
```
zacharycmontoya, line 947: This change will result in the last parent id being set to "0000000000000000" if it was somehow in the tracestate. Is that okay?

mabdinur: There is no valid scenario where a tracer propagates a tracestate header with `p:0000000000000000`. If a span is active `p:<span_id_in_hex>`. If a span is not active then `p` will not be added to the tracestate. `0000000000000000` is an internal state that occurs when a tracer supports W3C Phase 2 but did not extract a `p` value from the tracestate.

However if a tracer happens to generate an invalid tracestate we do not want to validate it here. Tracers should simply forward `p` value received in http headers to the backend via the `_dd.parent_id` tag.

zacharycmontoya: Sounds good to me, thanks for confirming

zacharycmontoya (approved):

PR #10093, opened 2024-08-05
feat(llmobs): flush buffer before appending event that would exceed payload limit

tomshen:
The Python SDK currently adds all spans in the buffer to the payload per periodic write(). However for extremely large spans, these may result in exceeding the 5 MB payload limit on the EVP public intake side.

This PR adds logic to check for the case where adding an event would cause the buffer to exceed the 5 MB payload size limit, and flushes the queue before adding the event to the buffer.

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/llmobs/_constants.py
```diff
@@ -34,6 +34,7 @@
 EVP_PROXY_AGENT_ENDPOINT = "{}/api/v2/llmobs".format(EVP_PROXY_AGENT_BASE_PATH)
 EVP_SUBDOMAIN_HEADER_NAME = "X-Datadog-EVP-Subdomain"
 EVP_SUBDOMAIN_HEADER_VALUE = "llmobs-intake"
+EVP_PAYLOAD_SIZE_LIMIT = 5_000_000  # 5MB
```
Yun-Kim, line 37: Correct me if I'm wrong (I'm not too familiar with EVP's exact limits) but shouldn't it be `5 << 20` (5 * 2^20) to get 5 MB? Or should we just stick with 5000000 (decimal system) to be safe?

ddtrace/llmobs/_writer.py
```diff
@@ -259,6 +262,12 @@ def stop(self, timeout=None):
             super(LLMObsSpanWriter, self).stop(timeout=timeout)
 
     def enqueue(self, event: LLMObsSpanEvent) -> None:
+        event_size = len(json.dumps(event))
+        for client in self._clients:
```
Yun-Kim, line 266: I assume this is just defensive in the case we have multiple clients, is that right?

ddtrace/llmobs/_writer.py
```diff
@@ -259,6 +262,12 @@ def stop(self, timeout=None):
...
+            if isinstance(client, LLMObsEventClient) and isinstance(client.encoder, LLMObsSpanEncoder):
+                if (client.encoder.buffer_size + event_size) > EVP_PAYLOAD_SIZE_LIMIT:
```
Yun-Kim, line 268: Do we need to use the encoder buffer lock while we check the buffer size?

tests/llmobs/test_llmobs_span_agentless_writer.py
```diff
@@ -31,6 +32,16 @@ def test_buffer_limit(mock_writer_logs, mock_http_writer_send_payload_response):
         )
 
 
+def test_flush_queue_when_event_cause_queue_to_exceed_payload_limit(
```
Yun-Kim, line 35: Can we get the same test for the agent proxy writer?

ddtrace/llmobs/_constants.py
tomshen: I think we can update to 5 << 20. The actual limit is 5.1MB, so this still gives us some extra space.

ddtrace/llmobs/_writer.py
tomshen: yes, this is just to be defensive

ddtrace/llmobs/_constants.py
```diff
@@ -34,6 +34,7 @@
 EVP_PROXY_AGENT_ENDPOINT = "{}/api/v2/llmobs".format(EVP_PROXY_AGENT_BASE_PATH)
 EVP_SUBDOMAIN_HEADER_NAME = "X-Datadog-EVP-Subdomain"
 EVP_SUBDOMAIN_HEADER_VALUE = "llmobs-intake"
+EVP_PAYLOAD_SIZE_LIMIT = 5 * (2**20)  # 5MB
```
Yun-Kim, line 37:
```suggestion
EVP_PAYLOAD_SIZE_LIMIT = 5 << 20  # 5MB
```

ddtrace/llmobs/_writer.py
```diff
@@ -170,6 +171,7 @@ def __len__(self):
     def _init_buffer(self):
         with self._lock:
             self._buffer = []
+            self.buffer_size = 0
```
Yun-Kim, line 174: Dumb question - do we need to reset the `buffer_size` once we flush?

tests/llmobs/test_llmobs_span_agentless_writer.py
```diff
@@ -31,6 +32,16 @@ def test_buffer_limit(mock_writer_logs, mock_http_writer_send_payload_response):
...
+    mock_writer_logs, mock_http_writer_send_payload_response
+):
+    with override_global_config(dict(_dd_api_key="foobar.baz", _dd_site=DATADOG_SITE)):
+        llmobs_span_writer = LLMObsSpanWriter(is_agentless=True, interval=1000, timeout=1)
+        llmobs_span_writer.enqueue(_large_event())
+        llmobs_span_writer.enqueue(_large_event())
+        mock_writer_logs.debug.assert_has_calls([mock.call("encode %d LLMObs span events to be sent", 1)])
```
Yun-Kim, line 42: Can we add an assertion that the debug log `"flushing queue because queuing next event will exceed EVP payload limit"` was called?

tests/llmobs/test_llmobs_span_agent_writer.py
```diff
@@ -26,6 +27,15 @@ def test_buffer_limit(mock_writer_logs, mock_http_writer_send_payload_response):
     )
 
 
+def test_flush_queue_when_event_cause_queue_to_exceed_payload_limit(
+    mock_writer_logs, mock_http_writer_send_payload_response
+):
+    llmobs_span_writer = LLMObsSpanWriter(is_agentless=False, interval=1000, timeout=1)
+    llmobs_span_writer.enqueue(_large_event())
+    llmobs_span_writer.enqueue(_large_event())
+    mock_writer_logs.debug.assert_has_calls([mock.call("encode %d LLMObs span events to be sent", 1)])
```
Yun-Kim, line 36: Can we add an assertion that the debug log `"flushing queue because queuing next event will exceed EVP payload limit"` was called?

Yun-Kim (commented):
Last couple comments, otherwise good to approve once addressed!

ddtrace/llmobs/_writer.py
tomshen: we call `_init_buffer` in `encode` (which is called when we flush)

Yun-Kim (approved):

PR #10094, opened 2024-08-05
chore: remove crash-by-design behaviors

emmettbutler:
This change satisfies the Single-Step Instrumentation General Availability requirement "[Python] No crash-by-design behaviours" by removing many `raise` statements from the library's code.

The choice of which statements to remove was based on a human audit of the results of `grep '\braise\b' -R ddtrace`. The `raise` statements that were left unchanged fall into one of three categories:

* reachable only via programming errors
* contained in non-tracing Product code
* re-raises of exceptions originating in integrated libraries

`raise`s not in these categories are generally replaced with `log.warning` paired with an immediate return from the containing function when possible. In some cases `print` is used instead of `log.warning` due to the ddtrace logger not yet being available from the typical bootstrapping flow.

This change also includes a bunch of conversions of type comments to Py3 type annotations. I updated these as I came across them in the files outputted by my `grep` command.

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/internal/sma.py
```diff
@@ -55,8 +46,7 @@ def set(self, count, total):
         :param total: The total quantity of the next bucket.
         :type total: :obj:`int`
         """
-        if count > total:
-            raise ValueError
+        count = min(count, total)
```
erikayasuda, line 49: Does this do the same thing as the `if` statement? If I have:
```
self.size = 2
self.counts = [1, 2, 3, 4]
self.totals = [1, 3, 5, 7]
```
Then I would need to call:
```
my_sma.set(5, 9) # correct
```
But this is checking for if I'm incorrect like this?
```
my_sma.set(5, 4) # incorrect
```
And do this (in essence)?
```
my_sma.set(4, 4)
```


(I think my understanding of this SMA implementation is not super clear honestly, so sorry if this is a stupid question 😅)

ddtrace/internal/writer/writer.py
```diff
@@ -276,16 +265,14 @@ def _put(self, data, headers, client, no_trace):
...
+    def _get_finalized_headers(self, count: int, client: WriterClientBase) -> dict:
         headers = self._headers.copy()
         headers.update({"Content-Type": client.encoder.content_type})  # type: ignore[attr-defined]
         if hasattr(client, "_headers"):
             headers.update(client._headers)
         return headers
 
-    def _send_payload(self, payload, count, client):
-        # type: (...) -> Response
+    def _send_payload(self, payload, count, client) -> Response:
```
erikayasuda, line 275: Should we type hint these too?
```suggestion
    def _send_payload(self, payload: bytes, count: int, client: WriterClientBase) -> Response:
```

ddtrace/internal/writer/writer.py
```diff
@@ -496,13 +479,14 @@ def __init__(
...
-                "Unsupported api version: '%s'. The supported versions are: %r"
-                % (self._api_version, ", ".join(sorted(WRITER_CLIENTS.keys())))
+        if self._api_version not in WRITER_CLIENTS:
+            log.warning(
+                "Unsupported api version: '%s'. The supported versions are: %r",
+                self._api_version,
+                ", ".join(sorted(WRITER_CLIENTS.keys())),
             )
+            self._api_version = sorted(WRITER_CLIENTS.keys())[-1]
+        client = WRITER_CLIENTS[self._api_version](buffer_size, max_payload_size)
```
erikayasuda, line 489: Should we still wrap this with a `try/except`?

ddtrace/internal/writer/writer.py
```diff
@@ -576,25 +560,21 @@ def _downgrade(self, payload, response, client):
             # These endpoints share the same encoding, so we can try sending the
             # same payload over the downgraded endpoint.
             return payload
-        raise ValueError()
 
-    def _send_payload(self, payload, count, client):
-        # type: (...) -> Response
+    def _send_payload(self, payload, count, client) -> Response:
```
erikayasuda, line 564: Type hint? I _think_ it should also be:
```suggestion
    def _send_payload(self, payload: bytes, count: int, client: WriterClientBase) -> Response:
```

tests/telemetry/test_telemetry.py
```diff
@@ -221,27 +221,6 @@ def process_trace(self, trace):
     ]["message"]
 
 
-def test_app_started_error_unhandled_tracer_exception(test_agent_session, run_python_code_in_subprocess):
```
erikayasuda, line 224: Does this code path not trigger the sampling rule parser error still?

ddtrace/_trace/_span_link.py
```diff
@@ -78,7 +79,7 @@ def kind(self, value):
 
     def _drop_attribute(self, key):
         if key not in self.attributes:
-            raise ValueError(f"Invalid key: {key}")
+            raise KeyError(f"Invalid key: {key}")
```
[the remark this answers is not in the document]
emmettbutler: It's only ever used in tests.

ddtrace/contrib/internal/aws_lambda/patch.py
```diff
@@ -118,8 +118,6 @@ def __call__(self, func, args, kwargs):
         try:
             self.response = self.func(*args, **kwargs)
             return self.response
-        except Exception:
-            raise
```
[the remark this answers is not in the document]
emmettbutler: I don't think this change alters the code's behavior at all, so whatever it was logging before it will still log after the change.

ddtrace/internal/sma.py
emmettbutler: Seems right to me. The `min` call in this change says "if count is larger than total, instead of raising an error, snap the count to the total".

erikayasuda (approved):

ddtrace/internal/writer/writer.py
emmettbutler: I don't think it's necessary because we are certain that `_api_version` is a valid key in `WRITER_CLIENTS` at this point.

emmettbutler:
@mabdinur @ZStriker19 could I have your review of this change?

emmettbutler:
@wconti27 @quinna-h @rachelyangdog could one of you please take a look at this change?

ddtrace/internal/sma.py
```diff
@@ -15,16 +15,12 @@ class SimpleMovingAverage(object):
...
+    def __init__(self, size: int) -> None:
         """
-        Constructor for SimpleMovingAverage.
-
         :param size: The size of the window to calculate the moving average.
         :type size: :obj:`int`
         """
-        if size < 1:
-            raise ValueError
+        size = max(1, size)
```
ZStriker19, line 23: Would we ever want to know if size < 1? Or is it fine to just always set it to a minimum of 1 even if something weird is happening and never know about it?

ZStriker19 (approved):
left a couple of questions but looks good.

ddtrace/internal/sma.py
emmettbutler: Given that the prior behavior was a crash, I don't really think so.

ddtrace/internal/remoteconfig/client.py
```diff
@@ -325,8 +317,7 @@ def is_subscriber_running(self, pubsub_to_check):
     def reset_products(self):
```
quinna-h, line 317: nit: do we also want return type here? Or are we only adding type annotations for the methods that previously had type comments

quinna-h (approved):

ddtrace/internal/remoteconfig/client.py
emmettbutler: A return type specification on this function would be ideal, but in this change I only converted existing type comments to annotations rather than adding new annotations.

PR #10102, opened 2024-08-06
chore(tracing): removes DD_PRIORITY_SAMPLING config

mabdinur:
## Background

In `ddtrace<1.19.0` setting `DD_PRIORITY_SAMPLING` to `False` would disable sampling rates set by the datadog agent. This was done by downgrading the trace agent endpoint from `v0.4`/`v0.5` to the legacy `v0.3` endpoint ([code](https://github.com/DataDog/dd-trace-py/blob/v1.18.8/ddtrace/internal/writer/writer.py#L527)).

In `ddtrace==1.19.0` tracer sampling was simplified and tracing components began to ignore the ``DD_PRIORITY_SAMPLING`` environment variable ([commit](https://github.com/DataDog/dd-trace-py/commit/828733bd9af6f83612cda9a2a5b1ff0aacd7cefb)). With this change setting `DD_PRIORITY_SAMPLING=False` no longer downgraded the trace_api to v0.3.

## Description

In an ideal world ``DD_PRIORITY_SAMPLING`` would have been deprecated in `v1.19.0` and then removed in `v2.0.0` (instead of just being ignored). However since this configuration has essentially done nothing for all `ddtrace~=2.0` releases it should be safe to remove. 

``DD_PRIORITY_SAMPLING`` is not documented in [python library docs](https://ddtrace.readthedocs.io/en/stable/configuration.html#configuration) or [datadog public docs](https://docs.datadoghq.com/tracing/trace_collection/library_config/python). Since this configuration is undocumented and is not used we can safely remove it.

## TODO

With this change we no longer need to support `v0.3` trace encoders and sending traces to the `v0.3/traces` agent endpoint.

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/internal/writer/writer.py
```diff
@@ -451,7 +451,6 @@ class AgentWriter(HTTPWriter):
     def __init__(
         self,
         agent_url,  # type: str
-        priority_sampling=False,  # type: bool
```
mabdinur, line 454: The `priority_sampling` parameter is never used. Since the AgentWriter is an internal component, it is safe to remove this argument.

ddtrace/internal/utils/http.py
```diff
@@ -266,10 +266,7 @@ def get_json(self):
                 # This typically happens when using a priority-sampling enabled
                 # library with an outdated agent. It still works, but priority sampling
                 # will probably send too many traces, so the next step is to upgrade agent.
-                log.debug(
-                    "Cannot parse Datadog Agent response. "
-                    "This occurs because Datadog agent is out of date or DATADOG_PRIORITY_SAMPLING=false is set"
-                )
+                log.debug("Cannot parse Datadog Agent response. " "This occurs because Datadog agent is out of date")
```
brettlangdon, line 269:
```suggestion
                log.debug("Cannot parse Datadog Agent response. This occurs because Datadog agent is out of date")
```

releasenotes/notes/removes-dd-prioritiy-sampling-config-5d07f4aedbfa4dc4.yaml
```diff
@@ -0,0 +1,4 @@
+---
+other:
+  - |
+    tracing: Removes the `DD_PRIORITY_SAMPLING` configuration option. This option was deprecated in v1.19.0.
```
brettlangdon, line 4:
```suggestion
    tracing: Removes the ``DD_PRIORITY_SAMPLING`` configuration option. This option was deprecated in v1.19.0.
```

brettlangdon: Do we instead want to state this as a deprecation and then state it already does nothing?

The more important thing to document is that the `priority_sampling=` arg to `Tracer.configure()` is being removed.

tests/integration/test_integration.py
```diff
@@ -28,7 +28,7 @@ def test_configure_keeps_api_hostname_and_port():
     assert tracer._writer.agent_url == "http://localhost:{}".format("9126" if AGENT_VERSION == "testagent" else "8126")
     tracer.configure(hostname="127.0.0.1", port=8127)
     assert tracer._writer.agent_url == "http://127.0.0.1:8127"
-    tracer.configure(priority_sampling=True)
+    tracer.configure()
```
brettlangdon, line 31: I think we need to pass in something that will reconfigure the writer to ensure it keeps the same url.

if we don't modify any arguments that would cause a writer recreation, then this test won't validate the expected behavior.

tests/tracer/test_writer.py
```diff
@@ -752,7 +752,7 @@ def test_writer_recreate_keeps_headers():
 
 
 @pytest.mark.parametrize(
-    "sys_platform, api_version, ddtrace_api_version, priority_sampling, raises_error, expected",
+    "sys_platform, api_version, ddtrace_api_version, _priority_sampling, raises_error, expected",
```
brettlangdon, line 755: why not remove it from here?

tests/tracer/test_writer.py
```diff
@@ -804,7 +804,14 @@ def test_writer_recreate_keeps_headers():
 )
 @pytest.mark.parametrize("writer_class", (AgentWriter,))
 def test_writer_api_version_selection(
-    sys_platform, api_version, ddtrace_api_version, priority_sampling, raises_error, expected, monkeypatch, writer_class
+    sys_platform,
+    api_version,
+    ddtrace_api_version,
+    _priority_sampling,
```
brettlangdon, line 810:
```suggestion
```

tests/tracer/test_writer.py
mabdinur: I was being lazy 😅. I will remove it

tests/integration/test_integration.py
mabdinur: Actually passing the priority_sampling argument does not trigger writer creation. This test has been broken for a while. I'll update it to use `appsec_enabled=True`. That should do the trick.

brettlangdon (approved):

rachelyangdog (approved):

PR #10103, opened 2024-08-06
chore(tracing): fix 128 bit trace id log injection

wconti27:
Internal API used by Logs Explorer page: `/api/ui/apm/has_traces/<trace-id>` only supports base10 64 bit trace ids and 128 bit trace ids formatted as 32 character length hex string. This PR fixes log injection to only inject these types. Previously, we were getting errors due to 128bit base10 integers being injected into logs, which resulted in "No traces found" error messages in Logs Explorer even though the trace exists. API Docs for endpoint found [here](https://docs.datadoghq.com/tracing/other_telemetry/connect_logs_and_traces/opentelemetry/?tab=python).

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/contrib/trace_utils.py
```diff
@@ -682,3 +683,17 @@ def _convert_to_string(attr):
         else:
             return ensure_text(attr)
     return attr
+
+
+def _get_trace_details_for_log_injection(span):
```
mabdinur, line 688: Can we add typing here

ddtrace/contrib/trace_utils.py
```diff
@@ -682,3 +683,17 @@ def _convert_to_string(attr):
...
+    # Note: logs only supports 32 length hex representation of trace_id for 128 bit ids, no base10 int support
+    trace_id = None
+    span_id = None
+    if span:
+        span_id = span.span_id
+        trace_id = span.trace_id
+        if trace_id and isinstance(trace_id, int) and trace_id > MAX_UINT_64BITS:
+            trace_id = "{:032x}".format(trace_id)
+
+    return trace_id, span_id
```
mabdinur, line 699: We can simplify this to the following. This will allow us to clean up some of the default values in our log integrations:
```suggestion
    if not span:
        return "0", "0"
    elif span.trace_id > MAX_UINT_64BITS and config._128_bit_trace_id_logging_enabled:
        return "{:032x}".format(span.trace_id), str(span.span_id)
    return str(span.trace_id), str(span.span_id)
```

ddtrace/_trace/tracer.py
```diff
@@ -416,17 +417,17 @@ def get_log_correlation_context(self) -> Dict[str, str]:
...
         if active:
-            trace_id = active.trace_id
-            if config._128_bit_trace_id_enabled and not config._128_bit_trace_id_logging_enabled:
-                trace_id = active._trace_id_64bits
+            trace_id, span_id = _get_trace_details_for_log_injection(active)
         return {
-            "trace_id": str(trace_id) if active else "0",
-            "span_id": str(active.span_id) if active else "0",
+            "trace_id": str(trace_id),
+            "span_id": str(span_id),
```
mabdinur, line 430: With the refactor below we do not need to convert these values to strings:

```suggestion
            "trace_id": trace_id,
            "span_id": span_id,
```

ddtrace/_trace/tracer.py
```diff
@@ -416,17 +417,17 @@ def get_log_correlation_context(self) -> Dict[str, str]:
         if self.enabled or self._apm_opt_out:
             active = self.context_provider.active()
 
+        trace_id = 0
+        span_id = 0
         if isinstance(active, Span) and active.service:
             service = active.service
         else:
             service = config.service
         if active:
```
mabdinur, line 426: We don't need to check if the span exists. This is handled by `_get_trace_details_for_log_injection`
```suggestion
```

ddtrace/_trace/tracer.py
```diff
@@ -416,17 +417,17 @@ def get_log_correlation_context(self) -> Dict[str, str]:
         if self.enabled or self._apm_opt_out:
             active = self.context_provider.active()
 
+        trace_id = 0
+        span_id = 0
```
mabdinur, line 421: This is handled by `_get_trace_details_for_log_injection`
```suggestion
```

ddtrace/contrib/logging/patch.py
```diff
@@ -83,11 +84,9 @@ def _w_makeRecord(func, instance, args, kwargs):
         span = _get_current_span(tracer=config.logging.tracer)
 
     if span:
-        trace_id = span.trace_id
-        if config._128_bit_trace_id_enabled and not config._128_bit_trace_id_logging_enabled:
-            trace_id = span._trace_id_64bits
+        trace_id, span_id = _get_trace_details_for_log_injection(span)
```
mabdinur, line 87: Question: Why cant we use `config.logging.get_log_correlation_context` in the logging integrations?

mabdinur: We can centralize all the log formatting logic in `tracer.get_log_correlation_context` and get rid of `_get_trace_details_for_log_injection`

ddtrace/contrib/trace_utils.py
wconti27: removed all this

ddtrace/contrib/trace_utils.py
wconti27: no, function is removed

ddtrace/contrib/logging/patch.py
wconti27: yup, done

ddtrace/_trace/tracer.py
```diff
@@ -406,27 +407,34 @@ def current_trace_context(self, *args, **kwargs) -> Optional[Context]:
...
-    def get_log_correlation_context(self) -> Dict[str, str]:
+    def get_log_correlation_context(self, active: Optional[Union[Context, Span]] = None) -> Dict[str, str]:
         """Retrieves the data used to correlate a log with the current active trace.
         Generates a dictionary for custom logging instrumentation including the trace id and
         span id of the current active span, as well as the configured service, version, and environment names.
         If there is no active span, a dictionary with an empty string for each value will be returned.
         """
-        active: Optional[Union[Context, Span]] = None
-        if self.enabled or self._apm_opt_out:
+        if active is None and (self.enabled or self._apm_opt_out):
```
mabdinur, line 416: What's the difference between _apm_opt_out and enabled? 
```
        if active is None and (self.enabled or not self._apm_opt_out):
```

ddtrace/_trace/tracer.py
```diff
@@ -406,27 +407,34 @@ def current_trace_context(self, *args, **kwargs) -> Optional[Context]:
...
         if active:
-            trace_id = active.trace_id
-            if config._128_bit_trace_id_enabled and not config._128_bit_trace_id_logging_enabled:
-                trace_id = active._trace_id_64bits
+            span_id = str(active.span_id)
+            # check if we are using 128 bit ids, and switch trace id to hex since backend needs hex 128 bit ids
+            if active.trace_id and active.trace_id > MAX_UINT_64BITS:
+                trace_id = "{:032x}".format(active.trace_id)
+            else:
+                trace_id = str(active.trace_id)
```
mabdinur, line 430: If active.trace_id is none this will raise an error. Trace id should honestly never be none here. If that happens then there was a programming error. We should raise a value error for this case.

ddtrace/_trace/tracer.py
```diff
@@ -406,27 +407,34 @@ def current_trace_context(self, *args, **kwargs) -> Optional[Context]:
...
         else:
             service = config.service
+
         if active:
-            trace_id = active.trace_id
-            if config._128_bit_trace_id_enabled and not config._128_bit_trace_id_logging_enabled:
-                trace_id = active._trace_id_64bits
+            span_id = str(active.span_id)
+            # check if we are using 128 bit ids, and switch trace id to hex since backend needs hex 128 bit ids
+            if active.trace_id and active.trace_id > MAX_UINT_64BITS:
```
mabdinur, line 427: Nice! So we are dropping support for the 128bit logging config. We should log a deprecation warning if this environment variable is set. We can do this in config.py. We should also add a release note.

ddtrace/settings/config.py
```diff
@@ -473,8 +473,6 @@ def __init__(self):
 
         self._128_bit_trace_id_enabled = asbool(os.getenv("DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED", True))
 
-        self._128_bit_trace_id_logging_enabled = asbool(os.getenv("DD_TRACE_128_BIT_TRACEID_LOGGING_ENABLED", False))
```
mabdinur, line 476: We should remove this config from our docs

ddtrace/_trace/tracer.py
wconti27: this was already there so i thought it may be important to keep

ddtrace/settings/config.py
wconti27: Its not documented.

ddtrace/_trace/tracer.py
wconti27: cleanup this up

ddtrace/_trace/tracer.py
mabdinur: That's fine we should probably fix this in a follow up PR. I am pretty confident it's a bug. It doesn't make sense to do log correlation when `apm_opt_out = True` and NOT do log correlation when `apm_opt_out = False`. 

Also `apm_opt_out` does not prevent spans from being sent to the agent. I believe the agent receives this flag as a span tag and then does something (maybe drops the spans 🤷). In the past we always injected span_ids and trace_ids into logs if tracing was enabled. We never took into account the span's sampling decision. I am not sure why we even need to check `apm_opt_out`.

ddtrace/contrib/logging/patch.py
```diff
@@ -77,20 +75,17 @@ def _w_makeRecord(func, instance, args, kwargs):
...
-            trace_id = span._trace_id_64bits
-        setattr(record, RECORD_ATTR_TRACE_ID, str(trace_id))
-        setattr(record, RECORD_ATTR_SPAN_ID, str(span.span_id))
-    else:
-        setattr(record, RECORD_ATTR_TRACE_ID, RECORD_ATTR_VALUE_ZERO)
-        setattr(record, RECORD_ATTR_SPAN_ID, RECORD_ATTR_VALUE_ZERO)
+        tracer = _get_tracer(tracer=config.logging.tracer)
+        if tracer:
+            span = tracer.current_span()
+            trace_details = tracer.get_log_correlation_context(active=span)
```
mabdinur, line 85: Do we need this? Can we replace this line with and remove the `active` argument from `get_log_correlation_context`.

```suggestion
            trace_details = tracer.get_log_correlation_context()
```

ddtrace/opentracer/tracer.py
```diff
@@ -366,11 +366,11 @@ def extract(self, format, carrier):  # noqa: A002
         self._dd_tracer.context_provider.activate(dd_span_ctx)
         return ot_span_ctx
 
-    def get_log_correlation_context(self):
-        # type: () -> Dict[str, str]
+    def get_log_correlation_context(self, active=None):
```
mabdinur, line 369: I don't think we need this argument (at least not yet). 

```suggestion
    def get_log_correlation_context(self):
```

mabdinur (approved):

ddtrace/contrib/logging/patch.py
wconti27: removed

ddtrace/opentracer/tracer.py
wconti27: removed

ddtrace/_trace/tracer.py
wconti27: sounds good

emmettbutler (approved):

PR #10104, opened 2024-08-06
chore(ci_visibility): add ITR support to v2 pytest plugin

romainkomorndatadog:
This adds ITR support to the in-development v2 of the `pytest` plugin as well as codeowners and source file information support.

The changes include:
- CI Visibility external API:
  - access to workspace path, codeowners, and CI Visibility enablement state
  - access to ITR test skipping status 
  - some bugfixes for core events 
- renaming CI parent items' `.children` attribute to `._children` since it is not a public attribute
- unrelated changes
  - moving `ddtrace.contrib.pytest.utils` to `_utils` (without a release note because all the methods within that module were already internal due to starting with a `_`)

Upon release, a breaking change will be that the plugin is skipping tests by suite, rather than by test.

No release note is added since this is still unsupported development code.

Note: this change works towards separating the contrib / public code side from our CIVisibility internals by leveraging the core API event bus, but some dependencies still reference `ddtrace.internal` directly.  

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/contrib/pytest/_plugin_v2.py
```diff
@@ -38,6 +50,69 @@
 _NODEID_REGEX = re.compile("^((?P<module>.*)/(?P<suite>[^/]*?))::(?P<name>.*?)$")
 
 
+def _handle_itr_should_skip(item, test_id) -> bool:
+    """Checks whether a test should be skipped, and returns a code collector if needed.
```
vitor-de-araujo, line 54: @romainkomorndatadog I did not catch what is meant by "returns a code collector" here, as it always returns a boolean.

romainkomorndatadog: That's because I'm terrible at updating comments when I change code. :(

ddtrace/contrib/pytest/_utils.py
```diff
@@ -104,3 +123,24 @@ def _is_pytest_8_or_later() -> bool:
...
 def _pytest_version_supports_itr() -> bool:
     return _get_pytest_version_tuple() >= ITR_MIN_SUPPORTED_VERSION
+
+
+def _pytest_marked_to_skip(item: pytest.Item) -> bool:
+    """Checks whether Pytest will skip an item"""
+    if item.get_closest_marker("skip") is not None:
+        return True
+
+    return any([True for marker in item.iter_markers(name="skipif") if marker.args[0] is True])
```
vitor-de-araujo, line 133: @romainkomorndatadog Can we rewrite this as:
```suggestion
    return any(marker.args[0] is True for marker in item.iter_markers(name="skipif"))
```
?

ddtrace/contrib/pytest/_utils.py
```diff
@@ -104,3 +123,24 @@ def _is_pytest_8_or_later() -> bool:
...
+def _is_test_unskippable(item: pytest.Item) -> bool:
+    """Returns True if a test has a skipif marker with value false and reason ITR_UNSKIPPABLE_REASON"""
+    return any(
+        [
+            True
+            for marker in item.iter_markers(name="skipif")
+            if marker.args[0] is False
+            and "reason" in marker.kwargs
+            and marker.kwargs["reason"] is ITR_UNSKIPPABLE_REASON
+        ]
```
vitor-de-araujo, line 145: @romainkomorndatadog ...I was going to propose a similar rewrite here, but I guess it's easier to read in this order than with the condition before the `for`. So feel free to ignore the previous one and close the conversation. :stuck_out_tongue:

vitor-de-araujo (approved):

ddtrace/contrib/pytest/_utils.py
romainkomorndatadog: I think we can actually even rewrite it as
```suggestion
    return any(marker.args[0] for marker in item.iter_markers(name="skipif"))
```
because `skipif` will skip if the first argument is truthy:
```
@pytest.mark.skipif(lambda: 'true', reason="true string")
```
results in
```
tests/test_views.py::test_remove_method_from_parent SKIPPED (true string)
```

I'll make the change in my soon-to-come follow-up.

ddtrace/contrib/pytest/_utils.py
romainkomorndatadog: I think this works as well:
```suggestion
    return any(
        (marker.args[0] is False and marker.kwargs.get("reason") is ITR_UNSKIPPABLE_REASON)
        for marker in item.iter_markers(name="skipif")
    )
```
Will also change in next PR.

PR #10107, opened 2024-08-06
chore(profiling): output pprof file with `DD_PROFILING_{EXPORT_LIBDD_ENABLED, OUTPUT_PPROF}`

taegyunkim:
Minimal and short-term change to export to file when `DD_PROFILING_EXPORT_LIBDD_ENABLED` and `DD_PROFILING_OUTPUT_PPROF` are set, similar to existing [PprofFileExporter](https://github.com/DataDog/dd-trace-py/blob/322e0013bdf3a9e7db42fa6467815e9a7bfecd75/ddtrace/profiling/exporter/file.py#L10), which is used when `DD_PROFILING_EXPORT_LIBDD_ENABLED` is turned off and `DD_PROFILING_OUTPUT_PPROF` is set. 

#9739 was a previous attempt to implement this in a more robust way. Ideally, we'd want this logic to be in libdatadog which can then be shared among language libraries. 

Tested on https://github.com/DataDog/prof-correctness/pull/46. Also checked that the output profile is lz4 compressed as libdatadog does that. 

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/internal/datadog/profiling/dd_wrapper/src/uploader.cpp
```diff
@@ -12,10 +16,33 @@ DdogCancellationTokenDeleter::operator()(ddog_CancellationToken* ptr) const
...
+    upload_seq++;
+}
+
+bool
+Datadog::Uploader::export_to_file(ddog_prof_EncodedProfile* encoded)
+{
+    // Write the profile to a file using the following format for filename:
+    // <output_filename>.<process_id>.<sequence_number>
+    std::ostringstream oss;
+    oss << output_filename << "." << getpid() << "." << upload_seq;
```
sanchda, line 34: Good idea.  I was originally going to suggest that runtime_id should go here, but none of our testing frameworks care about runtime_id--they only know about the PID.

ddtrace/internal/datadog/profiling/dd_wrapper/src/uploader.cpp
```diff
@@ -12,10 +16,33 @@ DdogCancellationTokenDeleter::operator()(ddog_CancellationToken* ptr) const
...
+    std::string filename = oss.str();
+    std::ofstream out(filename.c_str(), std::ios::binary);
+    if (!out.is_open()) {
+        errmsg = "Error opening output file " + filename;
+        std::cerr << errmsg << std::endl;
+        return false;
+    }
+
+    out.write(reinterpret_cast<const char*>(encoded->buffer.ptr), encoded->buffer.len);
```
sanchda, line 43: I've seen file writes fail in CI--we should check that the write succeeded (otherwise the error is somewhat opaque)

ddtrace/internal/datadog/profiling/dd_wrapper/src/uploader.cpp
```diff
@@ -12,10 +16,33 @@ DdogCancellationTokenDeleter::operator()(ddog_CancellationToken* ptr) const
...
+Datadog::Uploader::export_to_file(ddog_prof_EncodedProfile* encoded)
+{
+    // Write the profile to a file using the following format for filename:
+    // <output_filename>.<process_id>.<sequence_number>
+    std::ostringstream oss;
+    oss << output_filename << "." << getpid() << "." << upload_seq;
+    std::string filename = oss.str();
+    std::ofstream out(filename.c_str(), std::ios::binary);
+    if (!out.is_open()) {
+        errmsg = "Error opening output file " + filename;
```
sanchda, line 38: might be useful to print `strerr(errno)` here, since otherwise the fail diagnostics are tricky

ddtrace/internal/datadog/profiling/dd_wrapper/src/uploader.cpp
```diff
@@ -12,10 +16,33 @@ DdogCancellationTokenDeleter::operator()(ddog_CancellationToken* ptr) const
...
+        std::cerr << errmsg << std::endl;
+        return false;
+    }
+
+    out.write(reinterpret_cast<const char*>(encoded->buffer.ptr), encoded->buffer.len);
+    out.close();
```
sanchda, line 44: This should be handled by RAII.  It doesn't really matter, but if you take my other suggestion to change `out.write()`, then stylistically it'd make sense to write the close in two places, and at that point I'd say it's better to have it in zero places.

sanchda (approved):
Looks good.  Confirming that this looks like an end-to-end solution that implements the same config toggles as the old exporter.  I don't know whether this supports the same syntax or whether the files will always be the same, but I think that piece can get shaken out during testing if there are any downstream incompatibilities (I don't think there are).

ddtrace/internal/datadog/profiling/dd_wrapper/src/uploader.cpp
taegyunkim: Yeah, the current behavior is printing out pid by calling `os.getpid()`.

ddtrace/internal/datadog/profiling/dd_wrapper/src/uploader.cpp
taegyunkim: Done.

ddtrace/internal/datadog/profiling/dd_wrapper/src/uploader.cpp
taegyunkim: Checks for `fail()`

ddtrace/internal/datadog/profiling/dd_wrapper/src/uploader.cpp
taegyunkim: Oh, yes RAII should handle this and removed.

ddtrace/internal/datadog/profiling/dd_wrapper/include/uploader_builder.hpp
```diff
@@ -25,6 +25,7 @@ class UploaderBuilder
     static inline std::string profiler_version;
     static inline std::string url{ "http://localhost:8126" };
     static inline ExporterTagset user_tags{};
+    static inline std::string output_filename{ "" };
```
danielsn, line 28: Why static?

taegyunkim: This actually is a output_filename_prefix and it's followed by `<pid>.<counter>` where counter is incremented for each `UploaderBuilder::build` call.

taegyunkim: Chatted offline and agree that static member variables in here need not be static. Will follow up in a separate PR to revisit this.

PR #10124, opened 2024-08-07
chore(profiling): use lock-free queue to manage samples to avoid frequent alloc/free

taegyunkim:
When profiled with [ddprof](https://github.com/DataDog/ddprof), native profile shows ~525MiB from Allocated Memory view for dd-trace-py. 
Screenshot 2024-08-08 at 1 15 58 PM

This comes from the fact that `SampleManager` allocates a new `Sample` whenever it needs and frees right away. We can instead have a pool of `Sample`s to avoid this pattern. These `Sample`s are used by Stack/Memory/Lock Collectors. StackCollector MemoryCollector runs on separate threads, and LockCollector is a simple wrapper around python standard library. So the access to these `Sample`s need to be thread safe. 

#9617 is a draft PR using simple lock guarded vector of `Sample`s to solve the problem. However, using locks require us to carefully handle forking and could lead us to a deadlock when a thread holding the lock. So this PR attempts to avoid that problem by using a lock free queue available via libdatadog-ffi. 

We set the default capacity for pool to be 4. Even though this is a small number, as Python runs under GIL, only single thread can be run at a time and Collectors can create a handful number of `Sample` objects. Even though this is not the case and we need to allocate and free a lot of `Sample` objects, this PR is slightly better in a sense that we'll re-use some of the `Sample`s. I'll defer further investigation of setting default capacity for later. 

With this PR, I only see very minimal allocation happening from the first profile from ddprof, and the Allocated Memory decreases to 292 MiB from 525MiB. 

Screenshot 2024-08-08 at 2 05 28 PM
 

## Checklist

## Reviewer Checklist
- Title is accurate


r1viollet:
Thank you for pursuing this!

ddtrace/internal/datadog/profiling/dd_wrapper/src/synchronized_sample_pool.cpp
```diff
@@ -0,0 +1,93 @@
...
+    } else if (pop_result.tag == DDOG_ARRAY_QUEUE_POP_RESULT_ERR) {
+        auto err = pop_result.err;
+        std::string errmsg = err_to_msg(&err, "Failed to get sample from pool");
+        std::cerr << errmsg << std::endl;
+        ddog_Error_drop(&err);
+        return std::nullopt;
+    }
+    // To silence the warning about missing return statement
+    return std::nullopt;
+}
```
sanchda, line 70: Here's an idea for not returning within conditionals
```suggestion
std::optional<Sample*>
SynchronizedSamplePool::get_sample()
{
    std::optional<Sample*> result = std::nullopt;

    // It's actually ok to call ddog_ArrayQueue_* methods with a nullptr,
    // they will return an error result, but we already have printed out
    // an error message in the constructor, so check for nullptr here to
    // avoid spamming the error message.
    if (pool != nullptr) {
        ddog_ArrayQueue_PopResult pop_result = ddog_ArrayQueue_pop(pool.get());

        if (pop_result.tag == DDOG_ARRAY_QUEUE_POP_RESULT_OK) {
            result = static_cast<Sample*>(pop_result.ok);
        } else if (pop_result.tag == DDOG_ARRAY_QUEUE_POP_RESULT_ERR) {
            auto err = pop_result.err;
            std::string errmsg = err_to_msg(&err, "Failed to get sample from pool");
            std::cerr << errmsg << std::endl;
            ddog_Error_drop(&err);
        }
    }

    return result;
}
```

ddtrace/internal/datadog/profiling/dd_wrapper/src/synchronized_sample_pool.cpp
```diff
@@ -0,0 +1,93 @@
...
+        auto err = array_queue_new_result.err;
+        std::string errmsg = err_to_msg(&err, "Failed to create sample pool");
+        std::cerr << errmsg << std::endl;
+        ddog_Error_drop(&err);
+        pool = nullptr;
+    }
+}
+
+std::optional<Sample*>
+SynchronizedSamplePool::get_sample()
```
sanchda, line 47: super nit:  would it help anything if these interfaces more explicitly communicated the ownership status of the underlying data?  So using a verb like `checkout` (or `take`) instead of `get`.

(we do not follow this discipline elsewhere in the repo, we have no standards on what verbs to use in different , and the suggestion itself isn't even objectively helpful, so please take this as a suggestion and not an attempt at imposing a  requirement)

sanchda (approved):
I made some advisory comments.  Feel free to respond to or close any of them as you see fit, none of them block this PR.

I do feel a little bit anxious that the overall approach replaces what could be a very simple data structure in C++ with an opaque implementation across an FFI boundary, but I acknowledge that we live in a shared ecosystem and this might represent a global optima, even if it wouldn't be satisfying for a purist.

Overall good stuff, practical reduction of overhead, really impressed by your end-to-end vision in seeing the different pieces implemented across the stack--thank you!

ddtrace/internal/datadog/profiling/dd_wrapper/src/synchronized_sample_pool.cpp
```diff
@@ -0,0 +1,93 @@
...
+void
+sample_delete_fn(void* sample)
+{
+    delete static_cast<Sample*>(sample);
+}
+
+SynchronizedSamplePool::SynchronizedSamplePool()
+{
+    size_t capacity = 4;
+    const char* capacity_env = std::getenv("DD_PROFILING_SAMPLE_POOL_CAPACITY");
```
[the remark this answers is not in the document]
taegyunkim: Yeah, I also gave another thought and I prefer propagating default value from python side, will update shortly.

ddtrace/internal/datadog/profiling/dd_wrapper/src/synchronized_sample_pool.cpp
taegyunkim: I was thinking of Rust's `match(pop_result.tag)` and being extra pedantic to list out all cases, but this looks better. Thanks for the suggestion.

ddtrace/internal/datadog/profiling/dd_wrapper/src/sample_manager.cpp
```diff
@@ -30,13 +30,23 @@ Datadog::SampleManager::set_timeline(bool enable)
...
+    // Create a new Sample if we failed to get one.
     return new Datadog::Sample(type_mask, max_nframes); // NOLINT(cppcoreguidelines-owning-memory)
 }
 
 void
 Datadog::SampleManager::drop_sample(Datadog::Sample* sample)
 {
-    delete sample; // NOLINT(cppcoreguidelines-owning-memory)
+    sample->clear_buffers();
+    std::optional<Sample*> result_opt = sample_pool.return_sample(sample);
```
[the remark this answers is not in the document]
taegyunkim: The memory leak (losing `Sample`) due to `fork()` happens regardless of this PR. Will make a note of this as a comment.

ddtrace/internal/datadog/profiling/dd_wrapper/src/synchronized_sample_pool.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+void
+sample_delete_fn(void* sample)
+{
+    delete static_cast<Sample*>(sample);
+}
+
+SynchronizedSamplePool::SynchronizedSamplePool()
+{
+    size_t capacity = 4;
```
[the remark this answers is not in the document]
taegyunkim: Done.

ddtrace/internal/datadog/profiling/dd_wrapper/src/synchronized_sample_pool.cpp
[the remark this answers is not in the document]
taegyunkim: Done.

ddtrace/internal/datadog/profiling/dd_wrapper/src/sample_manager.cpp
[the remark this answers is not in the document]
sanchda: Right--agree.  I didn't make it clear in the comment, but this leaking behavior is very much the case right now as well.  If anything, this PR improves that.

ddtrace/internal/datadog/profiling/dd_wrapper/src/synchronized_sample_pool.cpp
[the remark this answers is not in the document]
sanchda: :bow: thank you

ddtrace/internal/datadog/profiling/dd_wrapper/src/synchronized_sample_pool.cpp
sanchda: Yeah, the implicit return in rust is 👨‍🍳 💋  for this type of thing

ddtrace/internal/datadog/profiling/dd_wrapper/src/synchronized_sample_pool.cpp
```diff
@@ -0,0 +1,79 @@
+#include "synchronized_sample_pool.hpp"
```
r1viollet, line 1: NIT: do we not require a header with license info in dd-trace-py ?

ddtrace/internal/datadog/profiling/dd_wrapper/include/sample_manager.hpp
```diff
@@ -19,12 +20,15 @@ class SampleManager
     static inline unsigned int max_nframes{ g_default_max_nframes };
     static inline SampleType type_mask{ SampleType::All };
     static inline std::mutex init_mutex{};
+    static inline size_t sample_pool_capacity{ g_default_sample_pool_capacity };
+    static inline std::unique_ptr<SynchronizedSamplePool> sample_pool{ nullptr };
```
r1viollet, line 24: do we want this to exist across several translation units ?
should we not have a single object that references the pool that we can reset upon forking ?

ddtrace/internal/datadog/profiling/dd_wrapper/src/synchronized_sample_pool.cpp
taegyunkim: No, [all these headers](https://github.com/DataDog/dd-trace-py/tree/taegyunkim/prof-10011-lockfree/ddtrace/internal/datadog/profiling/dd_wrapper/include) are missing license.

ddtrace/internal/datadog/profiling/dd_wrapper/include/sample_manager.hpp
taegyunkim: maybe? the underlying lock free queue doesn't have an api that lets the caller to take reference of the item in the container. All we do is `push` and `pop` basically.

PR #10126, opened 2024-08-07
chore(sampling): remove crash-by-design behaviors from sampling-related code

emmettbutler:
This change satisfies the Single-Step Instrumentation General Availability requirement "[Python] No crash-by-design behaviours" by removing many `raise` statements from the library's code.

The choice of which statements to remove was based on a human audit of the sampling-related results of `grep '\braise\b' -R ddtrace`. The `raise` statements that were left unchanged fall into one of these categories:

* reachable only via programming errors
* re-raises of exceptions originating in integrated libraries

`raise`s not in these categories are generally replaced with `log.warning` paired with an immediate return from the containing function when possible.

In a few cases, exceptions were used to indicate misconfigurations of sampling. This change replaces those exceptions with log messages and default return values.

This change also includes a bunch of conversions of type comments to Py3 type annotations. I updated these as I came across them in the files outputted by my `grep` command.

## Checklist

## Reviewer Checklist
- Title is accurate


tests/telemetry/test_telemetry.py
```diff
@@ -221,27 +221,6 @@ def process_trace(self, trace):
     ]["message"]
 
 
-def test_app_started_error_unhandled_tracer_exception(test_agent_session, run_python_code_in_subprocess):
-    env = os.environ.copy()
-    env["DD_SPAN_SAMPLING_RULES"] = "invalid_rules"
-
```
erikayasuda, line 227: Maybe we can keep this test if we add:
```
env["DD_TESTING_RAISE"] = True
```
So `config._raise` is set to True? And have a TODO to add another test to ensure without that config, there is no exception raised?

ddtrace/sampler.py
```diff
@@ -250,7 +245,7 @@ def __init__(
             self.rules = []
             # Validate that rules is a list of SampleRules
             for rule in rules:
-                if not isinstance(rule, SamplingRule):
+                if config._raise and not isinstance(rule, SamplingRule):
```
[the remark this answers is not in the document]
emmettbutler: thanks, that's a great catch.

tests/telemetry/test_telemetry.py
emmettbutler: This change replaces all of the exceptions from `DD_SPAN_SAMPLING_RULES` with warning logs, so regardless of that variable's value there won't be an exception raised. Because the intention of this PR is to make the library never raise an unhandled exception, the premise of this test isn't really valid anymore, which is why I chose to delete it.

erikayasuda (approved):

emmettbutler:
@mabdinur @ZStriker19 could I have your review of this change?

emmettbutler:
@wconti27 @quinna-h @rachelyangdog could one of you please take a look at this change?

quinna-h (approved):

PR #10129, opened 2024-08-07
chore(ci): properly set CMAKE_BUILD_PARALLEL_LEVEL for cibuildwheel

brettlangdon:
Previously we were setting the env var `CMAKE_BUILD_PARALLEL_LEVEL`, but it wasn't being passed into the containers used by cibuildwheel, so we were seeing no benefits.

This PR only updates the env var for linux since updating for macos and windows didn't seem to work, I need to investigate further, but I figured this is a good incremental change.

The benefit is linux builds that complete in around/under 4 minutes. It doesn't make a big change to the overall runtime of the whole build (until we can get windows/macos faster), but it does drop the "billable minutes" by 20-30 minutes (good show of the impact of the change).

We had parallel of 12 before, 24 wasn't that much faster, but we can just saturate the CPUs.

## Checklist

## Reviewer Checklist
- Title is accurate

### DIFF
--- a/.github/workflows/build_python_3.yml
+++ b/.github/workflows/build_python_3.yml
@@ -84,7 +84,6 @@ jobs:
         env:
           CIBW_SKIP: ${{ inputs.cibw_skip }}
           CIBW_PRERELEASE_PYTHONS: ${{ inputs.cibw_prerelease_pythons }}
-          CMAKE_BUILD_PARALLEL_LEVEL: 12
           CIBW_MUSLLINUX_I686_IMAGE: ghcr.io/datadog/dd-trace-py/pypa_musllinux_1_2_i686:latest
           CIBW_BEFORE_ALL: >
             if [[ "$(uname -m)-$(uname -i)-$(uname -o | tr '[:upper:]' '[:lower:]')-$(ldd --version 2>&1 | head -n 1 | awk '{print $1}')" != "i686-unknown-linux-musl" ]];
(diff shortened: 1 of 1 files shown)


.github/workflows/build_python_3.yml
```diff
@@ -93,7 +92,7 @@ jobs:
             fi
           CIBW_BEFORE_ALL_WINDOWS: rustup target add i686-pc-windows-msvc
           CIBW_BEFORE_ALL_MACOS: rustup target add aarch64-apple-darwin
-          CIBW_ENVIRONMENT_LINUX: "PATH=$HOME/.cargo/bin:$PATH"
+          CIBW_ENVIRONMENT_LINUX: PATH=$HOME/.cargo/bin:$PATH CMAKE_BUILD_PARALLEL_LEVEL=24
```
taegyunkim, line 95: how about just `CIBW_ENVIRONMENT: CMAKE_BUILD_PARALLEL_LEVEL=24`?
I wonder what are cmake versions on the images used by cibuildwheel, CMAKE_BUILD_PARALLEL_LEVEL is available from cmake 3.12
https://cmake.org/cmake/help/latest/envvar/CMAKE_BUILD_PARALLEL_LEVEL.html

brettlangdon: ugh... ok... so...

I tried that, and it caused other failures since I think it doesn't combine all the `CIBW_ENVIRONMENT_*` envs together (I think, I'd have to go back and look).

Also, if I add it as `CIBW_ENVIRONMENT_MACOS: ...` then I see very weird failures like... builds the wheel fine, but then testing it yells that the wheels isn't supported on that architecture... I think updating `CIBW_ENVIRONMENT*` envs is messing with something......

hence why I decided to just go for this now and defer to look into it again later.

taegyunkim: Cool, and where does the 24 come from?

taegyunkim (approved):

.github/workflows/build_python_3.yml
brettlangdon: A mix of arbitrarily large so that we can let cmake context switch during any IO, and David 😆 ("Spicy:  try juicing it to 24")

sanchda: > Cool, and where does the 24 come from?

I wanted to impress Brett by showing him that I know even bigger numbers than 12, so I went all out with the biggest one I know.

Also, I had a hunch that we can bump up the parallel level a bit more and eke out a bit more utilization from the CPUs, but I don't know that for certain.  My general intuition is that you usually cannot saturate a single core with a single worker, so you might as well take advantage of any blocking, off-CPU movements (paging, IO).  Obviously that doesn't scale, and there are diminishing returns.  If we wanted to be scientific, we'd select a parameter that minimized the sum of all build jobs and just periodically revisit it as our work (and the hardware) changes.

brettlangdon: From the testing 24 yielded faster builds than 12, but not by much, so seemed worth keeping, but not worth digging in too much further.

maybe sccache will help make things faster?

PR #10138, opened 2024-08-08
feat(llmobs): llmobs context provider

Yun-Kim:
**tl;dr: This PR adds an additional context provider to the `LLMObs` service to keep track of the active `LLMObs`-type span.**

This PR also adds a convenience `LLMObs.current_span()` method to get the current LLMObs type span, and uses this in our existing public methods that involve checking for the current span if not provided by the user as an argument (ex: `LLMObs.export_span(span=None, ...)`, `LLMObs.annotate(span=None, ...)`.

### Workflow

When an LLMObs span gets started, (i.e. `LLMObs._start_span(...)`, `BaseLLMIntegration.trace(submit_to_llmobs=True, ...)`, `BedrockIntegration.set_llmobs_tags(...)`), we look at the current active LLMObs span, use that to set the parent_id field of the started span, and then activate the started span as the new current active LLMObs span.

The LLMObsContextProvider updates the current active LLMObs span by iterating through the span's ancestor tree and stopping at the closest LLMObs-type span which is not finished.

For distributed tracing scenarios, we inject the request headers with an LLMObs parent ID. We then create a Context object using the extracted LLMObs parent ID header and activate that in the new process.

## Context

Spans submitted to LLM Observability are technically APM spans (but with enhanced metadata/metrics and in a different span structure) marked with an `llm` span_type. This means APM span/parent/trace IDs are all reused in LLMObs spans. 

**Problem**: if APM-specific spans (via integration/manual) are intermixed with LLMObs spans in a trace, this causes problems with traces submitted to LLM Observability, where spans may have `parent_id` values referencing spans that do not exist in LLM Observability storage, because those APM-specific spans are not ingested to LLM Observability. (example trace structure below)
```
Span 1 (LLMObs) --> Span 2 (APM) --> Span 3 (LLMObs)
# Span 3's parent_id is equal to span 2's span_id, but span 2 does not exist in LLM Obs ==> broken LLM Obs trace
```

### Initial Solution: Use Span Context

In an effort to reuse as much tracer-internals as possible, internal LLMObs trace processing involved traversing a span's parent tree to find the nearest LLMObs type ancestor span, and then using that span's span_id as the parent_id. For distributed cases, the LLMObs parent ID was injected into request headers, and then extracted into the underlying span `Context` object's `context._meta` dictionary field. 
This works fine for the most part, except for multithreading applications as the `Context._meta` field is a shallow copy and thus shared between threads, making it susceptible to race conditions where the propagated LLMObs parent ID would get overwritten by other threads starting.

### Current Approach: Manage our own LLMObs contexts

This approach is similar to the first, but instead of attempting to use the `Context._meta` field in a use case it was not designed for, we just create our own context provider for tracking active LLMObs span/contexts. Since we're relying on `ContextVar` which is thread-safe, this does not have the same problems as previously regarding multithreaded applications.

Since we are relying on the new `LLMObsContextProvider` to manage and update the active LLMObs span (which reuses the same logic of traversing the parent tree for the nearest LLMObs span), we can get rid of all of the special casing required for dealing with `Context`-propagated tags.

The only changes we need to make to shared library code is adding distributed injection code into `HTTPPropagator.inject()`, activating the propagated request headers into `trace_utils.activate_distributed_headers()`, and context propagation/activation in the threading patch code.

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/llmobs/_constants.py
```diff
@@ -3,8 +3,7 @@
 METADATA = "_ml_obs.meta.metadata"
 METRICS = "_ml_obs.metrics"
 ML_APP = "_ml_obs.meta.ml_app"
-PROPAGATED_PARENT_ID_KEY = "_dd.p.llmobs_parent_id"
-PARENT_ID_KEY = "_ml_obs.llmobs_parent_id"
+PARENT_ID_KEY = "_dd.p.llmobs_parent_id"
```
Yun-Kim, line 6: Unsure if we need to rename this to a different key name.

ddtrace/llmobs/_llmobs.py
```diff
@@ -239,15 +239,19 @@ def export_span(cls, span: Optional[Span] = None) -> Optional[ExportedLLMObsSpan
...
             return None
-        if span.span_type != SpanTypes.LLM:
-            log.warning("Span must be an LLMObs-generated span.")
-            return None
         return ExportedLLMObsSpan(span_id=str(span.span_id), trace_id="{:x}".format(span.trace_id))
 
+    def current_span(self) -> Optional[Span]:
+        """Returns the current active LLMObs span."""
+        if self.enabled is False:
+            return None
```
lievan, line 251: if we don't require `cls.enabled` to be true to start spans should we remove this check here as well? Since we've run into issues with customer code paths breaking when llmobs is turned off when customers are running unit tests for example

Yun-Kim:
Closing this due to annoying merge conflicts from upstream changes. Replacing this with #10767.

ddtrace/contrib/internal/futures/threading.py
```diff
@@ -32,5 +40,9 @@ def _wrap_execution(ctx: Optional[Context], fn, args, kwargs):
     variable because it's outside the asynchronous loop.
     """
     if ctx is not None:
-        ddtrace.tracer.context_provider.activate(ctx)
+        ddtrace.tracer.context_provider.activate(ctx[0])
```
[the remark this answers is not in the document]
Yun-Kim: The type check should be sufficient, `_wrap_execution()` is used directly only by `wrap_submit()` AFAIK.

ddtrace/llmobs/_llmobs.py
Yun-Kim: Fair point, will remove that check.

PR #10152, opened 2024-08-09
ci(tracer): make tracer suite depend on ci_visibility

vitor-de-araujo:
The tracer test suite tests `CIVisibilityWriter`, and therefore changes to ci_visibility must cause tracer tests to run. This avoids issues like https://github.com/DataDog/dd-trace-py/pull/10150 where a change in ci_visibility breaks a tracer test but does not cause the tests to be run.

## Checklist

## Reviewer Checklist
- Title is accurate


romainkomorndatadog (approved):

romainkomorndatadog:
I'm adding the `changelog-nochangelog` tag because I don't think backporting this makes sense:
- any backport would have to come from a PR where the tests ran (so the `tracer` suite would've run on changes to the `ci_visibility` component)
- we don't make fixes to the `ci_visibility` things all that often, so it's highly unlikely we'll backport something into the currently maintained releases

But also: "Famous last words."

PR #10165, opened 2024-08-09
chore(dep): replace vendored wrapt with package dependency

brettlangdon:
Since we added `opentelemetry-sdk` as a dependency of this package we are installing `wrapt` anyways (`opentelemetry-sdk` -> `opentelemetry-api` -> `Deprecated` -> `wrapt`).

The original reason we vendored `wrapt` was to ensure that the wrapt C extension was properly built and distributed with our library. `wrapt>=1.13.0` has been shipping pre-built wheels that are compatible with the version range we support.

Removing the vendored `wrapt` will help reduce our package size as well as the overall install size (removing a dependency that gets included twice).

## Checklist

## Reviewer Checklist
- Title is accurate


taegyunkim (approved):

gnufede (approved):

P403n1x87 (approved):

ddtrace/bootstrap/sitecustomize.py
```diff
@@ -78,6 +78,7 @@ def drop(module_name):
             "attr",
             "google",
             "google.protobuf",  # the upb backend in >= 4.21 does not like being unloaded
+            "wrapt",
```
erikayasuda, line 81: Just for my own sanity check...

We added `wrapt` here to make sure we keep the implicitly imported `wrapt` module from `opentelemetry-sdk`. If we don't add this here, we would remove it from `sys.modules` during unloading and it wouldn't be able to be used via `import wrapt` in the other files.

So if `wrapt` ends up _not_ getting installed by `opentelemetry-sdk` in the future, do we need to do anything? Or is that handled by having it in the `project.toml` file as a dependency?

erikayasuda (approved):

tests/internal/test_auto.py
```diff
@@ -1,6 +1,9 @@
+import sys
+
 import pytest
 
 
+@pytest.mark.skipif(sys.version_info < (3, 12, 5), reason="Python < 3.12.5 eagerly loads the threading module")
```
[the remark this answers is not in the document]
brettlangdon: "sort of", this should be broken on `main` but... isn't... but somehow moving to dependency `wrapt` causes this to always fail.

The reason is because Python 3.12 started eagerly loading `threading`, which means we cannot unload to save our own copy of the module. This is a known issue that is being fixed and released in Python 3.12.5

It does mean this test might fail when 3.12.5 is released.

ddtrace/bootstrap/sitecustomize.py
brettlangdon: The unloading is so we can keep our own copy of modules... however, in a _ton_ of places we do things like `if isinstance(thing, wrapt.FunctionWrapper)`, but if we have our own copy, some of that stuff may start failing. Instead, I think we definitely 100% want to use the same `wrapt` module for the whole process.

I didn't notice any issues, this is my hunch/instinct based on our usage.

wconti27 (approved):

PR #10166, opened 2024-08-09
chore(integrations): move httplib,httpx,jinja2,kafka,kombu to internal

rachelyangdog:
- Moves all integration internals in ddtrace/contrib/(integration name)/ to ddtrace/contrib/internal/(integration name)/ for httplib, httpx, jinja2, kafka, and kombu
- Ensures ddtrace/contrib/(integration name)/ and ddtrace/contrib/(integration name)/ continue to expose the same functions, classes, imports, and module level variables (via from ..internal.integration.module import * imports).
- Log a deprecation warning if internal modules in ddtrace/contrib/(integration name)/ and ddtrace/contrib/(integration name)/. Only patch and unpack methods should be exposed by these packages.
- https://github.com/DataDog/dd-trace-py/pull/9996- 

## Checklist

## Reviewer Checklist
- Title is accurate


mabdinur (approved):

ddtrace/contrib/httplib/__init__.py
```diff
@@ -58,9 +58,10 @@
 
 :ref:`Headers tracing <http-headers-tracing>` is supported for this integration.
 """
-from .patch import get_version
-from .patch import patch
-from .patch import unpatch
+
+from ..internal.httplib.patch import get_version
+from ..internal.httplib.patch import patch
+from ..internal.httplib.patch import unpatch
```
erikayasuda, line 64: Why do we not need to do
```
from . import patch as _ # noqa: F401, I001
```
for this one?

rachelyangdog: it allows users to import from `ddtrace.contrib.httplib.patch` directly

releasenotes/notes/move-integrations-to-internal-httplib-475d071a277b18cf.yaml
```diff
@@ -0,0 +1,12 @@
...
+  - |
+    httplib: Deprecates all modules in the ``ddtrace.contrib.httplib`` package. Use attributes exposed in ``ddtrace.contrib.httplib.__all__`` instead.
+  - |
+    httpx: Deprecates all modules in the ``ddtrace.contrib.httpx`` package. Use attributes exposed in ``ddtrace.contrib.httpx.__all__`` instead.
+  - |
+    jinja2: Deprecates all modules in the ``ddtrace.contrib.jinja2`` package. Use attributes exposed in ``ddtrace.contrib.jinja2.__all__`` instead.
+  - |
+    kafka: Deprecates all modules in the ``ddtrace.contrib.kafka`` package. Use attributes exposed in ``ddtrace.contrib.kafka.__all__`` instead.
+  - |
+    kombu: Deprecates all modules in the ``ddtrace.contrib.kombu`` package. Use attributes exposed in ``ddtrace.contrib.kombu.__all__`` instead.
```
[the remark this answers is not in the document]
rachelyangdog: I think we should keep this format because all the previous merged PRs have this format (for consistency)

ddtrace/contrib/httplib/__init__.py
mabdinur: Good catch. We should add it here. 

We honestly got lucky here. If we swap the order of these two imports [from ddtrace.contrib.httplib.patch import should_skip_request](https://github.com/DataDog/dd-trace-py/blob/da67b1f50f58539aab8f681d166413500797d8ad/tests/contrib/httplib/test_httplib.py#L14C30-L14C35) AND [from ddtrace.contrib.httplib import patch](https://github.com/DataDog/dd-trace-py/blob/da67b1f50f58539aab8f681d166413500797d8ad/tests/contrib/httplib/test_httplib.py#L12C1-L12C42) then most `httplib` tests will fail with the following error: `patch() E               TypeError: 'module' object is not callable`. The order we import the patch module and the patch function matter. We need to fix this module. 

We should also keep this mind for all of our integrations. We may not have explicit test coverage for this case in every integration 

Also it would we be nice if this module used required_modules helper in this integration. Something like this:

```
from ...internal.utils.importlib import require_modules


required_modules = ["http.client"]

with require_modules(required_modules) as missing_modules:
    if not missing_modules:
        from . import patch as _ # noqa: F401, I001
        ......
```

erikayasuda: Yeah I was wondering about that too. Is it just an older integration, so it pre-dates `require_modules()` and didn't get it?

mabdinur: Ya that was a miss on our part and I am afraid if we don't update it now we may never get to it

rachelyangdog: ok got it! will update and push soon

erikayasuda (approved):

PR #10171, opened 2024-08-09
ci(tracer): reduce the flakiness of the tracer tests

mabdinur:
- 85e63463e26cd20db968f4f26505875d78b4a62a Refactors all tracer tests that use `os.fork` to run in a process outside of pytest runner. Using `os.fork()` in pytest can lead to undefined behavior. 
- c294c9b76aa1415c55f07a93b4c283cd5b7d03f6 Ensures `Span.finish()` is called on all spans. Not finishing spans can lead to unexpected parenting in tests. It will also log a warning that might impact tests that validate stderr.
- https://github.com/DataDog/dd-trace-py/pull/10171/commits/69c8c40da7780ea0611770ebd38705d09daeca60 skips hour glass test
- Skips broken multiprocessing tests. PY312 dropped support for forking procceses in multithreaded applications: https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods . 
- Ensure tracer and severless tests do not rely on `DD_TRACE_AGENT_URL` being set to `""` in circleci configurations. These tests should always be run with the required configurations. 
- Skips PCR File container_info tests that always fails on MacOS. 
- Runs tracer tests with testagent. This ensures subprocess tests that submit trace payloads do not log 404 error logs. The presence of error and warning logs may cause subprocess tests to fail.

## Checklist

## Reviewer Checklist
- Title is accurate


.circleci/config.templ.yml
```diff
@@ -577,7 +577,7 @@ jobs:
     steps:
```
mabdinur, line 577:
```suggestion
    docker:
      - image: *ddtrace_dev_image
      - *testagent
    steps:
```

tests/tracer/test_rand.py
```diff
@@ -191,17 +202,26 @@ def test_threadsafe():
...
+            yield s.span_id, s.trace_id
+
     # Similar test to test_fork() above except we use the tracer API.
     # In this case we expect to never have collisions.
     if pid > 0:
         # parent
-        parent_ids_list = list(
-            chain.from_iterable((s.span_id, s.trace_id) for s in [tracer.start_span("s") for _ in range(100)])
-        )
+        parent_ids_list = list(chain.from_iterable(ids for ids in [get_ids() for _ in range(100)]))
```
erikayasuda, line 224: This is just to ensure span.finish() on each of these?

tests/tracer/test_tracer.py
```diff
@@ -754,8 +757,14 @@ def test_tracer_dogstatsd_url():
         assert str(e) == "Unknown url format for `foo://foobar:12`"
 
 
+@pytest.mark.skip(reason="Fails to Pickle RateLimiter in the Tracer")
```
erikayasuda, line 760: If making it a subprocess tests fails the test, should we just keep it as is instead? Or does that still make this test flaky, so we're silencing it for now?

tests/tracer/test_tracer.py
```diff
@@ -1040,8 +1053,13 @@ def _test_tracer_runtime_tags_fork_task(tracer, q):
     span.finish()
 
 
+@pytest.mark.skip(reason="Fails to Pickle RateLimiter in the Tracer")
```
erikayasuda, line 1056: Same question here

erikayasuda (approved):

.circleci/config.templ.yml
```diff
@@ -604,12 +604,12 @@ jobs:
...
 
   tracer:
+    <<: *machine_executor
     parallelism: 10
-    <<: *contrib_job_large
     steps:
       - run_test:
           pattern: "tracer"
-          trace_agent_url: ""
+          snapshot: true
```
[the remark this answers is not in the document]
mabdinur: Some of tracer tests submit traces to the agent. Right now these tests are silently logging 404 errors. These error logs caused the new subprocess tests to fail. By ensuring the tracer testsuite is run with an agent/testagent we resolve these errors. 

This also has the added benifit of running testagent checks on the tracer test suite. This surfaced this bug: https://github.com/DataDog/dd-trace-py/pull/10241

tests/tracer/runtime/test_container.py
```diff
@@ -253,7 +255,7 @@ def test_cgroup_info_from_line(line, expected_info):
             None,
         ),
         # PCF file
-        (
+        pytest.param(
```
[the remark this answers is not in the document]
mabdinur: This parameter always fails on M1s but passes in CI. With this change all tracer tests will pass locally.

tests/tracer/test_rand.py
mabdinur: Yup

tests/tracer/test_tracer.py
mabdinur: The default start method for the multiprocessing library is fork: https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods (on macos AND python>3.8 it is spawn). Forking is not supported by python and can lead to deadlocks in some multithread libraries (like ddtrace). 

Converting this test to a subprocess test solves one of these issues however there are still other issues that cause this test to be flaky. For now I think it's best to skip it

mabdinur:
Blocked by: https://github.com/DataDog/dd-trace-py/pull/10241

PR #10180, opened 2024-08-12
chore(ci_visibility): add tests for pytest plugin v2 and fix associated issues

romainkomorndatadog:
This adds a hatch-based tests for v2 of the `pytest` plugin and bundles a long list of changes/fixes are bundled into this commit to make the tests pass 

Somewhat in order of appearance in files modified by this commit:
- switching from `pytest_collection_modifyitems` to `pytest_collection_finish` to only process the tests that `pytest` will execute (post-filtering)
- tweaks `pytest_runtest_makereport` to finish events during during test teardown (otherwise events may be finished multiple times)
- introduces a core event hook to access the tracer as well as the item span (to support the `ddtracer` and `ddspan` fixtures)
- new API functionality
  - get_tags to fetch tag values from items
  - ITR events for checking whether an item was forced run
- fixes
  - uses pytest plugin config to properly name services and spans
  - properly sets ITR tags depending on ITR enablement
  - properly handles xfail scenarios
  - uses default names for module and suite if parsing the nodeid fails
  - adds module path to module discovery
  - properly specify resource on spans
  - don't crash by using `get_worskpace_path()` in session discovery
- adds per-item-class methods for telemetry event creation/finishing
- some comment/log message updates to fix bad copy/pasta
- add type hint return values (even though implicit is enough in some cases)
- test changes
  - introduces convenience function for filtering spans
  - deal with the fact that V2 does not generate separate spans for empty modules
  - deal with the fact that V2 always includes class names in tests
  - add V2 of snapshot tests (because snapshots can't have conditional values)

Since the v2 of the pytest plugin depends on the same files as the original version, and is only a temporary suite while the migration from v1 to v2 is completed, a new "v2" component is added to our suites spec. 

When the migration is complete, the original `riot`-based environment for `pytest` will be removed and the `hatch`-based v2 environment will be renamed in its place. The `v2` suite in `.suitespec.json` will also be deleted at that time.

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/contrib/pytest/_plugin_v2.py
```diff
@@ -56,13 +58,18 @@ def _handle_itr_should_skip(item, test_id) -> bool:
...
         return False
 
-    item_is_unskippable = CITest.is_item_itr_unskippable(item) or CISuite.is_item_itr_unskippable(test_id.parent_id)
+    suite_id = test_id.parent_id
 
-    if CISuite.is_item_itr_skippable(test_id.parent_id):
+    item_is_unskippable = CISuite.is_item_itr_unskippable(suite_id)
+
+    if CISuite.is_item_itr_skippable(suite_id):
         if item_is_unskippable:
```
vitor-de-araujo, line 66: @romainkomorndatadog  As I understand, this can only be true if the suite passes both `is_item_itr_skippable` and `is_item_itr_unskippable`. Can this happen? Why do we have both functions?

ddtrace/contrib/pytest/_utils.py
```diff
@@ -130,17 +144,21 @@ def _pytest_marked_to_skip(item: pytest.Item) -> bool:
...
     """Returns True if a test has a skipif marker with value false and reason ITR_UNSKIPPABLE_REASON"""
     return any(
-        [
-            True
-            for marker in item.iter_markers(name="skipif")
-            if marker.args[0] is False
-            and "reason" in marker.kwargs
-            and marker.kwargs["reason"] is ITR_UNSKIPPABLE_REASON
-        ]
+        (marker.args[0] is False and marker.kwargs.get("reason") is ITR_UNSKIPPABLE_REASON)
```
vitor-de-araujo, line 153: @romainkomorndatadog Any reason to use `is` instead of `==`?

ddtrace/internal/ci_visibility/api/ci_base.py
```diff
@@ -133,48 +146,58 @@ def __init__(
...
-        self._is_benchmark = False
-
-    def __repr__(self):
+    def __repr__(self) -> str:
         return f"{self.__class__.__name__}({self.item_id})"
 
-    def _add_all_tags_to_span(self):
+    @_require_span
+    def _add_all_tags_to_span(self) -> None:
+        if self._span is None:
```
vitor-de-araujo, line 154: @romainkomorndatadog Because of the decorator, we know for sure `self._span` is not None here, no?

ddtrace/contrib/pytest/_utils.py
romainkomorndatadog: Oh, wow, no, it should be `==` ... That's been wrong from the start.

Good thing Python does fun things like
```
>>> a = "datadog_itr_unskippable"
>>> b = "datadog_itr_unskippable"
>>> a is b
True
>>> b = "datadog_itr_unskippable_2"
>>> a is b
False
>>> b = "datadog_itr_unskippable"
>>> a is b
True
```

tests/contrib/pytest/test_pytest.py
```diff
@@ -25,6 +27,33 @@
...
+    target_type = _names_map[span_type][0]
+    target_name = _names_map[span_type][1] if name else None
+    return [
+        span
+        for span in spans
+        if span.get_tag("type") == target_type
+        and (
+            span.get_tag(target_name) == name
+            if name
+            else True and (span.get_tag("test.status") == status if status else True)
```
emmettbutler, line 52: Is there a way to write this condition that makes it a little clearer what it's trying to do?

ddtrace/contrib/pytest/_utils.py
emmettbutler: It's actually pretty important to use `==` here instead of `is`, because `==` tests for equality and `is` tests for identity (are these two the same object).

emmettbutler (approved):

tests/contrib/pytest/test_pytest.py
romainkomorndatadog: Yeah, this thing's getting a little unwieldy for a comprehension.

I'll rewrite it.

vitor-de-araujo: Also I think this gets parsed as `else (True and ...)`.

romainkomorndatadog: https://github.com/DataDog/dd-trace-py/pull/10180/files#diff-48a91ef11a9dbcd767fd3050fa39ca48a9bc8454d47d71b2afc35ebc630686f7R48-R59

I don't love it but.. at least it's more readable.

vitor-de-araujo (approved):

PR #9141, opened 2024-05-01
fix: pytest deadlock when using gevent

brettlangdon:
Fixes #8281

This fix resolves an issue when using ``pytest`` + ``gevent`` where the telemetry writer was eager initialized by ``pytest`` entrypoints loading of our plugin causing a potential dead lock.

## Reproduction

The specific case that we targeted solving with this PR:

```python
# test.py
import ddtrace

import gevent.monkey

gevent.monkey.patch_all()


def test_thing():
    pass
```

```shell
$ pytest test.py
$ pytest -p ddtrace -p ddtrace.pytest_bdd -p ddtrace.pytest_benchmark test.py
$ pytest -p no:ddtrace -p no:ddtrace.pytest_bdd -p no:ddtrace.pytest_benchmark test.py
```

Previously would dead-lock, but now will execute without problem.

## Trade offs

We no longer get telemetry enabled and sending data just from `import ddtrace`.

Moving telemetry writer starting to `ddtrace.bootstrap.preload` will mean that any unhandled exceptions that occur before the telemetry writer can be started will not get reported.

## Checklist

- [x] Change(s) are motivated and described in the PR description
- [x] Testing strategy is described if automated tests are not included in the PR
- [x] Risks are described (performance impact, potential for breakage, maintainability)
- [x] Change is maintainable (easy to change, telemetry, documentation)
- [x] [Library release note guidelines](https://ddtrace.readthedocs.io/en/stable/releasenotes.html) are followed or label `changelog/no-changelog` is set
- [x] Documentation is included (in-code, generated user docs, [public corp docs](https://github.com/DataDog/documentation/))
- [x] Backport labels are set (if [applicable](https://ddtrace.readthedocs.io/en/latest/contributing.html#backporting))
- [x] If this PR changes the public interface, I've notified `@DataDog/apm-tees`.

## Reviewer Checklist

- [x] Title is accurate
- [x] All changes are related to the pull request's stated goal
- [x] Description motivates each change
- [x] Avoids breaking [API](https://ddtrace.readthedocs.io/en/stable/versioning.html#interfaces) changes
- [x] Testing strategy adequately addresses listed risks
- [x] Change is maintainable (easy to change, telemetry, documentation)
- [x] Release note makes sense to a user of the library
- [x] Author has acknowledged and discussed the performance implications of this PR as reported in the benchmarks PR comment
[...]

PR #10196, opened 2024-08-13
chore(ci_visibility): revert sub-plugin strategy and move v2 functions to try-except blocks

romainkomorndatadog:
There are two changes in this commit:

1- remove "sub plugin" strategy introduced in #8999

The sub-plugin strategy was originally implemented to delay the import of `ddtrace`, however that import is (in the current scenario) inevitable because the `pyproject.toml` entry-points that tell `pytest` where to find plugins already import `ddtrace`, and the original motivation of the change in (#8281 and #4845) was instead addressed by #9141 .

The motivation to revert this change is that some hooks (eg: `pytest_load_inital_conftests`) added by the class-based plugins do not execute when the plugin is loaded in the `pytest_configure()` hook in the `plugin.py` file itself. All the methods in the plugin classes use the `@staticmethod` decorator and do not benefit from being in a class.

This change allows the versioned plugin files to implement hooks by virtue of being imported into the original plugin module.

2- move v2 functions to try-except blocks

Since CI Visibility should never affect `pytest`'s execution (or, at the very least, should never prevent it from running to completion), the hooks themselves should never raise exceptions. With this change, some functions simply wrap the entire body in a try/except block, while some others define a separate function that is then called in a try/except block.

A decorator was not used for this purpose because some hooks yield, while others don't, and logging relevant errors is also a little simpler.

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/contrib/pytest/_plugin_v2.py
```diff
@@ -118,30 +118,32 @@ def _handle_collected_coverage(test_id, coverage_collector) -> None:
...
-    def pytest_configure(config: pytest.Config) -> None:
+def pytest_configure(config: pytest.Config) -> None:
+    try:
         if is_enabled(config):
             unpatch_unittest()
             take_over_logger_stream_handler()
             enable_ci_visibility(config=dd_config.pytest)
         if _is_pytest_cov_enabled(config):
             patch_coverage()
+    except:  # noqa: E722            log.warning("encountered error during configure, disabling Datadog CI Visibility", exc_info=True)
```
vitor-de-araujo, line 136: @romainkomorndatadog Missing newline? :eyes:

romainkomorndatadog: You are as eagle-eyed as I am potato-eyed. :(

vitor-de-araujo (approved):

brettlangdon (approved):

PR #10210, opened 2024-08-13
chore(telemetry): Telemetry Additions

omerli:
## Checklist

## Reviewer Checklist
- Title is accurate

### DIFF
--- a/ddtrace/internal/telemetry/writer.py
+++ b/ddtrace/internal/telemetry/writer.py
@@ -335,8 +335,7 @@ class TelemetryWriter(PeriodicService):
         if self._client._agentless == enabled:
             return
 
-        with self._lock:
-            self._client = _TelemetryClient(enabled)
+        self._client = _TelemetryClient(enabled)
 
     def _is_running(self):
         # type: () -> bool
--- a/ddtrace/internal/telemetry/writer.py
+++ b/ddtrace/internal/telemetry/writer.py
@@ -48,6 +48,7 @@ from .constants import TELEMETRY_AGENT_HOST
 from .constants import TELEMETRY_AGENT_PORT
 from .constants import TELEMETRY_AGENT_URL
 from .constants import TELEMETRY_ANALYTICS_ENABLED
+from .constants import TELEMETRY_APM_PRODUCT
 from .constants import TELEMETRY_CLIENT_IP_ENABLED
 from .constants import TELEMETRY_CRASHTRACKING_ALT_STACK
 from .constants import TELEMETRY_CRASHTRACKING_AVAILABLE
@@ -117,7 +118,6 @@ from .metrics import RateMetric
 from .metrics_namespaces import MetricNamespace
 from .metrics_namespaces import NamespaceMetricType  # noqa:F401
 
-
 log = get_logger(__name__)
 
 
@@ -143,6 +143,7 @@ class _TelemetryClient:
         self._telemetry_url = self.get_host(config._dd_site, agentless)
         self._endpoint = self.get_endpoint(agentless)
         self._encoder = JSONEncoderV2()
+        self._agentless = agentless
 
         self._headers = {
             "Content-Type": "application/json",
@@ -264,6 +265,7 @@ class TelemetryWriter(PeriodicService):
         self._configuration_queue = {}  # type: Dict[str, Dict]
         self._lock = forksafe.Lock()  # type: forksafe.ResetObject
         self._imported_dependencies: Dict[str, Distribution] = dict()
+        self._product_enablement = {product.value: False for product in TELEMETRY_APM_PRODUCT}
 
         self.started = False
 
@@ -326,6 +328,15 @@ class TelemetryWriter(PeriodicService):
             self.stop()
         else:
             self.status = ServiceStatus.STOPPED
+
+    def enable_agentless_client(self, enabled=True):
+        # type: (bool) -> None
+
+        if self._client._agentless == enabled:
+            return
+
+        with self._lock:
+            self._client = _TelemetryClient(enabled)
 
     def _is_running(self):
         # type: () -> bool
@@ -430,7 +441,7 @@ class TelemetryWriter(PeriodicService):
             raise ValueError("Unknown configuration item: %s" % cfg_name)
         return name, value, item.source()
 
-    def app_started(self, register_app_shutdown=True):
+    def _app_started(self, register_app_shutdown=True):
         # type: (bool) -> None
         """Sent when TelemetryWriter is enabled or forks"""
         if self._forked or self.started:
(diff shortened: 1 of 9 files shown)


ddtrace/llmobs/_llmobs.py
```diff
@@ -167,10 +168,10 @@ def enable(
                     "DD_SITE is required for sending LLMObs data when agentless mode is enabled. "
                     "Ensure this configuration is set before running your application."
                 )
-            if not os.getenv("DD_INSTRUMENTATION_TELEMETRY_ENABLED"):
-                config._telemetry_enabled = False
-                log.debug("Telemetry disabled because DD_LLMOBS_AGENTLESS_ENABLED is set to true.")
-                telemetry.telemetry_writer.disable()
+            if telemetry.telemetry_writer.enable():
+                telemetry.telemetry_writer.enable_agentless_client()
+                telemetry.telemetry_writer.product_activated (TELEMETRY_APM_PRODUCT.LLMOBS, True)
```
mabdinur, line 173: We don't need to check if telemetry is enabled. As long as `DD_INSTRUMENTATION_TELEMETRY_ENABLED` is NOT set to `False` telemetry will be enabled when the first event is queued.
 
```suggestion
            telemetry.telemetry_writer.enable_agentless_client()
            telemetry.telemetry_writer.product_activated (TELEMETRY_APM_PRODUCT.LLMOBS, True)
```

ddtrace/internal/writer/writer.py
```diff
@@ -321,7 +321,6 @@ def _send_payload(self, payload, count, client):
 
     def write(self, spans=None):
         # Queues an app-started event before the first ci-visibility/llmobs/trace payload is sent
-        telemetry_writer.app_started()
```
mabdinur, line 324: We should also remove ` telemetry_writer.app_started()` from the profiler

ddtrace/internal/telemetry/writer.py
```diff
@@ -761,6 +803,10 @@ def _generate_logs_event(self, logs):
         self.add_event({"logs": list(logs)}, TELEMETRY_TYPE_LOGS)
 
     def periodic(self, force_flush=False, shutting_down=False):
+
```
mabdinur, line 806: nit: whitespace
```suggestion
```

ddtrace/internal/telemetry/writer.py
```diff
@@ -264,6 +265,7 @@ def __init__(self, is_periodic=True, agentless=None):
         self._configuration_queue = {}  # type: Dict[str, Dict]
         self._lock = forksafe.Lock()  # type: forksafe.ResetObject
         self._imported_dependencies: Dict[str, Distribution] = dict()
+        self._product_enablement = {product.value: False for product in TELEMETRY_APM_PRODUCT}
```
mabdinur, line 268: With this approach could we report that a product is disabled when it is actually enabled?

ddtrace/internal/telemetry/writer.py
```diff
@@ -327,6 +329,14 @@ def disable(self):
         else:
             self.status = ServiceStatus.STOPPED
 
+    def enable_agentless_client(self, enabled=True):
+        # type: (bool) -> None
+
+        if self._client._agentless == enabled:
+            return
+
+        self._client = _TelemetryClient(enabled)
```
mabdinur, line 338: What's the role of the agent when collection telemetry events? Should we make the default mode agent-less? Would we lose meaningful data?

ddtrace/internal/telemetry/writer.py
```diff
@@ -627,6 +646,29 @@ def _app_dependencies_loaded_event(self, newly_imported_deps: List[str]):
...
+        payload = {
+            "products": {
+                product.value: {
+                    "version": config.version,
+                    "enabled": enabled,
+                }
+            }
+        }
+
+        self.add_event(payload, "app-product-change")
```
mabdinur, line 670: We can run into a scenario where a product is enabled, disabled, and then enabled again. This will generate 3 app-product-change payloads. Can we generate the app-product-change payload in `TelemetryWriter.periodic(..)` (similar to how we handle integrations). This way we can batch multiple products into one payload.

```suggestion
```

Yun-Kim (commented):
I think I follow most of the changes in this PR, but I think this PR would benefit from a couple things:
1. Split out LLMObs telemetry into a separate PR, and keep this one focused on making app_started telemetry events more product-generic.
2. Can we add to the PR description so that it is easier to understand what this PR is achieving?

ddtrace/internal/writer/writer.py
omerli: done

ddtrace/internal/telemetry/writer.py
omerli: I think it make sense to start agentless if the API key is present during the tracer writer init time.
The only way we'd lose data is if the agent connection isn't initialized within 10s, but that's the same case as today and to my knowledge, it hasn't been a problem. If it is or ever becomes a problem, we can introduce some retry mechanism in the `periodic` function - something along the lines of adding a failure counter to each event and if we fail, we increment the counter and retry up to an agreed-upon threshold to prevent unbounded growth of event. 

For now I think I can just add my first suggestion if that sounds reasonable to you.

lmk

ddtrace/internal/telemetry/writer.py
omerli: Yes, that makes sense. I think this scenario is unlikely, but then again we cannot assume the logic flow of our users. 

I'll refactor and add an invalidation flag.

ddtrace/internal/telemetry/writer.py
mabdinur: Yes that sounds good. 

Also  in `TelemetryWriter._init__()` we can update the initialization of the agentless field to:  `agentless = config._dd_api_key is not None if agentless is None else agentless` (we should make this more readable 😓).   

With this approach we will always enable the agentless mode if `DD_API_KEY` is set. Otherwise we will default to sending telemetry to the agent. This should reduce the occurrence of dropped events. We can also do this in a follow up PR.

ddtrace/internal/telemetry/writer.py
```diff
@@ -537,12 +549,21 @@ def app_started(self, register_app_shutdown=True):
...
             self.remove_configuration("DD_APPSEC_SCA_ENABLED")
 
+        products = {
+            product: {
+                "version": _pep440_to_semver(),
+                "enabled": status
+            }
+            for product, status in self._product_enablement.items()
+        }
+
```
mabdinur, line 559: If we only want to report product change data once can we remove `app-product-change` and only send this products payload in app-started? 

```suggestion
        self._send_product_change_updates = False
```

PR #10214, opened 2024-08-14
chore(debugger): include locals/globals in evals

P403n1x87:
We make sure that local and global variables are available to expression evaluations on function probe exit.

## Checklist

## Reviewer Checklist
- Title is accurate

### DIFF
--- a/ddtrace/debugging/_signal/model.py
+++ b/ddtrace/debugging/_signal/model.py
@@ -1,4 +1,5 @@
 import abc
+from collections import ChainMap
 from dataclasses import dataclass
 from dataclasses import field
 from enum import Enum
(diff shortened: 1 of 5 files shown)


ddtrace/debugging/_signal/model.py
```diff
@@ -80,14 +81,16 @@ def _eval_condition(self, _locals: Optional[Dict[str, Any]] = None) -> bool:
 
         return False
 
-    def _enrich_args(self, retval, exc_info, duration):
-        _locals = list(self.args or _safety.get_args(self.frame))
+    def _enrich_locals(self, retval, exc_info, duration):
```
OmerRaviv, line 84: does the name `_enrich_locals` really makes sense for this method given the fact that it now always returns the arguments, the locals, the globals, and the @duration / @return / @exception psuedo variables?  Maybe we should call it something like `get_execution_context` or `capture_frame_state` or `build_debug_context` ?

OmerRaviv (approved):
LGTM

ddtrace/debugging/_signal/model.py
P403n1x87: Good point. I think this is tricky because we could potentially clash with other things. E.g. "execution context" could be mistaken for the _tracer_ execution context? I'll have a think and try to fix this in a follow-up.

PR #10224, opened 2024-08-14
chore: add sccache to testrunner docker image

taegyunkim:
Tested locally
1. modify docker-compose.yml 

```
     testrunner:
        build:
          context: ./docker
          dockerfile: Dockerfile
```

2. run tests using `riot run` 

## Checklist

## Reviewer Checklist
- Title is accurate

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -5,6 +5,7 @@ FROM debian:buster-20221219-slim
 
 ARG TARGETARCH
 ARG HATCH_VERSION=1.8.1
+ENV SCCACHE_VERSION=0.8.1
 
 # http://bugs.python.org/issue19846
 # > At the moment, setting "LANG=C" on a Linux system *fundamentally breaks Python 3*, and that's not OK.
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -5,6 +5,7 @@ FROM debian:buster-20221219-slim
 
 ARG TARGETARCH
 ARG HATCH_VERSION=1.8.1
+ENV SCCACHE_VERSION=0.8.1
```
brettlangdon, line 8: do we want just set `DD_USE_SCCACHE=1` here?

Do we also want to set a different/explicit sccache directory? since I assume we'll use a volume for the cache in the docker compose?

brettlangdon (approved):

docker/Dockerfile
taegyunkim: oh, thought it's turned on automatically if `sccache` binary is there. Good point. 

Hmm, yeah we'll be using the volume for the cache in the docker compose, instead of host one and possibly persisting over separate invocations of scripts/ddtest

PR #10235, opened 2024-08-14
chore(ci): add logging when pytest raised OSError on shutdown

brettlangdon:
This PR attempts to:

1. Grab more information about the consistent exception we are getting on pytest shutdown
2. Make this error no longer prevent a job from finishing

We monkeypatch the `_pytest.capture.FDCapture.done` method to wrap it in a `try/except` block and add logging when we get the exception, while also not re-raising the exception.

Since the job won't be failing anymore, we have to manually look for the error and debug on successful jobs.

Risks: if something with this change fails it may cause exceptions on non-internal jobs, but given the failure rate on internal jobs it can't be too much worse than what we have now.

## Checklist

## Reviewer Checklist
- Title is accurate

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -61,17 +61,19 @@ if os.environ.get("CI") == "true":
             except Exception as e:
                 import traceback
 
-                print("Failed to close FDCapture", e)
+                # Write to stderr since pytest captures and hides stdout by default
+                sys.stderr.write("Failed to close FDCapture: %s\n" % e)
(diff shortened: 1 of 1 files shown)


brettlangdon:
Is `print` going to be reliable here? would `sys.stderr` be better?

erikayasuda:
@brettlangdon Looks like @sanchda wrote on built off your PR 😂  I approved that one since it had additional logs that were helpful for seeing what the values were.

PR #10236, opened 2024-08-14
chore(integrations): deprecate public patch.py modules

mabdinur:
- Deprecates all modules that match the following format: `ddtrace/contrib/{integration_name}/patch.py` 
- Ensures deprecation warnings are not logged when integrations are patched
- 
## Checklist

## Reviewer Checklist
- Title is accurate


emmettbutler (changes requested):
Looks great, just needs a release note

emmettbutler (approved):

lievan (approved):

avara1986 (approved):

ddtrace/contrib/aiohttp/patch.py
```diff
@@ -1,4 +1,15 @@
...
-# TODO: deprecate and remove this module
+def __getattr__(name):
+    deprecate(
+        ("%s.%s is deprecated" % (__name__, name)),
+        category=DDTraceDeprecationWarning,
+    )
+
+    if name in globals():
+        return globals()[name]
+    raise AttributeError("%s has no attribute %s", __name__, name)
```
ZStriker19, line 15: Shouldn't we avoid raising errors basically always? Log instead?

wconti27 (approved):

ddtrace/contrib/aiohttp/patch.py
mabdinur: Here we are resecting the expected behavior of python modules. If a user accesses an attribute that doesn't exist (ex: ddtrace.contrib.aiohttp.munirsucks) then an attribute error should be raised. For all existing attributes a deprecation warning is logged.

PR #10250, opened 2024-08-16
ci(tracer): reduce the flakiness of the tracer tests

mabdinur:
- 85e63463e26cd20db968f4f26505875d78b4a62a Refactors all tracer tests that use `os.fork` to run in a process outside of pytest runner. Using `os.fork()` in pytest can lead to undefined behavior. 
- c294c9b76aa1415c55f07a93b4c283cd5b7d03f6 Ensures `Span.finish()` is called on all spans. Not finishing spans can lead to unexpected parenting in tests. It will also log a warning that might impact tests that validate stderr.
- https://github.com/DataDog/dd-trace-py/pull/10171/commits/69c8c40da7780ea0611770ebd38705d09daeca60 skips hour glass test
- Skips broken multiprocessing tests. PY312 dropped support for forking procceses in multithreaded applications: https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods . 
- Ensure tracer and severless tests do not rely on `DD_TRACE_AGENT_URL` being set to `""` in circleci configurations. These tests should always be run with the required configurations. 
- Skips PCR File container_info tests that always fails on MacOS. 
- Runs tracer tests with testagent. This ensures subprocess tests that submit trace payloads do not log 404 error logs. The presence of error and warning logs may cause subprocess tests to fail.

## Checklist

## Reviewer Checklist
- Title is accurate


.circleci/config.templ.yml
```diff
@@ -573,6 +573,14 @@ jobs:
       - run_test:
           pattern: "datastreams"
 
+  tracer:
+    <<: *machine_executor
+    parallelism: 10
+    steps:
+      - run_test:
+          pattern: "tracer"
+          snapshot: true
```
mabdinur, line 582: We need to support the `snapshot=True` configuration in `.gitlab/tests.yml` and then we can avoid adding the tracer test suite back to circleci.

```suggestion
  tracer:
```

erikayasuda (approved):

tests/tracer/test_tracer.py
```diff
@@ -1823,7 +1850,14 @@ def _target(span):
     assert len(spans) == 2
 
 
-def test_fork_manual_span_same_context(tracer):
+# Disable tracing to avoid sending traces to the agent
+# Datadog Agent is not available in CI
+@pytest.mark.subprocess(env={"DD_TRACE_ENABLED": "false"})
```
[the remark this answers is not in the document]
mabdinur: I experimented with running the test agent or a real Datadog agent in Gitlab but it felt like a weird thing to introduce in this PR. For now we can just avoid sending spans

tests/tracer/test_rand.py
```diff
@@ -88,6 +94,11 @@ def test_fork_no_pid_check():
 
 
 def test_fork_pid_check():
```
mabdinur, line 96:
```suggestion
@pytest.mark.subprocess()
def test_fork_pid_check():
```

ZStriker19 (approved):

PR #10260, opened 2024-08-16
ci: split the tracer suite up into ten smaller suites

emmettbutler:
This change attempts to resolve common `exit code -9` failures in the `tracer` test suite by splitting it into many smaller units. This is based on the hypothesis that the exit code indicates the process having been killed due to OOM.

## Checklist

## Reviewer Checklist
- Title is accurate


P403n1x87:
Could we have this split done automatically by the `scripts/gen_circleci_config.py` script instead? This would also avoid us having to manually update these suites when new sources are added.

emmettbutler:
@P403n1x87 sure! separate from this change, I'll look into having the script automatically split up large suites.

brettlangdon (approved):
nice

mabdinur:
We have quite a few tracer tests that generate spans but do not call span.finish. This is likely causing the memory issues described in the PR. This change should hopefully address this: https://github.com/DataDog/dd-trace-py/commit/c294c9b76aa1415c55f07a93b4c283cd5b7d03f6. 

I will get this change to a mergeable state ASAP

mabdinur:
> @mabdinur do you think there's a downside to this change that makes it not worth merging?

No concrete downsides. I am just afraid this change will make the tracer test suite harder to maintain. It would be nice if we could provide this parallelism without updating riot or suitespec.

tests/tracer/test_trace_utils.py
```diff
@@ -402,13 +402,16 @@ def test_set_http_meta(
...
     cookies,
-    appsec_enabled,
     span_type,
+    appsec_enabled,
 ):
     int_config.http.trace_headers(["my-header"])
     int_config.trace_query_string = True
     span.span_type = span_type
     with override_global_config({"_asm_enabled": appsec_enabled}):
+        from ddtrace.appsec import _handlers
```
ZStriker19, line 412: Was this just missing before and causing failures?

ZStriker19:
Not maybe a great reason, but a reason: The way this breaks up the tracer tests is kind of brutal with regards to rerunning them if they fail. There are now 90 tests for tracer so re-running if pieces fail sounds pretty cumbersome. Maybe we see if @mabdinur 's PR fixes the issue and if not, go this route?

PR #10270, opened 2024-08-17
chore(ci_visibility): add support for bitmap-based coverage data

romainkomorndatadog:
Per a new RFC, line-level coverage data is changing from using the current segment tuples to msgpack-encoded bytes (with little-endian ordering).

## CoverageLines

`CoverageLines` are introduced representation of coverage in the `ModuleCodeCollector`, providing a minimal, set-like API, and using a bytes-based internal storage (with bit-shifty fun to account for the desired little-endianness). A `from_list()` classmethod is added for convenience in test updates.

## Encoder changes

The new coverage format is stored using `Span.set_struct_tag()` with a dictionary similar to the existing segments-based one:
```python
{
  "files": [
    {
      "filename": "/src/flask/helpers.py",
      "bitmap": bytearray(b"..."),
    }
  ]
}
```

Paths are relative to the repo or workspace root, but prepended with a `/` to make them absolute (this is something that is already done by the backend, but doing it on the client means the backend is not modifying the tracer's payloads).

The encoder is temporarily updated to look for the `COVERAGE_TAG_NAME` in both the regular tags and struct tags.

## ModuleCodeCollector changes

The communication between threads and processes patched for coverage support is switched from JSON to pickling (so that bytes don't need to be converted to an array and then back to bytes). The pickling statements have the `nosec` comment because we know the source of the data. 

## Tests

Tests are updated to account for differences in plugin versions, and some utility methods are added to parse the two different coverage data formats into an intermediary format that can be checked for both without too much conditional hassle.

No release note is added because the internal code coverage collector and the v2 plugin are still unreleased, and the other changes are backwards-compatible.

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/internal/coverage/lines.py
```diff
@@ -0,0 +1,67 @@
+class CoverageLines:
+    def __init__(self, initial_size: int = 8):
```
vitor-de-araujo, line 2:
```suggestion
    def __init__(self, initial_size: int = 32):
```

ddtrace/internal/coverage/lines.py
```diff
@@ -0,0 +1,67 @@
...
+    def __copy__(self):
+        new_instance = CoverageLines()
+        new_instance._lines = self._lines.copy()
+        return new_instance
+
+    def __repr__(self):
+        return f"CoverageLines(num_lines={self._num_lines()})"
+
+    def _num_lines(self) -> int:
+        return sum([bin(byte).count("1") for byte in self._lines])
```
vitor-de-araujo, line 21:
```suggestion
        return sum(bin(byte).count("1") for byte in self._lines)
```

vitor-de-araujo (approved):

ddtrace/internal/coverage/lines.py
```diff
@@ -18,28 +28,30 @@ def __repr__(self):
...
     def add(self, line_number: int):
         lines_byte = line_number // 8
-        lines_bit = 2 ** (line_number % 8)
 
         if lines_byte >= len(self._lines):
             self._lines.extend(bytearray(lines_byte - len(self._lines) + 1))
 
+        # DEV this fun bit allows us to trick ourselves into little-endianness, which is what the backend wants to see
+        # in bytes
+        lines_bit = 2 ** (7 - (line_number % 8))
```
vitor-de-araujo, line 41: I still think it would be clearer to do `0b10000000 >> (line_number % 8)` (you can think about it visually without even thinking about bits, as moving the `1` to where you want it to be). But this is Fine™.

ddtrace/internal/coverage/lines.py
```diff
@@ -18,28 +28,30 @@ def __repr__(self):
...
-        return "".join([str(n) for n in self.to_list()])
+    def to_bytes_string(self) -> str:
+        return " ".join(bin(byte)[2:].zfill(8) for byte in self._lines)
 
     def to_list(self) -> list[int]:
         lines = []
         for _byte in self._lines:
-            for _bit in range(8):
+            for _bit in range(7, -1, -1):
                 if _byte & (1 << _bit):
```
vitor-de-araujo, line 55: We can also get rid of this trick if we do it with `>>`:
```suggestion
            for _bit in range(8):
                if _byte & (0b10000000 >> _bit):
```

ddtrace/internal/coverage/lines.py
```diff
@@ -0,0 +1,80 @@
...
+        # in bytes
+        lines_bit = 2 ** (7 - (line_number % 8))
+        self._lines[lines_byte] |= lines_bit
+
+    def add_lines(self, lines: list[int]):
+        for line in lines:
+            self.add(line)
+
+    def to_bytes_string(self) -> str:
+        return " ".join(bin(byte)[2:].zfill(8) for byte in self._lines)
```
vitor-de-araujo, line 49: We can also do this, but I don't know if it's an improvement:
```suggestion
        return " ".join(f"{byte:08b}" for byte in self._lines)
```

romainkomorndatadog: Resolving because I'm removing this.

ddtrace/contrib/pytest/_plugin_v2.py
```diff
@@ -96,25 +97,14 @@ def _handle_collected_coverage(test_id, coverage_collector) -> None:
...
-            file_path = file_path.resolve()
-
-        sorted_lines = sorted(covered_lines)
-
-        collapsed_ranges = collapse_ranges(sorted_lines)
-        file_segments = []
-        for file_segment in collapsed_ranges:
-            file_segments.append((file_segment[0], file_segment[1]))
-        coverage_data[file_path] = file_segments
+        # Collapse ranges to avoid sending a large amount of data
```
romainkomorndatadog, line 106:
```suggestion
```

ddtrace/internal/ci_visibility/api/ci_base.py
```diff
@@ -84,7 +84,7 @@ def _require_not_finished(func):
     @functools.wraps(func)
     def wrapper(self, *args, **kwargs):
         if self.is_finished():
-            log.warning("Method %s called on item %s, but it is already finished", func, self)
+            log.warning("Method %s called on item %s, but it is already finished", func, self, args)
```
romainkomorndatadog, line 87:
```suggestion
            log.warning("Method %s called on item %s, but it is already finished", func, self)
```

ddtrace/internal/module.py
```diff
@@ -214,20 +214,6 @@ def call_back(self, module: ModuleType) -> None:
         for callback in self.callbacks.values():
             callback(module)
 
-    def import_exception_call_back(self, module: ModuleType) -> None:
```
emmettbutler, line 217: Why is this removed?

emmettbutler (approved):

ddtrace/internal/coverage/lines.py
```diff
@@ -0,0 +1,72 @@
...
+    def _num_lines(self) -> int:
+        return sum(_bit_count(byte) for byte in self._lines)
+
+    def add(self, line_number: int):
+        lines_byte = line_number // 8
+
+        if lines_byte >= len(self._lines):
+            self._lines.extend(bytearray(lines_byte - len(self._lines) + 1))
+
+        # DEV this fun bit allows us to trick ourselves into little-endianness, which is what the backend wants to see
```
vitor-de-araujo, line 39: This fun bit, I see what you did there

tests/coverage/utils.py
```diff
@@ -12,4 +12,11 @@ def _get_relpath_dict(rootpath: str, dict_to_update: dict):
...
     relative path strings
     """
-    return {_get_relpath_str(Path(rootpath), Path(path)): lines for path, lines in dict_to_update.items()}
+    return {
+        _get_relpath_str(Path(rootpath), Path(path)): set(lines.to_sorted_list())
+        for path, lines in dict_to_update.items()
+    }
+
+
+def _get_lines_set_from_coverage_lines():
```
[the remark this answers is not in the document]
romainkomorndatadog: It extracts the set of lines from a `CoverageLines` object. It's a convenience function which is used in tests so that we don't have to rewrite every test if the underlying coverage representation changes.

PR #10315, opened 2024-08-21
feat(llmobs): support setting ml app & timestamp when submitting evaluation metrics

lievan:
The evaluation metric writer was still writing eval metrics to `/unstable` endpoint

Update the writer to use `/intake` endpoint, which also gives users the ability to set `timestamp_ms` and `ml_app` fields on an evaluation metric.

There are two important differences to the `/intake` endpoint
1. `ml_app` is required
2. `timestamp_ms` is required

These aren't breaking changes for users submitting evals on the SDK side since these values can be inferred by the SDK before we make the POST request to the new endpoint. The only change that impacts users is the ability to now supply those optional fields via `submit_evaluation`. If not supplied, we'll infer it.

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/llmobs/_writer.py
```diff
@@ -61,6 +61,7 @@ class LLMObsEvaluationMetricEvent(TypedDict, total=False):
     categorical_value: str
     numerical_value: float
     score_value: float
+    ml_app: str
```
sabrenner, line 64: do we want to add `timestamp_ms` here as well?

sabrenner (approved):
🚀

quinna-h (approved):

ddtrace/llmobs/_writer.py
lievan: Ah yes, thanks for catching

PR #10360, opened 2024-08-23
feat(tracer): add Span Pointer support

apiarian-datadog:
Span Pointers can be used in situations where Span Context cannot be propagated to try to create Span Links after the span is ingested. Though they have some significant limitations, they are sometimes the only way to try to link spans.

This PR adds the basic `Span.set_pointer` functionality. We will be following up with an initial S3 object implementation to support Serverless workflows.

## Checklist

## Reviewer Checklist
- Title is accurate

### DIFF
--- a/ddtrace/_trace/_span_link.py
+++ b/ddtrace/_trace/_span_link.py
@@ -35,6 +35,43 @@ def _id_not_zero(self, attribute_name, value):
         raise ValueError(f"{attribute_name} must be > 0. Value is {value}")
 
 
+def _span_link_shaped_dict(
+    *
+    trace_id: int,
+    span_id: int,
+    attributes: dict,
+    dropped_attributes_count: int,
+    tracestate: Optional[str],
+    flags: Optional[int]
+) -> dict:
+    d = {
+        "trace_id": "{:032x}".format(trace_id),
+        "span_id": "{:016x}".format(span_id),
+    }
+    if attributes:
+        d["attributes"] = {}
+        for k, v in attributes.items():
+            # flatten all values with the type list, tuple and set
+            for k1, v1 in flatten_key_value(k, v).items():
+                # convert all values to string
+                if isinstance(v1, str):
+                    d["attributes"][k1] = v1
+                elif isinstance(v1, bool):
+                    # convert bool to lowercase string to be consistent with json encoding
+                    d["attributes"][k1] = str(v1).lower()
+                else:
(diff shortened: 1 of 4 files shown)


ddtrace/_trace/span.py
```diff
@@ -634,6 +650,24 @@ def set_link(
...
+        pointers_key = PointersKey(hash, direction)
+
+        if pointers_key in self._pointers:
+            log.debug(
+                "Span %d already has a pointer %s. Overwriting existing pointer: %s",
+                self.span_id,
+                str(pointers_key),
+                str(self._pointers[pointers_key]),
+            )
+        self._pointers[pointers_key] = SpanPointer(kind=kind, direction=direction, hash=hash, extra_attributes=extra_attributes)
```
apiarian-datadog, line 669: since `SpanPointer`s are serialized to `span_links`, it might be nice to just store them in the `_links` property instead. But that runs up against the checks we have on `SpanLink`s (the trace and span ids must not be `0` or `None`, and we can only have one link per `span_id`). So we store them separately.

ddtrace/_trace/span.py
```diff
@@ -634,6 +650,24 @@ def set_link(
...
+    def set_pointer(
+        self,
+        kind: str,
+        direction: SpanPointerDirection,
+        hash: SpanPointerHash,
+        **extra_attributes: Any,
+    ) -> None:
+        pointers_key = PointersKey(hash, direction)
+
+        if pointers_key in self._pointers:
```
apiarian-datadog, line 662: this may actually not be necessary. though it may be strange to have multiple pointers to the same entity and direction, it's not necessarily a bug. so we might want to just keep track of the `SpanPointer`s as a list instead of a dictionary.

ddtrace/_trace/_span_link.py
```diff
@@ -35,6 +35,43 @@ def _id_not_zero(self, attribute_name, value):
         raise ValueError(f"{attribute_name} must be > 0. Value is {value}")
 
 
+def _span_link_shaped_dict(
```
apiarian-datadog, line 38: all this just to avoid the `_id_not_zero` checks =/

apiarian-datadog (commented):
a sketch to get the conversation started.

apiarian-datadog:
closing in favor of https://github.com/DataDog/dd-trace-py/pull/10594 and its followups

PR #10361, opened 2024-08-23
ci: move internal suite to gitlab

emmettbutler:
This change moves the `internal` test suite from circleci to gitlab for billing purposes. There are a handful of tests that fail locally under this new configuration, because I can't figure out how to configure a docker-compose alias for the ddagent service that matches the alias that gitlab seems to require. I think this is an acceptable price to pay for the benefit this change provides.

## Checklist

## Reviewer Checklist
- Title is accurate


emmettbutler:
@erikayasuda @brettlangdon this is reviewable

.gitlab/tests/core.yml
```diff
@@ -0,0 +1,6 @@
+internal:
+  extends: .test_base_riot
+  parallel: 4
```
brettlangdon, line 3:
```suggestion
```

redundant, unless you want to keep it at `4` to be explicit?

.gitlab/tests.yml
```diff
@@ -15,6 +15,26 @@ variables:
...
+      alias: ddagent
+      variables:
+        DD_HOSTNAME: ddagent
+        DD_REMOTE_CONFIGURATION_ENABLED: true
+        DD_SITE: datadoghq.com
+        DD_API_KEY: invalid_but_this_is_fine
+        DD_REMOTE_CONFIGURATION_KEY: invalid_but_this_is_fine
+        DD_REMOTE_CONFIGURATION_REFRESH_INTERVAL: 5s
+        DD_DOGSTATSD_NON_LOCAL_TRAFFIC: true
+    - name: ghcr.io/datadog/dd-apm-test-agent/ddapm-test-agent:v1.17.0
```
brettlangdon, line 29: do we need an alias here too? or does `localhost` work for this?

tests/internal/test_serverless.py
```diff
@@ -79,8 +78,6 @@ def test_not_azure_function_consumption_plan():
     assert in_azure_function() is False
 
 
-# DEV: Run this test in a subprocess to avoid messing with global sys.modules state
```
brettlangdon, line 82: is this still not true?

tests/internal/test_serverless.py
```diff
@@ -120,3 +117,5 @@ def find_spec(self, fullname, *args):
     import ddtrace
     import ddtrace.contrib.aws_lambda  # noqa:F401
     import ddtrace.contrib.psycopg  # noqa:F401
+
+    sys.meta_path = sys.meta_path[1:]
```
brettlangdon, line 121: `try/finally` ?

otherwise could this pollute the state of other jobs if an exception is raised?

.gitlab/tests/core.yml
```diff
@@ -0,0 +1,6 @@
...
+  variables:
+    SUITE_NAME: "internal"
+    SERVICES: "testagent"
```
brettlangdon, line 6: is this used for anything?

it seems like we are adding the agent and test agent to every riot job?

.gitlab/tests.yml
emmettbutler: The tests pass as is. Other suites i'm testing with suggest that it does need an alias, but not certain as of now.

tests/internal/test_serverless.py
emmettbutler: I removed the subprocess marker, so it no longer runs in a subprocess, making the comment irrelevant

.gitlab/tests/core.yml
emmettbutler: no, at present it's not

tests/internal/test_serverless.py
brettlangdon: yeah, but do we need it? the comment says to avoid polluting `sys.modules` with the imports, but it is ok that we are polluting `sys.modules`?

brettlangdon (approved):

PR #10389, opened 2024-08-26
feat: support baggage

rachelyangdog:
First PR introducing baggage support

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/internal/constants.py
```diff
@@ -9,13 +9,15 @@
 PROPAGATION_STYLE_B3_SINGLE = "b3"
 _PROPAGATION_STYLE_W3C_TRACECONTEXT = "tracecontext"
 _PROPAGATION_STYLE_NONE = "none"
-_PROPAGATION_STYLE_DEFAULT = "datadog,tracecontext"
+_PROPAGATION_STYLE_DEFAULT = "datadog,tracecontext,baggage"
```
mabdinur, line 12: Do we need to make this change? Baggage shouldn't be used to propagate distributed tracing headers. Adding it to propagation styles looks a bit odd

ddtrace/internal/opentelemetry/context.py
```diff
@@ -72,6 +85,18 @@ def get_current(self):
...
+
+        if isinstance(ddactive, DDSpan):
+            dd_baggage = ddactive.context._baggage
+        elif isinstance(ddactive, DDContext):
+            dd_baggage = ddactive._baggage
+        else:
+            dd_baggage = {}
+
+        # getting current active baggage
+        for key, value in dd_baggage.items():
```
mabdinur, line 97: Same here we should add test coverage for this

ddtrace/propagation/http.py
```diff
@@ -885,12 +887,57 @@ def _inject(span_context, headers):
         return headers
 
 
+class _BaggageHeader:
+    @staticmethod
+    def _encode_key(key):
+        safe_characters = "ABCDEFGHIJKLMNOPQRSTUVWXYZ" "abcdefghijklmnopqrstuvwxyz" "0123456789" "!#$%&'*+-.^_`|~"
```
mabdinur, line 893: We should make this a constant. We can define it on the class level

ddtrace/propagation/http.py
```diff
@@ -885,12 +887,57 @@ def _inject(span_context, headers):
...
+        return encoded
+
+    @staticmethod
+    def _inject(span_context, headers):
+        baggage_items = span_context._baggage.items()
+        if not baggage_items:
+            return
+
+        header_value = ",".join(
+            f"{_BaggageHeader._encode_key(str(key).strip())}={_BaggageHeader._encode_value(str(value).strip())}"
```
mabdinur, line 911: Should we raise an unhandled exception if value can not be converted to a string? How should we handle dicts,arrays,booleans,etc. The format needs to be consistent across languages (ex: encoding `True` vs `true`)

ddtrace/propagation/http.py
```diff
@@ -885,12 +887,57 @@ def _inject(span_context, headers):
...
+    @staticmethod
+    def _extract(headers):
+        header_value = headers.get("baggage")
+        if not header_value:
+            return None
+
+        baggage = {}
+        baggages = header_value.split(",")
+        for key_value in baggages:
+            key, value = key_value.split("=", 1)
```
mabdinur, line 926: What happens if key_value contains multiple `=` values? Is this behavior called out in the RFC?

ddtrace/propagation/http.py
```diff
@@ -885,12 +887,57 @@ def _inject(span_context, headers):
...
+            key = urllib.parse.unquote(key.strip())
+            value = urllib.parse.unquote(value.strip())
```
mabdinur, line 928: How should this deconding/unquoting handle errors?

ddtrace/internal/constants.py
rachelyangdog: The RFC says "The default values for these settings will change to include the baggage propagator: DD_TRACE_PROPAGATION_STYLE=datadog,tracecontext,baggage" So I was going off based on that

ddtrace/propagation/http.py
rachelyangdog: Keys cannot contain "=" so if there is an "=" it would be encoded.

rachelyangdog: In the case that there is "=" in value (which is allowed), I think the `split` function should split on the first "="

ddtrace/propagation/http.py
rachelyangdog: When "extracting baggage from propagation headers, they may encounter malformed header contents." When this occurs, we "should ignore the entire header." So instead of an error, we could potentially just ignore it? As of now, I think it would just error.

ddtrace/_trace/span.py
```diff
@@ -508,6 +509,15 @@ def _get_baggage_item(self, key: str) -> Optional[Any]:
         """Gets a baggage item from the span context of this span."""
         return self.context._get_baggage_item(key)
 
+    def _get_all_baggage_items(self) -> str:
```
mabdinur, line 512: Can we remove all the baggage getters and setters from the span object. We can use `span.context._get/set_baggage` everywhere

ddtrace/internal/opentelemetry/context.py
```diff
@@ -30,18 +31,28 @@ def attach(self, otel_context):
...
                 self._ddcontext_provider.activate(ddcontext)
             else:
+                ddcontext = None
                 log.error(
                     "Programming ERROR: ddtrace does not support activiting spans with the type: %s. Please open a "
                     "github issue at: https://github.com/Datadog/dd-trace-py and set DD_TRACE_OTEL_ENABLED=True.",
                     type(otel_span),
                 )
 
+            # get current open telemetry baggage and store it on the datadog context object
```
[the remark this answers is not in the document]
mabdinur: We should also make sure the opentelemetry baggage api getters and setters work as expected and create/set/remove values on datadog context objects: https://github.com/open-telemetry/opentelemetry-python/blob/bd51fcb7a3afb08aec975e7302c9cc36060bcacc/opentelemetry-api/src/opentelemetry/baggage/__init__.py#L53-#L117

ddtrace/propagation/http.py
```diff
@@ -885,12 +887,58 @@ def _inject(span_context, headers):
...
+        return encoded
+
+    @staticmethod
+    def _inject(span_context, headers):
+        baggage_items = span_context._baggage.items()
+        if not baggage_items:
+            return
+
+        header_value = ",".join(
+            f"{_BaggageHeader._encode_key(str(key).strip())}={_BaggageHeader._encode_value(str(value).strip())}"
```
mabdinur, line 912: Can we move `str(....).strip()`  to _encode_value and/or _encode_key

ddtrace/propagation/http.py
```diff
@@ -885,12 +887,58 @@ def _inject(span_context, headers):
...
+        encoded = urllib.parse.quote(key, safe=_BaggageHeader.safe_characters_key)
+        return encoded
+
+    def _encode_value(value):
+        encoded = urllib.parse.quote(value, safe=_BaggageHeader.safe_characters_value)
+        return encoded
+
+    @staticmethod
+    def _inject(span_context, headers):
+        baggage_items = span_context._baggage.items()
```
mabdinur, line 907: We should use the span_context getters and setters here.

ddtrace/propagation/http.py
```diff
@@ -885,12 +887,58 @@ def _inject(span_context, headers):
...
+class _BaggageHeader:
+    safe_characters_key = "ABCDEFGHIJKLMNOPQRSTUVWXYZ" "abcdefghijklmnopqrstuvwxyz" "0123456789" "!#$%&'*+-.^_`|~"
+    safe_characters_value = (
+        "ABCDEFGHIJKLMNOPQRSTUVWXYZ" "abcdefghijklmnopqrstuvwxyz" "0123456789" "!#$%&'()*+-./:<>?@[]^_`{|}~"
+    )
+
+    @staticmethod
+    def _encode_key(key):
+        encoded = urllib.parse.quote(key, safe=_BaggageHeader.safe_characters_key)
+        return encoded
```
mabdinur, line 899:
```suggestion
        return urllib.parse.quote(key, safe=_BaggageHeader.safe_characters_key)
```

ddtrace/propagation/http.py
```diff
@@ -885,12 +887,58 @@ def _inject(span_context, headers):
...
+
+        header_value = ",".join(
+            f"{_BaggageHeader._encode_key(str(key).strip())}={_BaggageHeader._encode_value(str(value).strip())}"
+            for key, value in baggage_items
+        )
+
+        headers["baggage"] = header_value
+
+    @staticmethod
+    def _extract(headers):
```
mabdinur, line 919: Can we add typing to this method

ddtrace/propagation/http.py
```diff
@@ -915,8 +963,19 @@ def _extract_configured_contexts_avail(normalized_headers):
...
             style_w_ctx = styles_w_ctx[contexts.index(context)]
+
+            # special case where baggage is first in propagation styles
+            # why would you do that? I don't know, but it's possible
+            if style_w_ctx[0] == _PROPAGATION_STYLE_BAGGAGE:
+                baggage_context = contexts[0]
+                contexts.append(baggage_context)
+                del contexts[0]
+                styles_w_ctx.append(_PROPAGATION_STYLE_BAGGAGE)
+                del style_w_ctx[0]
```
mabdinur, line 977: Should just short circuit here? 
```suggestion
                continue
```

mabdinur, line 977: Can we handle this outside of the for-loop? baggage should never be the `primary_context`

ddtrace/internal/constants.py
lucaspimentel: It was suggested by @zacharycmontoya (and I agree) that we follow OpenTelemetry here, where the baggage propagator is another propagator that is enabled/disabled using `OTEL_PROPAGATORS`, along with `tracecontext`, `b3`, etc.

See https://opentelemetry.io/docs/languages/sdk-configuration/general/#otel_propagators

ddtrace/propagation/http.py
lucaspimentel: Yup, when extracting headers, `key1=foo=bar` should be interpreted as:
```
key:   "key1"
value: "foo=bar"
```

ddtrace/propagation/http.py
rachelyangdog: In the RFC it says that users can also disable other propagation styles (tracecontext or datadog) and enable only the baggage propagator which means baggage could be the `primar_context` in certain cases. Considering this, I think we can leave the code inside the for-loop. Let me know your thoughts because I think there are may be better ways to go about this.

rachelyangdog: handle baggage separately entirely

ddtrace/propagation/http.py
```diff
@@ -1147,7 +1137,14 @@ def my_controller(url, headers):
                     context = HTTPPropagator._resolve_contexts(contexts, styles_w_ctx, normalized_headers)
                     if config.propagation_http_baggage_enabled is True:
                         _attach_baggage_to_context(normalized_headers, context)
-                    return context
+
+            # baggage
+            if config._propagation_style_extract and _PROPAGATION_STYLE_BAGGAGE in config._propagation_style_extract:
```
ZStriker19, line 1141: Is there ever an instance when we won't have a value for `config._propagation_style_extract`? It looks like the rest of the method trusts that at the very least it's an empty list. So I think we can drop that check.

ddtrace/propagation/http.py
```diff
@@ -1138,7 +1128,7 @@ def my_controller(url, headers):
                     context = propagator._extract(normalized_headers)  # type: ignore
                     if config.propagation_http_baggage_enabled is True:
```
ZStriker19, line 1128: Doe we need to make any changes to this behavior? How does your adding of baggage intersect with this?

ddtrace/propagation/http.py
```diff
@@ -1147,7 +1137,14 @@ def my_controller(url, headers):
                     context = HTTPPropagator._resolve_contexts(contexts, styles_w_ctx, normalized_headers)
                     if config.propagation_http_baggage_enabled is True:
                         _attach_baggage_to_context(normalized_headers, context)
-                    return context
+
+            # baggage
+            if config._propagation_style_extract and _PROPAGATION_STYLE_BAGGAGE in config._propagation_style_extract:
+                baggage_context = _BaggageHeader._extract(normalized_headers)
+                if baggage_context:
+                    context._baggage = baggage_context._baggage
```
ZStriker19, line 1144: If there are no valid headers for the other propagation styles specified I think we can end up in a situation where context is None, in which case this would throw an error. We should check if there's a context, if so add baggage, if not return the baggage context value I think?

ddtrace/propagation/http.py
rachelyangdog: Yeah, I originally didn't have the check for it, but Pylance was getting really angry at me saying that `config._propagation_style_extract` could be None.

ddtrace/propagation/http.py
rachelyangdog: I think there should be minimal interaction. If `config.propagation_http_baggage_enabled` is True, then the values will be merged into `_baggage`. The headers are different so they can be propagated separately. The idea is to not touch `config.propagation_http_baggage_enabled` because it would lead to a breaking change (RFC).

riotfile.py
```diff
@@ -2389,9 +2389,6 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
             name="opentelemetry",
             command="pytest {cmdargs} tests/opentelemetry",
             pys=select_pys(min_version="3.8"),
-            # DD_TRACE_OTEL_ENABLED must be set to true before ddtrace is imported
```
emmettbutler, line 2392: Did this stop being true?

tests/opentelemetry/test_context.py
```diff
@@ -3,12 +3,24 @@
 import time
 
 import opentelemetry
+
+# from opentelemetry.baggage import clear
```
emmettbutler, line 7: I recommend removing these changes, opening a separate draft PR if you want to keep track of them.

releasenotes/notes/baggage-support-be7eed26293f1216.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    tracing: Introduces support for Baggage. Baggage is enabled by default. The ``context.set_baggage_item``, ``context.get_baggage_item``, ``context.get_all_baggage_items``,``context.remove_baggage_item``, and ``context.remove_all_baggage_items`` are internal methods provided for manual modifications to baggage.
```
emmettbutler, line 4: Referencing OpenTelemetry and linking to public Baggage documentation here would make this release note easier to understand for folks with limited context.

tests/tracer/test_span.py
```diff
@@ -96,22 +96,48 @@ def test_set_tag_bool(self):
         assert s.get_tags() == dict(true="True", false="False")
         assert len(s.get_metrics()) == 0
 
-    def test_set_baggage_item(self):
+    def testset_baggage_item(self):
```
emmettbutler, line 99:
```suggestion
    def test_set_baggage_item(self):
```

ddtrace/internal/opentelemetry/context.py
```diff
@@ -88,3 +107,39 @@ def _ddcontext_provider(self):
...
+#     Remove baggage item from the current context.
+#     """
+#     context = func(*args, **kwargs)
+#     name: str = get_argument_value(args, kwargs, 0, "name") # type: ignore
+#     ddcontext = ddtracer.current_trace_context()
+#     if ddcontext:
+#         ddcontext.remove_baggage_item(name)
+#     return context
+
+# def wrap_remove_all_baggage(func, args, kwargs):
```
emmettbutler, line 137: I recommend removing these commented bits.

ddtrace/propagation/http.py
```diff
@@ -1024,7 +1095,6 @@ def parent_call():
 
     @staticmethod
     def extract(headers):
-        # type: (Dict[str,str]) -> Context
```
emmettbutler, line 1027: Why'd you choose to remove this?

ddtrace/propagation/http.py
```diff
@@ -1068,7 +1139,17 @@ def my_controller(url, headers):
                     context = HTTPPropagator._resolve_contexts(contexts, styles_w_ctx, normalized_headers)
                     if config.propagation_http_baggage_enabled is True:
                         _attach_baggage_to_context(normalized_headers, context)
-                    return context
+
+            # baggage
```
emmettbutler, line 1143:
```suggestion
```
If it's important to mark this code block as "baggage-related", consider doing so by putting it into a meaningfully-named function rather than using a comment, which will become outdated.

emmettbutler (approved):

tests/tracer/test_propagation.py
```diff
@@ -2918,6 +2932,60 @@ def test_span_links_set_on_root_span_not_child(fastapi_client, tracer, fastapi_t
...
+            PROPAGATION_STYLE_B3_SINGLE,
+            _PROPAGATION_STYLE_W3C_TRACECONTEXT,
+        ],
+        {
+            "baggage": {"foo": "bar"},
+            "trace_id": VALID_DATADOG_CONTEXT["trace_id"],
+            "span_id": VALID_DATADOG_CONTEXT["span_id"],
+        },
+        {
+            "baggage": "foo=bar",
```
ZStriker19, line 2979: In a similar way that we have  HTTP_HEADER_PARENT_ID, HTTP_HEADER_TRACE_ID, etc, could we set `_HTTP_HEADER_BAGGAGE: Literal["baggage"] = "baggage"` at the top and use that here and in the rest of the spots.

tests/tracer/test_propagation.py
```diff
@@ -2918,6 +2932,60 @@ def test_span_links_set_on_root_span_not_child(fastapi_client, tracer, fastapi_t
...
+            "baggage": {"foo": "bar"},
+        },
+        {
+            HTTP_HEADER_TRACE_ID: "13088165645273925489",
+            HTTP_HEADER_PARENT_ID: "8185124618007618416",
+            "baggage": "foo=bar",
+        },
+    ),
+    (
+        "baggage_order",
```
ZStriker19, line 2965:
```suggestion
        "baggage_order_first",
```

releasenotes/notes/baggage-support-be7eed26293f1216.yaml
```diff
@@ -0,0 +1,3 @@
+features:
+  - |
+    tracing: Introduces support for Baggage as defined by the [OpenTelemetry specification](https://opentelemetry.io/docs/specs/otel/baggage/api/). Baggage is enabled by default. You can manually modify baggage using methods like `context.set_baggage_item`, `context.get_baggage_item`, `context.get_all_baggage_items`, `context.remove_baggage_item`, and `context.remove_all_baggage_items`.
```
ZStriker19, line 3:
```suggestion
    tracing: Introduces support for Baggage as defined by the [OpenTelemetry specification](https://opentelemetry.io/docs/specs/otel/baggage/api/). Baggage is enabled by default. You can manually modify baggage using methods like ``context.set_baggage_item``, ``context.get_baggage_item``, ``context.get_all_baggage_items``, ``context.remove_baggage_item``, and ``context.remove_all_baggage_items``.
```

ZStriker19 (approved):

ddtrace/propagation/http.py
```diff
@@ -885,12 +890,70 @@ def _inject(span_context, headers):
         return headers
 
 
+class _BaggageHeader:
+    """Helper class to inject/extract Baggage Headers"""
+
+    safe_characters_key = "ABCDEFGHIJKLMNOPQRSTUVWXYZ" "abcdefghijklmnopqrstuvwxyz" "0123456789" "!#$%&'*+-.^_`|~"
+    safe_characters_value = (
+        "ABCDEFGHIJKLMNOPQRSTUVWXYZ" "abcdefghijklmnopqrstuvwxyz" "0123456789" "!#$%&'()*+-./:<>?@[]^_`{|}~"
+    )
```
mabdinur, line 899: nit:
```suggestion
    SAFE_CHARACTERS_KEY = "ABCDEFGHIJKLMNOPQRSTUVWXYZ" "abcdefghijklmnopqrstuvwxyz" "0123456789" "!#$%&'*+-.^_`|~"
    SAFE_CHARACTERS_VALUE = (
        "ABCDEFGHIJKLMNOPQRSTUVWXYZ" "abcdefghijklmnopqrstuvwxyz" "0123456789" "!#$%&'()*+-./:<>?@[]^_`{|}~"
    )
```

ddtrace/propagation/http.py
```diff
@@ -885,12 +890,70 @@ def _inject(span_context, headers):
...
+    def _encode_value(value: str) -> str:
+        return urllib.parse.quote(str(value).strip(), safe=_BaggageHeader.safe_characters_value)
+
+    @staticmethod
+    def _inject(span_context: Context, headers: Dict[str, str]) -> None:
+        baggage_items = span_context._baggage.items()
+        if not baggage_items:
+            return
+
+        if len(baggage_items) > DD_TRACE_BAGGAGE_MAX_ITEMS:
```
mabdinur, line 915: We should log a warning here. We don't want to fail silently

ddtrace/propagation/http.py
```diff
@@ -885,12 +890,70 @@ def _inject(span_context, headers):
...
+            return
+
+        header_value = ",".join(
+            f"{_BaggageHeader._encode_key(key)}={_BaggageHeader._encode_value(value)}" for key, value in baggage_items
+        )
+
+        buf = bytes(header_value, "utf-8")
```
mabdinur, line 922: This operation can raise an esception. We should wrap the encoding of the header in a try-except

tests/opentelemetry/test_context.py
```diff
@@ -3,12 +3,24 @@
...
 import ddtrace
 from ddtrace.constants import MANUAL_DROP_KEY
 from ddtrace.constants import MANUAL_KEEP_KEY
 
+# from ddtrace.contrib.pytest.plugin import ddspan
+from tests.opentelemetry.flask_app import otel  # noqa: F401
+
+
+# from tests.utils import flaky
+
```
mabdinur, line 23: can we clean up these imports
```suggestion
```

releasenotes/notes/baggage-support-be7eed26293f1216.yaml
```diff
@@ -0,0 +1,3 @@
+features:
+  - |
+    tracing: Introduces support for Baggage as defined by the [OpenTelemetry specification](https://opentelemetry.io/docs/specs/otel/baggage/api/). Baggage is enabled by default. You can manually modify baggage using methods like ``context.set_baggage_item``, ``context.get_baggage_item``, ``context.get_all_baggage_items``, ``context.remove_baggage_item``, and ``context.remove_all_baggage_items``.
```
mabdinur, line 3: We need to format the link using rst:
```suggestion
    tracing: Introduces support for Baggage as defined by the `OpenTelemetry specification <https://opentelemetry.io/docs/specs/otel/baggage/api/>`_. 
```

Also we should simplify the release note. Any additional details about this feature should be added to our docs page.

mabdinur:
> ## Benchmarks
> Benchmark execution time: 2024-10-21 20:08:37
> 
> Comparing candidate commit [2ca070a](https://github.com/DataDog/dd-trace-py/commit/2ca070a2f81d5dd829668f6cc49e0eca8c5ac801) in PR branch `rachel.yang/baggage` with baseline commit [d45ebca](https://github.com/DataDog/dd-trace-py/commit/d45ebcad1ce7eef35ccbba28b76053f35983e828) in branch `main`.
> 
> Found 0 performance improvements and 19 performance regressions! Performance is the same for 346 metrics, 53 unstable metrics.
> 
> #### scenario:httppropagationextract-b3_headers
> * 🟥 `execution_time` [+1.509µs; +1.604µs] or [+19.588%; +20.822%]
> 
> #### scenario:httppropagationextract-b3_single_headers
> * 🟥 `execution_time` [+1.537µs; +1.614µs] or [+23.140%; +24.291%]
> 
> #### scenario:httppropagationextract-full_t_id_datadog_headers
> * 🟥 `execution_time` [+1.824µs; +1.966µs] or [+11.583%; +12.489%]
> 
> #### scenario:httppropagationextract-invalid_priority_header
> * 🟥 `execution_time` [+1.229µs; +1.294µs] or [+25.377%; +26.717%]
> 
> #### scenario:httppropagationextract-invalid_span_id_header
> * 🟥 `execution_time` [+1.199µs; +1.262µs] or [+24.575%; +25.863%]
> 
> #### scenario:httppropagationextract-invalid_tags_header
> * 🟥 `execution_time` [+1.270µs; +1.331µs] or [+26.390%; +27.658%]
> 
> #### scenario:httppropagationextract-invalid_trace_id_header
> * 🟥 `execution_time` [+1.234µs; +1.284µs] or [+25.481%; +26.498%]
> 
> #### scenario:httppropagationextract-medium_header_no_matches
> * 🟥 `execution_time` [+1.223µs; +1.325µs] or [+14.946%; +16.198%]
> 
> #### scenario:httppropagationextract-medium_valid_headers_all
> * 🟥 `execution_time` [+1.215µs; +1.313µs] or [+12.622%; +13.637%]
> 
> #### scenario:httppropagationextract-tracecontext_headers
> * 🟥 `execution_time` [+2.066µs; +2.256µs] or [+8.660%; +9.454%]
> 
> #### scenario:httppropagationextract-valid_headers_all
> * 🟥 `execution_time` [+1.210µs; +1.271µs] or [+24.958%; +26.216%]
> 
> #### scenario:httppropagationextract-valid_headers_basic
> * 🟥 `execution_time` [+1.189µs; +1.237µs] or [+26.959%; +28.045%]
> 
> #### scenario:httppropagationextract-wsgi_invalid_priority_header
> * 🟥 `execution_time` [+1.270µs; +1.337µs] or [+26.163%; +27.538%]
> 
> #### scenario:httppropagationextract-wsgi_invalid_tags_header
> * 🟥 `execution_time` [+1.244µs; +1.302µs] or [+25.462%; +26.649%]
> 
> #### scenario:httppropagationextract-wsgi_invalid_trace_id_header
> * 🟥 `execution_time` [+1.208µs; +1.260µs] or [+24.820%; +25.889%]
> 
> #### scenario:httppropagationextract-wsgi_medium_header_no_matches
> * 🟥 `execution_time` [+1.120µs; +1.212µs] or [+13.234%; +14.326%]
> 
> #### scenario:httppropagationextract-wsgi_medium_valid_headers_all
> * 🟥 `execution_time` [+1.237µs; +1.348µs] or [+12.505%; +13.633%]
> 
> #### scenario:httppropagationextract-wsgi_valid_headers_all
> * 🟥 `execution_time` [+1.240µs; +1.291µs] or [+25.241%; +26.285%]
> 
> #### scenario:httppropagationextract-wsgi_valid_headers_basic
> * 🟥 `execution_time` [+1.218µs; +1.283µs] or [+27.500%; +28.982%]

@rachelyangdog this is worth looking into. This feature could introduce a significant performance regression

PR #10396, opened 2024-08-27
chore(ci_visibility): pytest: only compute dependency coverage once per suite

romainkomorndatadog:
This updates the coverage collection mechanism for the new pytest plugin version to only collect dependency coverage once per suite.

`get_coverage_data()` is introduced as an API for `CISuite` and `CITest` items to fetch coverage data from the internal API.
`ModuleCodeCollector` gets a new `get_import_coverage_for_paths()` classmethod that returns coverage for the given paths.

No release note is added since this is still internal/unreleased code.

## Checklist

## Reviewer Checklist
- Title is accurate

### DIFF
--- a/ddtrace/contrib/pytest/_plugin_v2.py
+++ b/ddtrace/contrib/pytest/_plugin_v2.py
@@ -90,7 +90,7 @@ def _start_collecting_coverage() -> ModuleCodeCollector.CollectInContext:
 
 def _handle_collected_coverage(test_id, coverage_collector) -> None:
     # TODO: clean up internal coverage API usage
-    test_covered_lines = ModuleCodeCollector._instance._get_covered_lines(include_imported=True)
+    test_covered_lines = coverage_collector.get_covered_lines()
     coverage_collector.__exit__()
 
     record_code_coverage_finished(COVERAGE_LIBRARY.COVERAGEPY, FRAMEWORK)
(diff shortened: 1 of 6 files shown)


ddtrace/internal/ci_visibility/api/ci_coverage_data.py
```diff
@@ -27,6 +27,9 @@ def __init__(self) -> None:
     def __bool__(self):
         return bool(self._coverage_data)
 
+    def get_data(self):
```
vitor-de-araujo, line 30: Missing type annotation :eyes:

ddtrace/contrib/pytest/_plugin_v2.py
```diff
@@ -270,6 +277,7 @@ def _pytest_runtest_protocol_post_yield(item, nextitem, coverage_collector):
         if CISuite.is_item_itr_skippable(suite_id) and not CISuite.was_forced_run(suite_id):
             CISuite.mark_itr_skipped(suite_id)
         else:
+            _handle_coverage_dependencies(suite_id)
             CISuite.finish(suite_id)
         if nextitem is None or (next_test_id is not None and next_test_id.parent_id.parent_id != module_id):
```
vitor-de-araujo, line 282: Not related to this PR, but didn't we have an issue where if the tests are randomized we can't rely on this order? What was the solution to this (or did I misunderstand the problem)?

vitor-de-araujo (approved):

ddtrace/contrib/pytest/_plugin_v2.py
romainkomorndatadog: We do, yes, but in this case, it only means that we would do this more expensive computation more times than necessary.

In the new API, the coverage data is accumulated until the suite is `.finish()`ed , and that only happens if all the suite's children are finished, or if the finish is forced (this is something I introduced to be part of the workaround with randomization, though it's not the whole solution itself).

What we could do is introduce something like `CISuite.has_unfinished_children() -> bool` that would return `False`.

I thought about it before but didn't because `.finish()` already does that check for items with children, so it was a little redundant.

PR #10409, opened 2024-08-27
chore(internal): revert improve generate OCI package size

brettlangdon:
Revert #10067 which introduces a symlink deduping script to try and cut down on the OCI package size.

There is an edge case where this causes resource consumption to sky rocket for our new k8s injection tests.

It is better to revert this for now to ensure our tests are stable, and then we can revisit the optimizations.

After this revert the image size will still be ok, but should not be forgotten.

## Checklist

## Reviewer Checklist
- Title is accurate


lib-injection/dedupe.py
```diff
@@ -1,57 +0,0 @@
-import collections
```
erikayasuda, line 1: nit: Should we just keep this script here, and re-add the step in the `prepare-oci-package.sh` file when we revisit? Or do you foresee the implementation changing drastically?

erikayasuda (approved):

erikayasuda:
@brettlangdon  Should we add a JIRA task for this in the backlog so we don't forget? 👀

brettlangdon:
> @brettlangdon Should we add a JIRA task for this in the backlog so we don't forget? 👀

there is one already

PR #10413, opened 2024-08-27
WIP: trimming public API to increase safety

rachelyangdog:
As more and more customers as well as internal teams use our APIs the risk exposure grows around APIs that were not thought through thoroughly with regards to reducing application crashing and need to handle errors.

Work done:
- evaluated every method to determine if it needed to be publicly available or not
- trim tracer.py, span.py, pin.py public API (made some methods private)
- increased safety for methods by adding different checks
- increased safety for methods by adding different defaults for return values (ex: not returning `None` for `current_trace_context`)
- created a BaseSpan class that Span and NonRecordingSpan inherits from
- created a [NonRecordingSpan](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-api/src/opentelemetry/trace/span.py#L514) class (OpenTelemetry)

Discussion: 
- trim span.py such that users cannot access any attributes other than `context` OR expand the  [NonRecordingSpan](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-api/src/opentelemetry/trace/span.py#L514) to return empty objects for methods currently implemented in span.py (such as `get_tags`)

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/_trace/pin.py
```diff
@@ -0,0 +1,237 @@
...
+
+        >>> conn = sqlite.connect('/tmp/user.db')
+        >>> # Override a pin for a specific connection
+        >>> pin = Pin.override(conn, service='user-db')
+        >>> conn = sqlite.connect('/tmp/image.db')
+    """
+
+    __slots__ = ["tags", "tracer", "_target", "_config", "_initialized"]
+
+    def __init__(
```
mabdinur, line 38: Can we revisit the design of the Pin class and its usage. Currently the Pin object is used in some integrations but is missing or half supported in others. It would be nice if we were consistent. We shouldn't do this in this PR but it's something to think about.

ddtrace/_trace/pin.py
```diff
@@ -0,0 +1,237 @@
...
+        self,
+        service=None,  # type: Optional[str]
+        tags=None,  # type: Optional[Dict[str, str]]
+        tracer=None,
+        _config=None,  # type: Optional[Dict[str, Any]]
+    ):
+        # type: (...) -> None
+        tracer = tracer or ddtrace.tracer
```
mabdinur, line 46: Can we ensure Pin always uses the global tracer? The ddtrace library does not support initializing multiple tracer objects. We can then deprecate the tracer parameter in Pin.init

```suggestion
        tracer = ddtrace.tracer
```

ddtrace/_trace/pin.py
```diff
@@ -0,0 +1,237 @@
...
+            if self.service:
+                ddtrace.config._add_extra_service(self.service)
+            return setattr(obj, pin_name, self)
+        except AttributeError:
+            log.debug("can't pin onto object. skipping", exc_info=True)
+
+    def onto(self, obj, send=True):
+        deprecate(
+            "The `pin.onto` method is deprecated and will be removed.",
+            message="""The `pin.onto` method is deprecated and will be removed.""",
```
mabdinur, line 174: We should also inform users why this method is being deprecated and if there is an alternative. Users could be getting value from this functionality.

mabdinur (commented):
We're making a few different deprecations in this PR. It would be nice to split them up. Good work so far!

PR #10423, opened 2024-08-28
chore(ci_visibility): don't log exception when codeowners can't be loaded

romainkomorndatadog:
Reverts the `Codeowners.__init__()` behavior introduced in #10094 which changed the exception being caught by the `CIVisibility` service when instantiating `Codeowners()`, resulting in undesirable log messages.

`Codeowners.__init__()` raising a `ValueError` exception, caught by the sole caller (`CIVisibility.__init__()`).

The `of()` behavior is maintained and the caller is updated to check for `[]` as an indication that no codeowner was found.

There is no release note as #10094 was only backported into `2.12`.

As a bonus, type hints are converted from comments to annotations.

## Checklist

## Reviewer Checklist
- Title is accurate

### DIFF
--- a/ddtrace/internal/ci_visibility/recorder.py
+++ b/ddtrace/internal/ci_visibility/recorder.py
@@ -239,8 +239,6 @@ class CIVisibility(Service):
 
         try:
             self._codeowners = Codeowners()
-        except ValueError:
-            log.warning("CODEOWNERS file is not available")
         except Exception:
             log.warning("Failed to load CODEOWNERS", exc_info=True)
 
@@ -683,8 +681,10 @@ class CIVisibility(Service):
             handles = cls._instance._codeowners.of(location)
             if handles:
                 span.set_tag(test.CODEOWNERS, json.dumps(handles))
-        except KeyError:
-            log.debug("no matching codeowners for %s", location)
+            else:
+                log.debug("no matching codeowners for %s", location)
+        except Exception:
+            log.debug("Failed to set codeowners for %s", location, exc_info=True)
 
     @classmethod
     def add_session(cls, session: CIVisibilitySession):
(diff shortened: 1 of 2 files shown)


brettlangdon (approved):

ddtrace/internal/ci_visibility/recorder.py
```diff
@@ -683,8 +683,10 @@ def set_codeowners_of(cls, location, span=None):
             handles = cls._instance._codeowners.of(location)
             if handles:
                 span.set_tag(test.CODEOWNERS, json.dumps(handles))
+            else:
+                log.debug("no matching codeowners for %s", location)
         except KeyError:
```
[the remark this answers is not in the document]
romainkomorndatadog: Actually, it should just be a bare `except`, really. `of()` [no longer raises an exception itself]( https://github.com/DataDog/dd-trace-py/pull/10094/files#diff-68384ccfde4c133ff7fb33b050476b2fa3efeac1ced057790ef2fe03df140777L195) since #10094 .

romainkomorndatadog (commented):
Added a3a6b2f78 to make things a little more resilient.

ddtrace/internal/codeowners.py
```diff
@@ -119,16 +118,20 @@ class Codeowners(object):
...
+
         path = path or self.location(cwd)
-        if path is not None:
-            self.path = path  # type: str
-        self.patterns = []  # type: List[Tuple[re.Pattern, List[str]]]
+
+        if path is None:
+            raise ValueError("CODEOWNERS file not found")
+
+        self.path = path
```
vitor-de-araujo, line 134: Execution only reaches this point if `path` is not `None`, right? So why is `self.path` optional?

vitor-de-araujo (approved):

ddtrace/internal/codeowners.py
romainkomorndatadog: I just wanted it to have a default value and so that `mypy` wouldn't complain when [I check for `None` in parse()](https://github.com/DataDog/dd-trace-py/pull/10423/files#diff-68384ccfde4c133ff7fb33b050476b2fa3efeac1ced057790ef2fe03df140777R148).

It's an abundance of caution for something impossible.

PR #9893, opened 2024-07-22
feat(langchain): tag tool call outputs for langchain chat models

sabrenner:
## What does this PR do?
[MLOB-1165]

**Note**: Most LOC are from requirements changes, adding cassettes for langchain-specific formatted tool calls, and snapshots.

Adds tagging for LCEL [tool/function calling](https://python.langchain.com/v0.1/docs/modules/model_io/chat/function_calling/). **Specifically, for tool call output.** The two common LangChain methods associated with this, `bind_tools` and `with_structured_output`, are responsible for making tool calls to the respective model and including the tool call results in the LangChain chat model response.

Tool calls are a typed property of the [`AiMessage` LangChain message type](https://github.com/langchain-ai/langchain/blob/1f01c0fd98d1cb768d950b8e46701bf949865af3/libs/core/langchain_core/messages/ai.py#L72). While tool calls from different providers/models are structured differently (ie Anthropic content blocks vs OpenAI tool calls, [ref](https://python.langchain.com/v0.2/docs/how_to/function_calling/#passing-tool-outputs-to-model) under the intro to this documentation), LangChain coalesces these differences under a common [ToolCall](https://github.com/langchain-ai/langchain/blob/1f01c0fd98d1cb768d950b8e46701bf949865af3/libs/core/langchain_core/messages/tool.py#L127-L154) struct. This is what we tag in this.

## Follow-up
1. **Tagging tool calls for LLM Observability will be a separate PR right after this one** (tests will re-use cassettes but that PR will validate we set the tool calls appropriately).
2. This PR covers the scope for `bind_tools`. However, `with_structured_output` actually appends an `OutputParser` to the chat model invocation to transform those tool calls into structured output for the chain. **Tracing and tagging `OutputParsers` will be a separate PR.**

### OpenAI
Before
Screenshot 2024-07-22 at 3 13 52 PM

After
Screenshot 2024-07-22 at 3 14 17 PM

### Anthropic
Before
Screenshot 2024-07-22 at 3 14 41 PM

After
Screenshot 2024-07-22 at 3 14 58 PM

## Checklist
- [x] PR author has checked that all the criteria below are met
- [x] The PR description includes an overview of the change
- [x] The PR description articulates the motivation for the change
- [x] The change includes tests OR the PR description describes a testing strategy
- [x] The PR description notes risks associated with the change, if any
- [x] Newly-added code is easy to change
- [x] The change follows the [library release note guidelines](https://ddtrace.readthedocs.io/en/stable/releasenotes.html)
[...]

PR #10435, opened 2024-08-28
feat(llmobs): capture tool call outputs from langchain chat models

sabrenner:
## What does this PR do
Follow-up to #9893

Captures tool calls returned as part of the chat completion from a supported chat model within LangChain. As noted on the referenced PR, LangChain standardizes tool calls from the model providers it has that support tool calling onto a `tool_calls` struct on the output message, with `name`, `id`, and `args` fields ([ref](https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/messages/tool.py#L131)), which attach to the LLMObs span event here as `name`, `tool_id`, and `arguments` respectively so it shows up properly in our UI.

## Follow up work
This is part of a series of changes to better capture use cases such as `with_structured_output`.  As a follow-up, LangChain `OutputParser`s will be traced to show the full "chain" from chat model &rarr; response with tool call &rarr; output parser &rarr; final output of the chain (as `with_structured_output` effectively turns a chat model into a chain with an output parser).

## UX
When LangChain and OpenAI are used (the LangChain chat model span is a `workflow`):
Screenshot 2024-08-28 at 2 39 19 PM

When just LangChain is used (the LangChain chat model span is an `llm`):
Screenshot 2024-08-28 at 2 40 00 PM

## Checklist

## Reviewer Checklist
- Title is accurate


releasenotes/notes/llmobs-langchain-chat-model-tool-calls-fbc5ced7dfb2ea89.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    LLM Observability: Added support for tagging tool calls returned as part of chat completions from langchain chat models.
```
Yun-Kim, line 4:
```suggestion
    LLM Observability: Added support for capturing tool calls returned from LangChain chat completions.
```

Yun-Kim (approved):
Small release note suggestion but LGTM!

PR #10466, opened 2024-08-30
ci: move two test suites to gitlab from circleci

emmettbutler:
This change moves the `ddtracerun` and `integration-snapshot` test suites from circleci to gitlab for billing purposes. It was necessary to adjust one of the ddtracerun tests to look for an optional environment variable specifying the location of the redis server, which had previously been `localhost` in both local and CI runs.

## Checklist

## Reviewer Checklist
- Title is accurate


.gitlab/tests.yml
```diff
@@ -37,6 +37,8 @@ variables:
         DD_POOL_TRACE_CHECK_FAILURES: true
         DD_DISABLE_ERROR_RESPONSES: true
         ENABLED_CHECKS: trace_content_length,trace_stall,meta_tracer_version_header,trace_count_header,trace_peer_service,trace_dd_service
+    - name: registry.ddbuild.io/redis:7.0.7
```
brettlangdon, line 40: we probably don't want this here. this will mean every job that uses riot will start a redis server.


instead we probably want to do something like:

```yaml
ddtracerun:
  extends: .test_base_riot
  services:
    - !reference [.test_base_riot, services]
    - name: registry.ddbuild.io/redis:7.0.7
      alias: redis
  variables:
    SUITE_NAME: "ddtracerun"
    TEST_REDIS_HOST: "redis"
```

wdyt?

brettlangdon: I think agent and test agent are _probably_ ok to keep as default services for all? maybe a tad wasteful, but probably generally true?

emmettbutler: makes sense to me

brettlangdon (approved):

PR #10476, opened 2024-09-02
feat(langchain): trace tool invocations

yahya-mouman:
This PR adds instrumentation to submit langchain tool.invoke/ainvoke tool spans to APM.

The tool spans sent to APM will contain the following I/O data and tags :
langchain.request.tool.name
langchain.request.tool.description
langchain.request.tool.metadata
langchain.request.tool.tags
langchain.request.input
langchain.request.config
langchain.response.output

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -974,6 +975,94 @@ def traced_similarity_search(langchain, pin, func, instance, args, kwargs):
...
 
+@with_traced_module
+def traced_base_tool_invoke(langchain, pin, func, instance, args, kwargs):
+    integration = langchain._datadog_integration
+    tool_input = get_argument_value(args, kwargs, 0, "input")
+    config = get_argument_value(args, kwargs, 1, "config", optional=True)
+
+    span = integration.trace(
+        pin,
+        "%s.%s.%s.%s" % (func.__module__, func.__class__.__name__, func.__name__, func.__self__.name),
```
Yun-Kim, line 986: I think this might be overkill on the span resource naming. Ideally we only need to know the name of the tool being invoked.
IMO it's unnecessary to add `method.invoke` as part of the span name, the tool name itself should make it fairly obvious that this is a tool invocation. Same with the underlying module, not sure if this gives users much helpful information that they don't already have.
Can we get just the tool's name as the resource? i.e. `Circumference calculator`

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -974,6 +975,94 @@ def traced_similarity_search(langchain, pin, func, instance, args, kwargs):
...
+        interface_type="tool",
+    )
+
+    tool_output = None
+    try:
+        tool_attributes = ["name", "description", "metadata", "tags"]
+
+        for attribute in tool_attributes:
```
Yun-Kim, line 994:
```suggestion
        for attribute in ("name", "description", "metadata", "tags"):
```

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -974,6 +975,94 @@ def traced_similarity_search(langchain, pin, func, instance, args, kwargs):
...
+                for idx, tag in enumerate(value):
+                    span.set_tag_str("langchain.request.tool.%s.%d" % (attribute, idx), str(value))
+            elif value is not None:
+                span.set_tag_str("langchain.request.tool.%s" % attribute, str(value))
+
+        if integration.is_pc_sampled_span(span):
+            if tool_input:
+                span.set_tag_str("langchain.request.input", integration.trunc(str(tool_input)))
+            if config:
+                span.set_tag_str("langchain.request.config", json.dumps(config))
```
Yun-Kim, line 1009: This can always be tagged since it is not a direct I/O, doesn't need to be in the `is_pc_sampled_span()` block.

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -974,6 +975,94 @@ def traced_similarity_search(langchain, pin, func, instance, args, kwargs):
...
+        tool_output = func(*args, **kwargs)
+        if tool_output is not None:
+            if integration.is_pc_sampled_span(span):
```
Yun-Kim, line 1012: Can we combine these two conditionals for simplicity?

releasenotes/notes/feat-submit-langchain-tool-spans-e7f76291dfa1efe5.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    Langchain : Adds support for tracing tool invocations.
```
Yun-Kim, line 4:
```suggestion
    langchain : Adds support for tracing tool invocations.
```
For integrations, we don't capitalize the library name in release notes. (not sure why, just standard practice)

tests/contrib/langchain/test_langchain_community.py
```diff
@@ -1274,3 +1274,65 @@ def test_faiss_vectorstore_retrieval(langchain_community, langchain_openai, requ
             retriever = faiss.as_retriever()
         with request_vcr.use_cassette("openai_retrieval_embedding.yaml"):
             retriever.invoke("What was the message of the last test query?")
+
+
+@pytest.mark.snapshot(
+    ignores=["meta.langchain.request.tool.description"],
```
Yun-Kim, line 1280: Why do we need to ignore the tool description tag in the snapshot test?

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -1101,6 +1190,8 @@ def patch():
         )
         wrap("langchain_core", "runnables.base.RunnableSequence.batch", traced_lcel_runnable_sequence(langchain))
         wrap("langchain_core", "runnables.base.RunnableSequence.abatch", traced_lcel_runnable_sequence_async(langchain))
+        wrap("langchain_core", "tools.BaseTool.invoke", traced_base_tool_invoke(langchain))
+        wrap("langchain_core", "tools.BaseTool.ainvoke", traced_base_tool_ainvoke(langchain))
```
Yun-Kim, line 1194: Looks like older versions of langchain-core (`v0.1.x`) are failing this as the `tools` module is not accessible directly through `langchain_core`. I suggest adding this line at line 1171:
`from langchain_core.tools import BaseTool  # noqa:F401`
This lets us import the necessary module/class before wrapping it.

ddtrace/contrib/internal/langchain/patch.py
yahya-mouman: Isn't a free field like this prone to grow large as well similar to the input ?

tests/contrib/langchain/test_langchain_community.py
yahya-mouman: Due to some subtle differences in langchain core versions the description might be different.
```
"Circumference calculator(radius: float) -> float - Use this tool when you need to calculate a circumference using the radius of a circle"
```
In the other versions it's just 
```
"Use this tool when you need to calculate a circumference using the radius of a circle"
```

So instead of using two cassettes for each case i just ignored the field

ddtrace/contrib/internal/langchain/patch.py
yahya-mouman: I'm guessing it's for memory optimization ?

ddtrace/contrib/internal/langchain/patch.py
yahya-mouman: My bad I forgot i had the invoke and ainvoke and only updated one of them 🦆

tests/contrib/langchain/test_langchain_community.py
Yun-Kim: Ahh I see. Thanks for clarifying, let's add a small comment next to the ignore line mentioning this!

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -974,6 +975,88 @@ def traced_similarity_search(langchain, pin, func, instance, args, kwargs):
...
+        for attribute in ("name", "description", "metadata", "tags"):
+            value = getattr(instance, attribute, None)
+            if isinstance(value, dict):
+                for key, meta_value in value.items():
+                    span.set_tag_str("langchain.request.tool.%s.%s" % (attribute, key), str(value))
+            elif isinstance(value, list):
+                for idx, tag in enumerate(value):
+                    span.set_tag_str("langchain.request.tool.%s.%d" % (attribute, idx), str(value))
+            elif value is not None:
+                span.set_tag_str("langchain.request.tool.%s" % attribute, str(value))
```
Yun-Kim, line 1001: I think this might be too much nesting. I'm find with handling `name` and `description` in the same for loop since they are simple strings, but can we handle `metadata` (dict) and `tags` (list of strings) separately?

Yun-Kim (commented):
Last comment, otherwise LGTM

ddtrace/contrib/internal/langchain/patch.py
Yun-Kim: Technically it's free but since it's not direct I/O I think we can be fine with that.

ddtrace/contrib/internal/langchain/patch.py
yahya-mouman: Yes i can do that !
for future references why is nesting capped ? Isn't complexity exactly the same since we iterate on a fixed length list ?

Yun-Kim: Yeah complexity/performance is not affected, it's just Python styling best practices to avoid excessive nesting for readability

tests/contrib/langchain/test_langchain_community.py
```diff
@@ -1274,3 +1274,67 @@ def test_faiss_vectorstore_retrieval(langchain_community, langchain_openai, requ
...
+    )
+
+    calculator.invoke("2")
+
+
+@pytest.mark.asyncio
+@pytest.mark.snapshot(
+    # tool description is generated differently is some langchain_core versions
+    ignores=["meta.langchain.request.tool.description"],
+    token="tests.contrib.langchain.test_langchain_community.test_base_tool_ainvoke",
```
Yun-Kim, line 1314: Just noticed that the snapshot test files are essentially identical. Can we delete the async snapshot test file and reuse the sync snapshot file for both ainvoke/invoke tests?
```suggestion
    token="tests.contrib.langchain.test_langchain_community.test_base_tool_invoke",
```

Yun-Kim (approved):
Small nit but otherwise LGTM!

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -974,6 +975,88 @@ def traced_similarity_search(langchain, pin, func, instance, args, kwargs):
...
+    try:
+        for attribute in ("name", "description"):
+            value = getattr(instance, attribute, None)
+            if value is not None:
+                span.set_tag_str("langchain.request.tool.%s" % attribute, str(value))
+
+        for key, meta_value in getattr(instance, "metadata", {}).items():
+            span.set_tag_str("langchain.request.tool.metadata.%s" % key, str(meta_value))
+        for idx, tag in getattr(instance, "tags", []):
+            span.set_tag_str("langchain.request.tool.tags.%d" % idx, str(value))
```
Yun-Kim, line 1042: Seeing some discrepancies between the sync and async functions (they should be identical minus async syntax)

yahya-mouman: Ah sorry about that forgot to update ainvoke as CI was failing for invoke due to values being None

PR #10503, opened 2024-09-03
feat(gemini): trace google gemini Integration

Yun-Kim:
Adds support for tracing Google's Gemini Python SDK `generate_content` and `generate_content_async` methods (also indirectly traces `send_message()`).

This PR also moves `_get_attr()` convenience helper from the anthropic integration to the shared `ddtrace.llmobs._utils.py` module.

### Features:
- Traces `generate_content/generate_content_async()` and indirectly `send_message/send_message_async()` which uses `generate_content` under the hood
- Handles streamed responses
- Captures input/outputs (truncation included, function call/response messages included), token usage, input configuration options, and model name.

### Testing:
The Gemini Python SDK does not use requests/httpx to submit requests but instead via GRPC. The vcrpy testing framework does not capture this so I wrote a mock client to store/return mock responses instead of sending actual requests to Google. This testing framework was inspired by the [Gemini Python SDK's testing strategy](https://github.com/google-gemini/generative-ai-python/blob/9407dcde5666ba58831227d3acf8bd0e3f3b4f81/tests/test_generative_models.py#L41).

### Next steps (future PRs)
- Wrap `send_message()` not to trace, but to attach a hash/session_id for all subsequent send_message calls
- Integrate with LLMObs

## Checklist

## Reviewer Checklist
- Title is accurate


releasenotes/notes/feat-google-gemini-d5ee30b1d711bc08.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    google_generativeai: This introduces tracing support for Google Gemini API ``generate_content`` calls.
```
erikayasuda, line 4: nit:
```suggestion
    google_generativeai: Introduces tracing support for Google Gemini API ``generate_content`` calls.
```

riotfile.py
```diff
@@ -2685,6 +2685,16 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
                 "anthropic": [latest, "~=0.28"],
             },
         ),
+        Venv(
+            name="google_generativeai",
+            command="pytest {cmdargs} tests/contrib/google_generativeai",
+            pys=select_pys(min_version="3.9"),
+            pkgs={
+                "pytest-asyncio": latest,
+                "google-generativeai": ["~=0.7.2"],
```
erikayasuda, line 2694: nit: Do we not need `latest` here?

docs/spelling_wordlist.txt
```diff
@@ -96,6 +96,7 @@ exec
 fastapi
 formatter
 gRPC
+generativeai
```
erikayasuda, line 99: What is this file for?

ddtrace/contrib/google_generativeai/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+~~~~~~~~~~~~~~~~~~~~~~
+
+To configure the Gemini integration on a per-instance basis use the
+``Pin`` API::
+
+    import google.generativeai as genai
+    from ddtrace import Pin, config
+
+    Pin.override(genai, service="my-gemini-service")
+"""  # noqa: E501
```
erikayasuda, line 79: nit:
```suggestion
"""
# noqa: E501
```

ddtrace/contrib/internal/google_generativeai/patch.py
```diff
@@ -0,0 +1,113 @@
...
+    return generations
+
+
+def patch():
+    if getattr(genai, "_datadog_patch", False):
+        return
+
+    genai._datadog_patch = True
+
+    Pin().onto(genai)
```
erikayasuda, line 96: Just a sanity check on my end: Are we using `Pin` here because it's common to have multiple processes/contexts where an instance of `genai` gets accessed?

erikayasuda (approved):
Just some nits and clarifying questions

docs/spelling_wordlist.txt
Yun-Kim: Any words in release notes / docs that are not in the traditional English dictionary should be added here, otherwise sphinx will flag/fail this in the build_docs job: https://sphinxcontrib-spelling.readthedocs.io/en/latest/customize.html#managing-lists-of-correctly-spelled-words-and-ignoring-words

riotfile.py
Yun-Kim: 0.7.2 is the latest version, wanted to avoid unnecessary double testing. I will instead replace this with latest

ddtrace/contrib/internal/google_generativeai/patch.py
Yun-Kim: Kind of, this way we can use the same tracer instance - this is just standard practice following the other integrations. Do you think we should avoid using Pin here?

ddtrace/contrib/google_generativeai/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+Global Configuration
+~~~~~~~~~~~~~~~~~~~~
+
+.. py:data:: ddtrace.config.google_generativeai["service"]
+
+   The service name reported by default for Gemini requests.
+
+   Alternatively, you can set this option with the ``DD_SERVICE`` or ``DD_ANTHROPIC_SERVICE`` environment
```
lievan, line 38:
```suggestion
   Alternatively, you can set this option with the ``DD_SERVICE`` or ``DD_GOOGLE_GENERATIVEAI_SERVICE`` environment
```

ddtrace/contrib/google_generativeai/__init__.py
```diff
@@ -0,0 +1,92 @@
+"""
+The Gemini integration instruments the Google Gemini Python API to traces for requests made to Google models.
+
+All traces submitted from the Gemini integration are tagged by:
+
+- ``service``, ``env``, ``version``: see the `Unified Service Tagging docs <https://docs.datadoghq.com/getting_started/tagging/unified_service_tagging>`_.
+- ``genai.request.model``: Google model used in the request.
```
lievan, line 7: The `genai` prefix is supposed to be specific to this integration right? Should we change it to `google_genai.request.model` to clarify? It's verbose, but imo clarity > verbosity

ddtrace/contrib/internal/google_generativeai/_utils.py
```diff
@@ -0,0 +1,219 @@
...
+        finally:
+            self._dd_span.finish()
+
+
+def _extract_model_name(instance):
+    """Extract the model name from the instance."""
+    model_name = getattr(instance, "model_name", "")
+    if not model_name or not isinstance(model_name, str):
+        return ""
+    if "/" in model_name:
```
lievan, line 53: Is this processing for the model name necessary? I assume users would just want the full model name

ddtrace/contrib/internal/google_generativeai/_utils.py
```diff
@@ -0,0 +1,219 @@
...
+    """
+    contents = get_argument_value(args, kwargs, 0, "contents")
+    generation_config = kwargs.get("generation_config", {})
+    system_instruction = getattr(instance, "_system_instruction", None)
+    stream = kwargs.get("stream", None)
+
+    generation_config_dict = None
+    if isinstance(generation_config, dict):
+        generation_config_dict = generation_config
+    elif generation_config is not None:
```
lievan, line 162: in the cases is `generation_config` not a dictionary? I see some examples have generation_config of type `genai.GenerationConfig`. Should we explicitly check for that instead?

This way if someone passes in an invalid type for `generation_config` we won't just be dumping the `.__dict__` attributes onto the span

ddtrace/contrib/google_generativeai/__init__.py
yahya-mouman: I agree with this one. Feels like genai would easily be mistaken for the general term rather than the gemini case.

ddtrace/contrib/internal/google_generativeai/_utils.py
yahya-mouman: I do agree that processing is good to be able to target cases based on normalized model names. However we lose information about if it's a proprietary or what provider it has (ex: tfhub/EfficientNetV2 vs Proprietary/EfficientNetV2). So I'm suggesting if we can also extract the model provider along with the name ?

ddtrace/contrib/internal/google_generativeai/_utils.py
```diff
@@ -0,0 +1,217 @@
...
+
+def _tag_request_content_part(span, integration, part, part_idx, content_idx):
+    """Tag the generation span with request content parts."""
+    text = getattr(part, "text", "")
+    function_call = getattr(part, "function_call", None)
+    function_response = getattr(part, "function_response", None)
+    if isinstance(part, dict):
+        text = part.get("text", "")
+        function_call = part.get("function_call", None)
+        function_response = part.get("function_response", None)
```
yahya-mouman, line 79:
```suggestion
    if isinstance(part, dict):
        text = part.get("text", "")
        function_call = part.get("function_call", None)
        function_response = part.get("function_response", None)
    else:
        text = getattr(part, "text", "")
        function_call = getattr(part, "function_call", None)
        function_response = getattr(part, "function_response", None)
```

Not 100% sure but isn't it better this way ? we would only execute the getattr if the part is not a dict.
Also wouldn't getattr on a dict return an attribute not found exception that we aren't catching ?

yahya-mouman: Nvm i just noticed we have a default None or empty string for the case where it's not found.

ddtrace/contrib/internal/google_generativeai/patch.py
erikayasuda: No not necessarily. I just had a conversation with someone a while ago that Pin is not always necessary, so I always tend to dig into how Pin gets used in new integrations. Either way, now that we abstracted implementation details in `internal/` it's not a blocker, just my own curiosity 🙂

ddtrace/contrib/internal/google_generativeai/_utils.py
```diff
@@ -0,0 +1,217 @@
...
+class TracedGenerateContentResponse(BaseTracedGenerateContentResponse):
+    def __iter__(self):
+        try:
+            for chunk in self.__wrapped__.__iter__():
+                yield chunk
+        except Exception:
+            self._dd_span.set_exc_info(*sys.exc_info())
+            raise
+        else:
+            tag_response(self._dd_span, self.__wrapped__, self._dd_integration, self._model_instance)
```
sabrenner, line 29: just for my own edification - i'm assuming `self.__wrapped__` is getting the accumulated chunks from the stream? is that all handled under the hood with `wrapt`? it just seems a bit different as to how we handle it in the `openai` integration.

ddtrace/contrib/internal/google_generativeai/_utils.py
Yun-Kim: Will add a comment in the code to clarify why we're doing this processing!

ddtrace/contrib/internal/google_generativeai/_utils.py
Yun-Kim: Will be adding a convenience method to extract this instead!

ddtrace/contrib/internal/google_generativeai/patch.py
```diff
@@ -0,0 +1,116 @@
...
+
+def get_version():
+    # type: () -> str
+    return getattr(genai, "__version__", "")
+
+
+@with_traced_module
+def traced_generate(genai, pin, func, instance, args, kwargs):
+    integration = genai._datadog_integration
+    stream = kwargs.get("stream", False)
```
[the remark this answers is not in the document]
Yun-Kim: Yes, we currently dont capture the tools/tool_configs provided in the request for any of our LLMObs integrations. We can consider that as a future step.

ddtrace/contrib/internal/google_generativeai/_utils.py
Yun-Kim: Yup, the google gemini SDK actually collects the chunks and joins them together into a regular response once the stream is finished, so we don't need to do any of the joining fortunately 😄

lievan (approved):

Issue #10344, opened 2024-08-22
dd-trace-py is only reporting get-socket in APM spans


guyzmo:
### Summary of problem

I set up dd-trace-py following the documentation, I did `patch(pymongo=True)`.
In the APM dashboard, I have traces for the db with `pymongo.get_socket` but I have no other information.

What I'm expecting is to have the type of command/operator that's being used, and the full query text.

### Which version of dd-trace-py are you using?

Currently I'm running 2.9.2 because of the latency bug, but I'm about to upgrade to 2.10.6.

### Which version of pip are you using?

I set up my project using rye, but I deploy using pip for python 3.10.2.

### Which libraries and their versions are you using?

<details>
  <summary>`pip freeze`</summary>

```
altair==5.3.0
anyio==4.4.0
attrs==24.2.0
beanie==1.11.6
blinker==1.8.2
bytecode==0.15.1
cachecontrol==0.14.0
cachetools==5.4.0
cattrs==23.2.3
certifi==2024.7.4
cffi==1.17.0
chardet==5.2.0
charset-normalizer==3.3.2
click==8.1.7
cloudevents==1.11.0
contourpy==1.2.1
cryptography==43.0.0
cycler==0.12.1
ddsketch==3.0.1
ddtrace==2.10.6
deprecated==1.2.14
deprecation==2.1.0
dnspython==2.6.1
email-validator==2.2.0
envier==0.5.2
et-xmlfile==1.1.0
exceptiongroup==1.2.2
fastapi==0.111.1
fastapi-cli==0.0.5
firebase-admin==6.5.0
flask==3.0.3
flexpolyline==0.1.0
fonttools==4.53.1
functions-framework==3.5.0
geographiclib==2.0
geopandas==1.0.1
geopy==2.3.0
gitdb==4.0.11
gitpython==3.1.43
google-api-core==2.19.1
google-api-python-client==2.140.0
google-auth==2.33.0
google-auth-httplib2==0.2.0
google-auth-oauthlib==1.2.1
google-cloud-appengine-logging==1.4.5
google-cloud-audit-log==0.2.5
google-cloud-bigquery==3.25.0
google-cloud-core==2.4.1
google-cloud-firestore==2.17.0
google-cloud-logging==3.11.1
google-cloud-pubsub==2.19.0
google-cloud-storage==2.18.1
google-cloud-tasks==2.16.5
google-crc32c==1.5.0
google-events==0.12.0
google-resumable-media==2.7.1
[...]

PR #10516, opened 2024-09-05
fix(pymongo): use bytecode wrapping to trace pymongo clients

mabdinur:
Resolves: https://github.com/DataDog/dd-trace-py/issues/10344

- Uses bytecode wrapping instead of wrapt's objectproxys to instrument mongo clients and socket connections
- Previously the pin object used in the MongoClient was different from the ones used to trace socket connections.  Updates usages of the Pin object to ensure sever, socket, and client objects use the same Pin. This sublte fix required updating the span counts in tests (now the DummyTracer used in tests is expected to capture one more span per trace).
- Updates the mongoengine integration to simply use the pymongo integration. This integration no longer does anything special and can be removed.
- Improves the test coverage of the pymongo integration. Ensures every pymongo version range with custom patching has test coverage.

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/contrib/internal/pymongo/client.py
```diff
@@ -88,6 +114,9 @@ def __getddpin__(self):
 
 @contextlib.contextmanager
 def wrapped_validate_session(wrapped, instance, args, kwargs):
+    # The function is exposed in the public API, but it is not used in the codebase.
+    # TODO: Remove this function when ddtrace.contrib.pymongo.client is removed.
```
mabdinur, line 118:
```suggestion
    # TODO(mabdinur): Remove this function when ddtrace.contrib.pymongo.client is removed.
```

mabdinur, line 118:
```suggestion
    # TODO(mabdinur): Remove this function when ddtrace.contrib.pymongo.client is removed.
```

releasenotes/notes/munir-pymongo-fix-patching-545f3947644c91cd.yaml
```diff
@@ -0,0 +1,4 @@
+---
+fixes:
+  - |
+    pymongo: Ensures instances of the ``pymongo.MongoClient`` can be patch after pymongo is imported.
```
emmettbutler, line 4:
```suggestion
    pymongo: Ensures instances of the ``pymongo.MongoClient`` can be patched after pymongo is imported.
```

ddtrace/contrib/internal/mongoengine/patch.py
```diff
@@ -1,5 +1,9 @@
+# TODO(mabdinue): Remove the pymongoengine integration, this integration does nothing special
```
emmettbutler, line 1:
```suggestion
# TODO(mabdinur): Remove the mongoengine integration, this integration does nothing special
```

emmettbutler (approved):

ddtrace/contrib/internal/pymongo/client.py
```diff
@@ -88,6 +114,9 @@ def __getddpin__(self):
 
 @contextlib.contextmanager
 def wrapped_validate_session(wrapped, instance, args, kwargs):
+    # The function is exposed in the public API, but it is not used in the codebase.
+    # TODO(<owner>): Remove this function when ddtrace.contrib.pymongo.client is removed.
```
mabdinur, line 118:
```suggestion
    # TODO(mabdinur): Remove this function when ddtrace.contrib.pymongo.client is removed.
```

ddtrace/contrib/internal/pymongo/client.py
```diff
@@ -45,6 +46,31 @@
...
+        return ddtrace.Pin.get_from(client._topology)
+
+    client.__setddpin__ = functools.partial(__setddpin__, client)
+    client.__getddpin__ = functools.partial(__getddpin__, client)
+
+    # Default Pin
+    ddtrace.Pin(service=_DEFAULT_SERVICE).onto(client)
+
+
+# TODO(<owner>): Remove TracedMongoClient when ddtrace.contrib.pymongo.client is removed from the public API.
```
mabdinur, line 73:
```suggestion
# TODO(mabdinur): Remove TracedMongoClient when ddtrace.contrib.pymongo.client is removed from the public API.
```

ddtrace/contrib/internal/pymongo/client.py
```diff
@@ -45,6 +46,31 @@
 _DEFAULT_SERVICE = schematize_service_name("pymongo")
 
 
+def trace_mongo_client_init(func, args, kwargs):
+    # Call MongoClient.__init__
```
mabdinur, line 50:
```suggestion
```

tests/contrib/pymongo/test.py
```diff
@@ -842,4 +864,4 @@ def test_server_info(self):
 
         assert len(spans) == 2
         self.check_socket_metadata(spans[0])
-        assert spans[1].name == "pymongo.cmd"
+        assert spans[1].name == "pymongo.cmd"
```
mabdinur, line 867:
```suggestion
        assert spans[1].name == "pymongo.cmd"

```

quinna-h (approved):

ddtrace/contrib/internal/pymongo/client.py
```diff
@@ -105,213 +91,192 @@
...
+def _trace_server_send_message_with_response(func, args, kwargs):
+    server_instance = get_argument_value(args, kwargs, 0, "self")
+    operation = get_argument_value(args, kwargs, 1, "operation")
+
+    span = _datadog_trace_operation(operation, server_instance)
+    if span is None:
+        return func(*args, **kwargs)
+    with span:
+        result = func(*args, **kwargs)
+        if result:
```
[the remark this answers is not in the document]
mabdinur: I will not refactor this code in this PR. I will keep this logic as is

ddtrace/contrib/internal/mongoengine/patch.py
```diff
@@ -1,5 +1,9 @@
+# TODO(mabdinur): Remove the pymongoengine integration, this integration does nothing special
+# it just uses the pymongo integration and creates unnecessary pin objects
```
erikayasuda, line 2: nit: Do we want to mark it as deprecated then? And fully remove in the next major version EOY?

ddtrace/contrib/internal/pymongo/client.py
```diff
@@ -45,47 +46,32 @@
 _DEFAULT_SERVICE = schematize_service_name("pymongo")
 
 
+# TODO(mabdinur): Remove TracedMongoClient when ddtrace.contrib.pymongo.client is removed from the public API.
```
erikayasuda, line 49: nit: Same question here about marking as deprecated

tests/contrib/mongoengine/test.py
```diff
@@ -42,8 +40,9 @@ def test_insert_update_delete_query(self):
 
         # ensure we get a drop collection span
         spans = tracer.pop()
-        assert len(spans) == 1
-        span = spans[0]
+        assert len(spans) == 2
```
erikayasuda, line 43: Sorry dumb question, why did this change to 2? 😅

tests/contrib/pymongo/test.py
```diff
@@ -1,7 +1,7 @@
 # stdlib
+# 3p
```
erikayasuda, line 2: nit: So this only applies for `time`? Since we're using our local wrapped `pymongo`?

erikayasuda: Also didn't know we did this 😅

erikayasuda (approved):
Just a few questions for my own clarification. LGTM!

ddtrace/contrib/internal/mongoengine/patch.py
mabdinur: This is a weird one. We want to deprecate this integrations public api (patch, unpatch, get_versions). 

We will still provide support for this integration we just don't need the code in ddtrace.contrib.internal.mongoengine to do it. I am not sure how to communicate this. Customer's shouldn't be calling patch, unpatch, or get_versions anyway. 

Maybe log a deprecation warning in `ddtrace.patch_all()` when mongoengine=True.

ddtrace/contrib/internal/pymongo/client.py
mabdinur: We already have deprecation warnings for this. With @rachelyangdog's work this class will be removed from the public api in v3.0.0.

erikayasuda: Oh oops, you're right this is the internal patch.py

tests/contrib/mongoengine/test.py
mabdinur: I made a subtle change to the pin object. I made sure the SAME pin object is applied to client, server, and socket connections. Previously the pin object used in this test had two different tracers for client and server connections. So this test was only configured to capture and assert the number of client spans. This looked wrong to me.

Now this test captures all the spans generated by the pymongo client and server objects in one tracer.

tests/contrib/pymongo/test.py
mabdinur: This is from linting this really old file. I didn't add this comment. We can remove it.

mabdinur, line 2:
```suggestion
```

ddtrace/contrib/internal/mongoengine/patch.py
erikayasuda: Yeah that makes sense. If we've already marked the integration as deprecated on the public side, then we can probably just deprecate the internal code along with it

PR #10525, opened 2024-09-05
fix(core): add level of indirection for provider.py contextvars

sanchda:
Whenever a contextvar is reassociated, it causes the underlying HAMT data structure to clone a node.  This clone operation requires de-referencing stored Python objects, which can cause segmentation faults if other libraries mis-manage the reference counts for their objects, causing them to be GC'd.

This patch stores a single wrapper object into the contextvar, then manipulates a reference within that wrapper in order to propagate our desired information.  In our normal testing fixture, we cause as many as 69 realloc (and clone) events in a single process (I deduce this by patching cpython itself to produce a log).  With this patch, that number is down to 1, and it doesn't originate from this provider.py

I have a standalone reproduction for the noted behavior [here](https://github.com/sanchda/systems_experiments/tree/main/contextvar_segfault).  The repro isn't very clever about how it manages the lifetimes of GC'd objects--the issues we see in the wild are a little bit more subtle, since they don't segfault during normal scope cleanup (unlike mine).

## Checklist

## Reviewer Checklist
- Title is accurate


emmettbutler:
interesting and promising

taegyunkim:
Could you run `hatch run lint:fmt` to format and trigger the rest of circle ci?

ddtrace/_trace/provider.py
```diff
@@ -14,9 +14,31 @@
...
+    """
+
+    def __init__(self, name: str):
+        self._context_var = contextvars.ContextVar[Optional[Union[Context, Span]]](name, default=None)
+
+    def get(self) -> Optional[Union[Context, Span]]:
+        return self._context_var.get()
+
+    def set(self, value: Optional[Union[Context, Span]]) -> None:
+        self._context_var.set(value)
```
r1viollet, line 37: It seems like we removed the wrapper along the way ?
The aim was to have `self._context_var.get().wrapper_set(value)`
I am not sure why we got rid of it ?

Yun-Kim (commented):
Just to help my understanding, would you be able to update the PR description with a small snippet example of a reassoc event that would trigger a segfault/corruption? Thanks!

sanchda:
> Just to help my understanding, would you be able to update the PR description with a small snippet example of a reassoc event that would trigger a segfault/corruption? Thanks!

Unfortunately, _any_ `ContextVar.set()` operation will trigger a re-association.  The sad truth of the matter is that CPython's implementation of ContextVars bleeds a lot of common state together.  I have a standalone reproduction [here](https://github.com/sanchda/systems_experiments/tree/main/contextvar_segfault).  I'll link it in the PR description.

r1viollet:
I agree with the code, but I still can not understand why I get a different type of context within greenlets.
I now get an actual context (and not a span). I am not sure what is the expected behaviour.

r1viollet (approved):
LGTM

Yun-Kim (commented):
This PR makes sense to me but I must admit I'm not nuanced in contextvars internals enough to give a more thoughtful review here. Is this not an issue with CPython's implementation of ContextVars, and if so, should it be fixed upstream?

r1viollet:
> This PR makes sense to me but I must admit I'm not nuanced in contextvars internals enough to give a more thoughtful review here. Is this not an issue with CPython's implementation of ContextVars, and if so, should it be fixed upstream?

I agree there could be an effort to understand what objects have lifetime issues. Though we still have to mitigate in the short term.

sanchda:
I dropped this PR for a while because we shipped a patch to some affected customers and it didn't directly fix the issue we saw.  However, it did reveal more information about where bad objects were being consumed.  I count that as a net win.

So, I'd like to summarize the benefits of this PR as the following
* Fewer clone operations means better performance
* Protects us from cpython global state
* ... and in so doing, makes crashes more actionable


@Yun-Kim:
> This PR makes sense to me but I must admit I'm not nuanced in contextvars internals enough to give a more thoughtful review here. Is this not an issue with CPython's implementation of ContextVars, and if so, should it be fixed upstream?

I don't know whether I'd call it an issue in the implementation or the design.  There are a few problems here.
* CPython directly uses the system allocator--it doesn't manage objects through the kind of indirection that the JVM does, so it's entirely possible for misbehaving code to convert a strong ref (e.g., a contextvar) into a weak ref, causing issues to propagate.  Solving this problem in a durable way would be extremely difficult.
* Even if the CPython maintainers did treat this as an implementation issue, they will not backport such a fix to versions which we currently support (e.g., 3.9).  That means we're on the hook for dealing with those versions safely.  I think this PR does that.

r1viollet:
Should we merge this to progress and get better stacks on associated crashes ?

wconti27 (approved):

wconti27 (changes requested):
Going to let @emmettbutler handle this review

releasenotes/notes/fix-contextvar-cloning-49adaf7fdf36e8fb.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    Fixes an issue where normal tracer operation could cause items stored by other libraries to be dereferenced by the
```
emmettbutler, line 4:
```suggestion
    Fixes an issue where normal library use could cause items stored by other libraries to be dereferenced by the
```

ddtrace/_trace/provider.py
```diff
@@ -14,9 +14,46 @@
...
-_DD_CONTEXTVAR: contextvars.ContextVar[Optional[Union[Context, Span]]] = contextvars.ContextVar(
-    "datadog_contextvar", default=None
-)
+class ContextVarManager:
+    """
+    In the implementation of ContextVar, when a key is re-associated with a new value the underlying HAMT must clone
+    a level of the tree in order to maintain immutability. This operation requires de-referencing pointers to Python
+    objects stored in the Context, which typically includes objects not created or managed by this library. It's
+    possible for such objects to have mis-managed reference counts (speculatively:  in order to convert their
+    ContextVar storage from a strong to a weak reference. When such objects are de-referenced--as they would be when
```
emmettbutler, line 23:
```suggestion
    ContextVar storage from a strong to a weak reference). When such objects are de-referenced--as they would be when
```

emmettbutler (approved):

sanchda:
@wconti27 thanks for bringing @emmettbutler into the discussion--can you toggle your review?  Merges are currently blocked since your last review required changes.  🙇

wconti27 (approved):

ZStriker19 (approved):

taegyunkim:
Whenever I see failing tests in CI for this PR, I've been pushing retry button. I've pressed the button enough number of times to believe those are actual failures.

sanchda:
> Whenever I see failing tests in CI for this PR, I've been pushing retry button. I've pressed the button enough number of times to believe those are actual failures.

but what if it passes if I retry just one more time (jk, checking)

PR #10538, opened 2024-09-05
chore(profiling): venvs to test profiler v2

taegyunkim:
- Created venvs with name `profile-v2`, it only runs tests in `tests/profiling-v2` with environment variable `DD_PROFILING_EXPORT_LIBDD_ENABLED` set. We plan to add `DD_PROFILING_STACK_V2_ENABLED` soon. 
- Updated gitlab to run for `profile$|profile-v2`
- Moved some tests to `tests/profiling-v2`

## Checklist

## Reviewer Checklist
- Title is accurate

PROF-10349


emmettbutler (approved):

sanchda (approved):
Yeah, I think this is the right way to go.  In particular, if we don't follow this strategy, then we'll need to compartmentalize individual tests in their own subprocesses, which will make it harder to deduce global state issues we don't explicitly test for.  Plus this way we can write the test in a way that is more similar to how an end-user application will be instrumented.

PR #10607, opened 2024-09-10
chore(rq): integrated core api to rq integration

jessicagamio:
Implementing the  Core API into the existing rq integration.

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/contrib/rq/__init__.py
```diff
@@ -134,12 +135,17 @@ def traced_queue_enqueue_job(rq, pin, func, instance, args, kwargs):
...
+            "rq.queue.enqueue_job", provider="rq", direction=SpanDirection.OUTBOUND
+        ),
+        pin=pin,
         service=trace_utils.int_service(pin, config.rq),
         resource=resource,
         span_type=SpanTypes.WORKER,
-    ) as span:
+        call_key="queue.enqueue_job",
+    ) as ctx, ctx[ctx["call_key"]] as span:
         span.set_tag_str(COMPONENT, config.rq.integration_name)
```
emmettbutler, line 149: You'll want to replace these `set_tag_str` calls with a `dispatch` call. That's fine to do in a separate pull request if you prefer.

ddtrace/contrib/rq/__init__.py
```diff
@@ -202,7 +217,7 @@ def traced_perform_job(rq, pin, func, instance, args, kwargs):
     finally:
         # Force flush to agent since the process `os.exit()`s
         # immediately after this method returns
-        pin.tracer.flush()
+        ctx["pin"].tracer.flush()
```
emmettbutler, line 220: This change looks incorrect to me given that `ctx` doesn't exist anymore at this point in the code.

emmettbutler (changes requested):
Aside from the one note I left, this looks great. Nice work so far. Please also add a brief description to the pull request.

ddtrace/contrib/rq/__init__.py
jessicagamio: @emmettbutler  sorry didn't see these updates. I added the span tags under "tags" in the context. The only time i used the dispatch was for `worker.perform.job`  where the job.status and job.origin tags needed to be added after the job is performed in order to match the snapshot trace.

Is it advised to always add the tags via the handler?

ddtrace/_trace/trace_handlers.py
```diff
@@ -691,6 +691,28 @@ def _on_botocore_patched_bedrock_api_call_success(ctx, reqid, latency, input_tok
     span.set_tag_str("bedrock.usage.completion_tokens", output_token_count)
 
 
+def _on_rq_queue_enqueue_job_is_async(ctx, job):
```
mabdinur, line 694: Can we make the function name more generic so we can reuse this function in other integrations.

```suggestion
def _on_propagate_context(ctx, job):
```

ddtrace/_trace/trace_handlers.py
```diff
@@ -691,6 +691,28 @@ def _on_botocore_patched_bedrock_api_call_success(ctx, reqid, latency, input_tok
...
+    span = ctx[ctx["call_key"]]
```
mabdinur, line 695: Is ctx always expected to have call_key. Do we need to check if the key exists and then log a warning/error if it doesn't. Ideally we should never crash a customers application

ddtrace/_trace/trace_handlers.py
```diff
@@ -691,6 +691,28 @@ def _on_botocore_patched_bedrock_api_call_success(ctx, reqid, latency, input_tok
...
+    """sets job.status and job.origin span tags after job is performed"""
+    # get_status() returns None when ttl=0
+    span = ctx[ctx["call_key"]]
+    span.set_tag_str("job.status", job.get_status() or "None")
+    span.set_tag_str("job.origin", job.origin)
+    if job.is_failed:
+        span.error = 1
+
+
+def _on_rq_end_of_traced_method_in_fork(ctx, job):
```
mabdinur, line 709: Again can we make this function name more general. We can potentially reuse this function in other integrations

ddtrace/contrib/rq/__init__.py
```diff
@@ -134,37 +136,43 @@ def traced_queue_enqueue_job(rq, pin, func, instance, args, kwargs):
...
-
-        span.set_tag_str("queue.name", instance.name)
-        span.set_tag_str("job.id", job.get_id())
-        span.set_tag_str("job.func_name", job.func_name)
-
+        call_key="queue.enqueue_job",
+        tags={
+            COMPONENT: config.rq.integration_name,
+            SPAN_KIND: SpanKind.PRODUCER,
+            "queue.name": instance.name,
```
mabdinur, line 152: I know we don't have this now but it would be nice if we had constants for these keys

ddtrace/contrib/rq/__init__.py
```diff
@@ -134,37 +136,43 @@ def traced_queue_enqueue_job(rq, pin, func, instance, args, kwargs):
...
+            "job.id": job.get_id(),
+            "job.func_name": job.func_name,
+        },
+    ) as ctx, ctx[ctx["call_key"]] as span:
         # If the queue is_async then add distributed tracing headers to the job
         if instance.is_async and config.rq.distributed_tracing_enabled:
-            HTTPPropagator.inject(span.context, job.meta)
+            core.dispatch("rq.queue.enqueue_job", [ctx, job])
```
mabdinur, line 159: Can we pass config.rq to this handler. We should move the details of checking if distributing tracing is enabled to the core api. Like how it's handled in _start_span

ddtrace/contrib/rq/__init__.py
```diff
@@ -174,35 +182,31 @@ def traced_perform_job(rq, pin, func, instance, args, kwargs):
...
-                if job.is_failed:
-                    span.error = 1
+                # call _after_perform_job handler for job status and origin
+                core.dispatch("rq.worker.perform_job", [ctx, job])
+
     finally:
         # Force flush to agent since the process `os.exit()`s
         # immediately after this method returns
-        pin.tracer.flush()
+
```
mabdinur, line 207: Nit white space 
```suggestion
```

ddtrace/_trace/trace_handlers.py
```diff
@@ -691,6 +691,28 @@ def _on_botocore_patched_bedrock_api_call_success(ctx, reqid, latency, input_tok
     span.set_tag_str("bedrock.usage.completion_tokens", output_token_count)
 
 
+def _on_rq_queue_enqueue_job_is_async(ctx, job):
+    span = ctx[ctx["call_key"]]
+    HTTPPropagator.inject(span.context, job.meta)
+
+
+def _on_rq_after_perform_job(ctx, job):
```
mabdinur, line 699: We can make this function more general. It can accept a list of key value pairs and an isfailed flag that is used to determine whether span.error=1 should be set. This will allow us to reuse this method in other integrations

mabdinur (commented):
Let some nuts but overall this looks really good. Great job!

ddtrace/contrib/rq/__init__.py
emmettbutler: @jessicagamio it's not - your new approach of using the `tags` kwarg on `context_with_data` is better.

ddtrace/_trace/trace_handlers.py
jessicagamio: @emmettbutler  from what I see `call_key` isn't really being verified to exist in other handlers..it just is expected to exist. Is there an integration example where this is not included? 

I've been using `span=ctx[ctx["call_key"]]` to inject the  span.context or add span tags...is there an alternative way of doing this without call_key?

ddtrace/_trace/trace_handlers.py
jessicagamio: @mabdinur hopefully i succeeded in this?

mabdinur: Yup this looks awesome!

mabdinur (approved):
Tests are passing: https://gitlab.ddbuild.io/DataDog/apm-reliability/dd-trace-py/-/jobs/642791801.

This change looks good to me. Thanks for sticking with it

ddtrace/_trace/trace_handlers.py
emmettbutler: `ctx["call"]` accesses the [default](https://github.com/DataDog/dd-trace-py/blob/e692716e8211b50555066e18c17285d4c45c4985/ddtrace/_trace/trace_handlers.py#L121) attribute that stores the span on contexts that have the [start_span](https://github.com/DataDog/dd-trace-py/blob/e692716e8211b50555066e18c17285d4c45c4985/ddtrace/_trace/trace_handlers.py#L852) handler attached to them. This is slightly more "safe" than using `call_key`, but only marginally so because both access data set conditionally elsewhere in the code. `if ctx.get_item("call", None) is not None` is the safest way to access a possibly-extant span stored on an `ExecutionContext`.

emmettbutler (changes requested):
This looks great. Seems like you've got your head around how the Core API works. There are a few bits that should be changed, but the main body of this change is ready to go.

ddtrace/contrib/rq/__init__.py
```diff
@@ -174,35 +186,32 @@ def traced_perform_job(rq, pin, func, instance, args, kwargs):
...
+                # call _after_perform_job handler for job status and origin
+                span_tags = {"job.status": job.get_status() or "None", "job.origin": job.origin}
+                job_failed = job.is_failed
+                core.dispatch("rq.worker.perform_job", [ctx, job_failed, span_tags])
+
     finally:
         # Force flush to agent since the process `os.exit()`s
         # immediately after this method returns
-        pin.tracer.flush()
+        core.context_with_data("rq.worker.after.perform.job")
```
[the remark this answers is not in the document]
jessicagamio: Prior to this I was reusing "rq.worker.perform"  job for two different handlers and this was causing conflict. The solution was to create a separate event for this dispatch handler. If you have any suggestions on how I can better handle this scenario I'd appreciate it.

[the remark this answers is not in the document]
emmettbutler: As far as I can tell, this line creates a context and then immediately discards it. There's nothing listening to the events emitted by the creation of that context, so no spans are created. From my read of the code, deleting this line would not change the behavior.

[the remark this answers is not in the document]
jessicagamio: Ah...okay so i don't have to "register" this event as an argument in context_with_data, that was a misunderstanding on my part.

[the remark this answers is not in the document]
jessicagamio: @emmettbutler  please let me know if the changes made here are acceptable to merge this pr

emmettbutler (approved):

PR #10612, opened 2024-09-10
ci: update which packages to select for update

quinna-h:
Skip over pinned modules when checking for outdated packages. 

Since the modules are pinned, we won't actually update the latest that we use from the riotfile, so it is not necessary to include them in the list of candidates for updating a package version.

Future items:
- Figure out a more robust solution for mapping the suite name to the integration name (currently hardcoded in a `dict`)

## Checklist

## Reviewer Checklist
- Title is accurate
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -


emmettbutler (commented):
If I understand this series of workflows properly, [this line](https://github.com/DataDog/dd-trace-py/blob/2e9d3b15977b14e94bb546f718bec3de91985876/scripts/freshvenvs.py#L70) in `freshvenvs.py` is the original source of the list of packages that might need updates. Instead of setting up another workflow and committing another text file to source control, I think it's possible to achieve the goal of avoiding updates for certain packages by putting the `update_pinned.py` logic directly in the `freshvenvs` script.

quinna-h:
> If I understand this series of workflows properly, [this line](https://github.com/DataDog/dd-trace-py/blob/2e9d3b15977b14e94bb546f718bec3de91985876/scripts/freshvenvs.py#L70) in `freshvenvs.py` is the original source of the list of packages that might need updates. Instead of setting up another workflow and committing another text file to source control, I think it's possible to achieve the goal of avoiding updates for certain packages by putting the `update_pinned.py` logic directly in the `freshvenvs` script.

Addressed! A concern I had was how frequently are we running `scripts/freshvenvs` vs updating the riotfile (the thought being that parsing through all the `Venvs` in riotfile is time consuming), but on second thought doing this once a week is fine.

scripts/freshvenvs.py
```diff
@@ -8,16 +8,35 @@
...
 
+import riotfile
 
-sys.path.append(str(pathlib.Path(__file__).parent.parent.resolve()))
 
 CONTRIB_ROOT = pathlib.Path("ddtrace/contrib")
 
 
+# Venv suite name -> integration name
+module_suite_mapping = {
```
emmettbutler, line 26: It's important to keep the difference between modules and packages clear in this code. I think this would more accurately be called `suite_to_package`, a package being a thing that's downloaded from PyPI.

scripts/freshvenvs.py
```diff
@@ -67,10 +86,22 @@ def _get_riot_envs_including_any(modules: typing.Set[str]) -> typing.Set[str]:
...
+        pinned = _iterate_venvs(v, package)
+        if pinned:
+            current_pinned.add(package)
 
+    packages = {m for m in modules if "." not in m and m not in current_pinned}
+    return packages
 
-def _get_version_extremes(package_name: str) -> typing.Tuple[str, str]:
+
+def _get_version_extremes(package_name: str) -> typing.Tuple[Optional[str], Optional[str]]:
```
emmettbutler, line 104: Nice catch

scripts/freshvenvs.py
```diff
@@ -124,6 +155,28 @@ def _versions_fully_cover_bounds(bounds: typing.Tuple[str, str], versions: typin
...
+    """
+    DFS through the Venv, as it may have nested Venvs.
+    Returns whether the package was pinned or not.
+
+    If package is in module_suite_mapping, remap it.
+    """
+    package = module_suite_mapping.get(package, package)
+
+    if package in venv.pkgs:
+        if "" in venv.pkgs[package]:
```
emmettbutler, line 168: Renaming `""` to `LATEST = ""` at the global level would make this code clearer.

scripts/freshvenvs.py
```diff
@@ -8,16 +8,35 @@
...
 from pip import _internal
 
+import riotfile
 
-sys.path.append(str(pathlib.Path(__file__).parent.parent.resolve()))
 
 CONTRIB_ROOT = pathlib.Path("ddtrace/contrib")
 
 
+# Venv suite name -> integration name
```
emmettbutler, line 25: Is this comment accurate? It's probably better to name the variable such that this comment becomes redundant.

scripts/freshvenvs.py
```diff
@@ -124,6 +155,28 @@ def _versions_fully_cover_bounds(bounds: typing.Tuple[str, str], versions: typin
     return versions[0] >= Version(upper_bound)
 
 
+def _iterate_venvs(venv: riotfile.Venv, package: str) -> bool:
```
emmettbutler, line 158: Given the use of `package` on line 165, it seems to me that the second parameter to this function is treated as a "suite name" rather than a "package name". Maybe `suite_name` would be a more descriptive name?

scripts/freshvenvs.py
emmettbutler: `"" in "any string"` returns `True`, so if `venv.pkgs[package]` is a string (as it is [here](https://github.com/DataDog/dd-trace-py/blob/b488c99d2b8a76e637ef46079e20e7a21e0748a1/riotfile.py#L227)), we'll erroneously enter this `if` block.

scripts/freshvenvs.py
```diff
@@ -124,6 +155,28 @@ def _versions_fully_cover_bounds(bounds: typing.Tuple[str, str], versions: typin
...
+    """
+    package = module_suite_mapping.get(package, package)
+
+    if package in venv.pkgs:
+        if "" in venv.pkgs[package]:
+            return False
+
+    if venv.venvs:
+        for child_venv in venv.venvs:
+            is_pinned = _iterate_venvs(child_venv, package)
```
emmettbutler, line 173: "pinned" is a tricky concept in this context, because there are lots of patterns we can use in the riotfile other than `latest` that specify a range of versions rather than one specific version (examples: "~=1.2.3", ">=2.3.4"). You could sidestep that trickiness by renaming `_iterate_venvs` to `_venv_sets_latest_for_package` or similar, and inverting the return values.

scripts/freshvenvs.py
```diff
@@ -67,10 +86,22 @@ def _get_riot_envs_including_any(modules: typing.Set[str]) -> typing.Set[str]:
 
 
 def _get_packages_implementing(modules: typing.Set[str]) -> typing.Set[str]:
```
emmettbutler, line 88: Because this function now does filtering, consider renaming it to make that clear to the caller (possibly `_get_updateable_packages_implementing`)

emmettbutler (changes requested):
This is looking great. I left a bunch of comments about clarity and readability, plus one that appears to be a logical error. Let me know when you'd like another review!

scripts/freshvenvs.py
quinna-h: With the way the `Venv`s are interpreted, the string ends up becoming a list of a single string. Example [here](https://github.com/DataDog/dd-trace-py/blob/b488c99d2b8a76e637ef46079e20e7a21e0748a1/riotfile.py#L2619) if we print the `package` and `venv.pkgs[package]`:

langchain
['==0.0.192']

scripts/freshvenvs.py
quinna-h: Thank you for the catch and calling that out!

scripts/freshvenvs.py
```diff
@@ -66,11 +84,24 @@ def _get_riot_envs_including_any(modules: typing.Set[str]) -> typing.Set[str]:
...
+    """Return all packages that can be updated and have contribs implemented for them"""
+    all_venvs = riotfile.venv.venvs
+    version_constrained_modules = set() # set of modules with pinned versions or a version range
+
+    for v in all_venvs:
+        package = v.name
+        if package not in modules:
+            continue
+        has_latest = _venv_sets_latest_for_package(v, package)
+        if not has_latest:
```
emmettbutler, line 97:
```suggestion
        if not _venv_sets_latest_for_package(v, package):
```

scripts/freshvenvs.py
```diff
@@ -66,11 +84,24 @@ def _get_riot_envs_including_any(modules: typing.Set[str]) -> typing.Set[str]:
...
+            version_constrained_modules.add(package)
```
emmettbutler, line 98:
```suggestion
            modules -= package
```

scripts/freshvenvs.py
```diff
@@ -66,11 +84,24 @@ def _get_riot_envs_including_any(modules: typing.Set[str]) -> typing.Set[str]:
...
+
+    packages = {m for m in modules if "." not in m and m not in version_constrained_modules}
```
emmettbutler, line 100:
```suggestion
    packages = {m for m in modules if "." not in m}
```

scripts/freshvenvs.py
```diff
@@ -124,9 +155,31 @@ def _versions_fully_cover_bounds(bounds: typing.Tuple[str, str], versions: typin
...
+    package = suite_to_package.get(suite_name, suite_name)
+
+    if package in venv.pkgs:
+        if LATEST in venv.pkgs[package]:
+            return True
+
+    if venv.venvs:
+        for child_venv in venv.venvs:
+            includes_latest = _venv_sets_latest_for_package(child_venv, package)
+            if includes_latest:
```
emmettbutler, line 174:
```suggestion
            if _venv_sets_latest_for_package(child_venv, package):
```

emmettbutler (approved):

PR #10617, opened 2024-09-10
ci: migrate opentelemetry test suite from circleci to gitlab

erikayasuda:
This PR migrates the opentelemetry test suite from CircleCI to GitLab for cost purposes.

## Checklist

## Reviewer Checklist
- Title is accurate


.gitlab/tests/contrib.yml
```diff
@@ -1,3 +1,12 @@
+aiohttp:
```
erikayasuda, line 1: All integration changes EXCEPT for opentelemetry here are just reshuffling for alphabetizing the order.

tests/opentelemetry/test_context.py
```diff
@@ -89,7 +89,7 @@ def _subprocess_task(parent_span_context, errors):
         ot_tracer._tracer.flush()
 
 
-@pytest.mark.snapshot(ignores=["meta.tracestate"])
+@pytest.mark.snapshot(ignores=["meta.tracestate", "meta._dd.parent_id"])
```
erikayasuda, line 92: Doing this to have these tests pass, but it looks like we have another `parent_id` field that looks pre-set. Maybe we can do the same with this value instead of ignoring it? 😅

emmettbutler (approved):

tests/opentelemetry/test_context.py
mabdinur: We should ignore it or update the trace normalization rules in the testagent. `_dd.parent_id` is set using a remote span's span_id. Span_ids are not consistent across test runs, so this field doesn't belong in a snapshot file.

erikayasuda: By "remote span" do you mean any span that doesn't share the same process? In that case I think this is correct since the tests that regenerated a new snapshot are the ones that test across process or forking

erikayasuda:
Closing this for now. Will possibly revisit.

PR #10594, opened 2024-09-10
chore: [SVLS-5261] change Span._links to a list

apiarian-datadog:
We are going to be using the `Span._links` structure to store Span Pointers soon. Span Pointers will have look like SpanLinks but will have `(trace_id=0, span_id=0)` so we cannot keep them in the old `_links` dictionary keyed by `span_id`.

## Checklist
- [x] PR author has checked that all the criteria below are met
- The PR description includes an overview of the change
- The PR description articulates the motivation for the change
- The change includes tests OR the PR description describes a testing strategy
- The PR description notes risks associated with the change, if any
- Newly-added code is easy to change
- The change follows the [library release note guidelines](https://ddtrace.readthedocs.io/en/stable/releasenotes.html)
- The change includes or references documentation updates if necessary
- Backport labels are set (if [applicable](https://ddtrace.readthedocs.io/en/latest/contributing.html#backporting))

## Reviewer Checklist
- [x] Reviewer has checked that all the criteria below are met 
- Title is accurate
- All changes are related to the pull request's stated goal
- Avoids breaking [API](https://ddtrace.readthedocs.io/en/stable/versioning.html#interfaces) changes
- Testing strategy adequately addresses listed risks
- Newly-added code is easy to change
- Release note makes sense to a user of the library
- If necessary, author has acknowledged and discussed the performance implications of this PR as reported in the benchmarks PR comment
- Backport labels are set in a manner that is consistent with the [release branch maintenance policy](https://ddtrace.readthedocs.io/en/latest/contributing.html#backporting)

PR #10621, opened 2024-09-10
feat(tracer): [SVLS-5261] private Span Pointers

apiarian-datadog:
Span Pointers are stored alongside Span Links. The API is private for now, so we don't have much documentation and all the new classes are prefixed with underscores.

## Checklist

## Reviewer Checklist
- Title is accurate

### DIFF
--- a/ddtrace/_trace/_span_link.py
+++ b/ddtrace/_trace/_span_link.py
@@ -37,6 +37,14 @@ from ddtrace.internal.utils.formats import flatten_key_value
 log = get_logger(__name__)
 
 
+class SpanLinkKind(Enum):
+    """
+    A collection of standard SpanLink kinds. It's possible to use others, but these should be used when possible.
+    """
+
+    SPAN_POINTER = "span-pointer"  # Should not be used on normal SpanLinks.
+
+
 def _id_not_zero(self, attribute_name, value):
     if not value > 0:
         raise ValueError(f"{attribute_name} must be > 0. Value is {value}")
@@ -129,7 +137,6 @@ class SpanLink:
 # Span Pointers are currently private, so let's put them here for now
 
 
-_SPAN_LINK_KIND_SPAN_POINTER = "span-pointer"
 _SPAN_POINTER_SPAN_LINK_TRACE_ID = 0
 _SPAN_POINTER_SPAN_LINK_SPAN_ID = 0
 
@@ -158,7 +165,7 @@ class _SpanPointer(SpanLink):
             },
         )
 
-        self.kind = _SPAN_LINK_KIND_SPAN_POINTER
+        self.kind = SpanLinkKind.SPAN_POINTER.value
 
     def __post_init__(self):
         # Do not want to do the trace_id and span_id checks that SpanLink does.
--- a/ddtrace/_trace/_span_link.py
+++ b/ddtrace/_trace/_span_link.py
@@ -27,6 +27,7 @@ SpanLinks can be set using :meth:`ddtrace.Span.link_span(...)` Ex::
 import dataclasses
 from enum import Enum
 from typing import Any
+from typing import Dict
 from typing import Optional
 
 from ddtrace.internal.utils.formats import flatten_key_value
@@ -140,7 +141,7 @@ class _SpanPointer(SpanLink):
         pointer_kind: str,
         pointer_direction: _SpanPointerDirection,
         pointer_hash: str,
-        extra_attributes: Optional[dict[str, Any]] = None,
+        extra_attributes: Optional[Dict[str, Any]] = None,
     ):
         super().__init__(
             trace_id=_SPAN_POINTER_SPAN_LINK_TRACE_ID,
--- a/ddtrace/_trace/_span_link.py
+++ b/ddtrace/_trace/_span_link.py
@@ -25,6 +25,8 @@ SpanLinks can be set using :meth:`ddtrace.Span.link_span(...)` Ex::
 """
 
 import dataclasses
+from enum import Enum
+from typing import Any
 from typing import Optional
 
 from ddtrace.internal.utils.formats import flatten_key_value
@@ -70,8 +72,8 @@ class SpanLink:
         self.attributes["link.name"] = value
 
     @property
-    def kind(self):
-        return self.attributes["link.kind"]
+    def kind(self) -> Optional[Any]:
+        return self.attributes.get("link.kind")
 
     @kind.setter
     def kind(self, value):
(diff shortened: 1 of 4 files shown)


apiarian-datadog:
Built on top of https://github.com/DataDog/dd-trace-py/pull/10594 for now.

ddtrace/_trace/span.py
```diff
@@ -631,11 +634,28 @@ def set_link(
...
+        self._set_link_object(
+            _SpanPointer(
+                pointer_kind=pointer_kind,
+                pointer_direction=pointer_direction,
+                pointer_hash=pointer_hash,
+                extra_attributes=extra_attributes,
+            )
+        )
+
+    def _set_link_object(self, link: Union[SpanLink, _SpanPointer]) -> None:
```
ZStriker19, line 655: I'm a bit worried about this method name being confusing since it's only applicable for span pointers. Could we change it to something like `_set_pointer_link` or `append_pointer_link`?

ddtrace/_trace/_span_link.py
```diff
@@ -70,8 +73,8 @@ def name(self, value):
         self.attributes["link.name"] = value
 
     @property
-    def kind(self):
-        return self.attributes["link.kind"]
+    def kind(self) -> Optional[Any]:
```
ZStriker19, line 76:
```suggestion
    def kind(self) -> Optional[str]:
```
Is it not always a string? Maybe out of scope.

ZStriker19 (approved):

ddtrace/_trace/span.py
apiarian-datadog: well, it takes *either* a span link or a span pointer. how about `_set_link_or_pointer` instead?

ddtrace/_trace/_span_link.py
apiarian-datadog: it technically *should* be a string. we can mark it as such on the setter, too? but if we're not going to throw an exception when someone says `link.kind = {"something": "funky"}` i'd prefer to stick with `Any`?

apiarian-datadog: ok, changed this and the setter to mark them as expecting an `str`

ddtrace/_trace/_span_link.py
```diff
@@ -117,3 +120,45 @@ def __str__(self) -> str:
...
+                "ptr.hash": pointer_hash,
+                **(extra_attributes or {}),
+            },
+        )
+
+        self.kind = _SPAN_LINK_KIND_SPAN_POINTER
+
+    def __post_init__(self):
+        if self.trace_id != _SPAN_POINTER_SPAN_LINK_TRACE_ID:
+            raise ValueError(f"span pointer trace_id must be {_SPAN_POINTER_SPAN_LINK_TRACE_ID}")
```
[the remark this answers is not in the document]
apiarian-datadog: changed to a `log.warning` and an override

ddtrace/_trace/_span_link.py
```diff
@@ -117,3 +124,55 @@ def __str__(self) -> str:
             f"trace_id={self.trace_id} span_id={self.span_id} attributes={attrs_str} "
             f"tracestate={self.tracestate} flags={self.flags} dropped_attributes={self._dropped_attributes}"
         )
+
+
+# Span Pointers are currently private, so let's put them here for now
```
apiarian-datadog, line 129: as this api gets more interesting and we add actual span linking functionality, we are going to be moving this code out to `ddtrace/_trace/_span_pointers.py` (I have a version of this on another branch now, but getting it ported here would be a bit of a hassle. let me know if anyone feels strongly that this should start in that other file, otherwise we'll leave it to the next PR)

apiarian-datadog: here's that next PR: https://github.com/DataDog/dd-trace-py/pull/10639/

ddtrace/_trace/span.py
```diff
@@ -631,11 +634,28 @@ def set_link(
...
+    def _add_span_pointer(
+        self,
+        pointer_kind: str,
+        pointer_direction: _SpanPointerDirection,
+        pointer_hash: str,
+        extra_attributes: Optional[Dict[str, Any]] = None,
+    ) -> None:
+        # This is a Private API for now.
+
+        self._set_link_or_append_pointer(
```
mabdinur, line 646: I am not sure if this has been discussed but can we avoid storing span pointers and "regular" span links in the same  span property? This will allow us to only serialize span pointers when sending trace payloads to severless trace endpoints (in the jsonencoder). I don't think we should send span pointers to the Datadog Agent (at least in the near term).

We can also achieve this functionality by removing span pointers from Span._links in the MsgPackEncoders (in _encoding.pyx). Either works for me.

ddtrace/_trace/span.py
```diff
@@ -631,11 +634,28 @@ def set_link(
...
+            _SpanPointer(
+                pointer_kind=pointer_kind,
+                pointer_direction=pointer_direction,
+                pointer_hash=pointer_hash,
+                extra_attributes=extra_attributes,
+            )
+        )
+
+    def _set_link_or_append_pointer(self, link: Union[SpanLink, _SpanPointer]) -> None:
```
mabdinur, line 655: Do we need a union here? _SpanPointers should also be a SpanLink
```suggestion
    def _set_link_or_append_pointer(self, link: SpanLink) -> None:
```

ddtrace/_trace/span.py
mabdinur: Since this API is private we don't need to address this issue in this PR.

mabdinur (approved):
left some nits overall LGTM

ddtrace/_trace/span.py
apiarian-datadog: yup, but i wanted to extra clear that we can store either a span link or a span pointer here. currently span pointers are implemented as span links, but i'd argue that's an implementation detail.

ddtrace/_trace/_span_link.py
```diff
@@ -117,3 +124,55 @@ def __str__(self) -> str:
...
+                "ptr.dir": pointer_direction.value,
+                "ptr.hash": pointer_hash,
+                **(extra_attributes or {}),
+            },
+        )
+
+        self.kind = _SPAN_LINK_KIND_SPAN_POINTER
+
+    def __post_init__(self):
+        if self.trace_id != _SPAN_POINTER_SPAN_LINK_TRACE_ID:
```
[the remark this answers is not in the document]
apiarian-datadog: not technically, i just trust nobody. happy to remove it if you feel it's too cautious =]

ddtrace/_trace/span.py
mabdinur: Awesome! I just wanted to double check that trace intake was aware of this change. If they can support span links with trace_id and span_id equal to zero then that's perfect. We can keep this change as is.

ddtrace/_trace/_span_link.py
```diff
@@ -117,3 +124,55 @@ def __str__(self) -> str:
             f"trace_id={self.trace_id} span_id={self.span_id} attributes={attrs_str} "
             f"tracestate={self.tracestate} flags={self.flags} dropped_attributes={self._dropped_attributes}"
         )
+
+
+# Span Pointers are currently private, so let's put them here for now
+
+
+_SPAN_LINK_KIND_SPAN_POINTER = "span-pointer"
```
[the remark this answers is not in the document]
apiarian-datadog: done

ddtrace/_trace/_span_link.py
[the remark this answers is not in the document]
apiarian-datadog: removed this.

ddtrace/_trace/span.py
apiarian-datadog: i've actually added a similar typehint to the `_links` attribute. please let me know if you would prefer i went the other way.

mabdinur: Doesn't really matter to me. We can keep it.

quinna-h (approved):

PR #10638, opened 2024-09-11
feat(llmobs): support tagging integration-generated spans

lievan:
This PR introduces `AnnotationContext` and `AnnotationContextAsync`, which are context managers that allow users to modify integration generated spans.

Create an `AnnotationContext` object by calling LLMObs.annotation_context(). On entering this context, a function is registered on span startup that modifies the span with data passed into annotation_context()

This PR only exposes a `tags` argument for `annotation_context`. This is a good candidate as a first argument because,
1. customers have asked for it
2. span kind agnostic
3. there's simple merging logic for tags

We also modify the `annotate` function to allow for tagging non-IO span data which does not depend on span kind so annotate can be re-used within the annotation context

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/llmobs/_llmobs.py
```diff
@@ -226,6 +227,21 @@ def disable(cls) -> None:
...
+        """
+        Sets specified attributes on all spans created while the returned AnnotationContext is active.
+
+        :param tags: Dictionary of JSON serializable key-value tag pairs to set or update on the LLMObs span
+                     regarding the span's context.
+        """
+
+        def mini_annotate(span):
+            if tags:
+                cls._tag_span_tags(span, tags)
```
Yun-Kim, line 241: Why not just use `LLMObs.annotate(span, tags=tags)`?

ddtrace/llmobs/_llmobs.py
```diff
@@ -226,6 +227,21 @@ def disable(cls) -> None:
 
         log.debug("%s disabled", cls.__name__)
 
+    @classmethod
+    def annotation_context(cls, tags: Optional[Dict[str, Any]] = None) -> AnnotationContext:
+        """
+        Sets specified attributes on all spans created while the returned AnnotationContext is active.
```
Yun-Kim, line 233:
```suggestion
        Sets specified attributes on all LLMObs spans created while the returned AnnotationContext is active.
```

releasenotes/notes/llmobs-annotation-context-ece8f33a805adce3.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    llmobs: This introduces the `LLMObs.annotation_context` method which enables users to modify certain attributes of integration generated spans.
+    `LLMObs.annotation_context` returns a context manager. While the context manager is active, all LLM Obs spans created will have the specified
+    data passed into `LLMObs.annotation_context`.
```
Yun-Kim, line 6:
```suggestion
     LLM Observability: Introduces the `LLMObs.annotation_context()` context manager method which allows modifying or overriding the tags of integration generated LLM Observability spans created while the conext manager is active.
```
In Python release notes, the standard is to refer to our product as `LLM Observability` instead of `llmobs`. Also let's be extra clear that this is currently for tags only.

ddtrace/llmobs/utils.py
```diff
@@ -21,6 +21,18 @@
 Message = TypedDict("Message", {"content": str, "role": str}, total=False)
 
 
+class AnnotationContext:
+    def __init__(self, _tracer, _annotation_callback):
+        self._tracer = _tracer
+        self._annotate_prompt = _annotation_callback
+
+    def __enter__(self):
+        self._tracer.on_start_span(self._annotate_prompt)
```
Yun-Kim, line 30: > There's a limitation here - we don't know the span kind right when the span is started since this value is set as a tag after-the-fact. This is problematic for Langchain spans, where the span kind may be conditional based on some post-processing logic

Assuming you're talking about `span.span_type`, why is this problematic for langchain spans? I believe `_on_start_span()` hooks run after `tracer._start_span(span_type=...)` has begun.

ddtrace/llmobs/_llmobs.py
lievan: `annotate` requires the `SPAN_KIND` tag to be set, which isn't available on span start (since tags are set after spans are started)

ddtrace/llmobs/utils.py
lievan: Sorry i mean the span kind (llm, retrieval, embedding, etc.) which is set as a tag, not span kind

sabrenner (approved):
LGTM 

one thought: did we also want to offer an async version of the `AnnotationContext`? ie `LLMObs.annotation_context_async()`? not sure if something like that is needed but thought i'd raise it anyways

ddtrace/llmobs/_llmobs.py
Yun-Kim: I think I'd prefer to just move up the `metadata/tags/metrics` tagging in the annotate() function to be before the span kind check, since the span kind check is only necessary for I/O

ddtrace/llmobs/utils.py
Yun-Kim: I'm still a tad confused, why is not knowing `span_kind` a limitation for settings tags in the annotation context? Or are you referring to future considerations when we will add prompt/I/O options to annotation contexts?

ddtrace/llmobs/utils.py
```diff
@@ -21,6 +21,18 @@
 Message = TypedDict("Message", {"content": str, "role": str}, total=False)
 
 
+class AnnotationContext:
```
Yun-Kim, line 24: +1 to Sam's comment, we'll need to provide an async version of this as well

ddtrace/llmobs/_llmobs.py
lievan: That makes sense!

ddtrace/llmobs/utils.py
lievan: Yea, i was referring to future considerations when the span kind is necessary. But for the data was need to annotate right now using context (prompt, name, tags) i don't think it is

releasenotes/notes/llmobs-annotation-context-ece8f33a805adce3.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+     LLM Observability: Introduces the `LLMObs.annotation_context()` context manager method which allows modifying or overriding the tags of integration generated LLM Observability spans created while the context manager is active.
```
Yun-Kim, line 4:
```suggestion
     LLM Observability: Introduces the `LLMObs.annotation_context()` and `LLMObs.annotate_context_async()` context manager methods which allows modifying or overriding the tags of integration generated LLM Observability spans created while the context manager is active.
```

ddtrace/llmobs/_llmobs.py
```diff
@@ -226,6 +228,30 @@ def disable(cls) -> None:
...
+        Do not use nested annotation contexts to override the same tags since the order in which annotations
+        are applied is non-deterministic.
+
+        :param tags: Dictionary of JSON serializable key-value tag pairs to set or update on the LLMObs span
+                     regarding the span's context.
+        """
+        return AnnotationContext(cls._instance.tracer, lambda span: cls.annotate(span, tags=tags))
+
+    @classmethod
+    def annotation_context_async(cls, tags: Optional[Dict[str, Any]] = None) -> AnnotationContextAsync:
```
Yun-Kim, line 244: Does this need to be `async def ...`? Just to be safe, can we test out with an actual async example (an async openai call?)

lievan: I don't think so, since the function is just meant to return the async context manager which will then be used by the user using `async with`

Will send you the code snippet i used to test offline!

ddtrace/llmobs/utils.py
```diff
@@ -21,6 +21,24 @@
...
```
Yun-Kim, line 24: Can we move this to `llmobs/_utils.py`? I'd rather users only add annotation contexts via `LLMObs.annotation_context()`

Yun-Kim (commented):
Small request on making the AnnotationContext class private, thanks!

Yun-Kim (approved):

radsimu-spektr:
how can one use the custom tags in order to filter/search traces?

PR #10642, opened 2024-09-11
feat(tracer): implement AWS payload tagging for request/response

bouwkast:
## Overview

This pull request adds the ability to expand AWS request/response payloads as span tags.
This matches our lambda offerings and provides useful information to developers when debugging communication between various AWS services.
This is based on the AWS Payload Tagging RFC and this implementation in [dd-trace-node](https://github.com/DataDog/dd-trace-js/pull/4309) and this implementation in [dd-trace-java](https://github.com/DataDog/dd-trace-java/pull/7312).

This feature is _disabled_ by default.

When activated this will produce span tags such as:

```
 "aws.request.body.PublishBatchRequestEntries.0.Id": "1",
 "aws.request.body.PublishBatchRequestEntries.0.Message": "ironmaiden",
 "aws.request.body.PublishBatchRequestEntries.1.Id": "2",
 "aws.request.body.PublishBatchRequestEntries.1.Message": "megadeth"
 "aws.response.body.HTTPStatusCode": "200",
```

## Configuration

There are five new configuration options:

- `DD_TRACE_CLOUD_REQUEST_PAYLOAD_TAGGING`:
  - `""` by default to indicate that AWS request payload expansion is **disabled** for _requests_.
  - `"all"` to define that AWS request payload expansion is **enabled** for _requests_ using the default `JSONPath`s for redaction logic.
  - a comma-separated list of user-supplied `JSONPath`s to define that AWS request payload expansion is **enabled** for _requests_ using the default `JSONPath`s and the user-supplied `JSONPath`s for redaction logic.
- `DD_TRACE_CLOUD_RESPONSE_PAYLOAD_TAGGING`:
  - `""` by default to indicate that AWS response payload expansion is **disabled** for _responses_.
  - `"all"` to define that AWS response payload expansion is **enabled** for _responses_ using the default `JSONPath`s for redaction logic.
  - a comma-separated list of user-supplied `JSONPath`s to define that AWS request payload expansion is **enabled** for _responses_ using the default `JSONPath`s and the user-supplied `JSONPath`s for redaction logic.
- `DD_TRACE_CLOUD_PAYLOAD_TAGGING_MAX_DEPTH` (not defined in RFC but done to match NodeJS): 
  - sets the depth after which we stop creating tags from a payload
  - defaults to a value of `10`
- `DD_TRACE_CLOUD_PAYLOAD_TAGGING_MAX_TAGS` (to match Java implementation)
  - sets the maximum number of tags allowed to be expanded
  - defaults to a value of `758`
- `DD_TRACE_CLOUD_PAYLOAD_TAGGING_SERVICES` (to match Java implementation)
  - a comma-separated list of supported AWS services
  - defaults to ` s3,sns,sqs,kinesis,eventbridge`

## Other

- [`jsonpath-ng` has been vendored](https://github.com/h2non/jsonpath-ng/blob/master/jsonpath_ng/jsonpath.py)
  - [`ply` has been vendored (v3.11) (dependency of `jsonpath-ng`)](https://github.com/dabeaz/ply/releases/tag/3.11)

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/_trace/utils_botocore/aws_payload_tagging.py
```diff
@@ -0,0 +1,192 @@
...
+            if isinstance(obj, bytes):
+                return True
+
+        return False
+
+    def _redact_json(self, data: Dict[str, Any], span: Span) -> None:
+        """
+        Redact sensitive data in the JSON payload based on default and user-provided JSONPath expressions
+        """
+        # TODO should we cache these paths and generated lists?
```
bouwkast, line 58: This is me not knowing python very well, the `Node.js` implementation uses a special JSON file to store these and read from these, Java doesn't.

ddtrace/_trace/utils_botocore/aws_payload_tagging.py
```diff
@@ -0,0 +1,192 @@
...
+            # S3
+            "$..SSECustomerKey",
+            "$..CopySourceSSECustomerKey",
+            "$..RestoreRequest.OutputLocation.S3.Encryption.KMSKeyId"
+        ]
+        if user_paths and user_paths != "all": # "all" is a special value that just enables the expansion and uses defaults
+            return request_defaults + self._get_redaction_paths_default() + user_paths.split(',')
+        return request_defaults + self._get_redaction_paths_default()
+
+    def _tag_object(self, span: Span, key: str, obj: Any, depth: int = 0) -> None:
```
bouwkast, line 144: Via: https://github.com/DataDog/datadog-lambda-python/blob/2b85536bbc24cef46bd61381ec62f57011e0633f/datadog_lambda/tag_object.py#L15

ddtrace/_trace/utils_botocore/span_tags.py
```diff
@@ -14,6 +15,7 @@
 from ddtrace.internal.constants import COMPONENT
 from ddtrace.internal.utils.formats import deep_getattr
 
+payload_tagging_endpoints =  {"s3", "sns", "sqs", "kinesis", "eventbridge"}
```
bouwkast, line 18: Unsure if these should be stored elsewhere 🤷

ddtrace/_trace/utils_botocore/span_tags.py
```diff
@@ -31,6 +33,11 @@ def set_botocore_patched_api_call_span_tags(span: Span, instance, args, params,
         if params and not config.botocore["tag_no_params"]:
             aws._add_api_param_span_tags(span, endpoint_name, params)
 
+        if params and config.botocore["payload_tagging_request"] is not None and config.botocore["payload_tagging_request"].replace(" ", "") != "":
+            if endpoint_name in payload_tagging_endpoints:
+                payload_tagger = AWSPayloadTagging() # TODO where do I put this?
```
bouwkast, line 38: Is it okay to create a new object here? It seems wrong for python?

ddtrace/_trace/utils_botocore/span_tags.py
```diff
@@ -54,6 +61,11 @@ def set_botocore_response_metadata_tags(
         return
     response_meta = result["ResponseMetadata"]
 
+    if config.botocore["payload_tagging_response"] is not None and config.botocore["payload_tagging_response"].replace(" ", "") != "":
+        if span.get_tag("aws_service") in payload_tagging_endpoints:
```
bouwkast, line 65: I'm unsure if calling `span.get_tag` here is a good idea

tests/contrib/botocore/test.py
```diff
@@ -3765,3 +3763,260 @@ def test_schematized_unspecified_service_secretsmanager_v1(self):
 
             assert span.service == DEFAULT_SPAN_SERVICE_NAME
             assert span.name == "aws.secretsmanager.request"
+
+
+    # @pytest.mark.snapshot
```
bouwkast, line 3770: I'd like to add these as snapshot tests, but I can't find documentation/examples of scrubbing out various fields (e.g. dates)

tests/contrib/botocore/test.py
bouwkast, line 1: Full disclosure, I copy/pasted these tests from elsewhere in these tests

ddtrace/contrib/internal/botocore/patch.py
```diff
@@ -104,10 +104,14 @@ def _load_dynamodb_primary_key_names_for_tables() -> Dict[str, Set[str]]:
         "empty_poll_enabled": asbool(os.getenv("DD_BOTOCORE_EMPTY_POLL_ENABLED", default=True)),
         "dynamodb_primary_key_names_for_tables": _load_dynamodb_primary_key_names_for_tables(),
         "add_span_pointers": asbool(os.getenv("DD_BOTOCORE_ADD_SPAN_POINTERS", default=True)),
+        "payload_tagging_request": os.getenv("DD_TRACE_CLOUD_REQUEST_PAYLOAD_TAGGING", default=""),
+        "payload_tagging_response": os.getenv("DD_TRACE_CLOUD_RESPONSE_PAYLOAD_TAGGING", default=""),
+        "payload_tagging_max_depth": int(os.getenv("DD_TRACE_CLOUD_PAYLOAD_TAGGING_MAX_DEPTH", 10)),
+        "payload_tagging_max_tags": int(os.getenv("DD_TRACE_CLOUD_PAYLOAD_TAGGING_MAX_TAGS", 758)),
+        "payload_tagging_services": set(os.getenv("DD_TRACE_CLOUD_PAYLOAD_TAGGING_SERVICES", default="s3,sns,sqs,kinesis,eventbridge").split(",")),
```
bouwkast, line 111: This `.split` feels wrong - I can't find other examples of something like this in other `patch.py` files.

tests/contrib/botocore/test.py
bouwkast: Awesome thanks and done

ddtrace/contrib/internal/botocore/patch.py
bouwkast: Ah thanks, done!

ddtrace/contrib/internal/botocore/patch.py
```diff
@@ -104,10 +104,14 @@ def _load_dynamodb_primary_key_names_for_tables() -> Dict[str, Set[str]]:
         "empty_poll_enabled": asbool(os.getenv("DD_BOTOCORE_EMPTY_POLL_ENABLED", default=True)),
         "dynamodb_primary_key_names_for_tables": _load_dynamodb_primary_key_names_for_tables(),
         "add_span_pointers": asbool(os.getenv("DD_BOTOCORE_ADD_SPAN_POINTERS", default=True)),
+        "payload_tagging_request": os.getenv("DD_TRACE_CLOUD_REQUEST_PAYLOAD_TAGGING", default=""),
+        "payload_tagging_response": os.getenv("DD_TRACE_CLOUD_RESPONSE_PAYLOAD_TAGGING", default=""),
+        "payload_tagging_max_depth": int(os.getenv("DD_TRACE_CLOUD_PAYLOAD_TAGGING_MAX_DEPTH", 10)),
+        "payload_tagging_max_tags": int(os.getenv("DD_TRACE_CLOUD_PAYLOAD_TAGGING_MAX_TAGS", 758)),
```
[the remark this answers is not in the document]
bouwkast: I did add a comment here, but they are magic numbers defined in the RFC as well.

ddtrace/_trace/utils_botocore/span_tags.py
```diff
@@ -54,6 +59,11 @@ def set_botocore_response_metadata_tags(
         return
     response_meta = result["ResponseMetadata"]
 
+    if config.botocore["payload_tagging_response"] is not None and config.botocore["payload_tagging_response"].replace(" ", "") != "":
+        if span.get_tag("aws_service") in config.botocore.get("payload_tagging_services"):
+            payload_tagger = AWSPayloadTagging() # TODO where do I put this?
```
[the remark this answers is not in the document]
bouwkast: Yep thanks exactly!

I've done that, but I have it always being instantiated - wasn't sure if managing it via the config was necessary.

ddtrace/_trace/utils_botocore/span_tags.py
```diff
@@ -54,6 +59,11 @@ def set_botocore_response_metadata_tags(
         return
     response_meta = result["ResponseMetadata"]
 
+    if config.botocore["payload_tagging_response"] is not None and config.botocore["payload_tagging_response"].replace(" ", "") != "":
```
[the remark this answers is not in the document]
bouwkast: If `config.botocore["payload_tagging_response"]` is `None` in that rewritten way that'll throw, wasn't really sure how to handle that, so I have left this as is

ddtrace/_trace/utils_botocore/aws_payload_tagging.py
```diff
@@ -0,0 +1,188 @@
...
```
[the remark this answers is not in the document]
bouwkast: I've added some comments around this

ddtrace/_trace/utils_botocore/aws_payload_tagging.py
```diff
@@ -0,0 +1,188 @@
...
+        """
+        # TODO should we cache these paths and generated lists?
+        request_redaction = self._get_redaction_paths_request(config.botocore.get("payload_tagging_request"))
+        response_redaction = self._get_redaction_paths_response(config.botocore.get("payload_tagging_response"))
+        for path in request_redaction + response_redaction:
+            expression = parse(path)
+            for match in expression.find(data):
+                match.context.value[match.path.fields[0]] = "redacted"
+
+    def _get_redaction_paths_default(self) -> list:
```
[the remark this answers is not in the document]
bouwkast: Thanks, made these constants, and then attempted to validate/cache them once

ddtrace/_trace/utils_botocore/aws_payload_tagging.py
bouwkast: Nope, after they are configured they are set, I attempted to cache them.

I don't think I can use `@cached` here as the `data` will always be different?

ddtrace/_trace/utils_botocore/aws_payload_tagging.py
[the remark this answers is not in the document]
tabgok: Much better, thanks!

ddtrace/_trace/utils_botocore/span_tags.py
```diff
@@ -14,6 +15,7 @@
 from ddtrace.internal.constants import COMPONENT
 from ddtrace.internal.utils.formats import deep_getattr
 
+_PAYLOAD_TAGGER=AWSPayloadTagging()
```
tabgok, line 18: nit: Missing a space.  The linter is also failing.

Hint: you can run `hatch -v run lint:fmt` to auto-format your files (check hatch.toml for other commands)

ddtrace/_trace/utils_botocore/aws_payload_tagging.py
```diff
@@ -110,31 +100,26 @@ def expand_payload_as_tags(self, span: Span, result: Dict[str, Any], key: str) -
...
 
     def _should_try_string(self, obj: Any) -> bool:
-        try:
-            if isinstance(obj, str) or isinstance(obj, unicode):
-                return True
-        except NameError:
-            if isinstance(obj, bytes):
-                return True
-
+        if isinstance(obj, str) or isinstance(obj, bytes):
```
tabgok, line 103: nit: you can make this a one-liner

tabgok (approved):

ddtrace/_trace/utils_botocore/aws_payload_tagging.py
```diff
@@ -0,0 +1,237 @@
...
+                self.current_tag_count += 1
+                span.set_tag(key, str(obj)[:5000])
+            return
+        if isinstance(obj, (int, float, Decimal)):
+            self.current_tag_count += 1
+            span.set_tag(key, str(obj))
+            return
+        if isinstance(obj, list):
+            for k, v in enumerate(obj):
+                self._tag_object(span, f"{key}.{k}", v, depth)
```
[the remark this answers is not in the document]
bouwkast: Could you explain a bit more? Or give an example showing what could happen here as I'm not really following thanks

ddtrace/_trace/utils_botocore/span_tags.py
```diff
@@ -31,6 +35,13 @@ def set_botocore_patched_api_call_span_tags(span: Span, instance, args, params,
         if params and not config.botocore["tag_no_params"]:
             aws._add_api_param_span_tags(span, endpoint_name, params)
 
+        if (
+            config.botocore["payload_tagging_request"]
+            and config.botocore["payload_tagging_request"].replace(" ", "")
```
erikayasuda, line 40: nit: Could we maybe do this clean up in `patch.py`?

tests/contrib/botocore/test.py
```diff
@@ -3765,3 +3774,259 @@ def test_schematized_unspecified_service_secretsmanager_v1(self):
...
+                "nine": {"DataType": "String", "StringValue": "nine"},
+                "ten": {"DataType": "String", "StringValue": "ten"},
+            }
+            self.sqs_client.send_message(
+                QueueUrl=self.sqs_test_queue["QueueUrl"], MessageBody="world", MessageAttributes=message_attributes
+            )
+            spans = self.get_spans()
+            assert spans
+            span = spans[0]
+            assert len(spans) == 1
```
erikayasuda, line 3801: We should probably switch the ordering here?

tests/contrib/botocore/test.py
```diff
@@ -3765,3 +3774,259 @@ def test_schematized_unspecified_service_secretsmanager_v1(self):
...
+                {"Data": json.dumps({"foo": "bar"}), "PartitionKey": partition_key},
+            ]
+
+            Pin.get_from(client).clone(tracer=self.tracer).onto(client)
+
+            with self.tracer.trace("kinesis.manual_span"):
+                client.create_stream(StreamName=stream_name, ShardCount=1)
+                client.put_records(StreamName=stream_name, Records=data)
+
+            spans = self.get_spans()
```
erikayasuda, line 4022: nit:
```suggestion
            spans = self.get_spans()
            assert len(spans) == 3, "Did not receive expected number of spans"
```

ddtrace/contrib/internal/botocore/patch.py
```diff
@@ -104,6 +104,17 @@ def _load_dynamodb_primary_key_names_for_tables() -> Dict[str, Set[str]]:
...
+        "payload_tagging_response": os.getenv("DD_TRACE_CLOUD_RESPONSE_PAYLOAD_TAGGING", default=None),
+        "payload_tagging_max_depth": int(
+            os.getenv("DD_TRACE_CLOUD_PAYLOAD_TAGGING_MAX_DEPTH", 10)
+        ),  # RFC defined 10 levels (1.2.3.4...10) as max tagging depth
+        "payload_tagging_max_tags": int(
+            os.getenv("DD_TRACE_CLOUD_PAYLOAD_TAGGING_MAX_TAGS", 758)
+        ),  # RFC defined default limit - spans are limited past 1000
+        "payload_tagging_services": set(
+            os.getenv("DD_TRACE_CLOUD_PAYLOAD_TAGGING_SERVICES", default={"s3", "sns", "sqs", "kinesis", "eventbridge"})
+        ),
```
erikayasuda, line 117: Could we also update our [configuration docs](https://github.com/DataDog/dd-trace-py/blob/main/docs/configuration.rst) with these new configs (name, brief description, default value(s), which version it was introduced in, etc.)

tests/contrib/botocore/test.py
```diff
@@ -3765,3 +3774,259 @@ def test_schematized_unspecified_service_secretsmanager_v1(self):
 
             assert span.service == DEFAULT_SPAN_SERVICE_NAME
             assert span.name == "aws.secretsmanager.request"
+
+    @pytest.mark.snapshot(ignores=snapshot_ignores)
+    @mock_sqs
+    def test_aws_payload_tagging_sqs(self):
+        with self.override_config("botocore", dict(payload_tagging_request="all", payload_tagging_response="all")):
```
erikayasuda, line 3781: Could we also add test cases for when `payload_tagging_request` and/or `payload_tagging_response` is not equal to `all`?

ddtrace/_trace/utils_botocore/aws_payload_tagging.py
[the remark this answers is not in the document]
bouwkast: Ah I think what @ygree is mentioning is that the sub-key `k` appended to the `key` needs to have any `.` escaped and not the concatenation of `{key}.{k}`

So if `key` was `aws.request.body` and `k` was `foo.bar` it should be `foo\.bar`.
I'll add `escaped_sub_key = k.replace(".", "\\.")`

ddtrace/_trace/utils_botocore/span_tags.py
bouwkast: I actually don't think that trimming out whitespace is needed anymore so I'll opt to just remove it.

ddtrace/_trace/utils_botocore/aws_payload_tagging.py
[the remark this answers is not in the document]
erikayasuda: Isn't the sub-key `k` always just an integer though? Since we're iterating through the enumerated values of a list?

ddtrace/_trace/utils_botocore/aws_payload_tagging.py
```diff
@@ -0,0 +1,237 @@
...
+                    + self._REDACTION_PATHS_DEFAULTS
+                    + request_redaction.split(",")
+                )
+            return self._REQUEST_REDACTION_PATHS_DEFAULTS + self._REDACTION_PATHS_DEFAULTS
+
+        return []
+
+    def _tag_object(self, span: Span, key: str, obj: Any, depth: int = 0) -> None:
+        """
+        Recursively expands the given AWS payload object and adds the values as flattened Span tags.
```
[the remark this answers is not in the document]
erikayasuda: No I think it should be fine! This was mainly just out of my own curiosity, but also my lack of familiarity with what a "normal" tag count for AWS payloads entails. If the recursion depth stays within 1000 (I think that's the sys default), we shouldn't have any issues. Plus, `payload_tagging_max_depth` and `payload_tagging_max_tags` should help bound this too.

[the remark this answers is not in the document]
bouwkast: Okay sounds good :)

Was also talking with @marcotc and we came to a similar conclusion - at most we're thinking it'd be an upper limit of ~10 recursive calls, so should be good. 

But if it proves to be an issue I can follow up with a PR to do convert it to some iterative stack-based method.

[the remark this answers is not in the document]
marcotc: I recommend adding a comment to this method, capturing that we do not expect the payload to be arbitrarily deeply nested. Just to make sure that readers in the future understand that we have considered the recursion concern.

tests/contrib/botocore/test.py
```diff
@@ -3765,3 +3774,414 @@ def test_schematized_unspecified_service_secretsmanager_v1(self):
...
+            # testing for span error
+            self.reset()
+            try:
+                s3.list_objects(bucket="mybucket")
+            except Exception:
+                spans = self.get_spans()
+                assert spans
+                span = spans[0]
+                assert span.error == 1
+                assert span.resource == "s3.listobjects"
```
erikayasuda, line 4083: I'm assuming we always expect an Exception to be raised here (because the bucket doesn't exist)? If so, could we do:
```suggestion
            with pytest.raises(Exception):
                s3.list_objects(bucket="mybucket")
            spans = self.get_spans()
            assert spans
            span = spans[0]
            assert span.error == 1
            assert span.resource == "s3.listobjects"
```
If there's a specific type of error we should look for, that could replace the general `Exception` in the context manager.

tests/contrib/botocore/test.py
```diff
@@ -3765,3 +3774,414 @@ def test_schematized_unspecified_service_secretsmanager_v1(self):
...
```
erikayasuda, line 4046: Same comment/suggestion here:
```suggestion
            with pytest.raises(Exception):
                s3.list_objects(bucket="mybucket")
            spans = self.get_spans()
            assert spans
            span = spans[0]
            assert span.error == 1
            assert span.resource == "s3.listobjects"
```

tests/contrib/botocore/test.py
```diff
@@ -3765,3 +3774,414 @@ def test_schematized_unspecified_service_secretsmanager_v1(self):
...
+    @mock_s3
+    def test_aws_payload_tagging_s3_valid_config(self):
+        with self.override_config(
+            "botocore", dict(payload_tagging_request="$..bucket", payload_tagging_response="$..HTTPHeaders")
+        ):
+            s3 = self.session.create_client("s3", region_name="us-west-2")
+            Pin(service=self.TEST_SERVICE, tracer=self.tracer).onto(s3)
+
+            s3.list_buckets()
+            s3.list_buckets()
```
erikayasuda, line 4058: Why do we call this twice here? Seems like it's intentional given that we're checking for 2 spans right after this?

tests/contrib/botocore/test.py
```diff
@@ -3765,3 +3774,414 @@ def test_schematized_unspecified_service_secretsmanager_v1(self):
...
+            assert span.get_tag("aws.region") == region
+            assert span.get_tag("region") == region
+            assert span.get_tag("aws.operation") == "PublishBatch"
+            assert span.get_tag("params.MessageBody") is None
+            assert_is_measured(span)
+            assert_span_http_status_code(span, 200)
+            assert span.service == "test-botocore-tracing.sns"
+            assert span.resource == "sns.publishbatch"
+            trace_json = span.get_tag("params.MessageAttributes._datadog.StringValue")
+            assert trace_json is None
```
erikayasuda, line 3883: Just realized that most of these checks are probably redundant if the snapshot JSON file is the correct expected span output.

ddtrace/_trace/utils_botocore/aws_payload_tagging.py
```diff
@@ -164,18 +165,17 @@ def _get_redaction_paths_request(self) -> list:
...
+            return (
+                self._REQUEST_REDACTION_PATHS_DEFAULTS + self._REDACTION_PATHS_DEFAULTS + request_redaction.split(",")
+            )
 
         return []
 
     def _tag_object(self, span: Span, key: str, obj: Any, depth: int = 0) -> None:
         """
         Recursively expands the given AWS payload object and adds the values as flattened Span tags.
+        It is not expected that AWS Payloads will be deeply nested so the number of recursive calls should be low.
```
erikayasuda, line 178: 🎉

erikayasuda (commented):
Just a few nits. Also can we create a release note for this? You can generate the template by running `reno new brief-name-for-your-change`, and then follow the instructions generated in the file (guidelines [here](https://ddtrace.readthedocs.io/en/stable/releasenotes.html))? Other than that LGTM and I can approve :)

tests/contrib/botocore/test.py
bouwkast: Yeah, I could remove the `assert`s from these if that makes more sense - I essentially copy/pasted pre-existing tests here

tests/contrib/botocore/test.py
bouwkast: Thanks, swapped to

```
with pytest.raises(Exception):
                s3.list_objects(bucket="mybucket")
```

and removed the asserts as it is all snapshotted

tests/contrib/botocore/test.py
bouwkast: So I copied pre-existing tests from this file and compared their snapshots before/after this feature was added

I don't fundamentally know why though it is called twice :(

tests/contrib/botocore/test.py
bouwkast: Done

erikayasuda (approved):
LGTM 🎉

ZStriker19 (approved):

docs/configuration.rst
```diff
@@ -659,7 +659,46 @@ The following environment variables for the tracer are supported:
          if no records are returned.
       version_added:
          v2.6.0:
+    
+    DD_TRACE_CLOUD_REQUEST_PAYLOAD_TAGGING:
+      type: String
+      default: None
+      description: |
+         Enables AWS request payload tagging when set to `"all"` or a valid comma-separated list of `JSONPath`s.
```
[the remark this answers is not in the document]
bouwkast: Thanks!

ddtrace/_trace/utils_botocore/aws_payload_tagging.py
```diff
@@ -0,0 +1,242 @@
...
+
+from ddtrace import Span
+from ddtrace import config
+from ddtrace.vendor.jsonpath_ng import parse
+
+
+_MAX_TAG_VALUE_LENGTH = 5000
+
+
+class AWSPayloadTagging:
```
P403n1x87, line 16: Many methods of this class seem to ignore the `self` argument. This is maybe an indication that things could be written differently?

ddtrace/_trace/utils_botocore/aws_payload_tagging.py
```diff
@@ -0,0 +1,242 @@
...
+            return False  # not enabled
+
+        if paths == "all":
+            return True  # enabled, use the defaults
+
+        # otherwise validate that we have valid JSONPaths
+        for path in paths.split(","):
+            if path:
+                try:
+                    parse(path)
```
P403n1x87, line 125: We're throwing away the parsing result here. Any chance we can save it to avoid having to re-parse again?

P403n1x87:
I'm slightly concerned by the new vendored libraries and by the potential performance of this solution. I'd love to see some benchmarking/profiling to get an idea of what the CPU overhead looks like, if possible 🙏

ddtrace/_trace/utils_botocore/aws_payload_tagging.py
bouwkast: @P403n1x87 Maybe? I'm honestly unsure and would need help to determine what a better solution would look like.

ddtrace/_trace/utils_botocore/aws_payload_tagging.py
```diff
@@ -0,0 +1,242 @@
...
+            self.validated = True
+
+        if not self.request_redaction_paths and not self.response_redaction_paths:
+            return
+
+        if not result:
+            return
+
+        # we will be redacting at least one of request/response
+        redacted_dict = copy.deepcopy(result)
```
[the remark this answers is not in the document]
bouwkast: Per the RFC on this, this is intended to be a debug feature that is not run in production for customers
So it was decided that performance wasn't a goal.

But investigating that seems reasonable, I can make a task for someone to investigate that.

ddtrace/_trace/utils_botocore/aws_payload_tagging.py
bouwkast: Yeah that is a good idea, I'll make a task for it to cache the parsed expressions (when they are all validated) and then use that in `_redact_json`

juanjux (approved):

PR #10650, opened 2024-09-12
feat(llmobs): format tool calls in OpenAI streamed responses

Yun-Kim:
This PR extends the OpenAI integration for LLM Observability to properly format tool calls returned in OpenAI streamed responses. Previously we cast the entire tool call as a raw string of form `[tool: {name}] {"arg1": "...", ...}` which was not very readable for users (example below)
Screenshot 2024-09-12 at 4 39 38 PM

Now, we format the tool call as a JSON structure so that it is much more readable on the Datadog UI (example below)
Screenshot 2024-09-12 at 4 39 53 PM

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/contrib/internal/openai/utils.py
```diff
@@ -249,26 +275,15 @@ def _construct_message_from_streamed_chunks(streamed_chunks: List[Any]) -> Dict[
...
-                formatted_content += "\n\n[tool: {}]\n\n".format(getattr(tool_call.function, "name", ""))
-                idx = tool_call.index
-            function_args = getattr(tool_call.function, "arguments", "")
-            message["content"] += "{}".format(function_args)
-            formatted_content += "{}".format(function_args)
-
+            _construct_tool_call_from_streamed_chunk(message["tool_calls"], tool_call_chunk=tool_call)
+    if message["tool_calls"]:
+        message["tool_calls"].sort(key=lambda x: x.get("index", 0))
     message["content"] = message["content"].strip()
```
lievan, line 286: Should we do a check here to pop the `tool_calls` key if it is an empty list? Just to avoid a situation where it is accidentally displayed in the FE
```
if message.get("tool_calls") is not None and len(message.get("tool_calls")) == 0:
      message.pop("tool_calls")
```

ddtrace/contrib/internal/openai/utils.py
```diff
@@ -300,11 +318,12 @@ def _set_token_metrics(span, integration, response, prompts, messages, kwargs):
         usage = response[0].get("usage", {})
         prompt_tokens = getattr(usage, "prompt_tokens", 0)
         completion_tokens = getattr(usage, "completion_tokens", 0)
+        total_tokens = getattr(usage, "total_tokens", 0)
```
lievan, line 321: Why did we choose to add this here? And I think there is a line below that resets
`total_tokens = prompt_tokens + completion_tokens`

Yun-Kim: We can actually get the total tokens from the returned TokenUsage chunk in the response instead of having to calculate. We only do the actual calculation if we have to do the prompt/completion token computation ourselves.

lievan (approved):

Issue #10643, opened 2024-09-12
psycopg autotrace regression in 2.12.x


maman:
### Summary of problem

Regression on 2.12.x

### Which version of dd-trace-py are you using?

2.12.2

### Which version of pip are you using?

23.0.1

### Which libraries and their versions are you using?

<details>
  <summary>`pip freeze`</summary>
<pre>
anyio==4.4.0
async-timeout==4.0.3
asyncpg==0.29.0
bytecode==0.15.1
certifi==2024.8.30
charset-normalizer==3.3.2
click==8.1.7
datrie==0.8.2
ddtrace==2.12.2
Deprecated==1.2.14
envier==0.5.2
exceptiongroup==1.2.2
greenlet==3.1.0
h11==0.14.0
idna==3.8
importlib_metadata==8.4.0
Jinja2==3.1.3
MarkupSafe==2.1.5
opentelemetry-api==1.27.0
osmium==3.7.0
protobuf==5.28.1
psutil==5.9.8
psycopg==3.1.18
psycopg-binary==3.1.18
psycopg2-binary==2.9.9
Pygments==2.18.0
PyICU==2.12
python-dotenv==1.0.1
PyYAML==6.0.1
requests==2.32.3
sniffio==1.3.1
SQLAlchemy==2.0.29
starlette==0.38.5
typing_extensions==4.12.2
ujson==5.9.0
urllib3==2.2.2
uvicorn==0.30.6
wrapt==1.16.0
xmltodict==0.13.0
zipp==3.20.1
</pre>
</details>

### How can we reproduce your problem?

Install [nominatim](https://nominatim.org/release-docs/latest/admin/Installation/) and [setup it with a database](https://nominatim.org/release-docs/latest/admin/Import/)

Then, run nominatim with ddtrace-run such as:
```
ddtrace-run uvicorn --factory nominatim.server.starlette.server:run_wsgi
```

### What is the result that you get?

```
failed to import ddtrace module 'ddtrace.contrib.psycopg' when patching on import
Traceback (most recent call last):
  File "/opt/venv/lib/python3.10/site-packages/ddtrace/_monkey.py", line 165, in on_import
    imported_module = importlib.import_module(path)
  File "/usr/local/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
cons-dora-api 
[...]

PR #10658, opened 2024-09-13
fix(postgres): remove require modules to resolve circular imports

mabdinur:
Resolves: https://github.com/DataDog/dd-trace-py/issues/10643

This regression was introduced in the following refactor: https://github.com/DataDog/dd-trace-py/commit/f42fb50fcef8ef4dd00c168551a5e3b624c25878#diff-fb98ba188c5f65d1365839bb51414f38505f968db842f5ac6d9ea14424c8fc0e

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/contrib/psycopg/__init__.py
```diff
@@ -60,23 +60,18 @@
     cursor = db.cursor()
     cursor.execute("select * from users where id = 1")
 """
-from ddtrace.internal.utils.importlib import require_modules
+# Required to allow users to import from `ddtrace.contrib.psycopg.patch` directly
```
erikayasuda, line 63: So the cycle here is that we try to import `psycopg` when bootstrapping and instrumenting all integrations, and then within this `psycopg/__init__.py` file we attempt to import `require_modules`? Wouldn't this mean that this problem isn't exclusive to the `psycopg` integration, but any of the integrations that use `require_modules`?

erikayasuda (approved):
Should we also change the title to `chore(psycopg)` instead of `chore(django)`?

ddtrace/contrib/psycopg/__init__.py
mabdinur: Yes it seems like it but I am not sure why we don't see this issue in other integrations. We have been using require modules for almost a year. 

cc: @P403n1x87

PR #10662, opened 2024-09-13
chore(llmobs): implement skeleton code for ragas faithfulness evaluator

lievan:
This PR 
1. implements `EvaluatorRunner`, a periodic service for LLM Obs that is responsible for running evaluations.
2. implements one dummy ragas faithfulness evaluator. The job of an evaluator is to define a `run_and_submit_evaluation` function which takes a span, generates an evaluation metric for that span, and submit the evaluation metric using a stored reference to LLM Obs instance. 

We add the `_DD_LLMOBS_EVALUATORS` env var to detect which evaluators should be enabled. Right now, the only supported evaluation is `ragas_faithfulness`. If no evaluators are detected, the evaluator runner is not started. 

Within the trace processor, spans events—after being enqueued to the span writer—are enqueued to the evaluator runner.

#### Intended Usage
```
_DD_LLMOBS_EVALUATORS="ragas_faithfulness,ragas_.." DD_LLMOBS_ENABLED=true python3 app.py
```

### No user facing changes for this pr
No changelog since this PR only implements the internal skeleton code necessary for RAGAS evaluation integration. The environment variable to enable the ragas evaluator service is hidden (`_DD_LLMOBS_RAGAS_FAITHFULNESS_ENABLED`) and will be made public when we implement an actual faithfulness function.

### (Full e2e poc, which contains some differences)
See https://github.com/DataDog/dd-trace-py/pull/10431 for an idea of what the full e2e implementation of the ragas integration looks like.

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/settings/config.py
```diff
@@ -569,6 +569,7 @@ def __init__(self):
         self._llmobs_sample_rate = float(os.getenv("DD_LLMOBS_SAMPLE_RATE", 1.0))
         self._llmobs_ml_app = os.getenv("DD_LLMOBS_ML_APP")
         self._llmobs_agentless_enabled = asbool(os.getenv("DD_LLMOBS_AGENTLESS_ENABLED", False))
+        self._llmobs_ragas_faithfulness_enabled = asbool(os.getenv("_DD_LLMOBS_RAGAS_FAITHFULNESS_ENABLED", False))
```
Yun-Kim, line 572: I don't think I love having a boolean flag for every single RAGAS metric. What if we were to make this a comma-separated string option that specifies which metrics to enable? i.e. `DD_LLMOBS_RAGAS_METRICS="faithfulness"`

Also, do you think it might be a better idea to keep this a simple os environment variable instead of declaring it as a config option until we're more sure about the overall API? Changing/removing config options are not simple (technically it is changing the public API)

lievan: I like both of those ideas, we can
1. remove the config option
2. only read from `DD_LLMOBS_RAGAS_METRICS="faithfulness,coherence,etc.."`

ddtrace/llmobs/_trace_processor.py
```diff
@@ -43,8 +43,9 @@ class LLMObsTraceProcessor(TraceProcessor):
     Processor that extracts LLM-type spans in a trace to submit as separate LLMObs span events to LLM Observability.
     """
 
-    def __init__(self, llmobs_span_writer):
+    def __init__(self, llmobs_span_writer, evaluators=None):
         self._span_writer = llmobs_span_writer
+        self._evaluators = evaluators
```
Yun-Kim, line 48: If we're going to have evaluators be a part of the trace processor, we're going to need to ensure we do the same thing on `_child_after_fork()` as we do for the span writer, i.e. something like
```
self._evaluator = self._evaluator.recreate()
self._trace_processor._evaluator = self._evaluator
```

ddtrace/llmobs/_evaluations/ragas/faithfulness/evaluator.py
```diff
@@ -0,0 +1,85 @@
...
+from ddtrace.internal.logger import get_logger
+from ddtrace.internal.periodic import PeriodicService
+
+from ...._writer import LLMObsEvaluationMetricEvent
+
+
+logger = get_logger(__name__)
+
+
+class RagasFaithfulnessEvaluator(PeriodicService):
```
Yun-Kim, line 19: Does it make more sense to make a base RagasEvaluator class and then extend that with custom evaluate() functions?

ddtrace/llmobs/_evaluations/ragas/faithfulness/evaluator.py
```diff
@@ -0,0 +1,85 @@
...
+import time
+from typing import Dict
+from typing import Optional
+
+from ddtrace import config
+from ddtrace.internal import forksafe
+from ddtrace.internal.logger import get_logger
+from ddtrace.internal.periodic import PeriodicService
+
+from ...._writer import LLMObsEvaluationMetricEvent
```
Yun-Kim, line 13: Standard practice is to avoid relative imports

ddtrace/llmobs/_evaluations/ragas/faithfulness/evaluator.py
```diff
@@ -0,0 +1,85 @@
...
+    def __init__(self, interval: float, _evaluation_metric_writer=None, _llmobs_instance=None):
+        super(RagasFaithfulnessEvaluator, self).__init__(interval=interval)
+        self.llmobs_instance = _llmobs_instance
+        self._lock = forksafe.RLock()
+        self._buffer = []  # type: list[Dict]
+        self._buffer_limit = 1000
+
+        self._evaluation_metric_writer = _evaluation_metric_writer
+
+        self.name = "ragas.faithfulness"
```
Yun-Kim, line 31: We can set this as an attribute on the class itself, not on `__init__`

ddtrace/llmobs/_evaluations/ragas/faithfulness/evaluator.py
```diff
@@ -0,0 +1,85 @@
...
+logger = get_logger(__name__)
+
+
+class RagasFaithfulnessEvaluator(PeriodicService):
+    """Base class for evaluating LLM Observability span events"""
+
+    def __init__(self, interval: float, _evaluation_metric_writer=None, _llmobs_instance=None):
+        super(RagasFaithfulnessEvaluator, self).__init__(interval=interval)
+        self.llmobs_instance = _llmobs_instance
+        self._lock = forksafe.RLock()
```
Yun-Kim, line 25: I'm slightly concerned that this will lead to tons of new background threads being spawned/active each with their own locks, which increases the likelihood of introducing some race condition.

Is it possible to have a single evaluator service handle all of the evaluations? Is having a separate thread per individual evaluation metric viable? I'd like to chat some about this before we merge this

ddtrace/llmobs/_evaluations/ragas/faithfulness/evaluator.py
```diff
@@ -0,0 +1,85 @@
...
+
+logger = get_logger(__name__)
+
+
+class RagasFaithfulnessEvaluator(PeriodicService):
+    """Base class for evaluating LLM Observability span events"""
+
+    def __init__(self, interval: float, _evaluation_metric_writer=None, _llmobs_instance=None):
+        super(RagasFaithfulnessEvaluator, self).__init__(interval=interval)
+        self.llmobs_instance = _llmobs_instance
```
Yun-Kim, line 24: Why do we need a reference to LLMObs?

ddtrace/llmobs/_trace_processor.py
```diff
@@ -56,11 +57,16 @@ def process_trace(self, trace: List[Span]) -> Optional[List[Span]]:
...
+        span_event = None
         try:
             span_event = self._llmobs_span_event(span)
             self._span_writer.enqueue(span_event)
         except (KeyError, TypeError):
             log.error("Error generating LLMObs span event for span %s, likely due to malformed span", span)
+        finally:
+            if span_event and self._evaluators:
+                for evaluator in self._evaluators:
+                    evaluator.enqueue(span_event)
```
Yun-Kim, line 69:
```suggestion
            if not span_event:
                return
            for evaluator in self._evaluators:
                evaluator.enqueue(span_event)
```
I assume `self._evaluators` is always empty if not configured, so we can directly go into the for loop instead of checking

ddtrace/llmobs/_evaluations/ragas/faithfulness/evaluator.py
lievan: I was planning on doing that after more ragas evaluators were introduced (e.g. how we refactored the span writer into base writer + span writer + eval metric writer) - i didn't see the value of doing that now with only faithfulness

ddtrace/llmobs/_evaluations/ragas/faithfulness/evaluator.py
lievan: In the full implementation, we want to use the llmobs service to trace the ragas evaluations. However, we cannot directly LLMObs since it will result in a circular import. Since it's un-used right now I can also just remove it for this PR

ddtrace/llmobs/_evaluations/ragas/faithfulness/evaluator.py
lievan: This is a good point, will schedule a time to talk about this

ddtrace/llmobs/_evaluators/ragas/faithfulness.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class RagasFaithfulnessEvaluator:
+    LABEL = "ragas_faithfulness"
+    METRIC_TYPE = "score"
+
+    def __init__(self, llmobs_service):
+        self.llmobs_service = llmobs_service
+
+    def evaluate(self, span):
```
Yun-Kim, line 12: Can we add typing here? Not sure what `span` is referring to exactly (dict? Span?)

ddtrace/llmobs/_evaluators/runner.py
```diff
@@ -0,0 +1,80 @@
...
+        self.llmobs_service = llmobs_service
+        self.executor = futures.ThreadPoolExecutor()
+        self.evaluators = []
+
+        evaluator_str = os.getenv("_DD_LLMOBS_EVALUATORS")
+        if evaluator_str is not None:
+            evaluators = evaluator_str.split(",")
+            for evaluator in evaluators:
+                if evaluator in SUPPORTED_EVALUATORS:
+                    self.evaluators.append(SUPPORTED_EVALUATORS[evaluator](llmobs_service=llmobs_service))
```
Yun-Kim, line 38: Let's avoid nested indents when possible.
```
if evaluator_str is None:
    return

```

ddtrace/llmobs/_evaluators/runner.py
```diff
@@ -0,0 +1,80 @@
...
+        try:
+            self.run(events)
+        except RuntimeError as e:
+            logger.debug("failed to run evaluation: %s", e)
+
+    def run(self, spans):
+        batches_of_results = []
+
+        for evaluator in self.evaluators:
+            batches_of_results.append(self.executor.map(lambda span: evaluator.evaluate(span), spans))
```
Yun-Kim, line 73: any specific reason why we need to use multithreading in a periodic service which is already a background worker thread? Does multithreading make a large difference in the evaluation call performance?

ddtrace/llmobs/_evaluators/runner.py
```diff
@@ -0,0 +1,80 @@
...
+            events = self._buffer
+            self._buffer = []
+
+        try:
+            self.run(events)
+        except RuntimeError as e:
+            logger.debug("failed to run evaluation: %s", e)
+
+    def run(self, spans):
+        batches_of_results = []
```
Yun-Kim, line 70: Personally this is kind of a confusing name, does `evaluation_results` make more sense here?

ddtrace/llmobs/_evaluators/runner.py
Yun-Kim: Is `ThreadPoolExecutor.map()` a blocking function? I.e. are we guaranteed to wait on this before we hit line 75?

ddtrace/llmobs/_evaluators/runner.py
```diff
@@ -0,0 +1,80 @@
...
+
+        for evaluator in self.evaluators:
+            batches_of_results.append(self.executor.map(lambda span: evaluator.evaluate(span), spans))
+
+        results = []
+        for batch in batches_of_results:
+            for result in batch:
+                results.append(result)
+
+        return results
```
Yun-Kim, line 80: We are returning individual evaluations per span no longer grouped by evaluation metric here, right?

ddtrace/llmobs/_llmobs.py
```diff
@@ -88,7 +89,13 @@ def __init__(self, tracer=None):
             interval=float(os.getenv("_DD_LLMOBS_WRITER_INTERVAL", 1.0)),
             timeout=float(os.getenv("_DD_LLMOBS_WRITER_TIMEOUT", 5.0)),
         )
-        self._trace_processor = LLMObsTraceProcessor(self._llmobs_span_writer)
+
+        self._evaluator_runner = EvaluatorRunner(
```
Yun-Kim, line 93: Are we going to make this feature enabled by default? If a user doesn't care about ragas evaluations, does it make sense for us to keep a background thread constantly running?

ddtrace/llmobs/_llmobs.py
```diff
@@ -114,13 +121,23 @@ def _start_service(self) -> None:
...
+            log.debug("Error starting evaluator runner")
+
     def _stop_service(self) -> None:
         try:
             self._llmobs_span_writer.stop()
             self._llmobs_eval_metric_writer.stop()
         except ServiceStatusError:
             log.debug("Error stopping LLMObs writers")
 
+        try:
```
Yun-Kim, line 136: Do we need to recreate the evaluator runner on fork?

ddtrace/llmobs/_llmobs.py
```diff
@@ -156,7 +173,6 @@ def enable(
 
         if os.getenv("DD_LLMOBS_ENABLED") and not asbool(os.getenv("DD_LLMOBS_ENABLED")):
             log.debug("LLMObs.enable() called when DD_LLMOBS_ENABLED is set to false or 0, not starting LLMObs service")
-            return
```
Yun-Kim, line 159: Why are we removing this?

ddtrace/llmobs/_trace_processor.py
```diff
@@ -113,6 +120,7 @@ def _llmobs_span_event(self, span: Span) -> Dict[str, Any]:
             "status": "error" if span.error else "ok",
             "meta": meta,
             "metrics": metrics,
+            "ml_app": ml_app,
```
Yun-Kim, line 123: Why are we adding this here?

tests/llmobs/test_llmobs_service.py
```diff
@@ -1504,3 +1544,10 @@ def process_trace(self, trace):
         exit_code = os.WEXITSTATUS(status)
         assert exit_code == 12
         llmobs_service.disable()
+
+
+def test_llm_span_ragas_evaluator(LLMObsWithRagas):
```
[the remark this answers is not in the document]
Yun-Kim: What is the purpose of this test?

ddtrace/llmobs/_evaluators/runner.py
lievan: Since the each evaluation for each span is independent of each other i thought it would make sense to implement multi-threading here. I imagine this would make a performance difference for evaluations that require a lot of IO operations e.g. api calls to model providers but I have not done any benchmarks here yet.

This also has another benefit where uncaught crashes in one evaluation thread over a span wouldn't impact the evaluation of other spans in the same dequeued batch.

ddtrace/llmobs/_evaluators/runner.py
lievan: dont need to collect results here actually, deleting

ddtrace/llmobs/_llmobs.py
lievan: Your right I don't think it makes sense. I was following the pattern with eval metrics where the eval metric writer is started even though a user may not submit eval metrics but that's a different case. I'll add a check inside the runner to see there are any evaluators turned on and if not, we don't start the runner

ddtrace/llmobs/_trace_processor.py
lievan: It's not needed; removing!

ddtrace/llmobs/_evaluators/ragas/faithfulness.py
```diff
@@ -0,0 +1,36 @@
+import math
+import time
+from typing import Optional
+
+
+class RagasFaithfulnessEvaluator:
```
lievan, line 6: Once we introduce more evaluators we'll also introduce a base evaluator class, but there's no need for that right now.

tests/llmobs/conftest.py
```diff
@@ -125,3 +132,10 @@ def AgentlessLLMObs(mock_llmobs_span_agentless_writer, mock_llmobs_eval_metric_w
         llmobs_service.enable(_tracer=dummy_tracer)
         yield llmobs_service
         llmobs_service.disable()
+
+
+@pytest.fixture
```
lievan, line 137: this fixture doesn't do anything right now since `evaluate` is a dummy function anyway that just returns 1.0, but it will set up the testing well for when `evaluate` actually implements a faithfulness evaluation

ddtrace/llmobs/_evaluators/runner.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def recreate(self) -> "EvaluatorRunner":
+        return self.__class__(
+            interval=self._interval,
+            llmobs_service=self.llmobs_service,
+            evaluators=self.evaluators,
+        )
+
+    def on_shutdown(self):
+        self.executor.shutdown()
```
sabrenner, line 72: should we call `periodic` here to run/flush any evaluators as a blocking operation before shutdown? or, because evaluators seem to be a bit costly (seeing as they run in a thread pool), is it left out intentionally?

ddtrace/llmobs/_evaluators/runner.py
```diff
@@ -0,0 +1,97 @@
...
+        if not self.evaluators:
+            logger.debug("no evaluators configured, not starting %r", self.__class__.__name__)
+            return
+        super(EvaluatorRunner, self).start()
+        logger.debug("started %r to %r", self.__class__.__name__)
+        atexit.register(self.on_shutdown)
+
+    def stop(self, *args, **kwargs):
+        if not self.evaluators and self.status == service.ServiceStatus.STOPPED:
+            logger.debug("no evaluators configured, not starting %r", self.__class__.__name__)
```
sabrenner, line 60:
```suggestion
            logger.debug("no evaluators configured, not stopping %r", self.__class__.__name__)
```
i feel like this log could be confusing tho, i think maybe we're good to just call `super(..).stop()` if the status isn't already `STOPPED`?

sabrenner (commented):
looks good! just left a couple questions, but overall really cool to see this start to take shape 😄

ddtrace/llmobs/_evaluators/runner.py
lievan: I think its ok leave out a call to `periodic` on shutdown because evaluators may take a while to run on the finished spans. I'll add a debug log here to note how many evaluations we are dropping though

ddtrace/llmobs/_evaluators/runner.py
lievan: Hm yea good point. The check for `self.evaluators` doesn't do much and I don't think the log is very helpful

ddtrace/llmobs/_trace_processor.py
```diff
@@ -57,11 +58,17 @@ def process_trace(self, trace: List[Span]) -> Optional[List[Span]]:
...
         try:
             span_event = self._llmobs_span_event(span)
             self._span_writer.enqueue(span_event)
         except (KeyError, TypeError):
             log.error("Error generating LLMObs span event for span %s, likely due to malformed span", span)
+        finally:
+            if not span_event:
+                return
+            if self._evaluator_runner:
+                self._evaluator_runner.enqueue(span_event)
```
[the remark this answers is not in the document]
lievan: Great point. I can see that there may be instances where the write fails for some reason but we still submit an eval metric. 

I don't think a write failure for a span should block prevent an evaluation metric from being submitted though, since users want to have as much data as possible for "average faithfulness scores over a week" which doesn't depend on the eval metric being tied to a span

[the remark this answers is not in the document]
sabrenner: makes sense to me, thanks for the clarification!

sabrenner (approved):

ddtrace/llmobs/_evaluators/runner.py
lievan: Resolving to unblock development on other RAGAS PRs, I'm planning to measure performance differences when the evaluator is no longer just a dummy function!

PR #10669, opened 2024-09-16
chore(ci_visibility): use two-pass procedure for code instrumentation in Python 3.10

vitor-de-araujo:
As my project for Innovation Week, I experimented with a more efficient way to do code coverage instrumentation in Python 3.10 (the idea is to apply this logic to the other versions in the future; I did it on Python 3.10 because it's what we use in dogweb).

The idea is to make all jumps fixed-size by emitting a fixed number of `EXTENDED_ARG` instructions before each jump, regardless of whether they are strictly necessary or not. This allows the jump targets to be rewritten in-place, without worrying that a new jump target might increase the number of `EXTENDED_ARG`s needed, which in turn would change the offsets of following instructions, which would change the jump targets, which might increase the number of `EXTENDED_ARGS`, and so on.

In this way, the instrumentation can be written as a two-pass procedure: first the instrumented bytecode is emitted with placeholders for the jump targets, and then the placeholders are patched with the final offsets of the jump targets. The algorithm is further explained in a long comment in the code.

As a result of these changes, the time taken to instrument code while running the Dogweb delancie tests drops from 7.8 seconds to around 2.5 seconds (about 3x as fast). 

## Checklist

## Reviewer Checklist
- Title is accurate


emmettbutler (commented):
How's the test coverage of this functionality? Are there tests in place that will fail when it breaks?

romainkomorndatadog:
>How's the test coverage of this functionality? Are there tests in place that will fail when it breaks?

The `dd_coverage` hatch env has some tests. They're more "integration" tests than unit tests because they run in a subprocess and execute their own script, but they're good "it either works or it doesn't" tests (and they run in each version of Python that we support). 

image

That said, there's always room for improvement.

vitor-de-araujo:
Yeah, I've tried commenting out the instrumentation insertion and the tests fail, and likewise for the `seen_lines` update. But I can add more tests if needed.

romainkomorndatadog (approved):

PR #10672, opened 2024-09-16
feat(langchain): add support for streamed calls

sabrenner:
## What does this PR do?
Adds support for `.(a)stream(...)` calls on LangChain LCEL chains, chat models, and completion (LLM) models. It accomplishes this by:

1. starting the span when the `stream` function is called
2. as the stream is consumed, add the chunk to a list of chunks
3. when the stream is consumed, finish the span, and concatenate the results into an output string

There is one caveat to this, where it's possible the last step in a chain's `stream` call is a `JSONOutputParser`. In this case, the stream is already concatenated for us, and we use that result instead. 

A few additional notes:
- The async versions, `astream`, aren't actually async functions, they just return async generators. This is reflected in shared patching functions and test snapshots.
- This whole process is accomplished through a `shared_stream`, which returns a compatible iterable for sync and async stream managers. It utilizes `on_span_started` and `on_span_finished` functions to coordinate tags to add in the different cases, such as chain, chat, or llm.
- Since the `stream` methods do not invoke the underlying `generate` methods we trace, there's not easy path for code re-use. Thus, I tried to mimic the tags we add for the relevant `chain.invoke`, `model.generate`, and `llm.invoke` patched functions.

**Note:** The version of `vcrpy` which we pinned for reduced flakiness did not like streamed calls when using the LangChain library. As such, I introduced a new fixture that returns a stub for the HTTP transport the `OpenAI` client uses. Then, the client with that transport specified can be used on the `langchain_openai` instance. This approach uses text files with just the response data, which is why the "cassettes" added for the tests written aren't the usual `yaml` format.

## Limitations
There are a couple of limitations that need to be noted:
1. Because of [upstream issues](https://github.com/langchain-ai/langchain/issues/26719) with langchain not handling indices returned from streamed chunks from certain partner libraries (primarily OpenAI), we cannot properly aggregate streamed multiple choices by index onto tags.
2. We are not providing direct tracing support for `astream_events` in this PR. Although it uses `astream` under the hood, and it is tested that `astream_events` does not conflict with the support added in this PR, official tracing support will be added in a follow-up PR
3. Additionally, submitting streamed responses to LLM Observability will be in a direct follow-up PR as well.

## Checklist

## Reviewer Checklist
- Title is accurate
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -


ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -975,6 +976,165 @@ def traced_similarity_search(langchain, pin, func, instance, args, kwargs):
...
+@with_traced_module
+def traced_chat_stream(langchain, pin, func, instance, args, kwargs):
+    integration: LangChainIntegration = langchain._datadog_integration
+    llm_provider = instance._llm_type
+
+    def _on_span_started(span: Span):
+        chat_messages = get_argument_value(args, kwargs, 0, "input")
+        if not isinstance(chat_messages, list):
+            chat_messages = [chat_messages]
+        for message_idx, message in enumerate(chat_messages):
```
sabrenner, line 1034: I added all of this logic because, unlike `model.generate`, which just takes in a list of `BaseMessage` types (ie `HumanMessage`, `SystemMessage`, etc.), `stream` can take in:

1. a single string
2. a single dict
3. a list of strings
4. a list of dicts
5. a list of `BaseMessage` types
6. a `PromptValue` type, which has a `messages` property of `BaseMessage` types

Do we care about all of these different types? Or, do we just want to listify and `str` each element (this logic would also carry over to LLMObs spans, future PR)? It would make the code a lot simpler, but maybe the view of each tag not as nice.

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -975,6 +976,165 @@ def traced_similarity_search(langchain, pin, func, instance, args, kwargs):
...
+                elif isinstance(message, langchain_core.messages.BaseMessage):
+                    span.set_tag_str(
+                        "langchain.request.messages.%d.content" % (message_idx), integration.trunc(str(message.content))
+                    )
+                    span.set_tag_str(
+                        "langchain.request.messages.%d.role" % (message_idx), str(message.__class__.__name__)
+                    )
+                else:
+                    span.set_tag_str(
+                        "langchain.request.messages.%d.content" % (message_idx), integration.trunc(message)
```
Yun-Kim, line 1064:
```suggestion
                        "langchain.request.messages.%d.content" % (message_idx), integration.trunc(str(message))
```
Just to be safe, otherwise set_tag_str() will always crash

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -975,6 +976,165 @@ def traced_similarity_search(langchain, pin, func, instance, args, kwargs):
...
+                    )
+
+        for param, val in getattr(instance, "_identifying_params", {}).items():
```
Yun-Kim, line 1067: I think we can try reducing the nested indents here

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -975,6 +976,165 @@ def traced_similarity_search(langchain, pin, func, instance, args, kwargs):
...
+            if isinstance(val, dict):
+                for k, v in val.items():
+                    span.set_tag_str("langchain.request.%s.parameters.%s.%s" % (llm_provider, param, k), str(v))
+            else:
+                span.set_tag_str("langchain.request.%s.parameters.%s" % (llm_provider, param), str(val))
+
+    def _on_span_finished(span: Span, streamed_chunks, error: Optional[bool] = None):
+        if not error and integration.is_pc_sampled_span(span):
```
Yun-Kim, line 1075: Same here

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -975,6 +976,165 @@ def traced_similarity_search(langchain, pin, func, instance, args, kwargs):
...
+            content = "".join([str(chunk.content) for chunk in streamed_chunks])
+            span.set_tag_str("langchain.response.content", integration.trunc(content))
+
+            usage = getattr(streamed_chunks[-1], "usage_metadata", None)
```
Yun-Kim, line 1079: Is usage metadata always guaranteed to be the last chunk?

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -975,6 +976,165 @@ def traced_similarity_search(langchain, pin, func, instance, args, kwargs):
...
+            if usage:
+                for k, v in usage.items():
+                    span.set_tag_str("langchain.response.usage_metadata.%s" % k, str(v))
```
Yun-Kim, line 1082: Let's avoid the triple nesting here

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -975,6 +976,165 @@ def traced_similarity_search(langchain, pin, func, instance, args, kwargs):
...
+    )
+
+
+@with_traced_module
+def traced_llm_stream(langchain, pin, func, instance, args, kwargs):
+    integration: LangChainIntegration = langchain._datadog_integration
+    llm_provider = instance._llm_type
+
+    def _on_span_start(span: Span):
+        inp = get_argument_value(args, kwargs, 0, "input")
```
Yun-Kim, line 1105: Does LLM.stream() also accept multiple types of inputs that we need to account for?

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -975,6 +976,165 @@ def traced_similarity_search(langchain, pin, func, instance, args, kwargs):
...
+            inp = [inp]
+        if integration.is_pc_sampled_span(span):
+            for idx, prompt in enumerate(inp):
+                span.set_tag_str("langchain.request.prompts.%d" % idx, integration.trunc(str(prompt)))
+        for param, val in getattr(instance, "_identifying_params", {}).items():
+            if isinstance(val, dict):
+                for k, v in val.items():
+                    span.set_tag_str("langchain.request.%s.parameters.%s.%s" % (llm_provider, param, k), str(v))
+            else:
+                span.set_tag_str("langchain.request.%s.parameters.%s" % (llm_provider, param), str(val))
```
Yun-Kim, line 1116: We can probably reduce indentation here too

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -975,6 +976,165 @@ def traced_similarity_search(langchain, pin, func, instance, args, kwargs):
...
+
+    def _on_span_finished(span: Span, streamed_chunks, error: Optional[bool] = None):
+        if not error and integration.is_pc_sampled_span(span):
```
Yun-Kim, line 1119: Same here

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -1203,6 +1363,9 @@ def patch():
         wrap("langchain", "embeddings.OpenAIEmbeddings.embed_documents", traced_embedding(langchain))
     else:
         from langchain.chains.base import Chain  # noqa:F401
+        from langchain_core import messages  # noqa: F401
+        from langchain_core import output_parsers  # noqa: F401
+        from langchain_core import prompt_values  # noqa: F401
```
Yun-Kim, line 1368: I assume these are due to import errors on langchain's end (not being able to directly access them without importing first?)

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -1225,6 +1388,23 @@ def patch():
         )
         wrap("langchain_core", "runnables.base.RunnableSequence.batch", traced_lcel_runnable_sequence(langchain))
         wrap("langchain_core", "runnables.base.RunnableSequence.abatch", traced_lcel_runnable_sequence_async(langchain))
+
+        # streaming
```
Yun-Kim, line 1392:
```suggestion
```

ddtrace/contrib/internal/langchain/utils.py
```diff
@@ -0,0 +1,79 @@
...
+import sys
+
+
+class BaseTracedLangChainStreamResponse:
+    def __init__(self, generator, integration, span, on_span_finish):
+        self._generator = generator
+        self._dd_integration = integration
+        self._dd_span = span
+        self._on_span_finish = on_span_finish
+        self._chunks = []
```
Yun-Kim, line 11: Looking at langchain's [source code](https://github.com/langchain-ai/langchain/blob/63c3cc1f1f2b78eb588e56fb62a7dacb43a2e27f/libs/core/langchain_core/language_models/chat_models.py#L408-L422), it looks like they actually add the chunks for us in the stream() operation. Would it be possible for us to access/reuse that instead of performing the appending/processing ourselves?

ddtrace/contrib/internal/langchain/utils.py
```diff
@@ -0,0 +1,79 @@
...
+        try:
+            async for chunk in self._generator.__aiter__():
+                self._chunks.append(chunk)
+                yield chunk
+        except Exception:
+            self._dd_span.set_exc_info(*sys.exc_info())
+            self._dd_integration.metric(self._dd_span, "incr", "request.error", 1)
+            raise
+        finally:
+            self._on_span_finish(self._dd_span, self._chunks, error=bool(self._dd_span.error))
```
Yun-Kim, line 40: This is part of a larger refactor I'm pushing for, but can we get rid of the `error` arg in these handlers? This is already captured as part of `span.error` so I don't see why we need this to be passed explicitly

tests/contrib/langchain/conftest.py
```diff
@@ -162,3 +162,81 @@ def langchain_pinecone(ddtrace_config_langchain, mock_logs, mock_metrics, langch
             yield langchain_pinecone
         except ImportError:
             yield
+
+
+@pytest.fixture
+def streamed_response_responder():
+    # TODO (sam.brenner): clean this up a bit, make it more generic
```
Yun-Kim, line 169: Let's remove TODOs in code and move them to internal tickets instead

tests/contrib/langchain/test_langchain_community.py
```diff
@@ -1276,6 +1278,149 @@ def test_faiss_vectorstore_retrieval(langchain_community, langchain_openai, requ
             retriever.invoke("What was the message of the last test query?")
 
 
+@pytest.mark.snapshot(ignores=IGNORE_FIELDS)
+def test_streamed_chain(langchain_core, langchain_openai, streamed_response_responder):
+    client = streamed_response_responder(
+        module="openai",
+        client_class_key="OpenAI",
+        http_client_key="http_client",
+        client_path=["chat", "completions"],
```
Yun-Kim, line 1287: What does the chat/completions client_path do?

Yun-Kim (commented):
Great work! I added a few comments about styling (avoiding nesting when possible for both readability and ironically line count), and some clarification questions but this is great. Awesome job on setting up a test fixture instead of being bogged down by vcrpy!

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -975,6 +976,165 @@ def traced_similarity_search(langchain, pin, func, instance, args, kwargs):
...
 
 
+# TODO refactor some of these on_span_started/on_span_finished functions
+# that are used in other patched methods in this file into the utils module
+@with_traced_module
+def traced_chain_stream(langchain, pin, func, instance, args, kwargs):
+    integration: LangChainIntegration = langchain._datadog_integration
+
+    def _on_span_started(span: Span):
+        inputs = get_argument_value(args, kwargs, 0, "input")
```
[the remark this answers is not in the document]
sabrenner: good question, let me explore that more. since we're not concerned with older versions of langchain/non-LCEL, i'll likely ignore those and verify that the name `input` is consistent and that it shows as the first in `args`

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -975,6 +976,165 @@ def traced_similarity_search(langchain, pin, func, instance, args, kwargs):
...
+                # it's possible that the chain has a json output parser
+                # this will have already concatenated the chunks into a json object
+
+                # it's also possible the json output parser isn't the last step,
+                # but one of the last steps, in which case we won't act on it here
+                # TODO (sam.brenner) make this more robust
+                content = json.dumps(streamed_chunks[-1])
+            else:
+                # best effort to join chunks together
+                content = "".join([str(chunk) for chunk in streamed_chunks])
```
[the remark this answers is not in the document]
sabrenner: it's a bit tricky with chains, because there could be an output parser that _does_ turn the chunks into strings, or they could be straight up `BaseMessage` types (ie `AIMessage`). in our current LCEL `chain` patching, we just [stringify](https://github.com/DataDog/dd-trace-py/blob/baac738726a6b615cb3f79693b0b1f117b688f9d/ddtrace/contrib/internal/langchain/patch.py#L844) to avoid going case-by-case.

ddtrace/contrib/internal/langchain/patch.py
sabrenner: oop good catch thanks!

ddtrace/contrib/internal/langchain/patch.py
sabrenner: I believe so (if it's included), but I'll try and look to see if there's evidence it could show up in a different chunk

ddtrace/contrib/internal/langchain/patch.py
sabrenner: ah yes good catch, it actually does look like it accepts a union of different values (although, different from the set accepted by chat models). will update to more accurately reflect that!

ddtrace/contrib/internal/langchain/patch.py
sabrenner: tbh i just added them by default. let me try and remove them and see what happens

ddtrace/contrib/internal/langchain/utils.py
sabrenner: good find! that one looks specific to chat models, i'll check if their runnable sequence (chain) implements something similar.

ddtrace/contrib/internal/langchain/utils.py
sabrenner: yep can do

tests/contrib/langchain/test_langchain_community.py
sabrenner: accesses the `openai.chat.completion` path specifically to set as the client to use for the `langchain_openai.ChatOpenAI` or `langchain_openai.OpenAI`. had to look for [this line](https://github.com/langchain-ai/langchain/blob/63c3cc1f1f2b78eb588e56fb62a7dacb43a2e27f/libs/partners/openai/langchain_openai/chat_models/base.py#L516-L517) to verify this is what kind of client they were setting. i can add a comment clarifying this

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -974,6 +976,166 @@ def traced_similarity_search(langchain, pin, func, instance, args, kwargs):
     return documents
 
 
+# TODO refactor some of these on_span_started/on_span_finished functions
```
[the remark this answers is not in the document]
sabrenner: I will leave a note in the description of your "refactor langchain ticket" for both of the TODOs to make sure they get addressed 😄

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -975,6 +976,165 @@ def traced_similarity_search(langchain, pin, func, instance, args, kwargs):
...
+                if isinstance(message, dict):
+                    span.set_tag_str(
+                        "langchain.request.messages.%d.content" % (message_idx),
+                        integration.trunc(str(message.get("content", ""))),
+                    )
+                    span.set_tag_str(
+                        "langchain.request.messages.%d.role" % (message_idx),
+                        str(message.get("role", "")),
+                    )
+                elif isinstance(message, langchain_core.prompt_values.PromptValue):
```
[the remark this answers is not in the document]
sabrenner: yep that makes sense to me! tried it out and it worked. nice find! (also added checks for `langchain_core`)

ddtrace/contrib/internal/langchain/patch.py
sabrenner: looks like it works without them 🫡

ddtrace/contrib/internal/langchain/patch.py
sabrenner: this might be openai-specific. since we already have a check in the non-streamed chat model patching to check `_is_openai_chat_instance`, i'll add that here as well, and in that case I believe it is guaranteed to be on the last chunk

ddtrace/contrib/internal/langchain/utils.py
sabrenner: it looks like it would only be accessible through the `run_manager` instance, which relies on the callback configs. i think we've thought about patching callback configs to run our own callbacks (which would good for handling these kinds of events for span management), but i think it's a bit out of scope for this PR. otherwise, it doesn't look like that accumulated `generations` goes anywhere we'd be able to use it 😢

ddtrace/contrib/internal/langchain/patch.py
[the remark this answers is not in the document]
sabrenner: yep, it looks like this named argument has been in place since the end of the `0.0.x` line, so we should be good here.

ddtrace/contrib/internal/langchain/utils.py
```diff
@@ -0,0 +1,125 @@
...
+):
+    options = {
+        "pin": pin,
+        "operation_id": f"{instance.__module__}.{instance.__class__.__name__}",
+        "interface_type": interface_type,
+    }
+
+    options.update(extra_options)
+
+    span = integration.trace(**options)
```
Yun-Kim, line 86: Can we add a follow up item to submit llmobs spans from streamed responses?

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -975,6 +977,138 @@ def traced_similarity_search(langchain, pin, func, instance, args, kwargs):
...
+            if not isinstance(val, dict):
+                span.set_tag_str("langchain.request.%s.parameters.%s" % (llm_provider, param), str(val))
+                continue
+            for k, v in val.items():
+                span.set_tag_str("langchain.request.%s.parameters.%s.%s" % (llm_provider, param, k), str(v))
+
+    def _on_span_finished(span: Span, streamed_chunks):
+        if span.error or not integration.is_pc_sampled_span(span):
+            return
+        content = "".join([str(chunk.content) for chunk in streamed_chunks])
```
Yun-Kim, line 1050: Just to be safe, is `chunk` guaranteed to have the `content` attribute?

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -975,6 +977,138 @@ def traced_similarity_search(langchain, pin, func, instance, args, kwargs):
...
+        span.set_tag_str("langchain.response.content", integration.trunc(content))
+
+        usage = streamed_chunks and getattr(streamed_chunks[-1], "usage_metadata", None)
+        if not usage:
```
Yun-Kim, line 1054: should also return if `usage` is not a dict type (would break in 1056 otherwise)

ddtrace/contrib/internal/langchain/patch.py
Yun-Kim: meh, as long as we're not making up roles when they don't exist in `llm.generate()` cases, I'd rather push for a unified way of tagging llms vs chat models, i.e. treat them all as messages. I think this is fine

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -975,6 +977,138 @@ def traced_similarity_search(langchain, pin, func, instance, args, kwargs):
...
+            if not isinstance(val, dict):
+                span.set_tag_str("langchain.request.%s.parameters.%s" % (llm_provider, param), str(val))
+                continue
+            for k, v in val.items():
+                span.set_tag_str("langchain.request.%s.parameters.%s.%s" % (llm_provider, param, k), str(v))
+
+    def _on_span_finished(span: Span, streamed_chunks):
+        if span.error or not integration.is_pc_sampled_span(span):
+            return
+        content = "".join([str(chunk) for chunk in streamed_chunks])
```
Yun-Kim, line 1094: Just for clarity: llm streams chunks are in the form of raw text?

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -1203,7 +1337,8 @@ def patch():
         wrap("langchain", "embeddings.OpenAIEmbeddings.embed_documents", traced_embedding(langchain))
     else:
         from langchain.chains.base import Chain  # noqa:F401
-        from langchain_core.tools import BaseTool  # noqa:F401
```
Yun-Kim, line 1206: Removing this didn't break patching?

tests/contrib/langchain/test_langchain_community.py
```diff
@@ -1276,6 +1278,169 @@ def test_faiss_vectorstore_retrieval(langchain_community, langchain_openai, requ
...
+        langchain.schema.HumanMessage(content=inp),
+    ]
+
+    for _ in chain.stream(input=messages):
+        pass
+
+
+# until we fully support `astream_events`, we do not need a snapshot here
+# this is just a regression test to make sure we don't throw
+async def test_astreamed_events_does_not_throw(langchain_openai, langchain_core, async_streamed_response_responder):
```
Yun-Kim, line 1426: So to confirm, we do not trace astream_events() (there is also no `stream_events()`, just `astream_events()`?), but it does in fact interfere with `stream()` so possibly can be affected by our patching of `stream()`?

Could we add a note in the PR description and a JIRA item to follow up on this?

ddtrace/contrib/internal/langchain/utils.py
```diff
@@ -0,0 +1,125 @@
+import inspect
+import sys
+
+
+class BaseTracedLangChainStreamResponse:
```
Yun-Kim, line 5: We don't need to base this off `wrapt.ObjectProxy`?

ddtrace/contrib/internal/langchain/utils.py
```diff
@@ -0,0 +1,125 @@
...
+    for input_idx, inp in enumerate(inputs):
+        if isinstance(inp, dict):
+            span.set_tag_str(
+                "langchain.request.messages.%d.content" % (input_idx),
+                integration.trunc(str(inp.get("content", ""))),
+            )
+            span.set_tag_str(
+                "langchain.request.messages.%d.role" % (input_idx),
+                str(inp.get("role", "")),
+            )
```
Yun-Kim, line 118: Let's only attach role if role is not None, in case of llm.stream() being called rather than chat_model.

releasenotes/notes/langchain-lcel-stream-calls-bff85c974a72cceb.yaml
```diff
@@ -0,0 +1,5 @@
+---
+features:
+  - |
+    langchain: Adds support for tracing ``stream`` calls on LCEL chains, chat completion models, or completion models.
+    Due to an upstream issue with the ``langchain`` library itself, streamed responses are not tagged correctly when the `n` parameter on the model is set to greater than 1.
```
Yun-Kim, line 5:
```suggestion
    Note that due to an upstream issue with the ``langchain`` library itself, streamed responses will not be tagged correctly based on the choice index when the underlying model is configured to return ``n>1`` choices.
```
Can we also link the associated github issue in this release note entry?

Yun-Kim (commented):
Looks good, mostly clarification questions! For easier tracking, can we add a limitations section to this PR description (i.e. n>1 response tagging, astream_events())? Thanks!

ddtrace/contrib/internal/langchain/utils.py
sabrenner: yep i have a ticket on our board!

ddtrace/contrib/internal/langchain/patch.py
sabrenner: yeah, `model.stream` for any generic `model` from a langchain partner library uses the base definition of `stream` in `langchain_core`, which returns an iterator of `BaseMessage` chunks, which have `content` on them. But I agree it's best to be safe, I'll use `getattr` with a default of the chunk itself and stringify

ddtrace/contrib/internal/langchain/patch.py
sabrenner: ah yes good catch!

ddtrace/contrib/internal/langchain/patch.py
sabrenner: similar to the chat model case, yeah their `llm.stream` definition specifies it's an [iterator of strings](https://github.com/langchain-ai/langchain/blob/f9ef688b3a0e3f576a76435e242090a76666974b/libs/core/langchain_core/language_models/llms.py#L515-L522).

ddtrace/contrib/internal/langchain/patch.py
sabrenner: ah i didn't mean to remove this line, thanks. but, it looks like all tests still passed. i'll put it back since it's out of scope for this PR, but maybe we can try removing it in a different one if we don't need it here

tests/contrib/langchain/test_langchain_community.py
sabrenner: yep that's my understanding of how it works, and yes no `stream_events`. I'll make a note + follow up ticket

ddtrace/contrib/internal/langchain/utils.py
sabrenner: i had tried that initially, as i took inspiration from the antropic integration. but, when doing that, i end up getting
```
AttributeError: 'generator' object has no attribute '_dd_integration'
```
so i think this is why i went with this approach

ddtrace/contrib/internal/langchain/patch.py
sabrenner: added a safety check in https://github.com/DataDog/dd-trace-py/pull/10672/commits/47a3f15fdc43b840d42c6749b0a9ed2681c68a19

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -975,6 +977,138 @@ def traced_similarity_search(langchain, pin, func, instance, args, kwargs):
...
+                span.set_tag_str("langchain.request.%s.parameters.%s" % (llm_provider, param), str(val))
+                continue
+            for k, v in val.items():
+                span.set_tag_str("langchain.request.%s.parameters.%s.%s" % (llm_provider, param, k), str(v))
+
+    def _on_span_finished(span: Span, streamed_chunks):
+        if span.error or not integration.is_pc_sampled_span(span):
+            return
+        content = "".join([str(getattr(chunk, "content", chunk)) for chunk in streamed_chunks])
+        span.set_tag_str("langchain.response.content", integration.trunc(content))
```
Yun-Kim, line 1051: Do we not want to tag role from the output messages as well?

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -975,6 +977,138 @@ def traced_similarity_search(langchain, pin, func, instance, args, kwargs):
...
+            # it's possible that the chain has a json output parser
+            # this will have already concatenated the chunks into a json object
+
+            # it's also possible the json output parser isn't the last step,
+            # but one of the last steps, in which case we won't act on it here
+            content = json.dumps(streamed_chunks[-1])
+        else:
+            # best effort to join chunks together
+            content = "".join([str(chunk) for chunk in streamed_chunks])
+        span.set_tag_str("langchain.response.content", integration.trunc(content))
```
Yun-Kim, line 1014: I think we can rename the tag name to `langchain.response.outputs` since this is the output of the chain, and not necessarily a llm/chat model, right?

Yun-Kim (approved):
Two last questions, I'll approve to unblock you but please address them (I'll defer to your judgement). Thanks!

ddtrace/contrib/internal/langchain/patch.py
sabrenner: ah yes good call

ddtrace/contrib/internal/langchain/patch.py
sabrenner: yeah that makes sense to me!

ddtrace/contrib/internal/langchain/patch.py
sabrenner: tagged it as `message_type` to be consistent with our normal `generate` patching.

PR #10674, opened 2024-09-16
ci: add test visibility instrumentation to GitLab with pytest plugin

romainkomorndatadog:
This makes changes to enable Test Visibility in GitLab using `--ddtrace`.

It relies on GitLab having an agent at `${HOST_IP}:8196` and introduces the undocumented and "private" `_CI_DD_AGENT_URL` environment variable to make the "outer" `CIVisibility` tracer submit data via the agent.

In the past, the tracer ran in agentless mode which had more or less the same effect, but did not affect snapshot data.

`riot` environments for some tests do not pass with telemetry enabled (the fact that it was previously not enabled in CircleCI was actually an accident), so the `DD_INSTRUMENTATION_TELEMETRY_ENABLED` is set to a falsy value in this change (but is enabled for some `tracer` tests where it is needed).

Using the GitLab host allows us to run without an API key in our environment which is arguably worth it.

## Checklist

## Reviewer Checklist
- Title is accurate


.circleci/config.templ.yml
```diff
@@ -186,6 +186,7 @@ commands:
                 environment:
                   DD_TRACE_AGENT_URL: << parameters.trace_agent_url >>
                   RIOT_RUN_RECOMPILE_REQS: "<< pipeline.parameters.riot_run_latest >>"
+                  DD_CIVISIBILITY_AGENTLESS_ENABLED: "true"
```
gnufede, line 189: If I'm not mistaken, yaml is string by default, no?
```suggestion
                  DD_CIVISIBILITY_AGENTLESS_ENABLED: true
```

.circleci/config.templ.yml
```diff
@@ -222,6 +223,7 @@ commands:
             - run:
                 environment:
                   RIOT_RUN_RECOMPILE_REQS: "<< pipeline.parameters.riot_run_latest >>"
+                  DD_CIVISIBILITY_AGENTLESS_ENABLED: "true"
```
gnufede, line 226: Ditto

gnufede (approved):
LGTM

.circleci/config.templ.yml
romainkomorndatadog: It is but I like my mistakes sprinkled throughout my PRs.

.gitlab/tests.yml
```diff
@@ -23,6 +27,7 @@ variables:
 .test_base_hatch_snapshot:
   extends: .test_base_hatch
   services:
+    - !reference [.test_base_hatch, services]
```
romainkomorndatadog, line 30:
```suggestion
```

.gitlab-ci.yml
```diff
@@ -18,8 +18,10 @@ variables:
   # DEV: we have a larger pool of amd64 runners, prefer that over arm64
   tags: [ "arch:amd64" ]
   timeout: 20m
+  services: []
```
romainkomorndatadog, line 21:
```suggestion
```

.gitlab/tests.yml
```diff
@@ -7,7 +7,11 @@ variables:
   # Hatch doesn't use pre-built wheels or venvs so we can start them right away
   needs: []
   parallel: 4
+  services:
+    - !reference [.testrunner, services]
```
romainkomorndatadog, line 11:
```suggestion
```

brettlangdon (approved):
changes to files owned by core team lgtm

I did not thoroughly review the changes to Test Visibility product, I am unsure the impact these changes will have on customers.

emmettbutler (approved):

riotfile.py
```diff
@@ -101,9 +101,6 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
         "_DD_CIVISIBILITY_USE_CI_CONTEXT_PROVIDER": "1",
         "DD_TESTING_RAISE": "1",
         "DD_REMOTE_CONFIGURATION_ENABLED": "false",
-        "DD_CIVISIBILITY_AGENTLESS_ENABLED": "1",
-        "DD_CIVISIBILITY_CODE_COVERAGE_ENABLED": "1",
-        "DD_CIVISIBILITY_ITR_ENABLED": "1",
```
erikayasuda, line 106: It looks like `DD_CIVISIBILITY_ITR_ENABLED` defaults to True so that was removed. 

But for `DD_CIVISIBILITY_AGENTLESS_ENABLED` and `DD_CIVISIBILITY_CODE_COVERAGE_ENABLED`, I just see these set to `1` in the `hatch.toml` file but not anywhere else. Are they unused?

.gitlab/tests.yml
```diff
@@ -8,6 +8,8 @@ variables:
   needs: []
   parallel: 4
   script:
+    - export PYTEST_ADDOPTS="${PYTEST_ADDOPTS} --ddtrace"
+    - export _DD_CIVISIBILITY_USE_CI_CONTEXT_PROVIDER=true
```
erikayasuda, line 12: Do we set this for hatch but not for riot because it's already set in [riotfile.py](https://github.com/DataDog/dd-trace-py/blob/main/riotfile.py#L101)?

ZStriker19 (approved):

.gitlab-ci.yml
```diff
@@ -20,6 +20,7 @@ variables:
   timeout: 20m
   before_script:
     - pyenv global 3.12 3.7 3.8 3.9 3.10 3.11 3.13-dev
+    - export _CI_DD_AGENT_URL=http://${HOST_IP}:8126/
```
erikayasuda, line 23: To clarify: the reason we need to add this new `_CI_DD_AGENT_URL` var is because with GitLab, we (currently) don't have a way to spin up a separate service/test agent, but we did with CircleCI?

erikayasuda (approved):
Sorry just a bunch of clarifying questions, otherwise LGTM if we can get some of the (I think) flaky tests to pass

.gitlab/tests.yml
romainkomorndatadog: Correct.

riotfile.py
romainkomorndatadog: `DD_CIVISIBILITY_AGENTLESS_ENABLED` is being removed because we don't want it in the `riotfile.py` (and we want it to be falsy anyway).

`DD_CIVISIBILITY_CODE_COVERAGE_ENABLED` is actually no longer a valid env var at all.

.github/workflows/testrunner.yml
```diff
@@ -14,6 +14,9 @@ on:
 jobs:
   build-and-publish:
     uses: ./.github/workflows/build-and-publish-image.yml
+    permissions:
+      contents: read
+      packages: write
```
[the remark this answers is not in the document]
romainkomorndatadog: It was not.

docker/Dockerfile
```diff
@@ -54,14 +54,16 @@ RUN \
       unixodbc-dev \
       wget \
       zlib1g-dev \
-  # Install Mariadb
+      awscli \
+  # For running datadog-ci CLI with npx
+  && apt-get install -y --no-install-recommends nodejs npm \
+  && npm install -g @datadog/datadog-ci \
+  # Install Mariadb last because mariadb_repo_setup currently breaks apt
```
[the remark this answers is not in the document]
romainkomorndatadog: I found https://www.reddit.com/r/mariadb/comments/1e5f38t/mariadb_repo_setup_is_broken_leads_to_unavailable/ and decided it's a known issue and, at best, we don't want it to break other potential `apt` usage, so I stuck it at the end.

.gitlab-ci.yml
romainkomorndatadog: In CircleCI, we were using agentless mode, so we didn't care what the agent URL was, but I decided it would be better to use an agent in GitLab since a- it's there, and b- it means we don't have to get an API key.

The `CIVisibility` tracer needs "its own" agent URL because otherwise it'll try to send traces to the test agent, in the snapshot test jobs (and the existing use of `_DD_CIVISIBILITY_USE_CI_CONTEXT_PROVIDER` already forces it to create a different tracer).

ddtrace/internal/ci_visibility/writer.py
```diff
@@ -106,11 +106,11 @@ def __init__(
             timeout = config._agent_timeout_seconds
         intake_cov_url = None
         if use_evp:
-            intake_url = agent.get_trace_url()
-            intake_cov_url = agent.get_trace_url()
+            intake_url = intake_url if intake_url else agent.get_trace_url()
+            intake_cov_url = intake_url
```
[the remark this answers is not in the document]
romainkomorndatadog: I like it this way because it makes it clear that we want to use the same URL as the `intake_url` when we have an agent, or when we're hardcoding the agentless URL.

PR #10676, opened 2024-09-16
fix(celery): close `celery.apply` spans even without after_task_publish, when using apply_async

wantsui:
The instrumentation for the Celery integration relies on various [Celery signals ](https://docs.celeryq.dev/en/stable/userguide/signals.html) in order to start and end the span when calling on `apply_async`.

The integration can fail if the expected signals don't trigger, which can lead to broken context propagation (and unexpected traces).

**Example:**
- dd-trace-py expects the signal `before_task_publish` to start the span then `after_task_publish` to close the span. If the `after_task_publish` signal never gets called (which can happen if a Celery exception occurs while processing the app), then the span won't finish.
- The same thing above can also happen to `task_prerun` and `task_postrun`.

**Solution**

This PR patches `apply_async` so that there is a check to see if there is a span lingering around and closes it when `apply_task` is called.

If an internal exception happens, the error will be marked on the `celery.apply` span.

To track this, I added new logs in debug mode:
> The after_task_publish signal was not called, so manually closing span

and 
> The task_postrun signal was not called, so manually closing span

There's a related PR https://github.com/DataDog/dd-trace-py/pull/10848 that works to improve how we extract information based on the protocols, that also affects when spans get closed or not.

Special Thanks:
- Thanks to @tabgok for going through this with me in great detail!
- @timmc-edx for helping us track it down!

[APMS-13158]

## Checklist

## Reviewer Checklist
- Title is accurate

APMS-13158

[APMS-13158]: https://datadoghq.atlassian.net/browse/APMS-13158?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ddtrace/contrib/internal/celery/app.py
```diff
@@ -96,3 +105,27 @@ def _traced_beat_inner(func, instance, args, kwargs):
             return func(*args, **kwargs)
 
     return _traced_beat_inner
+
+
+def _traced_apply_async_function(integration_config, fn_name, resource_fn=None):
```
tabgok, line 110: It likely won't be apparent to future readers why we're grabbing a span out of seemingly thin air and then closing it here. 

This is a perfect time for a comment explaining WHY we're doing this.  I see a good blurb in the test, but it should be added here as well.  I'd include something along the lines of "apply_async synchronously calls the before/after signals, so we make sure we close at the end"

releasenotes/notes/fix-celery-apply-async-span-close-b7a8db188459f5b5.yaml
```diff
@@ -0,0 +1,4 @@
+---
+fixes:
+  - |
+    tracing(celery): Fixes an issue where `celery.apply` spans didn't close if the after_task_publish or task_postrun signals didn't get called.
```
tabgok, line 4: nit: would add a little bit about why they weren't called (exception generated in handling)

tabgok (approved):

emmettbutler (approved):

ddtrace/contrib/internal/celery/utils.py
```diff
@@ -128,9 +128,12 @@ def retrieve_task_id(context):
...
     body = context.get("body")
-    if headers:
+    # Check if the id is in the headers, then check the body for it.
+    # If we don't check the id first, we could wrongly assume no task_id
+    # when the task_id is in the body.
+    if headers and 'id' in headers:
         # Protocol Version 2 (default from Celery 4.0)
         return headers.get("id")
-    else:
+    elif body and 'id' in body:
```
wantsui, line 137: @tabgok I believe this is the place we failed to extract ids correctly with the old logic, so it's another reason why the span doesn't get closed.

wantsui: Update: I'll resolve this issue in a separate PR: https://github.com/DataDog/dd-trace-py/pull/10848

emmettbutler (approved):

ddtrace/contrib/internal/celery/utils.py
```diff
@@ -131,6 +131,6 @@ def retrieve_task_id(context):
     if headers:
         # Protocol Version 2 (default from Celery 4.0)
         return headers.get("id")
-    else:
+    elif body:
```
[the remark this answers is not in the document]
wantsui: That makes sense, https://github.com/DataDog/dd-trace-py/pull/10848 ended up getting this logic (and another fix) so I removed the change in utils from this PR.

releasenotes/notes/fix-celery-apply-async-span-close-b7a8db188459f5b5.yaml
wantsui: fixed in the latest version of the release notes

tabgok (approved):

iherasymenko:
Linking to https://github.com/DataDog/dd-trace-py/issues/11624 as this PR introduced a regression.

PR #10682, opened 2024-09-17
chore: prefer `docker compose` instead of older `docker-compose`

erikayasuda:
With newer versions of Docker (20.10+), `docker compose` is used instead of `docker-compose`. Using `docker-compose` with the newer version of Docker will cause an error, and the `scripts/ddtest` will not run. This PR updates the script to check if `docker compose` is available, and if so will use that. Otherwise, it will default to using the original `docker-compose` command.

Docs have also been updated to use `docker compose` instead.

## Checklist

## Reviewer Checklist
- Title is accurate

### DIFF
--- a/docs/contributing-integrations.rst
+++ b/docs/contributing-integrations.rst
@@ -114,7 +114,7 @@ Many of the tests are based on "snapshots": saved copies of actual traces sent t
 
 1. Update the library and test code to generate new traces.
 2. Delete the snapshot file corresponding to your test at ``tests/snapshots/<snapshot_file>`` (if applicable).
-3. Use `docker-compose up -d testagent` to start the APM test agent, and then re-run the test. Use `--pass-env` as described
+3. Use `docker compose up -d testagent` to start the APM test agent, and then re-run the test. Use `--pass-env` as described
    `here <https://github.com/datadog/dd-apm-test-agent?tab=readme-ov-file#running-the-tests>`_ to ensure that your test run can talk to the test agent.
 
 Once the run finishes, the snapshot file will have been regenerated.
@@ -173,7 +173,7 @@ are not yet any expected spans stored for it, so we need to create some.
 
 .. code-block:: bash
 
-   $ docker-compose up -d testagent <container>
+   $ docker compose up -d testagent <container>
    $ scripts/ddtest
    > DD_AGENT_PORT=9126 riot -v run --pass-env <test_suite_name>
 
--- a/scripts/ddtest
+++ b/scripts/ddtest
@@ -10,7 +10,7 @@ then
 fi
 
 # retry docker pull if fails
-for i in {1..3}; do docker-compose pull -q testrunner && break || sleep 3; done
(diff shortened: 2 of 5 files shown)


scripts/ddtest
```diff
@@ -10,7 +10,7 @@ then
 fi
 
 # retry docker pull if fails
-for i in {1..3}; do docker-compose pull -q testrunner && break || sleep 3; done
```
sanchda, line 13: Do we know for a fact that it's safe to sunset `docker-compose` across the invocations of this script?  I ask out of ignorance, since I don't know the coverage for when `compose` was integrated into the `docker` CLI.

If we _don't_ know, may I propose something like the following?
```
get_docker_compose_command() {
    if docker compose version &> /dev/null; then
        echo "docker compose"
    elif docker-compose version &> /dev/null; then
        echo "docker-compose"
    else
        echo "Error: Neither 'docker compose' nor 'docker-compose' is available." >&2
        exit 1
    fi
}
```

I _think_ both `docker-compose` and `docker compose` provide a `version` parameter, which should make it straightforward to implement a fallback.  Something like

```
DOCKER_COMPOSE=$(get_docker_compose_command)
```

So then lines like this would become
```
for i in {1..3}; do $DOCKER_COMPOSE pull -q testrunner && break || sleep 3; done
```

(I wrote all of this out in a GH comment, so it is totally off-the-cuff and untested; just trying to help think of consequences)

mabdinur (approved):
LGTM can we also remove references to docker-compose in our contributing and SSI docs. We should also require docker compose v2 but I guess that's implied.

scripts/ddtest
erikayasuda: Good call out, thanks for the suggestion. I like this safer approach

scripts/ddtest
```diff
@@ -9,8 +9,14 @@ then
     CMD=bash
 fi
 
+# Check if 'docker compose' is available (Docker version 20.10+)
+if docker compose version &>/dev/null; then
```
sanchda, line 13: Yup, seems like a very practical failover.

docs/contributing-integrations.rst
```diff
@@ -114,7 +114,7 @@ Many of the tests are based on "snapshots": saved copies of actual traces sent t
 
 1. Update the library and test code to generate new traces.
 2. Delete the snapshot file corresponding to your test at ``tests/snapshots/<snapshot_file>`` (if applicable).
-3. Use `docker-compose up -d testagent` to start the APM test agent, and then re-run the test. Use `--pass-env` as described
+3. Use `docker compose up -d testagent` to start the APM test agent, and then re-run the test. Use `--pass-env` as described
```
sanchda, line 117: Makes total sense.  I agree that it's most sensible for documentation and internal guidance to use the newer, shinier `docker compose` instead of the old and crusty `docker-compose`.

sanchda (approved):
Looks like a pragmatic solution, LGTM--thanks!

emmettbutler (approved):

PR #10685, opened 2024-09-17
feat(tracer): [SVLS-5262] private Span Pointer Description

apiarian-datadog:
In order to help decouple code that calculates span pointers from the spans and pointers themselves, we make a private Span Pointer Description class.

## Checklist
- [ ] PR author has checked that all the criteria below are met
- The PR description includes an overview of the change
- The PR description articulates the motivation for the change
- The change includes tests OR the PR description describes a testing strategy
- The PR description notes risks associated with the change, if any
- Newly-added code is easy to change
- The change follows the [library release note guidelines](https://ddtrace.readthedocs.io/en/stable/releasenotes.html)
- The change includes or references documentation updates if necessary
- Backport labels are set (if [applicable](https://ddtrace.readthedocs.io/en/latest/contributing.html#backporting))

## Reviewer Checklist
- [ ] Reviewer has checked that all the criteria below are met 
- Title is accurate
- All changes are related to the pull request's stated goal
- Avoids breaking [API](https://ddtrace.readthedocs.io/en/stable/versioning.html#interfaces) changes
- Testing strategy adequately addresses listed risks
- Newly-added code is easy to change
- Release note makes sense to a user of the library
- If necessary, author has acknowledged and discussed the performance implications of this PR as reported in the benchmarks PR comment
- Backport labels are set in a manner that is consistent with the [release branch maintenance policy](https://ddtrace.readthedocs.io/en/latest/contributing.html#backporting)

PR #10687, opened 2024-09-17
chore: [SVLS-5262] Span Pointer standard hashing

apiarian-datadog:
Adding a standard hashing function for Span Pointers. This continues to be a private API for now.

## Checklist

## Reviewer Checklist
- Title is accurate

### DIFF
--- a/ddtrace/_trace/_span_pointer.py
+++ b/ddtrace/_trace/_span_pointer.py
@@ -1,5 +1,6 @@
 from enum import Enum
 from hashlib import sha256
+import random
 from typing import Any
 from typing import Dict
 from typing import NamedTuple
@@ -7,6 +8,10 @@ from typing import Optional
 
 from ddtrace._trace._span_link import SpanLink
 from ddtrace._trace._span_link import SpanLinkKind
+from ddtrace.internal.logger import get_logger
+
+
+log = get_logger(__name__)
 
 
 _SPAN_POINTER_SPAN_LINK_TRACE_ID = 0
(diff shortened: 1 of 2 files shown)


apiarian-datadog:
built on top of https://github.com/DataDog/dd-trace-py/pull/10685

ddtrace/_trace/_span_pointer.py
```diff
@@ -54,3 +55,31 @@ def __init__(
     def __post_init__(self):
         # Do not want to do the trace_id and span_id checks that SpanLink does.
         pass
+
+
+class _StandardHashingRules:
```
ZStriker19, line 60: This is a bit strange, but we can probably remove that part of the comment. 

One thing I don't understand is why we can't just have a method that uses a global value set to `self.desired_bits // self.bits_per_hex_digit` for the calculation. 

Or do we need this as a class because `"allows us to write tests for the chosen base_hashing_function instead of having to add runtime sanity checks."` ? 
However isn't the `base_hashing_function` just hardcoded here to use sha256 so it won't change and we could just always use that value?

I'm not actually against the way this is written, I'd just like to understand better.

ZStriker19 (approved):

ddtrace/_trace/_span_pointer.py
apiarian-datadog: an original version of this code said something like "assert num_bits_in_hash >= num_required_bits" but our tooling complained and it seemed silly to do the `desired_bits // bits_per_hex_digit` calculation and comparison every time the hashing function is called.

i wanted a way to encapsulate this all these "business rules" without having to use global variables or setup that hard-codes sha256 in both the production code and the tests.

now that i say that, if i loosen the constraint of "don't say `sha256` twice" i can write the `_standard_hashing_function` as a proper function and still write tests that validate that its output is the right size and such. will write that up tomorrow and see how that looks.

apiarian-datadog: done! i've rearranged the code to have a "simpler" though more obscure hashing function and more descriptive tests

ddtrace/_trace/_span_pointer.py
```diff
@@ -53,3 +54,11 @@ def __init__(
     def __post_init__(self):
         # Do not want to do the trace_id and span_id checks that SpanLink does.
         pass
+
+
+def _standard_hashing_function(*elements: bytes) -> str:
+    if not elements:
+        raise ValueError("elements must not be empty")
```
ZStriker19, line 61: We have to make sure that we do everything possible to avoid crashing user applications. With that in mind, can we switch to an error log and either return an empty string or some other value that would keep us from crashing down the line.

apiarian-datadog: can we keep the exception and agree to always catch it further up the stack?

apiarian-datadog: alright, per our discussion on slack, i'm going to make sure we don't throw exceptions.

ZStriker19 (approved):

PR #10694, opened 2024-09-17
fix(llmobs): safely handle non-json serializable arguments

Yun-Kim:
Safely handle non-JSON serializable tag arguments in LLMObs.annotate() and the OpenAI/LangChain/Bedrock/Anthropic integrations.
- LLMObs.annotate(): we previously just discarded the entire argument to annotate if the argument was non-JSON serializable. Now, we safely convert non-JSON serializable fields/objects to a default placeholder text, meaning users can still send *some* data even if some of it may be invalid.
- Same idea with each integration, we ensure we safely handle non-JSON serializable args and default to placeholder texts if necessary.
- We've moved all json.dumps() call into a private helper `safe_json()` which does the above for us.

Note: This PR removes some tests in `test_llmobs_service.py` regarding truly unserializable objects as this is highly unlikely, someone would have to go out of their way to make a truly unserializable object (i.e. override `__repr__()` with a non-json serializable value). We still catch these so any resulting crashes should not be from our code.

## Checklist

## Reviewer Checklist
- Title is accurate


sabrenner (commented):
lgtm! just left a small question, probably more for my own edification.

ddtrace/llmobs/_integrations/anthropic.py
```diff
@@ -71,18 +71,22 @@ def llmobs_set_tags(
 
         span.set_tag_str(SPAN_KIND, "llm")
         span.set_tag_str(MODEL_NAME, span.get_tag("anthropic.request.model") or "")
-        span.set_tag_str(INPUT_MESSAGES, json.dumps(input_messages))
-        span.set_tag_str(METADATA, json.dumps(parameters))
+        span.set_tag_str(
+            INPUT_MESSAGES, json.dumps(input_messages, skipkeys=True, default=_unserializable_default_repr)
+        )
+        span.set_tag_str(METADATA, json.dumps(parameters, skipkeys=True, default=_unserializable_default_repr))
```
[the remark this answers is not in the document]
Yun-Kim: Good catch, thanks! I think `skipkeys=True` is an extra defensive step, this should only be for cases where users add actual troublesome keys. This should be rare, but I'll add it for the other integrations to be safe

ddtrace/llmobs/_integrations/langchain.py
```diff
@@ -304,7 +315,7 @@ def _llmobs_set_meta_tags_from_similarity_search(
                 if isinstance(formatted_inputs, str):
                     span.set_tag_str(INPUT_VALUE, formatted_inputs)
                 else:
-                    span.set_tag_str(INPUT_VALUE, json.dumps(formatted_inputs))
+                    span.set_tag_str(INPUT_VALUE, json.dumps(formatted_inputs, default=_unserializable_default_repr))
```
vitor-de-araujo, line 318: Is the lack of `skip_keys=True` on this one intentional?

I was thinking that maybe we should have a utility function to do the `json.dumps(..., skipkeys=True, default=_unserializable_default_repr)` and use it everywhere instead of repeating this at multiple places.

vitor-de-araujo (approved):

ddtrace/llmobs/_integrations/langchain.py
Yun-Kim: Thanks for the catch + suggestion. I've made a utility helper following your suggestion, thanks!

ddtrace/llmobs/_llmobs.py
```diff
@@ -587,99 +586,78 @@ def _tag_embedding_io(cls, span, input_documents=None, output_text=None):
...
+        if not metadata:
+            return
         if not isinstance(metadata, dict):
             log.warning("metadata must be a dictionary of string key-value pairs.")
             return
-        try:
-            span.set_tag_str(METADATA, json.dumps(metadata, default=_unserializable_default_repr))
-        except TypeError:
-            log.warning("Failed to parse span metadata. Metadata key-value pairs must be JSON serializable.")
+        span.set_tag_str(METADATA, safe_json(metadata))
```
lievan, line 650: Are we missing a try/catch here?

lievan (approved):

PR #10713, opened 2024-09-19
fix(llmobs): avoid raising errors during llmobs integration span processing

Yun-Kim:
This PR does 2 things:

### User facing changes
- captures any integration-specific `_llmobs_set_tags()` method errors and logs the error instead of potentially crashing the user application.

### Non-user facing changes
Refactors the `BaseLLMIntegration` class and child classes to follow a cleaner and shared `llmobs_set_tags()` method, which internally try/catches an abstract method `_llmobs_set_tags()` instead (which is implemented by each integration). We also no longer need to check `integration.is_pc_sampled_llmobs(span)` since we don't currently do any sampling yet and we can handle it in the `llmobs_set_tags()` method if needed.
tldr: `_llmobs_set_tags()` is now an abstract method that needs to be implemented by all LLM integrations, and its function signature now takes in the following arguments/keyword arguments (same as `llmobs_set_tags()`):
- span: span to annotate
- args: list of args passed to the traced method
- kwargs: dict of keyword args passed to the traced method. If any integration requires additional data not contained by either args/kwargs (such as the model instance in Gemini or tool_input dictionary in langchain), we can pass it into the method using the kwarg dict.
- response: returned response from llm provider (streamed or non-streamed)
- operation: string denoting which LLM operation it is (eg. "completion", "chat", "embedding", "chain", "retrieval")

I did some refactoring to each integration to follow this new signature, which included merging logic for how we handle streamed responses, and additional required args (i.e. model instance, tool inputs).

Previously each integration did its own thing for `llmobs_set_tags()` with arbitrary args/kwargs, and it was difficult to maintain. Now that we have a strict function signature, future integrations should be simpler to create, and existing integrations should be easier to maintain.

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/contrib/internal/openai/_endpoint_hooks.py
```diff
@@ -189,7 +189,7 @@ class _CompletionHook(_BaseCompletionHook):
 
     def _record_request(self, pin, integration, span, args, kwargs):
         super()._record_request(pin, integration, span, args, kwargs)
-        if integration.is_pc_sampled_llmobs(span):
+        if integration.llmobs_enabled:
```
lievan, line 192: Do we need to set span kind here if there is [this](https://github.com/DataDog/dd-trace-py/blob/99f03d8016562b4a3ac98a92494bbfe862e0768d/ddtrace/llmobs/_integrations/openai.py#L150C1-L151C1) line that sets the span kind tag based on the operation name?

ddtrace/llmobs/_integrations/base.py
```diff
@@ -81,18 +85,15 @@ def llmobs_enabled(self) -> bool:
         return LLMObs.enabled
 
     def is_pc_sampled_span(self, span: Span) -> bool:
-        if span.context.sampling_priority is not None:
-            if span.context.sampling_priority <= 0:
-                return False
+        if span.context.sampling_priority and span.context.sampling_priority <= 0:
```
lievan, line 88: Assuming we don't want this to always evaluate to false if sampling priority is zero
```suggestion
        if span.context.sampling_priority is not None and span.context.sampling_priority <= 0:
```

ddtrace/llmobs/_integrations/base.py
```diff
@@ -81,18 +85,15 @@ def llmobs_enabled(self) -> bool:
...
         return self._span_pc_sampler.sample(span)
 
     def is_pc_sampled_log(self, span: Span) -> bool:
-        if span.context.sampling_priority is not None:
-            if span.context.sampling_priority <= 0:
-                return False
-
         if not self.logs_enabled:
             return False
+        if span.context.sampling_priority and span.context.sampling_priority <= 0:
```
lievan, line 95:
```suggestion
        if span.context.sampling_priority is not None and span.context.sampling_priority <= 0:
```

ddtrace/contrib/internal/openai/_endpoint_hooks.py
Yun-Kim: Great catch, thanks!

ddtrace/llmobs/_integrations/base.py
Yun-Kim: We only care about positive non-zero sampling priority values here, even if `span.context.sampling_priority == 0`, the first conditional overlaps with the other conditional `span.context.sampling_priority <= 0` so this should be fine

mabdinur (approved):
The fix looks good to me.  Moving forward should we try to remove the llm logic from `ddtrace/_trace/trace_handlers.py`?

ddtrace/llmobs/_integrations/base.py
Yun-Kim: Nevermind, missed what you were saying. Thanks!

lievan (approved):

Yun-Kim:
This PR is dependent on too many code changes from features (e.g. tool calling, streamed response support and refactors) that backporting it to 2.11, 2.12 and 2.13 requires a large amount of non-trivial code changes to allow it to be backported to those branches. We will leave this as a standalone fix that will be released in 2.14.

PR #10716, opened 2024-09-19
chore(tracer): core additions

christophe-papazian:
Small core additions:

- `discard_item` to discard items
- remove `traverse` optional argument of `get_item` and split the method and function in two version, `get_item` and `get_local_item` for efficiency reasons.
- use modern type annotations
- set `_CURRENT_CONTEXT` directly (without using an ephemeral None value that breaks type checking)
- add asyncio test for all of that with concurrent threads

## Checklist

## Reviewer Checklist
- Title is accurate


tests/internal/test_context_events_api.py
```diff
@@ -437,3 +437,44 @@ def make_another_context():
...
+
+    witness = object()
+
+    async def get_set_isolated(value: str):
+        with core.context_with_data(value):
+            core.set_item("key", value)
+            with core.context_with_data(value):
+                v = f"in {value}"
+                core.set_item("key", v)
+                await asyncio.sleep(random.random())
```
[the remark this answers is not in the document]
christophe-papazian: We don't need cryptographic level random for tests on concurrency.

ddtrace/internal/core/__init__.py
```diff
@@ -124,10 +126,6 @@ def _on_jsonify_context_started_flask(ctx):
 from .event_hub import reset as reset_listeners  # noqa:F401
 
 
-if TYPE_CHECKING:
```
emmettbutler, line 127: This is important for avoiding circular imports. I don't see any non-annotation uses of `Span`, so I expect that keeping this conditional will not break any other parts of your change.

ddtrace/internal/core/__init__.py
```diff
@@ -168,15 +165,17 @@ class ExecutionContext:
     __slots__ = ["identifier", "_data", "_parents", "_span", "_token"]
 
     def __init__(self, identifier, parent=None, span=None, **kwargs):
-        _deprecate_span_kwarg(span)
+        if span is not None:
```
emmettbutler, line 168: This is redundant with `_deprecate_span_kwarg`'s internal logic

ddtrace/internal/core/__init__.py
```diff
@@ -225,42 +224,49 @@ def context_with_data(cls, identifier, parent=None, span=None, **kwargs):
...
-        current = self
         while current is not None:
             if data_key in current._data:
                 return current._data.get(data_key)
-            if not traverse:
-                break
             current = current.parent
         return default
 
+    def get_local_item(self, data_key: str, default: Optional[Any] = None) -> Any:
```
emmettbutler, line 235: This is a clearer API. Good thinking.

ddtrace/internal/core/__init__.py
```diff
@@ -276,56 +282,68 @@ def __getattr__(name):
...
-_CONTEXT_CLASS = ExecutionContext
-
-
 def context_with_data(identifier, parent=None, **kwargs):
     return _CONTEXT_CLASS.context_with_data(identifier, parent=(parent or _CURRENT_CONTEXT.get()), **kwargs)
 
 
-def get_item(data_key, span=None):
-    # type: (str, Optional[Span]) -> Optional[Any]
+def get_root() -> ExecutionContext:
```
emmettbutler, line 299: This duplicates the functionality of this module's `__getattr__` function.

emmettbutler (approved):

ddtrace/internal/core/__init__.py
christophe-papazian: Annotations are supposed to be always defined. You can't not import them. But typings module should never create circular dependencies.

ddtrace/internal/core/__init__.py
christophe-papazian: Except at start where global context is not defined yet. If you remove that line, it will break at start.

ddtrace/internal/core/__init__.py
christophe-papazian: Oh yes, you can access the root through core.root indeed.
Removed.

ddtrace/internal/core/__init__.py
christophe-papazian: Found a fix by reversing logic order in the deprecate function

gnufede (approved):

quinna-h (approved):

PR #10720, opened 2024-09-19
chore(crashtracking): add python version to crashtracker tags

sanchda:
During escalations it's currently difficult to find the exact Python version relevant for a defect.  This is a simple fix for that.

I'm recommending a backport because
1. The fix is simple
2. 2.11 and onward introduce crashtracker, and we're still dealing with crashes from those versions

## Checklist

## Reviewer Checklist
- Title is accurate


brettlangdon (approved):
lgtm

this will also include any beta suffixes as well, e.g. `3.13.0b3+` is that ok with the spec for runtime version? (I'd like to keep it if we can)

sanchda:
> this will also include any beta suffixes as well, e.g. `3.13.0b3+` is that ok with the spec for runtime version? (I'd like to keep it if we can)

My take is that higher-level processing pipelines can normalize an overly-verbose version name in a runtime-aware way if they need to, but that our responsibility at the data collection level is to provide a fully qualified version string, identifying as precisely as possible the origin of the runtime build.

In future PRs, I'll be doing things like hashing libpython.so and other runtime assets during a crash, so that during investigation we can immediately qualify whether we're looking at the same thing a given user was utilizing at time of crash.

In my ideal dream state, our backend systems will actually have a roster of CPython/libc assets available for crash- or query-time analysis.  But that's a dream.  Or a nightmare I guess depending :shrug:

PR #10752, opened 2024-09-22
chore(ci_visibility): create client for test visibility backend API

romainkomorndatadog:
This creates a new client for the Test Visibility API (used currently for settings and ITR data, but soon to be used for new test and auto flake retries information).

This is mostly a refactor (eg: the majority of the original logic has been cut and pasted into a new class), with some notable changes:
- skippable tests/suites are now part of an `ITRData` dataclass, under the CIVisibility instance's `_itr_data` attribute
- checking for test or suite skippability status is done by getting a `InternalTestId` or a `TestSuiteId`, and looking for it in `_itr_data.skippable_items`
- fetching tests now depends on a new `_itr_skipping_ignore_parameters` configuration item (this is being introduced for backwards compatibility because parameters were previously not being checked)
- the `TestVisibilityAPISettings` replaces the `_CIVisibilitySettings` `namedtuple`
- oodles of tests are updated to match the new behaviors 

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/internal/ci_visibility/_api_client.py
```diff
@@ -0,0 +1,521 @@
...
+    module_id = TestModuleId(skippable_suite["attributes"]["configurations"]["test.bundle"])
+    return TestSuiteId(module_id, skippable_suite["attributes"]["suite"])
+
+
+def _parse_covered_files(covered_files_data: t.Dict[str, str]) -> t.Optional[t.Dict[str, CoverageLines]]:
+    covered_files = {}
+    parse_errors = 0
+    for covered_file, covered_lines_bytes in covered_files_data.items():
+        try:
+            covered_lines = CoverageLines.from_bytearray(bytearray(b64decode(covered_lines_bytes)))
```
vitor-de-araujo, line 152: I was going to say that we already have a [`from_b64_string` method](https://github.com/DataDog/dd-trace-py/blob/ceea872654b11cdd2549d92fbca48561076ecfe4/ddtrace/internal/test_visibility/coverage_lines.py#L89C9-L89C24) in `CoverageLines`, but it turns out it's missing the `b64decode()` there (which I think is wrong).

romainkomorndatadog: 😓 that's because I started adding that method and the decided it wasn't worth adding as a new to the API.

ddtrace/internal/ci_visibility/_api_client.py
```diff
@@ -0,0 +1,521 @@
...
+def _parse_skippable_suites(
+    skippable_suites_data: t.List[_SkippableResponseDataItem],
+) -> t.Set[_SKIPPABLE_ITEM_ID_TYPE]:
+    suites_to_skip: t.Set[_SKIPPABLE_ITEM_ID_TYPE] = set()
+    count_unparsed_suites = 0
+    for skippable_suite in skippable_suites_data:
+        try:
+            suite_id = _get_suite_id_from_skippable_suite(skippable_suite)
+            suites_to_skip.add(suite_id)
+        except Exception:  # noqa: E722
```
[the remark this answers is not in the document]
romainkomorndatadog: Good point. I think I just lack enough of an opinion to pick one. :D Maybe just `except:` ?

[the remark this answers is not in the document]
vitor-de-araujo: Well, the code quality thingy seems to complain about bare `except:` but not about `except Exception:` (but then again, it may be updated to complain about `except Exception:` too at some point, so this is not a strong argument). One potential point in favor of `except Exception` is that it makes you stop and thing "do I really need to catch every exception?" (but then again, in the tracer we often do want that).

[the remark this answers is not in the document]
romainkomorndatadog: > One potential point in favor of except Exception is that it makes you stop and thing "do I really need to catch every exception?" 

Doesn't `except:` make you stop and think "do I really need ignore everything?"?

ddtrace/internal/ci_visibility/_api_client.py
```diff
@@ -0,0 +1,524 @@
...
+        headers = _BASE_HEADERS.copy()
+        headers.update(self._get_headers())
+        return headers
+
+    def _do_request(self, method: str, endpoint: str, payload: str, timeout: t.Optional[float] = None) -> Response:
+        timeout = timeout if timeout is not None else self._timeout
+        headers = self._get_final_headers()
+        url = combine_url_path(self._base_url, endpoint)
+
+        conn: t.Optional[t.Union[HTTPSConnection | HTTPConnection | UDSHTTPConnection]] = None
```
vitor-de-araujo, line 253: Isn't `|` the same as `t.Union`?

romainkomorndatadog: Yeah, this was copy/paste'd from the original, but what I really want is `ddtrace.internal.utils.http.ConnectionType` anyway. Will update.

ddtrace/internal/ci_visibility/_api_client.py
```diff
@@ -0,0 +1,524 @@
...
+            response = Response.from_http_response(resp)
+            return response
+        finally:
+            if conn is not None:
+                conn.close()
+
+    def fetch_settings(self) -> TestVisibilityAPISettings:
+        """Fetches settings from the test visibility API endpoint
+
+        This raises encountered exceptions because fetch_settings may be used multiple times during a session.
```
vitor-de-araujo, line 280: I did not get why raising exceptions is useful because the function may be used multiple times :thinking:

ddtrace/internal/ci_visibility/utils.py
```diff
@@ -139,3 +139,8 @@ def take_over_logger_stream_handler(remove_ddtrace_stream_handlers=True):
     root_logger.setLevel(min(root_logger.level, ci_visibility_handler.level))
 
     log.debug("logger setup complete")
+
+
+def combine_url_path(*args: str):
+    """Combine URL path segments."""
+    return "/".join(str(segment).strip("/") for segment in args)
```
vitor-de-araujo, line 146: Nice solution :ok_hand: 

The only potential problem I see is that it will remove a trailing `/` from an URL, but in our case I think it's not a problem.

vitor-de-araujo (approved):

ddtrace/internal/ci_visibility/utils.py
romainkomorndatadog: It's not a problem for our current usage but... I'm going to add a comment that it should be used with care. 😬

ddtrace/internal/ci_visibility/_api_client.py
```diff
@@ -0,0 +1,524 @@
...
+                error_type = ERROR_TYPES.BAD_JSON
+                return None
+
+            correlation_id = meta.get("correlation_id")
+            if correlation_id is None:
+                log.debug("Skippable tests response missing correlation_id")
+            else:
+                log.debug("Skippable tests response correlation_id: %s", correlation_id)
+
+            covered_files_data = skippable_response["meta"].get("coverage")
```
quinna-h, line 445: nit: should we use `meta.get("coverage")` to be consistent with previous

quinna-h, line 445:
```suggestion
            covered_files_data = meta.get("coverage")
```

quinna-h (approved):

tests/ci_visibility/api/test_internal_test_visibility_api.py
```diff
@@ -44,76 +44,81 @@ def test_api_is_item_itr_skippable_false_when_skipping_disabled_test_level(self)
             itr_enabled=True,
             skipping_enabled=False,
             suite_skipping_mode=False,
-            skippable_items={
-                "skippable_module/suite.py": ["skippable_test"],
-            },
+            skippable_items=_make_fqdn_test_ids([("skippable_module", "suite.py", "skippable_test")]),
         ):
-            CIVisibility.enable()
+            try:
```
erikayasuda, line 49: nit: Should we use a try/finally in a test? I guess there's no equivalent of `tearDown()` in pytest, so this is probably the cleanest way 🤔

erikayasuda (approved):

tests/ci_visibility/api/test_internal_test_visibility_api.py
romainkomorndatadog: We could use a a fixture but I've got vague plans to rewrite the entire test setup/teardown fixture to be more pleasant than this so I'm going to leave it as is for now.

PR #10767, opened 2024-09-23
feat(llmobs): llmobs-specific context manager

Yun-Kim:
[MLOB-1342]

## Summary
Public facing changes:
- Any LLMObs method (`export_span(), annotate()`) that allow an optional span argument will now default to finding the current active LLMObs span rather than the current active APM span.
- ~Adds multithreading (futures.multithreading) support for LLMObs. Previously multithreaded apps would result in broken traces.~ UPDATE: will be moving multithreading support to a separate PR.

Private changes:
- LLMObs has its own context provider which keeps track of the active LLM-type span (generated by both LLMObs._start_span() and LLM integrations)
- HTTPPropagation now adds LLMObs parent ID as a field on the request headers directly, rather than through the context object.
- Adds private helper method `LLMObs._instance.current_span()`, returns the current active LLMObs-generated (integration, SDK) span.
- Adds private helper method `LLMObs._instance._current_trace_context()`, returns current LLMObs context (which can represent both a span or a distributed span)
- Adds a new field to the LLMObs span event struct, `_dd` which is a str/str dictionary containing the span/trace IDs of the APM span to correlate with. Currently these are the same span/trace IDs as the LLMObs span/trace ID, but this unlocks future steps of using independent span/trace IDs.
 
## Previous behavior
LLMObs spans are based on APM spans, except LLMObs spans' parenting involves only other LLMObs spans. So with a potential trace structure containing a mixture of APM-specific and LLMObs spans, like:
```
Span A (LLMObs span) --> Span B (Apm-specific span) --> Span C (LLMObs span)
```
LLMObs only cares about the LLMObs spans, where span C's parent is the root span, even though in APM it would be span B. Combined with distributed tracing and multithreading, this makes it not so easy to determine that "correct" (read LLMObs) parenting tree for traces submitted to LLM Observability.

### Problems with previous approach

Previously we worked around this by traversing the span's local parent tree and finding the next LLM-type span on both span start and finish for non-distributed cases, and for distributed cases we would attach the parent ID on the span context's meta field to be propagated in distributed request headers. However attaching things to the span context meta was not suitable long-term due to a couple factors:
1. Context objects are not thread-safe: in a multithreading case with n>1 child threads creating their own spans, the parent ID stored in the context object could be overwritten during thread execution, therefore incorrectly propagating parent IDs.
2. Context objects store trace-specific information, and are not designed for our use case where we skip spans here and there in the trace. This also leads to edge cases that were handled with [ugly workaround code](https://github.com/DataDog/dd-trace-py/blob/3bffd02a071db91a6fdc56ae12b496dc84ff8abf/ddtrace/llmobs/_llmobs.py#L312-L317):
<details>
  <summary><i>Example ugly workaround</i></summary>
  <b>Any meta fields set on the context object gets propagated as span tags on all subsequent spans in the trace on span start time, except for the spans in the first service of a trace which get propagated at span finish time. Fixing this resulted in overriding these span tags on span start and more checks on span finish.</b>
</details>

## Current approach
Instead of being dependent on a Context object that doesn't quite fit our use case and trying to make it fit our use case, we simply keep track of our own active LLMObs span/context:
- `LLMObsContextProvider` handles keeping track of the current active LLMObs span via `active()` and `activate()`
- Instead of traversing a span's local ancestor tree to solve for a span's llmobs parent ID, we just use `LLMObsContextProvider._activate_llmobs_span()` and set the llmobs parent ID as a tag at span start time.
(called by `LLMObs._start_span()` and `BaseLLMIntegration.trace(submit_to_llmobs=True)` and the bedrock integration).
- `LLMObs.inject_distributed_headers` now uses the LLMObsContextProvider to inject the active llmobs span's ID into span context and request headers
- `LLMObs.activate_distributed_headers()` now uses the LLMObsContextProvider to activate the extracted llmobs context to continue the trace in a distributed case.
- `trace_utils.activate_distributed_headers()` now includes automatic llmobs context activation if llmobs is enabled. I've used signal handling to avoid importing LLMObs entirely for non-LLMObs users (same for `HTTPPropagator.inject()`).

By keeping track of our own active LLMObs spans, spans submitted to LLM Observability have an independent set of span and parent IDs, even if the span and trace IDs are shared with APM spans for now. This is the first step to decoupling from tracer internals.

## Next steps
We can go further by generating LLMObs-specific span/trace IDs which are separate from APM. This will solve some edge cases with traces involving mixed APM/LLMObs spans.

## Checklist

## Reviewer Checklist
- Title is accurate

[MLOB-1342]: https://datadoghq.atlassian.net/browse/MLOB-1342?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


erikayasuda (approved):
apm-core files LGTM!

ddtrace/llmobs/_llmobs.py
```diff
@@ -807,14 +836,38 @@ def inject_distributed_headers(cls, request_headers: Dict[str, str], span: Optio
...
+        if not context:
+            context = HTTPPropagator.extract(request_headers)
+        if not context.trace_id or not context.span_id:
+            log.warning("Failed to extract trace/span ID from request headers.")
+            return
+        if PARENT_ID_KEY not in request_headers:
+            log.warning("Failed to extract LLMObs parent ID from request headers.")
+        _parent_id = request_headers.get(PARENT_ID_KEY)
+        if _parent_id is None:
+            return
```
sabrenner, line 863:
```suggestion
        if PARENT_ID_KEY not in request_headers:
            log.warning("Failed to extract LLMObs parent ID from request headers.")
            return
        _parent_id = request_headers.get(PARENT_ID_KEY)
```

sabrenner (approved):
LGMT! Just a couple small nits, but feel free to ignore if they're intentional. Really cool stuff, walked through the logic a couple times and seems OK to me.

tests/llmobs/_utils.py
```diff
@@ -193,6 +193,7 @@ def _llmobs_base_span_event(
         "meta": {"span.kind": span_kind},
         "metrics": {},
         "tags": _expected_llmobs_tags(span, tags=tags, error=error, session_id=session_id),
+        "_dd": {"span_id": str(span.span_id), "trace_id": "{:x}".format(span.trace_id)},
```
lievan, line 196: Does this represents APM trace & span id for the trace link?

lievan (approved):

tests/llmobs/_utils.py
Yun-Kim: Yes it does 👍

ddtrace/llmobs/_llmobs.py
```diff
@@ -352,23 +352,46 @@ def export_span(cls, span: Optional[Span] = None) -> Optional[ExportedLLMObsSpan
...
+            if span.span_type != SpanTypes.LLM:
+                log.warning("Span must be an LLMObs-generated span.")
+                return None
+            return ExportedLLMObsSpan(span_id=str(span.span_id), trace_id="{:x}".format(span.trace_id))
+        except (TypeError, AttributeError):
+            log.warning("Failed to export span. Span must be a valid Span object.")
             return None
-        return ExportedLLMObsSpan(span_id=str(span.span_id), trace_id="{:x}".format(span.trace_id))
+
+    def current_span(self) -> Optional[Span]:
```
Yun-Kim, line 370: Should we make this private? No great reasons to make it public.

ddtrace/llmobs/_llmobs.py
```diff
@@ -352,23 +352,46 @@ def export_span(cls, span: Optional[Span] = None) -> Optional[ExportedLLMObsSpan
...
+        """Returns the currently active LLMObs-generated span.
+        Note that there may be an active span represented by a context object
+        (i.e. a distributed trace) which will not be returned by this method.
+        """
+        active = self._llmobs_context_provider.active()
+        return active if isinstance(active, Span) else None
+
+    def current_trace_context(self) -> Optional[Context]:
```
Yun-Kim, line 378: Should we make this private? No great reasons to make it public.

mabdinur (approved):

Yun-Kim:
Gotta look at if relying on contextvars means we don't need to actively inject llmobs propagation into the futures threading integration.

ddtrace/propagation/http.py
```diff
@@ -1090,9 +1090,9 @@ def parent_call():
                 headers[_HTTP_BAGGAGE_PREFIX + key] = span_context._baggage[key]
 
         if config._llmobs_enabled:
-            from ddtrace.llmobs._utils import _inject_llmobs_parent_id
+            from ddtrace.llmobs import LLMObs
 
-            _inject_llmobs_parent_id(span_context)
+            LLMObs._inject_llmobs_context(headers)
```
Yun-Kim, line 1095: Instead of injecting into the headers, let's add it into the context and let the injection mechanisms propagate that into the headers.

ddtrace/contrib/internal/futures/threading.py
```diff
@@ -13,24 +15,34 @@ def _wrap_submit(func, args, kwargs):
...
     variable because it's outside the asynchronous loop.
     """
-    if ctx is not None:
-        ddtrace.tracer.context_provider.activate(ctx)
+    if ctx[0] is not None:
+        ddtrace.tracer.context_provider.activate(ctx[0])
+    if ctx[1] is not None and config._llmobs_enabled:
+        from ddtrace.llmobs import LLMObs
+
+        LLMObs._instance._llmobs_context_provider.activate(ctx[1])
```
Yun-Kim, line 47: Explore using signals instead of importing!

ddtrace/llmobs/_integrations/bedrock.py
```diff
@@ -36,9 +34,7 @@ def _llmobs_set_tags(
         operation: str = "",
     ) -> None:
         """Extract prompt/response tags from a completion and set them as temporary "_ml_obs.*" tags."""
-        if span.get_tag(PROPAGATED_PARENT_ID_KEY) is None:
-            parent_id = _get_llmobs_parent_id(span) or "undefined"
-            span._set_ctx_item(PARENT_ID_KEY, parent_id)
+        LLMObs._instance._activate_llmobs_span(span)
```
Yun-Kim, line 37: Note: this is an existing issue and we'll need to fix it after this PR. But because of the way the bedrock integration uses shared tracing handlers, we don't actually get to reuse the `BaseLLMIntegration.trace()` helper to activate the span immediately when it's started. We'll need to fix it to get everything context activation-wise in one place.

ddtrace/contrib/internal/futures/threading.py
Yun-Kim: Following up on a separate PR.

Kyle-Verhoog (approved):
great work @Yun-Kim!

Yun-Kim:
Closing in favor of #12236 due to branch conflicts with 3.x-staging.

PR #10796, opened 2024-09-24
fix(profiling): code provenance using libdatadog exporter

taegyunkim:
## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/internal/datadog/profiling/dd_wrapper/src/code_provenance.cpp
```diff
@@ -0,0 +1,178 @@
...
+    new (&get_instance().mtx) std::mutex(); // Recreate the mutex
+    get_instance().reset();                 // Reset the state
+}
+
+void
+Datadog::CodeProvenance::set_enabled(bool enable)
+{
+    std::lock_guard<std::mutex> lock(mtx);
+    enabled = enable;
+}
```
sanchda, line 27: It's possible, but not necessary, to improve this by making `enabled` a std::atomic.  Since it's a bool, it'll use a lock-free implementation on all the platforms we support, which will diminish the need for a lock.

In all fairness, there's almost certainly no need to do even _that_ much, but I think there's value in making it clear that an interface is intended for concurrent access.

Again--not necessary, since I don't think this is in a performance-critical path.  Plus there is something to be said about using uniform primitives across the interface, so maybe it's not even worth it.

ddtrace/internal/datadog/profiling/dd_wrapper/src/code_provenance.cpp
```diff
@@ -0,0 +1,178 @@
...
+Datadog::CodeProvenance::set_stdlib_path(std::string_view _stdlib_path)
+{
+    std::lock_guard<std::mutex> lock(mtx);
+    this->stdlib_path = _stdlib_path;
+}
+
+void
+CodeProvenance::add_packages(std::unordered_map<std::string_view, std::string_view> distributions)
+{
+    std::lock_guard<std::mutex> lock(mtx);
```
sanchda, line 53: If `enabled` was atomic, it'd be possible to move the lock guard to after the check.

ddtrace/internal/datadog/profiling/dd_wrapper/src/code_provenance.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+    if (!enabled) {
+        return "";
+    }
+
+    std::ostringstream out;
+
+    // TODO(taegyunkim): Use a JSON library to serialize this
```
sanchda, line 106: https://github.com/nlohmann/json is quite easy to use (not that there's any shortage of good json libraries!)

sanchda: I'm also OK with leaving it like this for now.  No need to overcomplicate our dependencies for one function.  :)

ddtrace/internal/datadog/profiling/dd_wrapper/src/uploader.cpp
```diff
@@ -68,20 +71,36 @@ Datadog::Uploader::upload(ddog_prof_Profile& profile)
...
-                                                      nullptr,
-                                                      nullptr);
+    std::vector<ddog_prof_Exporter_File> files_to_send = { {
+      .name = to_slice("auto.pprof"),
+      .file = ddog_Vec_U8_as_slice(&encoded->buffer),
+    } };
+
+    std::string json_str;
+
+    // DEV: re-consider this is ok to do so, as calling is_enabled/serialize will acquire
```
sanchda, line 81: Yeah, good point.  The check for enablement is non-atomic.

I guess the only way to solve this is actually internal to the CodeProvenance object itself.  You could have a `try_serialize_to_json_str()` or whatever that internally checks for enablement and returns an error or whatever if it isn't enabled.

In practice, I guess the worst-case scenario is that we lose some code provenance data under very weird circumstances?  So not a super big deal to solve.

sanchda (approved):
Good stuff!  I left some comments, but none of them are blockers for this PR.

emmettbutler (approved):

ddtrace/internal/datadog/profiling/dd_wrapper/src/code_provenance.cpp
taegyunkim: Updated to use `atomic`

ddtrace/internal/datadog/profiling/dd_wrapper/src/code_provenance.cpp
taegyunkim: Updated to use `atomic`

ddtrace/internal/datadog/profiling/dd_wrapper/src/uploader.cpp
taegyunkim: Oh, I was referring to the fact that this function is run with profile lock held and we're also going to take the lock in CodeProvenance too.

For your comment, `serialize_to_json_str()` already checks for `enabled` so code is simplified here.

PR #10808, opened 2024-09-25
feat(langchain): trace similarity search with vector and score

yahya-mouman:
This PR adds instrumentation to submit langchain : 
- similarity_search_with_score
- asimilarity_search_with_score
- similarity_search_with_score_by_vector
- asimilarity_search_with_score_by_vector
- similarity_search_by_vector
- asimilarity_search_by_vector
- similarity_search_by_vector_with_score
- asimilarity_search_by_vector_with_score

It also patches langchain-pinecone, langchain-qdrant and langchain-milvus

This is in response to the initial feature that traces similarity_search to further extend the support for it and highlight the scoring and the vector retrieval part.

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/llmobs/_integrations/langchain.py
```diff
@@ -287,11 +304,17 @@ def _llmobs_set_meta_tags_from_similarity_search(
...
+            else:
+                doc, score = d, None
+            document = Document(text=doc.page_content)
+            document["id"] = getattr(doc, "id", "")
+            metadata = getattr(doc, "metadata", {})
+            document["name"] = metadata.get("name", document["id"])
+            if score is not None:
+                document["score"] = score
+            documents.append(document)
         span.set_tag_str(OUTPUT_DOCUMENTS, safe_json(self.format_io(documents)))
```
sabrenner, line 318:
```suggestion
        span.set_tag_str(OUTPUT_DOCUMENTS, safe_json(documents))
```

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -905,10 +909,15 @@ def traced_similarity_search(langchain, pin, func, instance, args, kwargs):
         documents = func(*args, **kwargs)
         span.set_metric("langchain.response.document_count", len(documents))
         for idx, document in enumerate(documents):
+            if not isinstance(document, tuple):
```
lievan, line 912: Do we re-use `document` elsewhere? If not, maybe we can do
```
for idx, document_and_maybe_score in enumerate(documents):
    if not isinstance(document_and_maybe_score, tuple):
        document, score = document_and_maybe_score, None
    else:
        document, score = document_and_maybe_score
    span.set_tag_str(
        "langchain.response.document.%d.page_content" % idx, integration.trunc(str(document.page_content))
    )
...
```
For me that is a bit more clear

ddtrace/llmobs/_integrations/langchain.py
```diff
@@ -341,12 +357,18 @@ def _llmobs_set_meta_tags_from_similarity_search(
...
-            metadata = getattr(d, "metadata", {})
-            doc["name"] = metadata.get("name", doc["id"])
-            documents.append(doc)
-        span.set_tag_str(OUTPUT_DOCUMENTS, safe_json(self.format_io(documents)))
+            if isinstance(d, tuple):
+                doc, score = d
+            else:
+                doc, score = d, None
+            document = Document(text=doc.page_content)
+            document["id"] = getattr(doc, "id", "")
```
lievan, line 365:
```suggestion
            if getattr(doc, "id"):
                document["id"] = doc.id
```
And I think we should have the same logic for metadata & name

ddtrace/contrib/internal/langchain/patch.py
yahya-mouman: Hah I like the name ! Yes I don't believe it's used anywhere else.

ddtrace/llmobs/_integrations/langchain.py
yahya-mouman: Not sure I fully understand. The suggestion you added would make it so that we wouldn't have an id entry when we want to get it for the name substitute. Hence why we leave it as an empty string in this case.
Can you elaborate more on what is wrong with the current logic ?

ddtrace/llmobs/_integrations/langchain.py
```diff
@@ -84,6 +85,10 @@ def _llmobs_set_tags(
 
             is_workflow = LLMObs._integration_is_enabled(llmobs_integration)
 
+        if kwargs.get("is_workflow") is not None:
+            # when the traced function forces an is_workflow state
+            is_workflow = bool(kwargs.get("is_workflow"))
+
```
sabrenner, line 91: This fails tests because we sometimes override `kwargs` to be a list ([ref](https://github.com/DataDog/dd-trace-py/blob/2abc58a2971610cf2625a1934ac54d3da3c9814d/ddtrace/contrib/internal/langchain/patch.py#L688)). I re-discovered this by using the debugger and seeing that `kwargs` is a list. Two things we could do:
1. change the check to `if isinstance(kwargs, dict) and kwargs.get(...) is not None:`
2. pass in `is_workflow` separately from `kwargs`, which I would argue for. I see in the patching code we do `kwargs["is_workflow"] = True`, can we just pass in `is_workflow` as a keyword arg to `llmobs_set_tags` instead?

I'm personally in favor of the second

yahya-mouman: For the second are we okay with adding it even in the base integration template ?

ddtrace/llmobs/_integrations/langchain.py
sabrenner: ah, right. maybe there's another way we can check this? ie, a property on the span, like resource name, `interface_type`, etc? I think that could be a better way of doing it, instead of modifying kwargs. If we need to modify the `llmobs_set_tags` to take in optional `kwargs`, I don't think it's the worst thing, but if we can do it deterministically based on span properties first I think that's best. lmk what you think!

ddtrace/llmobs/_integrations/langchain.py
sabrenner: ah, right. maybe there's another way we can check this? ie, a property on the span, like resource name, interface_type, etc? I think that could be a better way of doing it, instead of modifying kwargs. If we need to modify the llmobs_set_tags to take in optional kwargs, I don't think it's the worst thing, but if we can do it deterministically based on span properties first I think that's best. lmk what you think!

yahya-mouman: I believe for now and for the sake of getting stuff shipped i'll stick to 1. I tried it and it works well. At the moment the only use case we have of force setting a span to a workflow is in the retrieval case. Maybe if we use it more then we can start thinking of establishing a more permanent spot for it.

sabrenner: I think that should be OK for now, especially since we're modifying `kwargs` _after_ the call is made. But I think we should follow up with a better solution if we can, we shouldn't try to modify inputs to a function unless we're wrapping one of the arguments (ie, some kind of function)

ddtrace/llmobs/_integrations/langchain.py
sabrenner: i think maybe it's that we don't want to set the document id if it doesn't exist on the returned document. we should only set it on our document object if it existed on the returned document, and not default to an empty string.

riotfile.py
```diff
@@ -2619,6 +2619,7 @@ def select_pys(min_version=MIN_PYTHON_VERSION, max_version=MAX_PYTHON_VERSION):
             },
             venvs=[
                 Venv(
+                    name="langchain_v0_langchain_community_v0",
```
sabrenner, line 2622: thanks for renaming these! should make isolated testing a bit easier 😅 🎉

tests/contrib/langchain/test_langchain_llmobs.py
```diff
@@ -527,6 +553,7 @@ def test_llmobs_chain(self, langchain_core, langchain_openai, mock_llmobs_span_w
             mock_tracer=mock_tracer,
             cassette_name="lcel_openai_chain_call.yaml",
         )
+        # TODO : test gives 1 span instead of 2
```
sabrenner, line 556:
```suggestion
```

tests/contrib/langchain/test_langchain_llmobs.py
```diff
@@ -551,6 +578,7 @@ def test_llmobs_chain_nested(self, langchain_core, langchain_openai, mock_llmobs
             mock_tracer=mock_tracer,
             cassette_name="lcel_openai_chain_nested.yaml",
         )
+        # TODO : figure out why it's 2 not 4
```
sabrenner, line 581:
```suggestion
```

sabrenner (commented):
gave it a first pass, left a couple questions on if we can also trace async methods in this PR, otherwise why not. also left a couple suggestions on style, but those aren't as blocking. 

overall, really good job on navigating the multitude of vectorstores and how the different methods interact with each other, and safe wrapping to account for it! 🎉

sabrenner:
also, looks like langchain tests are still failing because you might need to run `scripts/compile-and-prune-test-requirements`

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -917,25 +932,96 @@ def traced_similarity_search(langchain, pin, func, instance, args, kwargs):
...
+            )
+            span.set_tag_str(
+                "langchain.request.pinecone.index_name",
+                instance._index.configuration.server_variables.get("index_name", ""),
+            )
+            span.set_tag_str(
+                "langchain.request.pinecone.project_name",
+                instance._index.configuration.server_variables.get("project_name", ""),
+            )
+            api_key = instance._index.configuration.api_key.get("ApiKeyAuth", "")
```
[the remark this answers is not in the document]
yahya-mouman: Added that and updated the previous traced_similarity_search as well !

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -917,25 +932,96 @@ def traced_similarity_search(langchain, pin, func, instance, args, kwargs):
         integration.metric(span, "incr", "request.error", 1)
         raise
     finally:
+        # force the similarity search to be a workflow
+        kwargs["is_workflow"] = True
```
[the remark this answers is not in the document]
yahya-mouman: I believe it's safe unless there is a kwarg named is_workflow in the original list. If updating the original llmobs_set_tags interface with an optional field is fine i'm willing to do it within this PR as well.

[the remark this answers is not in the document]
sabrenner: i think we're good to leave it for now, unless you wanted to try adding it in here. we can leave it as a follow-up 😄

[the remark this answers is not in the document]
yahya-mouman: I gave it a try I'll push the changes now !

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -867,19 +868,23 @@ async def traced_lcel_runnable_sequence_async(langchain, pin, func, instance, ar
 
 @with_traced_module
 def traced_similarity_search(langchain, pin, func, instance, args, kwargs):
```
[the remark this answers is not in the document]
yahya-mouman: I spent a bit extra time to add the async versions of each raced method and as many tests as possible. It does inflate the test suite though since we don't have a way to execute an async version of a function without duplicating the code.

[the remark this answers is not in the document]
sabrenner: yeah, that's the unfortunate nature of adding async wrappers in Python, it's really gonna be a lot of duplicated code for like 1 line of change 😢 there are definitely use cases where users are using the async versions of these methods though, so I'd say it's good to support it.

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -900,42 +910,307 @@ def traced_similarity_search(langchain, pin, func, instance, args, kwargs):
...
+            span, args=args, kwargs=kwargs, response=documents, operation="retrieval", is_workflow_override=True
+        )
         span.finish()
         integration.metric(span, "dist", "request.duration", span.duration_ns)
         if integration.is_pc_sampled_log(span):
             integration.log(
                 span,
                 "info" if span.error == 0 else "error",
-                "sampled %s.%s" % (instance.__module__, instance.__class__.__name__),
+                "sampled %s.%s.%s" % (instance.__module__, instance.__class__.__name__, func.__name__),
```
sabrenner, line 949: we can likely get rid of these changes because logs are deprecated

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -900,42 +910,307 @@ def traced_similarity_search(langchain, pin, func, instance, args, kwargs):
...
+                span,
+                "info" if span.error == 0 else "error",
+                "sampled %s.%s.%s" % (instance.__module__, instance.__class__.__name__, func.__name__),
                 attrs={
                     "query": query,
                     "k": k or "",
                     "documents": [
-                        {"page_content": document.page_content, "metadata": document.metadata} for document in documents
+                        {"page_content": document.page_content, "metadata": document.metadata}
+                        for document, _ in documents_with_maybe_score
```
sabrenner, line 1049: ditto, we don't have to necessarily have these

tests/snapshots/tests.contrib.langchain.test_langchain_community.test_milvus_vectorstore_similarity_search.json
```diff
@@ -0,0 +1,152 @@
...
....0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]",
```
sabrenner, line 134: why does this have the whole embedding, instead of it being truncated?

tests/snapshots/tests.contrib.langchain.test_langchain_community.test_milvus_vectorstore_asimilarity_search.json
```diff
@@ -0,0 +1,178 @@
...
....0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]",
```
sabrenner, line 160: why does this have the whole embedding, instead of truncated?

tests/snapshots/tests.contrib.langchain.test_langchain_community.test_milvus_vectorstore_asimilarity_search_by_vector.json
```diff
@@ -0,0 +1,157 @@
...
....0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]",
```
sabrenner, line 81: why does this have the whole embedding, instead of truncated?

tests/snapshots/tests.contrib.langchain.test_langchain_community.test_milvus_vectorstore_asimilarity_search_by_vector.json
```diff
@@ -0,0 +1,157 @@
...
```
sabrenner, line 139: why does this have the whole embedding, instead of truncated?

sabrenner (commented):
I think almost everything looks good to me! Just have a couple questions about why some `langchain.request.embedding` tags aren't truncated properly (are they bad snapshots?). Otherwise, will approve once I have some clarification!

Another thing I wanna raise, which maybe we can address right after - `traced_similarity_search` and `traced_similarity_search_by_vector` (+ async versions) are largely the same, except for input tagging. Could we possibly consolidate these functions, maybe with an outer function which takes a variable to check against, ie:
```python
def traced_similarity (langchain, is_similarity_search_by_vector):
  @with_traced_module
  def _traced_similarity_search(langchain, pin, func, instance, args, kwargs):
    ... # do the function, but where they differ, use is_similarity_search_by_vector to differentiate

  return _traced_similarity_search(langchain)

# then in `wrap`, but conforming to your new safe_wrap logic:
...
wrap(langchain.vectorstores, 'similarity_search`, traced_similarity(langchain, False))
```
Or something like that. Maybe it wouldn't work out, but if you like how it looks, it could be a good way to clean up the code a bit. Or maybe it would make it more confusing 😅 either way, I don't think it needs to be done in this PR.

tests/contrib/langchain/cassettes/langchain_community/langchain_qdrant_similarity_search_with_score.yaml
```diff
@@ -0,0 +1,67 @@
+interactions:
```
Yun-Kim, line 1: Ditto for qdrant cassettes, looks like we can reuse cassettes instead of creating new ones for each test case

tests/contrib/langchain/test_langchain.py
Yun-Kim, line 1: We no longer keep up / add new support for langchain v0, let's revert the added tests/changes to this file.

tests/contrib/langchain/test_langchain_community.py
```diff
@@ -757,24 +758,114 @@ def test_chat_prompt_template_does_not_parse_template(langchain_openai, mock_tra
...
             )
             embed = langchain_openai.OpenAIEmbeddings(model="text-embedding-ada-002")
             index = pc.Index("langchain-retrieval")
             vectorstore = langchain_pinecone.PineconeVectorStore(index, embed, "text")
-            vectorstore.similarity_search("Who was Alan Turing?", 1)
+            vectorstore.similarity_search("Evolution", 1)
+
+
+@pytest.mark.asyncio
+@pytest.mark.snapshot
```
Yun-Kim, line 781: Let's use the same token name as the synchronous test to avoid creating excessive snapshot files.

tests/contrib/langchain/test_langchain_community.py
```diff
@@ -757,24 +758,114 @@ def test_chat_prompt_template_does_not_parse_template(langchain_openai, mock_tra
...
+                environment=os.getenv("PINECONE_ENV", "<not-a-real-env>"),
+            )
+            embed = langchain_openai.OpenAIEmbeddings(model="text-embedding-ada-002")
+            index = pc.Index("langchain-retrieval")
+            vectorstore = langchain_pinecone.PineconeVectorStore(index, embed, "text")
+            vectorstore.similarity_search_with_score("Evolution", 1)
+
+
+@pytest.mark.asyncio
+@pytest.mark.snapshot
```
Yun-Kim, line 824: Ditto on duplicate snapshots

tests/contrib/langchain/test_langchain_community.py
```diff
@@ -806,20 +897,272 @@ def test_pinecone_vectorstore_retrieval_chain(langchain_openai, request_vcr):
...
+                url=os.getenv("QDRANT_URL", "<not-a-real-url"), api_key=os.getenv("QDRANT_API_KEY", "<not-a-real-key>")
+            )
+            vectorstore = langchain_qdrant.QdrantVectorStore(
+                collection_name="test_collection_1536", client=client, embedding=embeddings, distance="Dot"
+            )
+            vectorstore.similarity_search("LangChain", k=1)
+
+
+@pytest.mark.asyncio
+@pytest.mark.snapshot
```
Yun-Kim, line 928: Ditto

tests/contrib/langchain/test_langchain_community.py
```diff
@@ -806,20 +897,272 @@ def test_pinecone_vectorstore_retrieval_chain(langchain_openai, request_vcr):
...
+                url=os.getenv("QDRANT_URL", "<not-a-real-key>"), api_key=os.getenv("QDRANT_API_KEY", "<not-a-real-key>")
+            )
+            vectorstore = langchain_qdrant.QdrantVectorStore(
+                collection_name="test_collection_1536", client=client, embedding=embeddings, distance="Dot"
+            )
+            vectorstore.similarity_search_with_score("LangChain", k=1)
+
+
+@pytest.mark.asyncio
+@pytest.mark.snapshot
```
Yun-Kim, line 981: Ditto

tests/contrib/langchain/test_langchain_community.py
```diff
@@ -806,20 +897,272 @@ def test_pinecone_vectorstore_retrieval_chain(langchain_openai, request_vcr):
...
+                url=os.getenv("QDRANT_URL", "<not-a-real-key>"), api_key=os.getenv("QDRANT_API_KEY", "<not-a-real-key>")
+            )
+            vectorstore = langchain_qdrant.QdrantVectorStore(
+                collection_name="test_collection_1536", client=client, embedding=embeddings, distance="Dot"
+            )
+            vectorstore.similarity_search_by_vector(embedding=vector, k=1)
+
+
+@pytest.mark.asyncio
+@pytest.mark.snapshot
```
Yun-Kim, line 1033: Ditto

tests/contrib/langchain/test_langchain_community.py
```diff
@@ -806,20 +897,272 @@ def test_pinecone_vectorstore_retrieval_chain(langchain_openai, request_vcr):
...
+            langchain_openai=langchain_openai,
+            langchain_core=langchain_core,
+            langchain_milvus=langchain_milvus,
+            pymilvus=pymilvus,
+        )
+        vectorstore.similarity_search("weather", k=1)
+
+
+@pytest.mark.asyncio
+@pytest.mark.snapshot(
```
Yun-Kim, line 1081: Ditto

tests/contrib/langchain/test_langchain_community.py
```diff
@@ -806,20 +897,272 @@ def test_pinecone_vectorstore_retrieval_chain(langchain_openai, request_vcr):
...
+            langchain_core=langchain_core,
+            langchain_milvus=langchain_milvus,
+            pymilvus=pymilvus,
+        )
+        vector = vectorstore.embeddings.embed_documents("weather")[0]
+        vectorstore.similarity_search_by_vector(embedding=vector, k=1)
+
+
+@pytest.mark.asyncio
+@pytest.mark.snapshot(
```
Yun-Kim, line 1127: Ditto

tests/snapshots/tests.contrib.langchain.test_langchain_community.test_milvus_vectorstore_asimilarity_search.json
Yun-Kim: I'm going to as far as suggest we drop tagging the embeddings entirely.

tests/snapshots/tests.contrib.langchain.test_langchain_community.test_milvus_vectorstore_asimilarity_search_by_vector.json
```diff
@@ -0,0 +1,157 @@
+[[
```
Yun-Kim, line 1: Do we really need unique snapshot tests for async versions of the same test?

tests/snapshots/tests.contrib.langchain.test_langchain_community.test_faiss_vectorstore_retrieval.json
```diff
@@ -10,14 +10,14 @@
     "error": 0,
```
Yun-Kim, line 10: I'm seeing faiss added as a dependency and faiss snapshot tests, but not seeing them in the actual test file as tests. Can we either add them in, or remove these additional files / dependencies?

tests/snapshots/tests.contrib.langchain.test_langchain_community.test_pinecone_vectorstore_asimilarity_search_with_score.json
```diff
@@ -0,0 +1,113 @@
+[[
```
Yun-Kim, line 1: I think I'm not understanding what the different tagging we do (do we?) between similarity search with vector / by score that requires separate snapshot tests entirely. Can you correct me here? If not, could we try minimizing the number of snapshot tests (and file footprint) we add to the ddtrace project?

Yun-Kim (commented):
Implementation looks good, just some suggestions and concerns regarding unnecessary cassette/snapshot file duplication.

PR #10809, opened 2024-09-25
fix(tracing): avoid assigning span's local root to self, so that the python GC doesn't kick in

erikayasuda:
## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/_trace/span.py
```diff
@@ -595,6 +595,22 @@ def context(self) -> Context:
             self._context = Context(trace_id=self.trace_id, span_id=self.span_id, is_remote=False)
         return self._context
 
+    @property
+    # This can't actually return None, but we say it can to satisfy type checking.
+    def _local_root(self) -> Optional["Span"]:
```
brettlangdon, line 600: I would consider making this `-> "Span"`, we'll need to then check the places in code that do `span._local_root is None` to see if they actually are expecting a `None` value, or if they are just checking it because the original definition of `_local_root` was `Optional["Span"]`

brettlangdon (approved):
@erikayasuda and I also discussed another direction we could take.

Since each span holds a reference to it's parent, any time we need to access the "local root" we can just traverse this linked list back to the root span (span whose `_parent` is None) to get the local root as well.

In extreme cases a "tall" "local" trace will only be 20-30 spans deep, so the overhead of traversing this should not be too expensive (citation/benchmarks needed).

With this approach, `def _local_root() -> "Span":` could be a simple traversal, and then we can avoid ever needing to "set" local root, and passing along `span._local_root = parent._local_root`, etc



This is not a required change, but an observation.

ddtrace/_trace/span.py
```diff
@@ -595,6 +595,22 @@ def context(self) -> Context:
...
+        if self._local_root_value is None:
+            return self
+        return self._local_root_value
+
+    @_local_root.setter
+    def _local_root(self, value: "Span") -> None:
+        if value is not self:
```
brettlangdon, line 607: do we need an `else` branch here? like, what if we have a `self._local_root_value` set already, but then we try to update with `span._local_root = span`... do we then need to do `else: self._local_root_value = None` ?

I know this case is highly unlikely, but given this code it is technically possible.

tests/tracer/test_tracer.py
```diff
@@ -2057,3 +2057,23 @@ def test_asm_standalone_configuration():
...
+    assert not any(str(obj).startswith("<Span") for obj in gc.get_objects())
+
+    # To check the exact nature of the objects and their references, use the following:
+
+    # for i, obj in enumerate(objects):
+    #     print("--------------------")
+    #     print(f"object {i}:", obj)
+    #     print("referrers:", [f"object {objects.index(r)}" for r in gc.get_referrers(obj)[:-2]])
+    #     print("referents:", [f"object {objects.index(r)}" if r in objects else r for r in gc.get_referents(obj)])
+    #     print("--------------------")
```
erikayasuda, line 2079: This was helpful to work with, thank you!

ddtrace/propagation/http.py
```diff
@@ -988,6 +988,7 @@ def parent_call():
             span_context = non_active_span.context
 
         if hasattr(ddtrace, "tracer") and hasattr(ddtrace.tracer, "sample"):
+            root_span: Optional[Span] = None
```
erikayasuda, line 991: Added this change to make mypy happy

ddtrace/_trace/span.py
```diff
@@ -595,6 +595,22 @@ def context(self) -> Context:
             self._context = Context(trace_id=self.trace_id, span_id=self.span_id, is_remote=False)
         return self._context
 
+    @property
+    def _local_root(self) -> "Span":
+        if self._local_root_value is None:
+            return self
+        else:
+            return self._local_root_value
```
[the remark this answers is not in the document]
erikayasuda: Thought I needed this to make mypy happy, but it was unhappy elsewhere. Resolved.

ddtrace/_trace/span.py
erikayasuda: Done 👍

taegyunkim (changes requested):
From profiling, we need to propagate `local_root.span_id` and `local_root.span_type` and this PR will change the behavior by returning early. 

1. https://github.com/DataDog/dd-trace-py/blob/057bb26193fe4ad713f6e5990d4a9679ae3c92b0/ddtrace/profiling/event.py#L104
2. https://github.com/DataDog/dd-trace-py/blob/057bb26193fe4ad713f6e5990d4a9679ae3c92b0/ddtrace/internal/datadog/profiling/ddup/_ddup.pyx#L308
3. https://github.com/DataDog/dd-trace-py/blob/057bb26193fe4ad713f6e5990d4a9679ae3c92b0/ddtrace/internal/processor/endpoint_call_counter.py#L41

I hope profiling tests catch those, if not I'm happy to contribute tests.

taegyunkim (changes requested):
These also need to be updated too, right?

1. https://github.com/DataDog/dd-trace-py/blob/057bb26193fe4ad713f6e5990d4a9679ae3c92b0/ddtrace/internal/core/__init__.py#L303
2. https://github.com/DataDog/dd-trace-py/blob/057bb26193fe4ad713f6e5990d4a9679ae3c92b0/ddtrace/debugging/_exception/replay.py#L125
3. https://github.com/DataDog/dd-trace-py/blob/057bb26193fe4ad713f6e5990d4a9679ae3c92b0/ddtrace/propagation/_database_monitoring.py#L80

erikayasuda:
> From profiling, we need to propagate `local_root.span_id` and `local_root.span_type` and this PR will change the behavior by returning early.

@taegyunkim  All tests were passing in a prior commit (no functionality was changed, just some mypy nits). Is the concern here that `_local_root` will return `None` instead of the actual Span? Given the property getter/setter they added, the behavior of calling `_local_root` on a Span should still be the same, we just have a new slot to prevent the cyclic reference. If the issue is elsewhere, we'll definitely take up your offer to contribute to add the necessary profiling tests 🙏

taegyunkim:
@erikayasuda Re-read your code and our tests. We do have tests checking `local_root.span_id` and `local_root.span_type` are propagated and they all pass, so should be ok. 

https://github.com/DataDog/dd-trace-py/blob/057bb26193fe4ad713f6e5990d4a9679ae3c92b0/tests/profiling_v2/collector/test_stack.py#L55-L56

Thanks for confirming!

taegyunkim (approved):
Thanks for the work

mabdinur:
> @erikayasuda and I also discussed another direction we could take.
> 
> Since each span holds a reference to it's parent, any time we need to access the "local root" we can just traverse this linked list back to the root span (span whose `_parent` is None) to get the local root as well.
> 
> In extreme cases a "tall" "local" trace will only be 20-30 spans deep, so the overhead of traversing this should not be too expensive (citation/benchmarks needed).
> 
> With this approach, `def _local_root() -> "Span":` could be a simple traversal, and then we can avoid ever needing to "set" local root, and passing along `span._local_root = parent._local_root`, etc
> 
> This is not a required change, but an observation.

This is a bit off topic but with partial flushing the root span can be finished, encoded, and sent to the agent while a child span is still active. Since spans are only encoded once, any modifications made to a finished Span may not be reflected in the payload sent to the Datadog Agent.  

We can end up in a scenario where `ddtrace.tracer.current_root_span()` returns a span that has ended, the span is updated (ex: tags are set) but those changes are never reflected in the Datadog UI. In most cases child spans are finished before parent spans but parent/root spans finishing before a child span is a common scenario in async executions. 

Should we solve this issue in this PR? Should we only keep a reference to local root spans that have not been encoded and sent to the Datadog Agent?

ddtrace/propagation/http.py
```diff
@@ -988,6 +988,7 @@ def parent_call():
             span_context = non_active_span.context
 
         if hasattr(ddtrace, "tracer") and hasattr(ddtrace.tracer, "sample"):
```
mabdinur, line 990: nit: do we need this check? ddtrace.tracer should always be defined and the tracer should always have a `sample` method

mabdinur (commented):
The memory usage benchmark scenarios look very promising. Can we also include screen captures of the results from the flaskrealworld app in the microbenchmarking platform?

erikayasuda:
@mabdinur 
> Should we only keep a reference to local root spans that have not been encoded and sent to the Datadog Agent?
Just to confirm, this is already the current behavior right? Or would we be introducing this new behavior of keeping references to a finished root span? If it's the latter, then I'd prefer to address it in this PR. Otherwise, I can own a separate PR to address that issue.

ddtrace/_trace/span.py
danielshaar: For what it's worth, I thought about this and only didn't include it because I figured there would never be a scenario in which a child span could turn into a root span. Seems totally fine to add to be defensive.

mabdinur:
> @mabdinur
> 
> > Should we only keep a reference to local root spans that have not been encoded and sent to the Datadog Agent?
> 
> Just to confirm, this is already the current behavior right? Or would we be introducing this new behavior of keeping references to a finished root span? If it's the latter, then I'd prefer to address it in this PR. Otherwise, I can own a separate PR to address that issue.

It's a separate issue. I can address it. I would like to propose making `Span` properties immutable after `Span.finish()` is called. Trying to modify a finished span should raise a warning.

ddtrace/_trace/span.py
```diff
@@ -595,6 +595,21 @@ def context(self) -> Context:
...
+    @property
+    def _local_root(self) -> "Span":
+        if self._local_root_value is None:
+            return self
+        return self._local_root_value
+
+    @_local_root.setter
+    def _local_root(self, value: "Span") -> None:
+        if value is not self:
+            self._local_root_value = value
```
erikayasuda, line 607:
```suggestion
    @_local_root.setter
    def _local_root(self, value: "Span") -> None:
        if value is not self:
            self._local_root_value = value
        else:
            self._local_root_value = None
```

ddtrace/propagation/http.py
erikayasuda: Good question 🤔 Honestly don't have enough historical context to back it up, but I'm assuming it used to _not_ always be the case? I can ask about this offline to the author for this code.

ddtrace/_trace/span.py
brettlangdon: I think from the code, we almost never set this, other than on initialization anyways?

Part of the reason I like the parent linked list traversal approach.. definitely higher overhead on reads, but probably simplifies a bunch of logic/code in places.

"like" - no strong opinion, no need to actually change that. this approach does work and covers every case Erika and I walked through.

brettlangdon:
@mabdinur 

> Should we solve this issue in this PR? Should we only keep a reference to local root spans that have not been encoded and sent to the Datadog Agent?

This change does not change any existing behavior or problems. This scenario exists today, and will continue to exist (no worse than today).

Especially since child spans will always hold reference to their parents, as long as child spans are still active (not GC'd at least) then all parents will not get GC'd as well.

Changing that finished spans get "locked" is new unrelated behavior change.

brettlangdon (approved):

ddtrace/_trace/span.py
danielshaar: Yeah, it's not set outside of initialization.

The parent traverse removes 1 line from tracer.py and changes the _local_root to a recursive function. Doesn't seem like much of a simplification?

On the flip-side, you introduce a limit of 1000 nested spans because of the recursive function call stack limit in Python. That could be fine in practice, but seems like an unnecessary restriction and worsening of performance, no?

danielshaar (approved):

mabdinur (approved):

PR #10843, opened 2024-09-27
feat: code origin for span

P403n1x87:
We add the feature of collecting code origin information for entry and exit span on traces to allow for a quick navigation to the relevant part of the code base.

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/debugging/_origin/span.py
```diff
@@ -0,0 +1,267 @@
...
+        super().__init__(f)
+
+        lines = linenos(f)
+        start_line = min(lines)
+        filename = str(Path(f.__code__.co_filename).resolve())
+        name = f.__qualname__
+        module = f.__module__
+        self.location = EntrySpanLocation(
+            name=name,
+            start_line=start_line,
```
shatzi, line 135: we no longer require start_line and end_lines. we can just put line number for entry and we use code tokenization now to find start and end lines.

So can remove this and no longer extract lines from the code.

ddtrace/debugging/_origin/span.py
```diff
@@ -0,0 +1,267 @@
...
+        span = ddtrace.tracer.current_span()
+        location = self.location
+        if root is None or span is None or root.get_tag("_dd.entry_location.file") is not None:
+            return self
+
+        # Add tags to the local root
+        for s in (root, span):
+            s.set_tag_str("_dd.code_origin.type", "entry")
+
+            s.set_tag_str("_dd.code_origin.frames.0.file", location.file)
```
shatzi, line 155: Should we make this as a const CODE_ORIGIN = "_dd.code_origin.frames" ?

shatzi (approved):
Thank you

wconti27 (approved):
LGTM

ddtrace/ext/__init__.py
```diff
@@ -22,3 +22,17 @@ class SpanKind(object):
...
+        SpanTypes.CACHE,
+        SpanTypes.CASSANDRA,
+        SpanTypes.ELASTICSEARCH,
+        SpanTypes.GRPC,
+        SpanTypes.HTTP,
+        SpanTypes.REDIS,
+        SpanTypes.SQL,
+        SpanTypes.WORKER,
+    }
+)
```
erikayasuda, line 38: I think I just lack context here, but why are these specifically `EXIT_SPAN_TYPES`? I see they're used in line 200 on `SpanCodeOriginProcessor.on_span_start()` to check the started span type, and if it's not one of these, we return early. Are these just the exit span types we're collecting the code origin data for currently, and it will expand later?

erikayasuda (approved):

ddtrace/ext/__init__.py
P403n1x87: Here we want to identify any span that is representing a call to an external service. AFAIK these are the only such span types that we currently support. This set could grow in the future if we introduce new types

PR #10845, opened 2024-09-27
chore(ci_visibility): add unique tests request to test vis API client

romainkomorndatadog:
This adds the unique tests request (`self.fetch_unique_tests()`) to the API client to support Early Flake Detection work.

A refactor wraps the client's `_do_request()` method is also wrapped in `_do_request_with_telemetry()` to DRY the three `_do_request()` callers. The telemetry calls are also refactored into a `api_request` telemetry module.

This also fixes the test verifying the skippable API request payload which had been shamefully copy/pasted and was actually not testing what it was supposed to.

## Checklist

## Reviewer Checklist
- Title is accurate

### DIFF
--- a/ddtrace/internal/ci_visibility/_api_client.py
+++ b/ddtrace/internal/ci_visibility/_api_client.py
@@ -88,13 +88,6 @@ class EarlyFlakeDetectionSettings:
 
 
 @dataclasses.dataclass(frozen=True)
-class UniqueTestsData:
-    test_modules: t.Dict[str, TestModuleId]
-    test_suites: t.Dict[str, TestSuiteId]
-    test_cases: t.Dict[str, InternalTestId]
-
-
-@dataclasses.dataclass(frozen=True)
 class TestVisibilityAPISettings:
     __test__ = False
     coverage_enabled: bool = False
@@ -313,10 +306,7 @@ class _TestVisibilityAPIClientBase(abc.ABC):
             try:
                 sw.stop()  # Stop the timer before parsing the response
                 response_bytes = len(response.body)
-                if isinstance(response.body, bytes):
-                    parsed = json.loads(response.body.decode())
-                else:
-                    parsed = json.loads(response.body)
+                parsed = json.loads(response.body)
                 return parsed
             except JSONDecodeError:
                 error_type = ERROR_TYPES.BAD_JSON
@@ -331,10 +321,10 @@ class _TestVisibilityAPIClientBase(abc.ABC):
         """
 
         metric_names = APIRequestMetricNames(
-            GIT_TELEMETRY.SETTINGS_COUNT,
-            GIT_TELEMETRY.SETTINGS_MS,
-            GIT_TELEMETRY.SETTINGS_RESPONSE,
-            GIT_TELEMETRY.SETTINGS_ERRORS,
+            count=GIT_TELEMETRY.SETTINGS_COUNT,
+            duration=GIT_TELEMETRY.SETTINGS_MS,
+            response_bytes=None,
+            error=GIT_TELEMETRY.SETTINGS_ERRORS,
         )
 
         payload = {
@@ -412,10 +402,10 @@ class _TestVisibilityAPIClientBase(abc.ABC):
             timeout = DEFAULT_ITR_SKIPPABLE_TIMEOUT
 
         metric_names = APIRequestMetricNames(
-            SKIPPABLE_TESTS_TELEMETRY.REQUEST,
-            SKIPPABLE_TESTS_TELEMETRY.REQUEST_MS,
-            SKIPPABLE_TESTS_TELEMETRY.RESPONSE_TESTS,
-            SKIPPABLE_TESTS_TELEMETRY.REQUEST_ERRORS,
+            count=SKIPPABLE_TESTS_TELEMETRY.REQUEST,
+            duration=SKIPPABLE_TESTS_TELEMETRY.REQUEST_MS,
+            response_bytes=SKIPPABLE_TESTS_TELEMETRY.RESPONSE_BYTES,
+            error=SKIPPABLE_TESTS_TELEMETRY.REQUEST_ERRORS,
         )
 
         payload = {
@@ -448,7 +438,7 @@ class _TestVisibilityAPIClientBase(abc.ABC):
 
         meta = skippable_response.get("meta")
         if meta is None:
-            log.debug("SKippable tests response did not contain metadata field, no tests will be skipped")
+            log.debug("Skippable tests response did not contain metadata field, no tests will be skipped")
             record_api_request_error(metric_names.error, ERROR_TYPES.BAD_JSON)
             return None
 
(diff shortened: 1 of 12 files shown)


ddtrace/internal/ci_visibility/_api_client.py
```diff
@@ -272,11 +286,57 @@ def _do_request(self, method: str, endpoint: str, payload: str, timeout: t.Optio
...
+                if response.status == 403:
+                    raise CIVisibilityAuthenticationException()
+                raise ValueError("API response status code: %d", response.status)
+            try:
+                sw.stop()  # Stop the timer before parsing the response
+                response_bytes = len(response.body)
+                if isinstance(response.body, bytes):
+                    parsed = json.loads(response.body.decode())
+                else:
+                    parsed = json.loads(response.body)
```
vitor-de-araujo, line 319: I've seen this repeated elsewhere. It looks like our `Response` class already has a [`get_json()` method](https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/internal/utils/http.py#L272) that kinda does the same thing (although it catches `ValueError` and `TypeError` and returns nothing in that case, and doesn't catch `JSONDecodeError`, which is a weird choice of exceptions to handle).

ddtrace/internal/ci_visibility/_api_client.py
```diff
@@ -293,56 +353,40 @@ def fetch_settings(self) -> TestVisibilityAPISettings:
...
+                    slow_test_retries_5s=attributes["early_flake_detection"]["slow_test_retries"]["5s"],
+                    slow_test_retries_10s=attributes["early_flake_detection"]["slow_test_retries"]["10s"],
+                    slow_test_retries_30s=attributes["early_flake_detection"]["slow_test_retries"]["30s"],
+                    slow_test_retries_5m=attributes["early_flake_detection"]["slow_test_retries"]["5m"],
+                )
+            else:
+                early_flake_detection = EarlyFlakeDetectionSettings()
         except KeyError:
-            record_settings(sw.elapsed() * 1000, error=ERROR_TYPES.UNKNOWN)
+            record_api_request_error(metric_names.error, ERROR_TYPES.UNKNOWN)
```
vitor-de-araujo, line 386: We don't record elapsed time anymore?

ddtrace/internal/ci_visibility/_api_client.py
```diff
@@ -272,11 +286,57 @@ def _do_request(self, method: str, endpoint: str, payload: str, timeout: t.Optio
...
+                if isinstance(response.body, bytes):
+                    parsed = json.loads(response.body.decode())
+                else:
+                    parsed = json.loads(response.body)
+                return parsed
+            except JSONDecodeError:
+                error_type = ERROR_TYPES.BAD_JSON
+                raise
+        finally:
+            record_api_request(metric_names, sw.elapsed() * 1000, response_bytes=response_bytes, error=error_type)
```
vitor-de-araujo, line 325: ...ah, we record it here now, right? So recording the request is separate from recording whether there were errors.

ddtrace/internal/ci_visibility/_api_client.py
```diff
@@ -366,102 +433,107 @@ def _do_skippable_request(self, timeout: float) -> t.Optional[_SkippableResponse
...
+            covered_files=covered_files,
+            skippable_items=items_to_skip,
+        )
 
-        covered_files: t.Optional[t.Dict[str, CoverageLines]] = None
+    def fetch_unique_tests(self) -> t.Optional[t.Set[InternalTestId]]:
+        metric_names = APIRequestMetricNames(
+            EARLY_FLAKE_DETECTION_TELEMETRY.REQUEST,
+            EARLY_FLAKE_DETECTION_TELEMETRY.REQUEST_MS,
+            EARLY_FLAKE_DETECTION_TELEMETRY.RESPONSE_BYTES,
```
vitor-de-araujo, line 485: This is the only one that has `RESPONSE_BYTES` in the name, are the other metrics used as the `response_bytes` parameter also number of bytes? Just to make sure.

ddtrace/internal/ci_visibility/_api_client.py
```diff
@@ -366,102 +433,107 @@ def _do_skippable_request(self, timeout: float) -> t.Optional[_SkippableResponse
...
+        if skippable_response is None:
+            # We did not fetch any data, but telemetry has already been recorded, and a warning has been logged
+            return None
 
-        except Exception:
-            log.warning("Error retrieving skippable test data, no tests will be skipped", exc_info=True)
-            error_type = ERROR_TYPES.UNKNOWN
+        meta = skippable_response.get("meta")
+        if meta is None:
+            log.debug("SKippable tests response did not contain metadata field, no tests will be skipped")
```
vitor-de-araujo, line 451: :ghost: 
```suggestion
            log.debug("Skippable tests response did not contain metadata field, no tests will be skipped")
```

ddtrace/internal/ci_visibility/_api_client.py
```diff
@@ -80,6 +87,13 @@ class EarlyFlakeDetectionSettings:
     faulty_session_threshold: float = 30.0
 
 
+@dataclasses.dataclass(frozen=True)
+class UniqueTestsData:
```
vitor-de-araujo, line 91: Where is this used? :eyes:

romainkomorndatadog: It is not, I thought it'd be useful and then decided not to use it and then (I think I can say "as usual", at this point), I forgot to remove it.

ddtrace/internal/ci_visibility/_api_client.py
romainkomorndatadog: It's vestigial from the original code, and IIRC, one of them was only there for compatibility with Python 2.7 anyway. 

I'll double check and remove one if that's the case. I'm not sure I want to use the `get_json()` method for the reason you mentioned.

ddtrace/internal/ci_visibility/_api_client.py
romainkomorndatadog: Yes, all the metrics are technically added separately and there's no association between them.

The only reason they were all in a single function call was because it is (or seems to be, at least) more logical.

ddtrace/internal/ci_visibility/_api_client.py
romainkomorndatadog: Eagle-eyed Vítor to the rescue. 😀

ddtrace/internal/ci_visibility/_api_client.py
```diff
@@ -80,6 +87,13 @@ class EarlyFlakeDetectionSettings:
...
+    test_modules: t.Dict[str, TestModuleId]
+    test_suites: t.Dict[str, TestSuiteId]
+    test_cases: t.Dict[str, InternalTestId]
```
vitor-de-araujo, line 94: "test case" here means individual tests? I ask because `unittest` usually uses a `BlaBlaTestCase` class for a test suite (if I understand correctly), so this might be confusing.

vitor-de-araujo (approved):

ddtrace/internal/ci_visibility/_api_client.py
romainkomorndatadog: They are but you made me (re)realize that we don't record the settings response bytes, so I have to make an exception for it. I shall refactor a bit.

ddtrace/internal/ci_visibility/_api_client.py
romainkomorndatadog: (resolving because I'm deleting)

PR #10851, opened 2024-09-27
fix(profiling): reverse locations for stack v2

taegyunkim:
Fixes an issue where flamegraph was upside down for stack v2, when `DD_PROFILING_STACK_V2_ENABLED`

## Checklist

## Reviewer Checklist
- Title is accurate


vitor-de-araujo (approved):
I don't have much context on this, but why did the order change from v1 to v2? I wonder if we should make the order the same in both instead of reversing it during flush.

sanchda:
> I don't have much context on this, but why did the order change from v1 to v2? I wonder if we should make the order the same in both instead of reversing it during flush.

It was discovered recently that the order of the frames was incorrect only in this situation.  v1 doesn't have this behavior during stack sampling, and other forms of sampling (memory, locks) also don't have it.

sanchda (approved):

sanchda:
>  why did the order change from v1 to v2?

Forgot to answer this question!

Stack v2 uses a different read mechanism for stacks.  Rather than using the Python interfaces, it traverses Python's vmstructs (internal runtime memory) directly.  The way it does this reverses the callstacks.  We should have reversed it all along.

PR #10861, opened 2024-09-30
chore(internal): remove time functions from compat module

vitor-de-araujo:
This PR removes 4 time-related functions from `ddtrace.internal.compat` that existed for compatibility with Python < 3.7, and replaces all uses with the corresponding functions from the builtin `time` module: `time_ns`, `monotonic`, `monotonic_ns`, `process_time_ns`. 

The original motivation for this was noticing that the fallback code for `time_ns` had the wrong multiplier [edit: it's not actually wrong, just written in a weird way], and then checking that we don't really need these functions anymore, as we only support Python >= 3.7.

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/profiling/event.py
```diff
@@ -16,7 +16,7 @@ class Event(object):
 
     __slots__ = ("timestamp",)
 
-    def __init__(self, timestamp=compat.time_ns()):
+    def __init__(self, timestamp=time.time_ns()):
```
vitor-de-araujo, line 19: Is this correct? `time.time_ns()` will only be evaluated once at module load time, not every time the constructor is called.

P403n1x87: Good shout. This shouldn't be a big issue because generally the profiler is a singleton and it is booted at start-up. The same _kinda_ applies to tests (modulo a few delays perhaps). However, definitely something to keep in mind, just in case.

ddtrace/debugging/_debugger.py
```diff
@@ -229,11 +229,11 @@ def _open_contexts(self) -> None:
             contexts.append(self._collector.attach(signal))
 
         # Save state on the wrapping context
-        self.set("start_time", compat.monotonic_ns())
+        self.set("start_time", time.monotonic_ns())
```
P403n1x87, line 232: Would we save much if we imported `monotonic_ns` and called it directly, instead of every time looking it up from the `time` module?

vitor-de-araujo: Is this something we should do just here, or on every place using `monotonic_ns`, or on every place using any of these functions?

P403n1x87: If it's not much hassle I think we can do it here. WDYT?

vitor-de-araujo: Done!

ddtrace/profiling/event.py
taegyunkim: oh, good catch. That sounds like a bug to me but I can't find any location that uses `timestamp`. I'll try removing it.

ddtrace/profiling/collector/_lock.py
```diff
@@ -117,12 +117,12 @@ def _acquire(self, inner_func, *args, **kwargs):
         if not self._self_capture_sampler.capture():
             return inner_func(*args, **kwargs)
 
-        start = compat.monotonic_ns()
+        start = time.monotonic_ns()
```
P403n1x87, line 120: We might want to do the same here to spare some compute cycles

ddtrace/profiling/event.py
taegyunkim: https://github.com/DataDog/dd-trace-py/pull/10864

taegyunkim (approved):

ddtrace/internal/utils/time.py
```diff
@@ -1,9 +1,9 @@
 from datetime import datetime
+import time
```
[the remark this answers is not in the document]
vitor-de-araujo: That's what I meant to ask [here](https://github.com/DataDog/dd-trace-py/pull/10861#discussion_r1781009288), if we should change all instances of `monotonic` or just that one, but I think the question was misunderstood. Should I go ahead and change all?

rachelyangdog (approved):

P403n1x87 (approved):

PR #10862, opened 2024-09-30
chore: introduce product plugin interface

P403n1x87:
We introduce a product plugin interface that allows us to implement dependency inversion for product life-cycle logic. This also allows us to encapsulate common support logic, such as the uWSGI master process forking support, so that individual products don't have to implement the same logic.

## Checklist

## Reviewer Checklist
- Title is accurate

### DIFF
--- a/ddtrace/internal/products.py
+++ b/ddtrace/internal/products.py
@@ -1,0 +1,178 @@
+import atexit
+from collections import defaultdict
+from collections import deque
+import sys
+import typing as t
+
+from ddtrace.internal import forksafe
+from ddtrace.internal.logger import get_logger
+from ddtrace.internal.uwsgi import check_uwsgi
+from ddtrace.internal.uwsgi import uWSGIConfigError
+from ddtrace.internal.uwsgi import uWSGIMasterProcess
+
+
+log = get_logger(__name__)
+
+if sys.version_info < (3, 10):
+    from importlib_metadata import entry_points
+else:
+    from importlib.metadata import entry_points
+
+try:
+    from typing import Protocol  # noqa:F401
+except ImportError:
+    from typing_extensions import Protocol  # type: ignore[assignment]
+
+
+class Product(Protocol):
+    requires: t.List[str]
+
+    def post_preload(self) -> None:
+        ...
+
+    def start(self) -> None:
+        ...
+
+    def restart(self, join: bool = False) -> None:
+        ...
+
+    def stop(self, join: bool = False) -> None:
+        ...
+
+    def at_exit(self, join: bool = False) -> None:
+        ...
+
+
+class ProductManager:
+    __products__: t.Dict[str, Product] = {}  # All discovered products
+
+    def __init__(self) -> None:
+        self._products: t.Optional[t.List[t.Tuple[str, Product]]] = None  # Topologically sorted products
+
+        for product_plugin in entry_points(group="ddtrace.products"):
+            name = product_plugin.name
+            log.debug("Discovered product plugin '%s'", name)
+
+            # Load the product protocol object
+            try:
+                product: Product = product_plugin.load()
+            except Exception:
+                log.exception("Failed to load product plugin '%s'", name)
+                continue
+
+            log.debug("Product plugin '%s' loaded successfully", name)
+
+            self.__products__[name] = product
+
+    def _sort_products(self) -> t.List[t.Tuple[str, Product]]:
+        # Data structures for topological sorting
+        q: t.Deque[str] = deque()  # Queue of products with no dependencies
+        g = defaultdict(list)  # Graph of dependencies
+        f = {}  # Remaining dependencies for each product
+
+        for name, product in self.__products__.items():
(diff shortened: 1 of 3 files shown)


ddtrace/internal/products.py
```diff
@@ -0,0 +1,178 @@
...
+
+        # Stop all products on exit
+        atexit.register(self.exit_products)
+
+    def run_protocol(self) -> None:
+        # uWSGI support
+        try:
+            check_uwsgi(worker_callback=forksafe.ddtrace_after_in_child)
+        except uWSGIMasterProcess:
+            # We are in the uWSGI master process, we should handle products in the
```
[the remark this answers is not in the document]
P403n1x87: This is uWSGI lingo

ddtrace/internal/products.py
```diff
@@ -0,0 +1,178 @@
...
+        ...
+
+
+class ProductManager:
+    __products__: t.Dict[str, Product] = {}  # All discovered products
+
+    def __init__(self) -> None:
+        self._products: t.Optional[t.List[t.Tuple[str, Product]]] = None  # Topologically sorted products
+
+        for product_plugin in entry_points(group="ddtrace.products"):
```
erikayasuda, line 52: So when we start implementing products, would we make them discoverable via `entry_points(group="ddtrace.products")` by adding them to the `pyproject.toml` file (like you did [here](https://github.com/DataDog/dd-trace-py/pull/10173/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711))?

ddtrace/internal/products.py
```diff
@@ -0,0 +1,178 @@
...
+    def stop_products(self, join: bool = False) -> None:
+        for name, product in reversed(self.products):
+            try:
+                product.stop(join=join)
+                log.debug("Stopped product '%s'", name)
+            except Exception:
+                log.exception("Failed to stop product '%s'", name)
+
+    def exit_products(self, join: bool = False) -> None:
+        for name, product in reversed(self.products):
```
erikayasuda, line 131: So we need to do this and `stop_products` in reverse because of the topological sorting?

ddtrace/internal/products.py
```diff
@@ -0,0 +1,178 @@
...
+        q: t.Deque[str] = deque()  # Queue of products with no dependencies
+        g = defaultdict(list)  # Graph of dependencies
+        f = {}  # Remaining dependencies for each product
+
+        for name, product in self.__products__.items():
+            product_requires = getattr(product, "requires", [])
+            if not product_requires:
+                q.append(name)
+            else:
+                f[name] = list(product_requires)
```
erikayasuda, line 78: Why do we need to type cast to `list` here again?

ddtrace/internal/products.py
```diff
@@ -0,0 +1,178 @@
...
+                continue
+
+            log.debug("Product plugin '%s' loaded successfully", name)
+
+            self.__products__[name] = product
+
+    def _sort_products(self) -> t.List[t.Tuple[str, Product]]:
+        # Data structures for topological sorting
+        q: t.Deque[str] = deque()  # Queue of products with no dependencies
+        g = defaultdict(list)  # Graph of dependencies
```
erikayasuda, line 70: nit: Would it make sense to make this a `set` instead? So that the `remove()` later on is less expensive? Probably not a huge deal since this will only run in when the service starts up

ddtrace/internal/products.py
```diff
@@ -0,0 +1,178 @@
...
+
+        return [(name, self.__products__[name]) for name in ordering if name not in f]
+
+    @property
+    def products(self) -> t.List[t.Tuple[str, Product]]:
+        if self._products is None:
+            self._products = self._sort_products()
+        return self._products
+
+    def start_products(self) -> None:
```
erikayasuda, line 106: Generally speaking, if products are being sorted based on dependency/topological order, does that mean earlier products failing to start up will cause problems for later products?

erikayasuda (approved):

ddtrace/internal/products.py
P403n1x87: Yes! 🙂

ddtrace/internal/products.py
P403n1x87: Correct

ddtrace/internal/products.py
P403n1x87: We are making a copy that we can modify in our sorting algorithm, leaving the original one untouched.

ddtrace/internal/products.py
P403n1x87: Ah good point. I think we actually need to do this for `f`, since we are just iterating over the values of `g`

ddtrace/internal/products.py
P403n1x87: Quite likely yes! We should probably store the outcome of the operation and prevent dependant products from starting (with some informative logging).

PR #10863, opened 2024-09-30
chore(telemetry): decouple telemetry writer from global config

mabdinur:
## Description

Currently, the telemetry writer relies on `ddtrace.config` to enable and send telemetry payloads. `ddtrace.config` is also used to generate and queue `app-started` and `app-client-configuration-changed` telemetry events. This creates a circular dependency that can be painful to manage. Currently we leverage inlined imports and hope that `ddtrace.config` is initialized before `ddtrace.internal.telemetry` is imported. This change resolves this circular dependency by refactoring components in`ddtrace.internal.telemetry` to access environment variables directly (instead of using ddtrace.config).

## Motiviation

This change will allow ddtrace contributors to queue telemetry configurations, metrics, and logs before the global config object is initialized.

## Changes
- Removes all references to `ddtrace.config` and all envier configuration objects from `ddtrace.internal.telemetry` package
- Ensures configuration telemetry is queued when configurations are set (currently this information is collected 10 seconds after ddtrace is first imported).
- Ensures the origin of a configuration value is recorded (possible values: remote_config, env_var, code, and unknown). Previously the origin of all configurations was set to unknown.
- Updates the name of telemetry configurations to match the name of environment variables (where possible).

## Checklist

## Reviewer Checklist
- Title is accurate


P403n1x87:
That's great! Did we consider a callback/core event hub alternative approach to break the circular dependency?

mabdinur:
> That's great! Did we consider a callback/core event hub alternative approach to break the circular dependency?

Good point. I though this approach could introduce a lot of complexity. We will need to ensure telemetry is started soon after we start queueing events to prevent memory leaks or data loss associated with unbounded metric and logs events.

I think the simplest approach is to decouple the `ddtrace.internal.telemetry` package from as many ddtrace modules as possible. This will allow us to start collecting telemetry via telemetry_writer in the vast majority of ddtrace modules. This PR doesn't fully achieve this goal but it's a step in that direction.

ddtrace/settings/config.py
```diff
@@ -561,20 +556,17 @@ def __init__(self):
...
-        self._ci_visibility_agentless_enabled = asbool(os.getenv("DD_CIVISIBILITY_AGENTLESS_ENABLED", default=False))
-        self._ci_visibility_agentless_url = os.getenv("DD_CIVISIBILITY_AGENTLESS_URL", default="")
-        self._ci_visibility_intelligent_testrunner_enabled = asbool(
-            os.getenv("DD_CIVISIBILITY_ITR_ENABLED", default=True)
-        )
-        self.ci_visibility_log_level = os.getenv("DD_CIVISIBILITY_LOG_LEVEL", default="info")
-        self._otel_enabled = asbool(os.getenv("DD_TRACE_OTEL_ENABLED", False))
+        self._ci_visibility_agentless_enabled = _get_config("DD_CIVISIBILITY_AGENTLESS_ENABLED", False, asbool)
+        self._ci_visibility_agentless_url = _get_config("DD_CIVISIBILITY_AGENTLESS_URL", "")
+        self._ci_visibility_intelligent_testrunner_enabled = _get_config("DD_CIVISIBILITY_ITR_ENABLED", False, asbool)
```
mabdinur, line 561:
```suggestion
        self._ci_visibility_intelligent_testrunner_enabled = _get_config("DD_CIVISIBILITY_ITR_ENABLED", True, asbool)
```

ddtrace/auto.py
```diff
@@ -19,3 +19,7 @@ def main():
 directly.
 """
 import ddtrace.bootstrap.sitecustomize  # noqa:F401
+from ddtrace.internal.telemetry import telemetry_writer
+
+
+telemetry_writer.add_configuration("ddtrace_auto_used", True, "unknown")
```
emmettbutler, line 25: Putting this in `auto.py` means that `import auto` falls out of sync with `import ddtrace.bootstrap.sitecustomize`. This isn't inherently a problem but it does change the meaning of `import auto` in a way that could surprise customers.

ddtrace/settings/_core.py
```diff
@@ -0,0 +1,50 @@
...
+from typing import Optional  # noqa:F401
+from typing import Union  # noqa:F401
+
+from envier.env import EnvVariable
+from envier.env import _normalized
+
+from ddtrace.internal.telemetry import telemetry_writer
+
+
+def report_telemetry(env):
```
emmettbutler, line 14:
```suggestion
def report_telemetry(env: Any) -> None:
```

emmettbutler (approved):

taegyunkim (approved):

taegyunkim:
LGTM, Approval for ddtrace/settings/profiling.py

tests/telemetry/test_writer.py
```diff
@@ -249,106 +263,222 @@ def test_app_started_event_configuration_override(
...
+            "[a-z0-9]{13}|gh[opsu]_[0-9a-zA-Z]{36}|ey[I-L][\\w=-]+\\.ey[I-L][\\w=-]+(?:\\.[\\w.+\\/=-]+)?|[\\-]"
+            "{5}BEGIN[a-z\\s]+PRIVATE\\sKEY[\\-]{5}[^\\-]+[\\-]{5}END[a-z\\s]+PRIVATE\\sKEY|ssh-rsa\\s*[a-z0-9\\/\\.+]{100,}",
+        },
+        {"name": "DD_APPSEC_RASP_ENABLED", "origin": "default", "value": "True"},
+        {"name": "DD_APPSEC_RULES", "origin": "default", "value": "None"},
+        {"name": "DD_APPSEC_SCA_ENABLED", "origin": "default", "value": "None"},
+        {"name": "DD_APPSEC_STACK_TRACE_ENABLED", "origin": "default", "value": "True"},
+        {"name": "DD_APPSEC_WAF_TIMEOUT", "origin": "default", "value": "5.0"},
+        {"name": "DD_CIVISIBILITY_AGENTLESS_ENABLED", "origin": "env_var", "value": "False"},
+        {"name": "DD_CIVISIBILITY_AGENTLESS_URL", "origin": "default", "value": ""},
```
mabdinur, line 319:
```suggestion
        {"name": "DD_CIVISIBILITY_AGENTLESS_URL", "origin": "default", "value": ""},
        {
            'name': "DD_CIVISIBILITY_EARLY_FLAKE_DETECTION_ENABLED",
            'origin': "default",
            'value': "True",
        },
```

avara1986 (approved):
As I mentioned, I think it would be better to use `envier` or `pydantic_settings` to access environment variables: 
https://github.com/DataDog/dd-trace-py/pull/10863/files#diff-0cb9315daf5ed5208de657467cc10a6b7dda616a161c9b5b1f8cff83c95894deR52

That way, all products follow the same approach. Other than that, LGTM!

rachelyangdog (approved):

P403n1x87 (approved):

ddtrace/settings/dynamic_instrumentation.py
```diff
@@ -131,3 +132,4 @@ class DynamicInstrumentationConfig(En):
 
 
 config = DynamicInstrumentationConfig()
+_report_telemetry(config)
```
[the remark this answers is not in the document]
mabdinur: Yup. This would be ideal. The main issue here is that most subclasses of `En` do not initialize the super class. So we will still need to make a code change in 10 different places 😞

ddtrace/auto.py
mabdinur: I am a fan of adding telemetry configurations where operations occur. It makes the code easier to read. This module will now trigger `ddtrace.bootstrap.sitecustomize` AND collect telemetry on how this operation was executed.

Instead we could put this logic in `ddtrace.bootstrap.sitecustomize`. We can check if `ddtrace.auto` is in sys.modules and then queue this configuration.

ddtrace/bootstrap/sitecustomize.py
```diff
@@ -158,7 +159,7 @@ def _(threading):
         else:
             log.debug("additional sitecustomize found in: %s", sys.path)
 
-    config._ddtrace_bootstrapped = True
+    telemetry_writer.add_configuration("ddtrace_bootstrapped", True, "unknown")
```
mabdinur, line 162:
```suggestion
    telemetry_writer.add_configuration("ddtrace_bootstrapped", True, "unknown")
    telemetry_writer.add_configuration("ddtrace_auto_used", "ddtrace.auto" in sys.modules, "unknown")
```

ddtrace/auto.py
mabdinur, line 25:
```suggestion
```

ddtrace/_trace/tracer.py
```diff
@@ -183,6 +183,10 @@ def _default_span_processors_factory(
     return span_processors, appsec_processor, deferred_processors
 
 
+class _TracerSingleton:
+    pass
+
+
```
mabdinur, line 189:
```suggestion
```

ddtrace/_trace/tracer.py
```diff
@@ -196,6 +200,7 @@ class Tracer(object):
     """
 
     SHUTDOWN_TIMEOUT = 5
+    _instance = None
```
mabdinur, line 203:
```suggestion
    _instance = None
```

mabdinur, line 203:
```suggestion
```

PR #10908, opened 2024-10-02
feat(llmobs): submit spans for streamed calls

sabrenner:
## What does this PR do?
Builds off of #10672, extending support for `llm.stream`, `chat_model.stream`, and `chain.stream` to LLM Observability. This PR tags those spans appropriately for LLM Observability, and marks them to submit to LLM Observability.

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -970,6 +968,9 @@ def _on_span_finished(span: Span, streamed_chunks):
         else:
             # best effort to join chunks together
             content = "".join([str(chunk) for chunk in streamed_chunks])
+        integration.llmobs_set_tags(span, args=args, kwargs=kwargs, response=content, operation="chain")
+        if span.error or not integration.is_pc_sampled_span(span):
```
lievan, line 972: Do we want to set llm obs tags before we do the `integration.is_pc_sampled_span`?

lievan, line 972: Another clarification question here about this - do we move the check for `span.error` after setting tags because the span may contain an error that happened mid-stream but we still want to capture the outputs that were returned?

sabrenner: good question. it looks like we recently changed `llmobs_set_tags` to run independently of `is_pc_sampled_span` ([example](https://github.com/DataDog/dd-trace-py/blob/d490d8ec05dffe06fabaeff98cd4b9ed0483f514/ddtrace/contrib/internal/langchain/patch.py#L250)). followed the same logic here, but i think even before we did something like `is_pc_sampled_llmobs`, which I think always returned `True`.

sabrenner: we re-check `span.error` in `llmobs_set_tags` for `llm`, `chat`, and `chain`, which i kept before adding output messages/value properties

lievan (approved):

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -1032,13 +1041,15 @@ def _on_span_finished(span: Span, streamed_chunks):
...
         provider=llm_provider,
+        model=model,
     )
 
 
 @with_traced_module
 def traced_llm_stream(langchain, pin, func, instance, args, kwargs):
     integration: LangChainIntegration = langchain._datadog_integration
     llm_provider = instance._llm_type
+    model = _extract_model_name(instance)
```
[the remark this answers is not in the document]
sabrenner: we were, but it (original streamed responses PR) never landed in a release so i felt it could just be included here instead of in a separate PR

ddtrace/contrib/internal/langchain/patch.py
```diff
@@ -1004,12 +1006,19 @@ def _on_span_started(span: Span):
                 span.set_tag_str("langchain.request.%s.parameters.%s.%s" % (llm_provider, param, k), str(v))
 
     def _on_span_finished(span: Span, streamed_chunks):
-        if span.error or not integration.is_pc_sampled_span(span):
+        joined_chunks = streamed_chunks[0]
+        for chunk in streamed_chunks[1:]:
+            joined_chunks += chunk  # base message types support __add__ for concatenation
```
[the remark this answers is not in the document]
sabrenner: i think it would involve typing it with `langchain_core`. while `langchain_core` _should_ be a defined module if this function is being traced, i always like to preface any use if `langchain_core`, etc. by checking if it exists first, which I'm not sure how to do with a type annotation. open to a suggestion though!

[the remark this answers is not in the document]
lievan: Hmm that makes sense, I think the comment here suffices then - thanks!

PR #10920, opened 2024-10-03
chore(ci_visibility): add Early Flake Detection retry logic to core API

romainkomorndatadog:
Introduces EFD retry logic to the `CIVisibility` service, as well as the internal APIs.

The functionality added covers:
- fetching unique tests from the Datadog API
- identifying tests as new and computing "new test" percentage for the purpose of faulty session detection
- checking whether a test should be retried
- new span tags and telemetry

Some notable changes on top of the additions above:
- EFD retries no longer get their own `TestId`s, and are instead tracked with on their original test's object and referred to by number
- Snapshot tests that were exercising the `TestId`s with retry numbers have been updated since the tests were removed
- `ITR_SKIPPING_LEVEL` is moved to the external API to reduce circular dependencies between the internal and external API

Other notes on likely future changes left out of scope for now:
- `InternalTestId` (which only existed for the purpose of the retry numbers) is no longer necessary
- "Unique tests" refers to an internal name for the endpoint, and "known tests" may be a more sensible name in this context

There is no release note since this is just an internal feature to-be-leveraged by `pytest` and `unittest` in future changes.

## Checklist

## Reviewer Checklist
- Title is accurate


tests/ci_visibility/api/README.md
```diff
@@ -0,0 +1,34 @@
...
+   1. Set expected environment variables (eg: `DD_API_KEY` and `DD_CIVISIBILIKTY_AGENTLESS_ENABLED`)
+   1. Run the script, eg: `python tests/ci_visibility/api/fake_runner_all_pass.py`
+
+#### As tests
+
+1. Choose a `riot` environment (eg: using `riot list ci_visibility`)
+1. Make sure the `testagent` is running (refer to contributor docs again)
+1. Run the test(s) (note: you may want to pass `-s` `riot run` to speed up tests)
+   1. All tests: `riot -v run 1b90fc9 -- -k FakeApiRunnersSnapshotTestCase`
+   1. Individual test: `riot -v run 1b90fc9 -- -k `
```
romainkomorndatadog, line 34:
```suggestion
   1. Individual test: `riot -v run 1b90fc9 -- -k test_manual_api_fake_runner_mix_fail_itr_test_level` 
```

ddtrace/internal/ci_visibility/api/_session.py
```diff
@@ -78,10 +92,45 @@ def _telemetry_record_event_finished(self):
             test_framework=self._session_settings.test_framework_metric_name,
             has_codeowners=self._codeowners is not None,
             is_unsupported_ci=self._session_settings.is_unsupported_ci,
+            early_flake_detection_abort_reason="faulty" if self.efd_is_faulty_session() else None,
```
vitor-de-araujo, line 95: We have `self._efd_is_faulty_session` and `self.efd_is_faulty_session()`. Above the first one was used, here the second one was. Should they be the same?

The method computes the property, right? Maybe we should make the method return the property value immediately if it's already set, and use the method everywhere.

ddtrace/internal/ci_visibility/api/_test.py
```diff
@@ -128,29 +153,122 @@ def finish_itr_skipped(self) -> None:
...
-                source_file_info=self._source_file_info,
-                initial_tags=self._tags,
-                is_early_flake_retry=True,
-            ),
+    def make_early_flake_retry_from_test(self, retry_number: int) -> "TestVisibilityTest":
+        if self._parameters is not None:
+            raise ValueError("Cannot create an early flake retry from a test with parameters")
+        retry_test = self.__class__(
+            self.name,
+            self._session_settings,
```
vitor-de-araujo, line 161: Is `parameters` missing intentionally?

I wonder if we can do better than just remembering to keep the field list in sync with this call.

ddtrace/internal/ci_visibility/api/_test.py
```diff
@@ -128,29 +153,122 @@ def finish_itr_skipped(self) -> None:
...
+            return False
+
+        if not self.is_new():
+            return False
+
+        if self._efd_initial_finish_time_ns is None:
+            log.debug("Early Flake Detection: efd_should_retry called but test has initial result")
+            return False
+
+        initial_duration = (self._efd_initial_finish_time_ns - self._span.start_ns) / 1e9
```
vitor-de-araujo, line 203: I think every duration variable should have the unit in the name (e.g., `initial_duration_seconds`), but then again the whole repo is like this. If we are consistent to use units in everything that _isn't_ seconds, that might be good enough, but I don't know if we currently are.

ddtrace/internal/ci_visibility/api/_test.py
```diff
@@ -128,29 +153,122 @@ def finish_itr_skipped(self) -> None:
...
+        if initial_duration <= 5:
+            return num_retries < efd_settings.slow_test_retries_5s
+        if initial_duration <= 10:
+            return num_retries < efd_settings.slow_test_retries_10s
+        if initial_duration <= 30:
+            return num_retries < efd_settings.slow_test_retries_30s
+        if initial_duration <= 300:
+            return num_retries < efd_settings.slow_test_retries_5m
+
+        self._efd_abort_reason = "slow"
```
vitor-de-araujo, line 216: Maybe create a (cursed) enum for EFD abort reasons.

ddtrace/internal/ci_visibility/recorder.py
```diff
@@ -471,8 +489,6 @@ def enable(cls, tracer=None, config=None, service=None):
             log.debug("%s already enabled", cls.__name__)
             return
 
-        _register_session_handlers()
```
vitor-de-araujo, line 474: This was already unnecessary before, since it's called at the toplevel, right?

tests/ci_visibility/api/README.md
```diff
@@ -0,0 +1,34 @@
...
+Refer to the `ddtrace` contributor documentation for how to update snapshot
+tests.
+
+### Running fake runners
+
+#### Manually
+ 
+1. Set up (and activate) an environment (eg: using `pip install ddtrace` or `riot shell`)
+1. Run the script:
+   1. Set expected environment variables (eg: `DD_API_KEY` and `DD_CIVISIBILIKTY_AGENTLESS_ENABLED`)
```
vitor-de-araujo, line 25: :ghost: 
```suggestion
   1. Set expected environment variables (eg: `DD_API_KEY` and `DD_CIVISIBILITY_AGENTLESS_ENABLED`)
```

vitor-de-araujo (approved):

ddtrace/internal/ci_visibility/recorder.py
```diff
@@ -429,6 +438,16 @@ def _fetch_tests_to_skip(self):
         except Exception:  # noqa: E722
             log.debug("Error fetching skippable items", exc_info=True)
 
+    def _fetch_unique_tests(self) -> Optional[Set[InternalTestId]]:
+        try:
+            if self._api_client is not None:
+                return self._api_client.fetch_unique_tests()
+            else:
+                log.warning("API client not initialized, cannot fetch unique tests")
```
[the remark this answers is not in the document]
vitor-de-araujo: Another rule that we might choose to kill with fire

ddtrace/internal/ci_visibility/api/_session.py
romainkomorndatadog: Yep. There's only one other place where we were looking at the attribute directly, so I updated that in e796da1000be1acf2980de8a3000061553bb1092 .

ddtrace/internal/ci_visibility/api/_test.py
```diff
@@ -54,28 +59,45 @@ def __init__(
...
 
-        # Currently unsupported
+        self._is_new = is_new
+
+        self._is_efd_retry = is_efd_retry
+        self._efd_retries: List[TestVisibilityTest] = []
+        self._efd_abort_reason: Optional[str] = None
+        self._efd_initial_finish_time_ns: Optional[int] = None
+
+        # Currently unsupported]
```
vitor-de-araujo, line 75: Oh, I missed this one the first time :P 
```suggestion
        # Currently unsupported
```

PR #10942, opened 2024-10-04
chore(di): trigger probes

P403n1x87:
We implement trigger probes. These allows triggering the capture of debug information along a trace, ensuring all the relevant probes are also triggered.

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/debugging/_session.py
```diff
@@ -0,0 +1,34 @@
...
+    def get_for_trace(cls) -> t.Optional["Session"]:
+        return SessionManager.get_session_for_trace()
+
+
+class SessionManager:
+    _session_trace_map: wkdict = wkdict()  # Trace context to Session mapping
+
+    @classmethod
+    def link_session_to_trace(cls, session, trace_context: t.Optional[t.Any] = None) -> None:
+        cls._session_trace_map[trace_context or tracer.current_root_span()] = session
```
shatzi, line 26: passing trace_context can break the get_session_for_trace as it only works for the root_spans. why do we want to let user specify this?

ddtrace/debugging/_session.py
```diff
@@ -0,0 +1,34 @@
...
+
+    @classmethod
+    def get_session_for_trace(cls) -> t.Optional[Session]:
```
shatzi, line 29: a trace can have multiple sessions assoiated to a trace.
we might want to have different API:
get_session_for_trace(cls, session_id) -> return a specific session (used by manual probes)
should_record_code_origin_snapshot() -> return true if one session has level > 2

ddtrace/debugging/_live.py
```diff
@@ -0,0 +1,22 @@
...
+from ddtrace.debugging._session import Session
+from ddtrace.internal import core
+
+
+def handle_distributed_context(context: t.Any) -> None:
+    debug_tag = context._meta.get("_dd.p.debug")
+    if debug_tag is None:
+        return
+
+    ident, _, level = debug_tag.partition(":")
```
shatzi, line 12: RFC require support for multiple sessions (separated by ",")

shatzi (commented):
very cool progress, thank you

ddtrace/contrib/trace_utils.py
```diff
@@ -574,6 +575,8 @@ def activate_distributed_headers(tracer, int_config=None, request_headers=None,
         # have a context with the same trace id active
         tracer.context_provider.activate(context)
 
+        dispatch("distributed_context.activated", (context,))
```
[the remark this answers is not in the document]
P403n1x87: That's what the core API expects, a tuple object 🙁

ddtrace/debugging/_origin/span.py
```diff
@@ -246,17 +251,29 @@ def enable(cls):
...
+        # Register the entrypoint wrapping for entry spans
+        core.on("service_entrypoint.patch", partial(wrap_entrypoint, cls.__uploader__.get_collector()))
+
     @classmethod
     def disable(cls):
         if cls._instance is None:
             return
 
+        # Unregister the entrypoint wrapping for entry spans
+        core.reset_listeners("service_entrypoint.patch", wrap_entrypoint)
```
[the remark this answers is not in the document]
P403n1x87: ah good catch!

ddtrace/debugging/_session.py
```diff
@@ -0,0 +1,34 @@
+from dataclasses import dataclass
+import typing as t
+from weakref import WeakKeyDictionary as wkdict
+
+from ddtrace import tracer
+
+
+@dataclass
+class Session:
+    ident: str
```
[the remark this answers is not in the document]
P403n1x87: That's pretty common in Python: https://docs.python.org/3/library/threading.html#threading.Thread.ident

ddtrace/debugging/_session.py
P403n1x87: Unfortunately the tracer propagates headers by activating a context object, which is not retrievable with the current tracer API (i.e. the root span of a distributed trace does not carry the distributed header information, which is attached to the context object that is used as a parent for the local root)

ddtrace/debugging/_origin/span.py
```diff
@@ -152,36 +154,39 @@ def __enter__(self):
...
-        #     # Create a snapshot
-        #     snapshot = Snapshot(
-        #         probe=location.probe,
-        #         frame=self.__frame__,
-        #         thread=current_thread(),
-        #         trace_context=root,
-        #     )
+        # Check if we have any level 2 debugging sessions running for the
+        # current trace
+        if any(s.level >= 2 for s in Session.from_trace()):
```
shatzi, line 159: I think this check should only happen on exit. as the trigger probe can mark the trace as captured in the middle of the span.

ddtrace/debugging/_probe/model.py
```diff
@@ -26,6 +27,7 @@
 
 DEFAULT_PROBE_RATE = 5000.0
 DEFAULT_SNAPSHOT_PROBE_RATE = 1.0
+DEFAULT_TRIGGER_PROBE_RATE = 1.0
```
shatzi, line 30: are we good with trace per second?

ddtrace/debugging/_session.py
```diff
@@ -0,0 +1,104 @@
...
+from ddtrace import tracer
+
+
+SessionId = str
+
+
+def _sessions_from_debug_tag(debug_tag: str) -> t.Generator["Session", None, None]:
+    for session in debug_tag.split("."):
+        ident, _, level = session.partition(":")
+        yield Session(ident=ident, level=int(level or 0))
```
shatzi, line 14: maybe we want default level to be different. level zero is same as not passing the session at all.

ddtrace/debugging/_signal/model.py
```diff
@@ -121,12 +122,17 @@ def _rate_limit_exceeded(self) -> bool:
...
+        exceeded = self.session is None and probe.limiter.limit() is RateLimitExceeded
         if exceeded:
             self.state = SignalState.SKIP_RATE
 
         return exceeded
 
+    @property
+    def session(self):
+        session_id = self.probe.tags.get("sessionId")
+        return Session.lookup(session_id) if session_id is not None else None
```
shatzi, line 134: if the session doesn't exists - the probe should not capture at all.

two points:
1. I think this code just make this manual probe as a normal probe if there is no sessions. 
2. if session level is 0 - it should also not capture.

ddtrace/debugging/_signal/trigger.py
```diff
@@ -0,0 +1,64 @@
...
+        # DEV: Unfortunately we don't have an API for this :(
+        self.trace_context._meta[f"_dd.ld.probe_id.{self.probe.probe_id}"] = "true"  # type: ignore[union-attr]
+
+    def enter(self, scope: t.Mapping[str, t.Any]) -> None:
+        self._link_session()
+
+    def exit(self, retval: t.Any, exc_info: ExcInfoType, duration: float, scope: t.Mapping[str, t.Any]) -> None:
+        session = self.session
+        if session is not None:
+            session.unlink_from_trace(self.trace_context)
```
shatzi, line 44: this is interesting...  we can record only during that method call. which is cool.

why not do this: 
1. it would hid the _pp.d.debug values set while the span is active.
2. we can't use trigger probe on method at some internal business logic method. 
3. it act differently from line-probes.

I would vote to remove this for now and maybe add this logic with some extra configuration

ddtrace/settings/code_origin.py
```diff
@@ -20,7 +20,7 @@ class SpanCodeOriginConfig(En):
         enabled = En.v(
             bool,
             "enabled",
-            default=False,
+            default=True,
```
shatzi, line 23: we are auto enable code origin for python?

tests/debugging/test_session.py
```diff
@@ -0,0 +1,11 @@
...
+from ddtrace.debugging._session import _sessions_to_debug_tag
+
+
+def test_tag_identity():
+    """
+    Test that the combination of _sessions_from_debug_tag and
+    _sessions_to_debug_tag is the identity function.
+    """
+    debug_tag = "session1:1,session2:2"
+    assert _sessions_to_debug_tag(_sessions_from_debug_tag(debug_tag)) == debug_tag
```
shatzi, line 11: test default level

tests/debugging/test_session.py
```diff
@@ -0,0 +1,11 @@
+from ddtrace.debugging._session import _sessions_from_debug_tag
+from ddtrace.debugging._session import _sessions_to_debug_tag
+
+
+def test_tag_identity():
+    """
+    Test that the combination of _sessions_from_debug_tag and
+    _sessions_to_debug_tag is the identity function.
+    """
+    debug_tag = "session1:1,session2:2"
```
shatzi, line 10: should be "." ?

I think this is a broken test

shatzi (approved):
Thank you!

emmettbutler (approved):

shatzi (approved):
I think the only thing that left is probe budget for probe with session

ddtrace/debugging/_origin/span.py
```diff
@@ -172,20 +181,11 @@ def __enter__(self):
...
-
-        return self
-
-    def _close_signal(self, retval=None, exc_info=(None, None, None)):
-        try:
-            snapshot: Snapshot = t.cast(Snapshot, self.get("snapshot"))
-        except KeyError:
-            # No snapshot was created
-            return
+            self.set("start_time", monotonic_ns())
```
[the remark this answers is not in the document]
P403n1x87: ah of course 🤦

ddtrace/debugging/_probe/model.py
```diff
@@ -247,6 +246,10 @@ class LogProbeMixin(AbstractProbeMixIn):
     take_snapshot: bool
     limits: CaptureLimits = field(compare=False)
 
+    @property
```
shatzi, line 249: I love this, as down the line we can provide the user to set the a probe limit on the UI

ZStriker19 (approved):

PR #10949, opened 2024-10-04
fix(botocore): bedrock cross-region inference model name does not throw

### LINKED ISSUE #10772: ValueError in boto3 Bedrock Patch for invoke_model with Cross-Region Inference Model IDs
### Summary of problem

The boto3 patch logic is broken for bedrock when using `invoke_model` with [cross-region inference](https://docs.aws.amazon.com/bedrock/latest/userguide/cross-region-inference-support.html). To use an inference profile, a region name is prepended to the modelId e.g.
- with cross-region inference: `us.anthropic.claude-3-5-sonnet-20240620-v1:0`
- without cross-region inference: `anthropic.claude-3-5-sonnet-20240620-v1:0`

and this seems to break part of the code that splits the model id into the provider and model name, but this works with boto3.

```python
File /python3.11/site-packages/ddtrace/contrib/internal/botocore/services/bedrock.py:321, in patched_bedrock_api_call(original_func, instance, args, kwargs, function_vars)
    319 params = function_vars.get("params")
    320 pin = function_vars.get("pin")
--> 321 model_provider, model_name = params.get("modelId").split(".")
    322 integration = function_vars.get("integration")
    323 submit_to_llmobs = integration.llmobs_enabled and "embed" not in model_name

ValueError: too many values to unpack (expected 2)
```

`2.12.2`

`24.2`

```
 name         : boto3                  
 version      : 1.34.34
 description  : The AWS SDK for Python
```

Execute the `bedrock-runtime` boto3 client `invoke_model` with a model id that uses the cross-region inference model id structure like: `us.anthropic.claude-3-5-sonnet-20240620-v1:0`.

It throws a `ValueError` exception and breaks our application code

It should not break our application code and should successfully make the request

sabrenner:
## What does this PR do?
Fixes #10772 

Model IDs with cross-region inference would throw because we assumed `modelID` would only have the `provider` and `model_name`, when it could have the region as well. This would result in:

```
File /python3.11/site-packages/ddtrace/contrib/internal/botocore/services/bedrock.py:321, in patched_bedrock_api_call(original_func, instance, args, kwargs, function_vars)
    319 params = function_vars.get("params")
    320 pin = function_vars.get("pin")
--> 321 model_provider, model_name = params.get("modelId").split(".")
    322 integration = function_vars.get("integration")
    323 submit_to_llmobs = integration.llmobs_enabled and "embed" not in model_name

ValueError: too many values to unpack (expected 2)
```

We are not tagging that cross-region inference in this PR, just resolving the error.

### Testing
To test this change, a singular `anthropic_message` test case and corresponding cassette was modified. This had an unfortunate side effect of affecting an LLMObs test as well. I can add a different test instead, but I believe that would require an additional 60+ line cassette which isn't totally needed, so I opted for this instead.

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/contrib/internal/botocore/services/bedrock.py
```diff
@@ -318,7 +318,11 @@ def handle_bedrock_response(
 def patched_bedrock_api_call(original_func, instance, args, kwargs, function_vars):
     params = function_vars.get("params")
     pin = function_vars.get("pin")
-    model_provider, model_name = params.get("modelId").split(".")
+    model_meta = params.get("modelId").split(".")
+    if len(model_meta) == 2:
+        model_provider, model_name = model_meta
+    else:
+        _, model_provider, model_name = model_meta  # cross-region inference
```
lievan, line 325: Approving to unblock the fix but should we do something like
```
if len(model_meta) == 2:

elif len(model_meta) == 3

else:
   log.warning('failed to parse model name')
   model_name = placeholder
   model_provider = placeholder
```
in a follow up PR?

lievan (approved):

ddtrace/contrib/internal/botocore/services/bedrock.py
sabrenner: ah hmm yeah, good question. i'm not sure what the standard is in general for un-parsable things we want to tag (not sure how common placeholders are, and if they serve any value), and i also don't think logging from an integration itself is standard. we'll follow-up on this next week 😄

sabrenner: and for this particular case, i think it can only be `provider.model` or `region.provider.model` ([ref](https://docs.aws.amazon.com/bedrock/latest/userguide/model-ids.html), [ref](https://docs.aws.amazon.com/bedrock/latest/userguide/cross-region-inference-support.html))

sabrenner: resolving for now, will follow up this week

PR #10987, opened 2024-10-09
chore(telemetry): update config names and values

mabdinur:
- Removes the name argument from `_ConfigItem.__init__(..)`. The name of a configuration will be set to the first environment variable in the envs field (ie the primary environment variable).
- Avoids sending arbitrary configuration names to instrumentation telemetry. When possible, the configuration name should match a well documented name (ex: a public environment variable).
- Avoids reporting integers, floats, booleans and None types as strings in instrumentation telemetry payloads

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/settings/config.py
```diff
@@ -198,10 +198,10 @@ def get_error_ranges(error_range_str):
 class _ConfigItem:
     """Configuration item that tracks the value of a setting, and where it came from."""
 
-    def __init__(self, name, default, envs):
+    def __init__(self, default, envs):
         # type: (str, Union[_JSONType, Callable[[], _JSONType]], List[Tuple[str, Callable[[str], Any]]]) -> None
```
mabdinur, line 202:
```suggestion
        # type: (Union[_JSONType, Callable[[], _JSONType]], List[Tuple[str, Callable[[str], Any]]]) -> None
```

ddtrace/internal/telemetry/writer.py
```diff
@@ -465,7 +465,11 @@ def remove_configuration(self, configuration_name):
     def add_configuration(self, configuration_name, configuration_value, origin="unknown"):
         # type: (str, Any, str) -> None
         """Creates and queues the name, origin, value of a configuration"""
-        if not isinstance(configuration_value, (bool, str, int, float, type(None))):
+        if isinstance(configuration_value, dict):
+            configuration_value = ",".join(":".join((k, str(v))) for k, v in configuration_value.items())
+        elif isinstance(configuration_value, (list, tuple)):
+            configuration_value = ",".join(str(v) for v in configuration_value)
+        elif not isinstance(configuration_value, (bool, str, int, float, type(None))):
```
erikayasuda, line 472: Dumb question: Why are there so many ways that configs can be represented?

ddtrace/settings/config.py
```diff
@@ -198,10 +198,10 @@ def get_error_ranges(error_range_str):
 class _ConfigItem:
     """Configuration item that tracks the value of a setting, and where it came from."""
 
-    def __init__(self, name, default, envs):
-        # type: (str, Union[_JSONType, Callable[[], _JSONType]], List[Tuple[str, Callable[[str], Any]]]) -> None
+    def __init__(self, default, envs):
+        # type: (Union[_JSONType, Callable[[], _JSONType]], List[Tuple[str, Callable[[str], Any]]]) -> None
         # _ConfigItem._name is only used in __repr__ and instrumentation telemetry
-        self._name = name
+        self._name = envs[0][0]
```
erikayasuda, line 204: >  The name of a configuration will be set to the first environment variable in the envs field (ie the primary environment variable).

Is this always guaranteed? Wouldn't it be better to always just have it passed in directly like before?

ddtrace/settings/config.py
```diff
@@ -248,16 +248,6 @@ def source(self) -> _ConfigSource:
...
 
-    def _telemetry_value(self) -> str:
-        val = self.value()
-        if val is None:
-            return ""
-        elif isinstance(val, (bool, int, float)):
-            return str(val).lower()
-        elif isinstance(val, dict):
-            return ",".join(":".join((k, str(v))) for k, v in val.items())
-        return str(val)
```
erikayasuda, line 259: So I saw:
> Avoids reporting integers, floats, booleans and None types as strings in instrumentation telemetry payloads

Not against this change, but was there a reason we did this initially instead of using the original types? Did something change to prompt this?

ddtrace/internal/telemetry/writer.py
mabdinur: Telemetry payloads are serialized as json,  json supports seralizing the following python types: `None, int, float, str, bool, List[_JSONType], Dict[str, _JSONType]`. `configuration_value` is not one of these types (ex: a set) then we will fail to generate and send telemetry events.

erikayasuda (approved):

ddtrace/settings/config.py
mabdinur: Yup but this felt redundant. Also this could result in some configurations having names that are not equal to the environment variable. In my opinion all configurations should always be named after their primary environment variable. This will improve consistency across languages. We should avoid arbitrary names as much as possible.

ddtrace/settings/config.py
mabdinur: Java currently reports Null values as empty strings and stringify booleans (current behavior) while golang/dotnet/nodejs/ruby report nil values and booleans in their expected json representation. Here I am aligning with how the majority of languages handle telemetry values. 

I have no idea why Java was implemented in this way.

PR #11012, opened 2024-10-14
chore(core): core update for span management

christophe-papazian:
Simplify the core "call_key" mechanism with a simple "span" attribute for the core context (based on a new _innerspan attribute).

- `ctx.set_item(ctx.get_item("call_key"), span)` become `ctx.span = span`
- `span = ctx.get_item(ctx["call_key"])` become `span = ctx.span`

Add a "span_key" mechanism to automatically add the span to the context data for simple access in case we need to access a span by name on a non current context.

* Technical Notes *
I wanted to ensure that the inner span for a context is only set once. This is not the case, so I removed the condition. In the previous version, different spans could be assigned to the call_key during context life. As the logic was preserved in this PR, this is also the case with ctx.span.

APPSEC-55053

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/internal/core/__init__.py
```diff
@@ -273,10 +273,28 @@ def root(self):
...
+        if self._inner_span is None:
+            raise ValueError("No span set on ExecutionContext")
+        return self._inner_span
+
+    @span.setter
+    def span(self, value: "Span") -> None:
+        # if self._inner_span is not None:
+        #     if self._inner_span is not value:
+        #         raise ValueError(f"Cannot overwrite ExecutionContext span {self._inner_span.name} with {value.name}")
+        #     return
```
gnufede, line 287:
```suggestion
```

gnufede (approved):

emmettbutler (approved):

ddtrace/internal/core/__init__.py
```diff
@@ -273,10 +273,24 @@ def root(self):
             current = current.parent
         return current
 
+    @property
+    def span(self) -> "Span":
+        if self._inner_span is None:
+            raise ValueError("No span set on ExecutionContext")
```
ZStriker19, line 279: Are we catching this anywhere? I think we want to avoid raising errors all the way up if possible right?

ZStriker19: Oh in get_span, got it.

ZStriker19 (approved):

PR #11024, opened 2024-10-14
fix(profiling): stack v2 properly tracks asyncio tasks

taegyunkim:
## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/profiling/_asyncio.py
```diff
@@ -1,12 +1,15 @@
 # -*- encoding: utf-8 -*-
 from functools import partial
 import sys
+import threading
```
P403n1x87, line 4: I think we're fine here because this module would be loaded early at boot, but just a note that with module cloning we need to make sure that this is a reference to _our_ copy of `threading`, and not the one that the application will end up using.

P403n1x87 (approved):

sanchda (approved):

ddtrace/profiling/_asyncio.py
taegyunkim: Is there a way to refer to our copy of `threading`?

sanchda: ddtrace/profiling/_threading.pyx has a good example, I think.  It loads the module defined by ddtrace/internal/_unpatched.py to pick up on the one "we" import if possible (line 53 of _threading.pyx)

taegyunkim: Oh, right. I was looking for that but couldn't find it. Thanks for the pointer.

taegyunkim: Just ended up doing `from ddtrace.internal._unpatched import _threading as ddtrace_threading`, please see the updated code.

sanchda: Yeah, that looks good to me.

PR #11031, opened 2024-10-15
chore(iast): remove DD_IAST_ENABLED env var usage

gnufede:
Code Security: Remove internal usage of `DD_IAST_ENABLED` env var using the `asm_config` value or `_is_iast_enabled()` instead.

Also, since the Python runtime isn't expected to change during execution, we can cache the result of `_is_python_version_supported()`

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/appsec/_iast/__init__.py
```diff
@@ -73,15 +70,12 @@ def ddtrace_iast_flask_patch():
...
+        from ddtrace.appsec._iast._ast.ast_patching import _should_iast_patch
+        from ddtrace.appsec._iast._loader import _exec_iast_patched_module
 
-        if _is_python_version_supported():
-            from ddtrace.appsec._iast._ast.ast_patching import _should_iast_patch
-            from ddtrace.appsec._iast._loader import _exec_iast_patched_module
-
-            log.debug("IAST enabled")
-            ModuleWatchdog.register_pre_exec_module_hook(_should_iast_patch, _exec_iast_patched_module)
+        log.debug("IAST enabled")
```
avara1986, line 77: why do you remove `if _is_python_version_supported():`?

gnufede: Because `_is_iast_enabled()` already checks it

avara1986 (approved):

.circleci/config.templ.yml
```diff
@@ -510,14 +510,9 @@ jobs:
     <<: *machine_executor
     parallelism: 5
     steps:
-      - when:
-          condition:
-            matches: { pattern: "main", value: << pipeline.git.branch >> }
```
erikayasuda, line 515: Should we still scope when these tests run? Maybe `main` + any release branches?

ddtrace/bootstrap/preload.py
```diff
@@ -6,6 +6,7 @@
 import os  # noqa:I001
 
 from ddtrace import config  # noqa:F401
+from ddtrace.appsec._iast._utils import _is_iast_enabled
```
erikayasuda, line 9: I saw in a different file there's a DEV comment saying:
> DEV: These imports are here to avoid _ast.ast_patching import in the top level because they are slow and affect serverless startup time

I know this is a different import, but is importing this and using `_is_iast_enabled` faster than just looking up the env var directly here? If it is please disregard this comment, just wasn't sure 🫣

erikayasuda (approved):

.circleci/config.templ.yml
gnufede: Sorry, this was just a test to verify those tests in the CI before merging my PR, I'll revert this change now

ddtrace/bootstrap/preload.py
gnufede: Great question! The goal of this PR is to allow at some point to use `configure` to enable IAST, regardless of env vars.
In any case, I just timed them to be sure:
```
$ ddtrace-run -d python -m timeit "import os; os.getenv('DD_IAST_ENABLED', False)"
500000 loops, best of 5: 574 nsec per loop
```
vs
```
$ ddtrace-run -d python -m timeit "from ddtrace.appsec._iast._utils import _is_iast_enabled; _is_iast_enabled()"
500000 loops, best of 5: 466 nsec per loop
```

PR #11095, opened 2024-10-19
refactor: use telemetry in product protocol

P403n1x87:
We ensure that every service that is enabled/disabled via the new product plugin interface sends telemetry event update automatically. It won't be necessary for each product using the interface to manually call into the telemetry service.

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/internal/products.py
```diff
@@ -119,6 +120,7 @@ def start_products(self) -> None:
             try:
                 product.start()
                 log.debug("Started product '%s'", name)
+                telemetry_writer.product_activated(name, True)
```
P403n1x87, line 123: Can we pass arbitrary strings?

mabdinur: This is the list of supported product names: https://github.com/DataDog/instrumentation-telemetry-api-docs/blob/main/GeneratedDocumentation/ApiDocs/v2/SchemaDocumentation/Schemas/products.md. 

But I am not sure if unsupported names are stored or dropped by intake.

P403n1x87: Can we find out what happens? This way we would be reporting all services that are activated. If the agent simply drops the services it doesn't know about that would be fine

mabdinur: The unsupported product names are ignored. [Here](https://github.com/DataDog/dd-go/blob/e931cb2967813bb038a0de0344784064e1161013/trace/apps/tracer-telemetry-intake/telemetry-payload/app_started.go#L57) we deserialize the json payload into a struct. If the name of the product is not defined in the struct it will be ignored.

mabdinur (approved):

emmettbutler (approved):

shatzi (approved):

PR #11108, opened 2024-10-21
chore(profiling): run native tests in ci

taegyunkim:
1. Use `pytest-cpp` to discover and run profiling C++ tests 
	a. `pytest-cpp` will find executables that start with `test_`
	b. all tests are put into `ddtrace/internal/datadog/profiling/test` and `tests/profiling_v2/native_test` will symlink to it for discovery. 
3. `DD_PROFILING_NATIVE_TESTS` flag is used to build native tests
	a. needed to set this in `build_base_venv`
	b. setup.py includes native test binaries when the flag is set
	c. gitlab will also propagate those artifacts

## Checklist

## Reviewer Checklist
- Title is accurate


setup.py
```diff
@@ -53,6 +53,8 @@
 CRASHTRACKER_DIR = HERE / "ddtrace" / "internal" / "datadog" / "profiling" / "crashtracker"
 STACK_V2_DIR = HERE / "ddtrace" / "internal" / "datadog" / "profiling" / "stack_v2"
 
+BUILD_PROFILING_NATIVE_TESTS = os.getenv("DD_PROFILING_NATIVE_TESTS", "0").lower() in ("1", "yes", "on", "true")
```
sanchda, line 56: I think this is fine for now, but we are in a weird situation where we want to make sure the same assets that are tested are the ones we distribute.  That would mean optional flags like this become necessary, and we need to decide on how we structure things.

That said, apart from running tests in a new job, with a new runner, and new configuration, I think your solution here is the only other option.

Just flagging this for awareness, I don't think there's anything to be done in this PR.

sanchda (approved):
Generally speaking this looks good and I really like how it works within the framework of our existing tests.

I left some comments about the use of `RPATH`, since this is loader-facing change which will affect how resolution occurs in customer environments.  In practice I don't think this actually matters, so I'm not blocking the PR based on this feedback, but the general guidance is to minimize dependency resolution at dynamic load-time.

ddtrace/internal/datadog/profiling/dd_wrapper/test/CMakeLists.txt
```diff
@@ -22,11 +22,18 @@ function(dd_wrapper_add_test name)
     add_ddup_config(${name})
 
     gtest_discover_tests(${name})
+
+    set_target_properties(${name} PROPERTIES INSTALL_RPATH "$ORIGIN/..")
```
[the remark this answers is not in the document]
taegyunkim: Without this line, the test binaries can't find libdd_wrapper.so, that's why I needed to add this one.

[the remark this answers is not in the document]
sanchda: 🤦 sorry, I didn't realize this was the `test` directory.  I deserve ridicule.

ddtrace/internal/datadog/profiling/stack_v2/CMakeLists.txt
```diff
@@ -96,16 +96,22 @@ endif()
...
 target_link_libraries(${EXTENSION_NAME} PRIVATE dd_wrapper Threads::Threads)
 
+# to discover dd_wrapper at runtime
+set(RPATH_DIRS "$ORIGIN/..")
+
 if(Python3_LIBRARIES)
     target_link_libraries(${EXTENSION_NAME} PRIVATE ${Python3_LIBRARIES})
+
+    # to discover Python at runtime
+    set(RPATH_DIRS "${RPATH_DIRS}:${Python3_RUNTIME_LIBRARY_DIRS}")
```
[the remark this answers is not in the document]
taegyunkim: It turns out that this isn't really necessary to run the tests using pytest. Will just revert this file.

ddtrace/contrib/pytest/_plugin_v1.py
```diff
@@ -582,6 +594,10 @@ def pytest_runtest_protocol(item, nextitem):
         yield
         return
 
+    if PYTEST_CPP_ENABLED and isinstance(item, pytest_cpp.plugin.CppItem):
```
romainkomorndatadog, line 597: I don't love the idea of modifying our `pytest` integration just so we can use a plugin in our CI.

Why do `pytest_cpp` tests need to be ignored?

taegyunkim: `pytest_cpp.plugin.CppItem` objects don't have member attributes as expected in our `pytest` integration, and I had to add bunch of lines checking for `None` in the code to avoid failures like 
https://gitlab.ddbuild.io/DataDog/apm-reliability/dd-trace-py/-/jobs/699959248

The changes looked like [`e2f518a` (#11108)](https://github.com/DataDog/dd-trace-py/pull/11108/commits/e2f518aaed11295955ad694dac9d6ecbf54999cd)

I wanted to minimize the diff for this PR and reverted above changes and added this line.

romainkomorndatadog: At first glance, I think most of those will be fixed with the next version of the `pytest` integration (which you can enable with `_DD_CIVISIBILITY_USE_PYTEST_V2=1` in your environment).

I'm actually looking to make the next version of the plugin be the default in `dd-trace-py` for dogfooding reasons, so it'd be a good test.

taegyunkim: Good to know and will try that, thanks!

taegyunkim: It works, thank you.

PR #11153, opened 2024-10-24
refactor(di): encapsulate common signal code

P403n1x87:
We encapsulate common signal boilerplate code to simplify the coding of signals and ensure homogeneous behaviour among them.

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/debugging/_expressions.py
```diff
@@ -375,24 +376,24 @@ class DDExpression:
     __compiler__ = dd_compile
 
     dsl: str
-    callable: Callable[[Dict[str, Any]], Any]
+    callable: Callable[[Mapping[str, Any]], Any]
 
-    def eval(self, _locals):
+    def eval(self, scope: Mapping[str, Any]) -> Any:
```
shatzi, line 381: this is nice renaming.

ddtrace/debugging/_exception/replay.py
```diff
@@ -188,7 +190,9 @@ def on_span_exception(
                         )
 
                         # Capture
-                        snapshot.line()
+                        snapshot.line(ChainMap(frame.f_locals, frame.f_globals))
```
shatzi, line 193: we call this ChainMap multiple locations - I think it better to have an util method e.g; createScopeFromFrame(frame) -> to avoid problems down the line if we decide to limit or remove globals...

or maybe line() should just get the frame object instead.

ddtrace/debugging/_probe/model.py
```diff
@@ -156,22 +156,26 @@ def location(self):
         return (maybe_stringify(self.resolved_source_file), self.line)
 
 
-class ProbeEvaluateTimingForMethod(str, Enum):
+class ProbeEvalTiming(str, Enum):
     DEFAULT = "DEFAULT"
-    ENTER = "ENTER"
```
shatzi, line 161: scary...

ddtrace/debugging/_signal/context.py
```diff
@@ -0,0 +1,95 @@
...
+
+    - If the probe has no timing, it defaults to the ENTRY timing.
+    """
+
+    def __init__(self, signal: Signal) -> None:
+        self.signal = signal
+        self._timing = ProbeEvalTiming.ENTRY
+
+        probe = signal.probe
+        if isinstance(probe, TimingMixin):
```
shatzi, line 36: nit: would easier to read if there is else under this if with `self._timing = ProbeEvalTiming.ENTRY`

ddtrace/debugging/_signal/context.py
```diff
@@ -0,0 +1,95 @@
...
+    def enter(self) -> None:
+        signal = self.signal
+        probe = signal.probe
+
+        if self._timing is not ProbeEvalTiming.ENTRY:
+            return
+
+        scope = ChainMap(signal.args, signal.frame.f_globals)
+        if isinstance(probe, ProbeConditionMixin) and not signal._eval_condition(scope):
+            return
```
shatzi, line 49: if this is a bug?

we can have a probe with a condition, the condition is false. it return without setting signal.state
at the exit() - we check if signal == None - go ahead and capture.

so, probe with eval==ENTRY will capture always on exit. I think we should mark this as SKIP_COND?

ddtrace/debugging/_debugger.py
```diff
@@ -412,13 +426,28 @@ def _dd_debugger_hook(self, probe: Probe) -> None:
                 log.error("Unsupported probe type: %r", type(probe))
                 return
 
-            signal.line()
+            scope = ChainMap(actual_frame.f_locals, actual_frame.f_globals)
```
shatzi, line 429: would love to see this also get moved under SignalContext - so there is only one place in our code that deal with those Mixin-s, what do you think?

ddtrace/debugging/_exception/replay.py
```diff
@@ -188,7 +190,9 @@ def on_span_exception(
...
+
+                        snapshot.state = SignalState.DONE
 
                         # Collect
                         self.__uploader__.get_collector().push(snapshot)
```
shatzi, line 198: where does the rate-limit of exception replay happen?

shatzi (approved):
Awesome stuff man!!!

I would like to see everything encapsulated under SignalContext, and remove the last few places like in debugger.py that still check for RateLimit and Condtion Mixins.

ddtrace/debugging/_signal/context.py
P403n1x87: `_eval_condition` sets the state https://github.com/DataDog/dd-trace-py/blob/1d33870b84922eaf89136b75e7e90a048f3fa688/ddtrace/debugging/_signal/model.py#L83 so we should be good

ddtrace/debugging/_debugger.py
P403n1x87: Hmm but we don't need a context for a line probe 🤔

ddtrace/debugging/_exception/replay.py
P403n1x87: rate-limiting is done at the very beginning https://github.com/DataDog/dd-trace-py/blob/1d33870b84922eaf89136b75e7e90a048f3fa688/ddtrace/debugging/_exception/replay.py#L156-L157

ddtrace/debugging/_probe/model.py
P403n1x87: The RFC mentions `ENTRY` as accepted value so I'm aligning with that

ddtrace/debugging/_exception/replay.py
P403n1x87: The arguments are all different so I don't think we can make a single helper in this case 🙁

ddtrace/debugging/_debugger.py
```diff
@@ -426,28 +421,12 @@ def _dd_debugger_hook(self, probe: Probe) -> None:
...
-                    return
-
-                try:
-                    signal.line(scope)
-                except Exception:
-                    log.exception("Failed to execute line probe %r", probe)
-                    return
-
-                signal.state = SignalState.DONE
+            self._probe_registry.set_emitting(probe)
```
shatzi, line 426: I think we miss a check if signal.state is SignalState.DONE before set emitting.

shatzi (approved):
nice work!

PR #11182, opened 2024-10-25
feat(llmobs): add token metrics for langchain llm/chat model spans when appropriate, color as appropriate

sabrenner:
## Checklist

## Reviewer Checklist
- Title is accurate

### DIFF
--- a/ddtrace/llmobs/_integrations/langchain.py
+++ b/ddtrace/llmobs/_integrations/langchain.py
@@ -14,6 +14,7 @@ from ddtrace.internal.utils import get_argument_value
 from ddtrace.llmobs import LLMObs
 from ddtrace.llmobs._constants import INPUT_DOCUMENTS
 from ddtrace.llmobs._constants import INPUT_MESSAGES
+from ddtrace.llmobs._constants import INPUT_TOKENS_METRIC_KEY
 from ddtrace.llmobs._constants import INPUT_VALUE
 from ddtrace.llmobs._constants import METADATA
 from ddtrace.llmobs._constants import METRICS
@@ -21,8 +22,10 @@ from ddtrace.llmobs._constants import MODEL_NAME
 from ddtrace.llmobs._constants import MODEL_PROVIDER
 from ddtrace.llmobs._constants import OUTPUT_DOCUMENTS
 from ddtrace.llmobs._constants import OUTPUT_MESSAGES
+from ddtrace.llmobs._constants import OUTPUT_TOKENS_METRIC_KEY
 from ddtrace.llmobs._constants import OUTPUT_VALUE
 from ddtrace.llmobs._constants import SPAN_KIND
+from ddtrace.llmobs._constants import TOTAL_TOKENS_METRIC_KEY
 from ddtrace.llmobs._integrations.base import BaseLLMIntegration
 from ddtrace.llmobs._utils import safe_json
 from ddtrace.llmobs.utils import Document
@@ -73,8 +76,8 @@ class LangChainIntegration(BaseLLMIntegration):
 
         is_workflow = False
 
+        llmobs_integration = "custom"
         if model_provider:
-            llmobs_integration = "custom"
             if model_provider.startswith(BEDROCK_PROVIDER_NAME):
                 llmobs_integration = "bedrock"
             elif model_provider.startswith(OPENAI_PROVIDER_NAME):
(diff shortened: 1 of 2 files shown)


releasenotes/notes/llmobs-langchain-token-metrics-and-coloring-21eed854130b235c.yaml
```diff
@@ -0,0 +1,5 @@
+---
+features:
+  - |
+    LLM Observability: When not using a provider integration (OpenAI, Anthropic, or Bedrock) with the LangChain integration, token metrics will be appended to the LLM Observability ``llm`` span.
+    LLM Observability: When ``chat_model.with_structured_output(..., method="json_mode")`` is used, or ``response_format={"type": "json_object"}`` is passed into the chat model invocation, the LLM Observability will be an ``llm`` span instead of a ``workflow`` span.
```
sabrenner, line 5:
```suggestion
    LLM Observability: When ``chat_model.with_structured_output(..., method="json_mode")`` is used, or ``response_format={"type": "json_object"}`` is passed into the chat model invocation, the LLM Observability span will be an ``llm`` span instead of a ``workflow`` span.                 
```

ddtrace/llmobs/_integrations/langchain.py
```diff
@@ -97,7 +103,8 @@ def _llmobs_set_tags(
         elif operation == "tool":
             self._llmobs_set_meta_tags_from_tool(span, tool_inputs=kwargs, tool_output=response)
 
-        span.set_tag_str(METRICS, safe_json({}))
+        if span.get_tag(METRICS) is None:
+            span.set_tag_str(METRICS, safe_json({}))
```
Yun-Kim, line 107: Do we need to set metrics if empty? If not, I think we can just get rid of this.

ddtrace/llmobs/_integrations/langchain.py
```diff
@@ -441,6 +490,42 @@ def record_usage(self, span: Span, usage: Dict[str, Any]) -> None:
...
             self.metric(span, "incr", "tokens.total_cost", total_cost)
 
+    def check_token_usage_chat_or_llm_result(self, result):
+        """Checks for token usage on the top-level ChatResult or LLMResult object"""
+        llm_output = getattr(result, "llm_output", None) or {}  # in case it is explicitly set to None
+        token_usage = (
+            llm_output.get("token_usage", {})
+            or llm_output.get("usage_metadata", {})
+            or llm_output.get("usage", {})
+            or {}  # in case one is explicitly set to None
```
Yun-Kim, line 500: Are these mutually exclusive? i.e. llm_output will only have one of these filled while the others are None or don't exist? If so, thoughts on chaining the get default value instead of the OR operation?

```python
llm_output.get("token_usage", llm_output.get("usage_metadata", llm_output.get("usage", {})))
```

releasenotes/notes/llmobs-langchain-token-metrics-and-coloring-21eed854130b235c.yaml
```diff
@@ -0,0 +1,5 @@
+---
+features:
+  - |
+    LLM Observability: When not using a provider integration (OpenAI, Anthropic, or Bedrock) with the LangChain integration, token metrics will be appended to the LLM Observability ``llm`` span.
+    LLM Observability: When ``chat_model.with_structured_output(..., method="json_mode")`` is used, or ``response_format={"type": "json_object"}`` is passed into the chat model invocation, the LLM Observability span will be an ``llm`` span instead of a ``workflow`` span.
```
Yun-Kim, line 5:
```suggestion
    LLM Observability: When langchain's ``chat_model.with_structured_output(..., method="json_mode")`` is used, or ``response_format={"type": "json_object"}`` is passed into a langchain chat model invocation, the LLM Observability span will be an ``llm`` span instead of a ``workflow`` span.                 
```

Yun-Kim (approved):
Looks good for the most part, just a couple suggestions. Thanks @sabrenner!

ddtrace/llmobs/_integrations/langchain.py
sabrenner: good question, probably not? I had just adjusted the logic from what we had before. I'll remove and see if it breaks tests

ddtrace/llmobs/_integrations/langchain.py
sabrenner: Ah good point, it should only ever have one, the name just depends on the partner library. Let me change this logic (can maybe do this in a couple other spots), and see that it still looks good.

ddtrace/llmobs/_integrations/langchain.py
sabrenner: looks like it's fine to remove

ddtrace/llmobs/_integrations/langchain.py
sabrenner: i changed the logic a bit, but still added an extra `dict` check to be safe, since we're technically getting these from partner libraries, seemingly without standardization, I just wanna be extra safe we don't crash. 

Also, for some reason, sometimes these fields are explicitly set to `None`, so they don't fall back on the default value of `dict.get(..., default)`, so I added those extra `is not None` checks to be safe. 

tl;dr check everything to be safe with langchain partner libraries 😅

PR #11206, opened 2024-10-29
fix(profiling): remove usage of `ensure_binary_or_empty` from ddup

taegyunkim:
`ensure_binary_or_empty()` calls can incur significant memory allocations on CPython versions before 3.12 for stack v1. See below images, and all services shown here are with stack v1 with libdatadog exporter (`DD_PROFILING_EXPORT_LIBDD_ENABLED` or `DD_PROFILING_TIMELINE_ENABLED`). 

Python 3.11.x 237MiB/713MiB (33% of the profile) 
image

Python 3.9.x 209MiB/595MiB (35%)
image

Older CPython versions seem to have a less efficient implementation of `str.encode()` which is used by `ensure_binary_or_empty()`. `str.encode()` is implemented in C so we don't show any frame below `ensure_binary()` in above images as Python profiler can't profile native code. 

This function is used across all profilers (stack, memory, and lock). Though enabling stack v2 implementation via `DD_PROFILING_STACK_V2_ENABLED` could remove most of it, since the function is still going to be used for memory and lock profilers, we fix it here. 

We use CPython `PyUnicode_AsUTF8AndSize` to retrieve raw pointer for the given Python `str` object, then create `std::string_view` to pass that over to the `Sample`. We don't propagate any information if the string is not in UTF-8. This behavior is ok for now as libdatadog exporter only accepts utf-8, using `std::str::from_utf8`. 

See below image for comparison of memory allocations before and after this change with Python 3.11.x. The relative and absolute amount is different from above examples, but still show a sizable reduction. 

Before: 252MiB/1.26GiB (20%)
image

After: 0MiB/1.02GiB, memory allocation from `ensure_binary_or_empty()` is just completely gone
image

## Checklist

## Reviewer Checklist
- Title is accurate


emmettbutler (approved):

sanchda (approved):
Holy 💩 this is an extremely heroic fix--amazing!

sanchda:
I looked over a couple of instances of the conversion and it looks fine, but I confess that I didn't inspect every single one carefully.  I'm somewhat assuming that if it passes tests, then things are generally correct enough.

PR #11235, opened 2024-10-30
fix(profiling): clear finished thread ids from `ThreadSpanLinks`

taegyunkim:
## Checklist

## Reviewer Checklist
- Title is accurate


P403n1x87:
Would it be possible for us to do something like https://github.com/P403n1x87/echion/blob/f1ce747d23a631ef48dade4012cd2dc6dc8f1a4d/echion/monkey/threading.py#L27? This way we could untrack (at least pure Python) threads as soon as they are done

taegyunkim:
> Would it be possible for us to do something like [P403n1x87/echion@`f1ce747`/echion/monkey/threading.py#L27](https://github.com/P403n1x87/echion/blob/f1ce747d23a631ef48dade4012cd2dc6dc8f1a4d/echion/monkey/threading.py#L27)? This way we could untrack (at least pure Python) threads as soon as they are done

Oh yeah, we do have `stack_v2.unregister_thread()`, we can clear the mapping from there for the specific thread. Thanks for pointing that out!

P403n1x87 (approved):

releasenotes/notes/profiling-span-link-leak-e7eb1aca3362cd2e.yaml
```diff
@@ -0,0 +1,8 @@
+---
+fixes:
+  - |
+    profiling: when a Python thread finishes, frees memory used for mapping
```
emmettbutler, line 4:
```suggestion
    profiling: when a Python thread finishes, this change frees memory used for mapping
```

emmettbutler (approved):

taegyunkim:
Added an integration test checking `unregister_thread()` is called when thread completes.

PR #11270, opened 2024-11-01
chore(tracing): update default service naming algorithm to use an inferred base service

wconti27:
## Overview

Updates the service naming algo to default to an "inferred base service" when `DD_SERVICE` is not set. The PR aims to remove any uses of `'unnamed-python-service'` as well as when service is `''`

For schema version `v1`, the tracer `config.service` previously defaulted to `unnamed-python-service`. If an inferred project name is found, the inferred name will be used instead. 

For schema version `v0`, the tracer `config.service` previously defaulted to `None` if no service was set. This has also been replaced, but with the caveat that we must check if `config.service == config._inferred_service` when setting service on an "Internal Integration" (ie a server internal to the customer application) via `trace_utils.int_service(...)`. If it does equal, we use the default integration service name, to replicate previous behavior where `config.service` would be `None`. We do this for `v0` because many internal integrations have a default service name that is currently used in place, and we don't want to change any of these names (ie: FastAPI as as service name) and break customers. 

Addresses [INPLAT-127](https://datadoghq.atlassian.net/browse/INPLAT-127)

## Inferred Service Naming Logic (comes from the SSI project) Algorithm:

1. **Detection Logic**:
   - The main function, `detect_service`, processes the command-line arguments and checks for executables. At the time being these executable checks are limited to `python`. 
   - Adds 1 detector classes, `PythonDetector`. We can add more in the future for Flask and other popular executables if desired.

2. **PythonDetector**:
   - Searches for the nearest top-level directory containing a `__init__.py` file or a regular `.py` file.
   - If the `-m` flag is encountered, it captures the following argument as a module name.
   - Iterates up the directory tree until it identifies a corresponding package structure, returning names in dot notation.

## Testing

The PR required consistent naming for the service naming algorithm. All pytest subprocess fixtures used to run tests were modified to add a `ddtrace_subprocess_dir` within the temp file name. This name will be used now as the fallback if `DD_SERVICE` is unset during subprocess tests. Makes snapshot updates much easier. 

## Checklist

## Reviewer Checklist
- Title is accurate

[INPLAT-127]: https://datadoghq.atlassian.net/browse/INPLAT-127?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ddtrace/internal/schema/processor.py
```diff
@@ -7,6 +7,7 @@
 
 class BaseServiceProcessor(TraceProcessor):
     def __init__(self):
+        # breakpoint()
```
P403n1x87, line 10:
```suggestion
```

ddtrace/internal/schema/processor.py
```diff
@@ -22,4 +23,5 @@ def process_trace(self, trace):
         return trace
 
     def _update_dd_base_service(self, span):
+        # breakpoint()
```
P403n1x87, line 26:
```suggestion
```

ddtrace/settings/_inferred_service.py
```diff
@@ -0,0 +1,182 @@
...
+
+            should_skip_arg = prev_arg_is_flag or has_flag_prefix or is_env_variable
+
+            if module_flag:
+                return ServiceMetadata(arg), True
+
+            if not should_skip_arg:
+                wd = self.working_dir_from_envs(self.ctx.envs)
+                abs_path = self.abs_path(arg, wd)
+                if not pathlib.Path(abs_path).exists():
```
P403n1x87, line 44:
```suggestion
                abs_path = Path(arg).resolve()
                if not abs_path.exists():
```

ddtrace/settings/_inferred_service.py
```diff
@@ -0,0 +1,182 @@
...
+                    return ServiceMetadata(""), False  # File not found
+                stripped = abs_path
+                if not pathlib.Path(stripped).is_dir():
+                    stripped = str(pathlib.Path(stripped).parent)
```
P403n1x87, line 48:
```suggestion
                stripped = abs_path
                if not stripped.is_dir():
                    stripped = stripped.parent
```

ddtrace/settings/_inferred_service.py
```diff
@@ -0,0 +1,182 @@
...
+        return os.getcwd()
+
+    def abs_path(self, a, wd):
+        # Converts the given path to an absolute path
+        return str(pathlib.Path(wd) / a)
+
+    # deduce_package_name is walking until a `__init__.py` is not found.
+    # All the dir traversed are joined then with `.` 
+    def deduce_package_name(self, fp):
+        up = str(pathlib.Path(fp).parent)
```
P403n1x87, line 72: I don't think we need to make these many conversions to `str` here given that we are reconverting them to `Path` anyway

ddtrace/settings/_inferred_service.py
```diff
@@ -0,0 +1,182 @@
...
+
+class GunicornDetector:
+    def __init__(self, ctx):
+        self.ctx = ctx
+
+    def expected_command_name(self):
+        return 'gunicorn'
+
+    def detect(self, args):
+        # breakpoint()
```
P403n1x87, line 109:
```suggestion
```

P403n1x87 (commented):
Can we have an high-level description of the logic in the PR description, please? Some examples might be good too 🙏

ddtrace/settings/_inferred_service.py
wconti27: done

ddtrace/settings/_inferred_service.py
wconti27: removed the str conversions

ddtrace/settings/config.py
```diff
@@ -471,11 +472,14 @@ def __init__(self):
 
         self.env = _get_config("DD_ENV", self.tags.get("env"))
         self.service = _get_config("DD_SERVICE", self.tags.get("service", DEFAULT_SPAN_SERVICE_NAME))
+        self._inferred_service = detect_service(sys.argv)
```
tabgok, line 475: "Inferred service" is a name collision that will add some confusion, we should probably name it something else.

It collides with "inferred service" as part of peer service when the library is talking (receiving or sending) to a third-party system.

Maybe `auto_named_service` or `_inferred_base_service` or something similar?

ddtrace/contrib/trace_utils.py
```diff
@@ -372,12 +372,15 @@ def int_service(pin, int_config, default=None):
         return cast(str, int_config.service_name)
 
     global_service = int_config.global_config._get_service()
-    if global_service:
+    if global_service and global_service != int_config.global_config._inferred_service:
```
tabgok, line 375: This seems weird, are we expecting or seeing the integration config (`int_config`) to have a global config with an inferred service that isn't the global config's (`config`) service?

ddtrace/settings/_inferred_service.py
```diff
@@ -0,0 +1,199 @@
...
+            else:
+                return self.parse_name_from_wsgi_app(args[-1]), True
+        return "", False
+
+    def parse_name_from_wsgi_app(self, wsgi_app):
+        name, _, _ = wsgi_app.partition(":")
+        return name
+
+
+def detect_service(args, detector_classes=[PythonDetector, GunicornDetector]):
```
tabgok, line 160: You should cache the result (unless we think it should be dynamic)

wconti27: done.

ddtrace/contrib/trace_utils.py
wconti27: I put a comment explaining why. TLDR is `global_service = config.service`, and if no `DD_SERVICE` is set, then config.service defaults to our inferred base service. However, for v0, we would still have `global_service=None` for this function, so we would be breaking a ton of customer service names without this check.

tabgok: ah, ok, this is a wierd redirection but its also used elsewhere.

tabgok (approved):

ddtrace/contrib/trace_utils.py
wconti27: Yeah it is weird logic to add, but necessary to not break customers. I also added the same logic to the `_get_service` function as well.

ddtrace/contrib/trace_utils.py
```diff
@@ -372,12 +372,19 @@ def int_service(pin, int_config, default=None):
...
+    # use the inferred base service value, and instead use the integration default service. If we
+    # didn't do this, we would have a massive breaking change from adding inferred_base_service.
+    if global_service and global_service != int_config.global_config._inferred_base_service:
         return cast(str, global_service)
 
     if "_default_service" in int_config and int_config._default_service is not None:
         return cast(str, int_config._default_service)
 
+    if default is None and global_service:
+        return cast(str, global_service)
```
P403n1x87, line 386: I see it is also done in other places within this function, but I don't quite see the need for `cast` here 🤔 . Anyway, this can be addressed in a follow-up PR.

ddtrace/settings/_inferred_base_service.py
```diff
@@ -0,0 +1,210 @@
...
+            if has_flag_prefix and arg == "-m":
+                module_flag = True
+
+            prev_arg_is_flag = has_flag_prefix
+
+        return ServiceMetadata(""), False
+
+    def abs_path(self, a, wd):
+        # Converts the given path to an absolute path
+        return pathlib.Path(wd) / a
```
P403n1x87, line 76:
```suggestion
```

ddtrace/settings/_inferred_base_service.py
```diff
@@ -0,0 +1,210 @@
...
+
+    # deduce_package_name is walking until a `__init__.py` is not found.
+    # All the dir traversed are joined then with `.`
+    def deduce_package_name(self, fp):
```
P403n1x87, line 80:
```suggestion
    def deduce_package_name(self, fp: pathlib.Path) -> str:
```

ddtrace/settings/_inferred_base_service.py
```diff
@@ -0,0 +1,210 @@
...
+        up = pathlib.Path(fp).parent
+        current = fp
+        traversed = []
+
+        while current != up:
+            if not pathlib.Path(current, INIT_PY).exists():
```
P403n1x87, line 86:
```suggestion
            if not (current / INIT_PY).exists():
```

ddtrace/settings/_inferred_base_service.py
```diff
@@ -0,0 +1,210 @@
...
+                break
+            traversed.insert(0, pathlib.Path(current).name)
```
P403n1x87, line 88:
```suggestion
            traversed.insert(0, current.name)
```

ddtrace/settings/_inferred_base_service.py
```diff
@@ -0,0 +1,210 @@
...
+            current = up
+            up = pathlib.Path(current).parent
```
P403n1x87, line 90:
```suggestion
            up = current.parent
```

ddtrace/settings/_inferred_base_service.py
```diff
@@ -0,0 +1,210 @@
...
+
+        return ".".join(traversed), len(traversed) > 0
+
+    # findNearestTopLevel returns the top level dir containing a .py file starting walking up from fp
+    def find_nearest_top_level(self, fp):
```
P403n1x87, line 95:
```suggestion
    def find_nearest_top_level(self, fp: pathlib.Path) -> str:
```

ddtrace/settings/_inferred_base_service.py
```diff
@@ -0,0 +1,210 @@
...
+        up = pathlib.Path(fp).parent
```
P403n1x87, line 96:
```suggestion
        up = fp.parent
```

ddtrace/settings/_inferred_base_service.py
```diff
@@ -0,0 +1,210 @@
...
+        current = fp
+        last = current
+
+        while current != up:
+            if not fnmatch.filter(os.listdir(current), ALL_PY_FILES):
+                break
+            last = current
+            current = up
+            up = pathlib.Path(current).parent
```
P403n1x87, line 105:
```suggestion
            up = current.parent
```

ddtrace/settings/_inferred_base_service.py
```diff
@@ -0,0 +1,210 @@
...
+
+        return pathlib.Path(last).name
```
P403n1x87, line 107:
```suggestion
        return last.name
```

ddtrace/settings/_inferred_base_service.py
P403n1x87, line 1: It would be great if we could add type hints 🙏

wconti27: done

ddtrace/settings/_inferred_base_service.py
```diff
@@ -157,12 +155,12 @@ def extract_gunicorn_name_from(self, args):
...
         return "", False
 
-    def parse_name_from_wsgi_app(self, wsgi_app):
+    def parse_name_from_wsgi_app(self, wsgi_app: str) -> str:
         name, _, _ = wsgi_app.partition(":")
         return name
 
 
-def detect_service(args, detector_classes=[PythonDetector, GunicornDetector]):
+def detect_service(args: List[str], detector_classes=[PythonDetector, GunicornDetector]) -> Union[None, str]:
```
P403n1x87, line 163:
```suggestion
def detect_service(args: List[str], detector_classes=[PythonDetector, GunicornDetector]) -> Optional[str]:
```

ddtrace/settings/_inferred_base_service.py
```diff
@@ -0,0 +1,227 @@
...
+            return ServiceMetadata(self.parse_name_from_wsgi_app(wsgi_app)), True
+
+        name, ok = self.extract_gunicorn_name_from(args)
+        if ok:
+            return ServiceMetadata(name), True
+
+        return ServiceMetadata("gunicorn"), True
+
+    def extract_gunicorn_name_from(self, args: List[str]) -> Tuple[str, bool]:
+        skip = False
```
raphaelgavache, line 136: wouldn't the gunicorn integration provide a better / more reliable name?

wconti27: AFAIK we don't have an actual `gunicorn` integration. we patch `gevent` which is used by `gunicorn`, so I don't think we have any special service naming in place for gunicorn other than what's provided by the patch which is nothing specific to gunicorn. Possibly @tabgok or @mabdinur may know more here.

ddtrace/contrib/trace_utils.py
wconti27: Agreed

Yun-Kim (commented):
Can we get some more documentation (at least in the PR description and docstrings in the `_inferred_base_service.py` functions) on how service names are being inferred? This would be helpful for reviewers.

Yun-Kim (approved):
LGTM from team mlobs

tabgok (approved):

ddtrace/settings/_inferred_base_service.py
```diff
@@ -0,0 +1,281 @@
...
+        return "gunicorn"
+
+    def detect(self, args: List[str]) -> Tuple[ServiceMetadata, bool]:
+        """
+        Extracts and returns the Gunicorn application name from various potential sources.
+
+        This function first checks for the application name in the environment variable
+        specified by `GUNICORN_CMD_ARGS`. If found, it extracts the name from it.
+        Next, it looks for the WSGI application name in the environment variable
+        `WSGI_APP_ENV`. If that is not present, it tries to extract the name from
```
mabdinur, line 142: Should we handle the service naming on the integration level?  Web frameworks (ex: [Flask](https://flask.palletsprojects.com/en/stable/tutorial/factory/#the-application-factory). [django](https://docs.djangoproject.com/en/5.1/topics/settings/#envvar-DJANGO_SETTINGS_MODULE), [fastapi](https://fastapi.tiangolo.com/advanced/settings/#the-env-file)) have their own way of setting application names. It's not necessarily done by environment variable. Some web frameworks also have a programatic API for setting application names. This will approach will completely miss this.  

Also the ddtrace library does not provide tracing for http servers directly (ex: gunicorn, Uvicorn, Uwsgi, hypercorn, Waitress). We only support tracing for the web frameworks run on these servers. This change introduces a new layer of instrumentation that increases the complexity of our integrations.

ddtrace/settings/config.py
```diff
@@ -745,6 +749,15 @@ def _get_service(self, default=None):
         :type default: str
         :rtype: str|None
         """
+
+        # We check if self.service != _inferred_base_service since config.service
```
mabdinur, line 753: Can we centralize this logic

romainkomorndatadog (approved):
LGTM for the CI Vis pieces (but I find it a little curious that the diff for `pytest` snapshots is that the `_dd.base_service` tag is disappearing).

ddtrace/contrib/internal/aiobotocore/patch.py
```diff
@@ -116,7 +117,7 @@ async def _wrapped_api_call(original_func, instance, args, kwargs):
         schematize_cloud_api_operation(
             "{}.command".format(endpoint_name), cloud_provider="aws", cloud_service=endpoint_name
         ),
-        service=schematize_service_name(service),
+        service=ext_service(None, config.aiobotocore, default=schematize_service_name(service)),
```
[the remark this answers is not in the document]
wconti27: So the aiobotocore logic for service name is messy currently, these changes were necessary to ensure the same tests pass for schema version v1 without any changes, which would be breaking. We don't include the pin within this call since we would always default to the `pin.service` which in this case would be `aws.[service]`. However, for current V1 behavior, aiobotocore defaults to the default service name, this can be confirmed by looking at the current test cases.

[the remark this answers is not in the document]
wconti27: Fixed

ddtrace/contrib/trace_utils.py
```diff
@@ -372,12 +372,19 @@ def int_service(pin, int_config, default=None):
         return cast(str, int_config.service_name)
 
     global_service = int_config.global_config._get_service()
-    if global_service:
+    # We check if global_service != _inferred_base_service since global service (config.service)
+    # defaults to _inferred_base_service when no DD_SERVICE is set. In this case, we want to not
+    # use the inferred base service value, and instead use the integration default service. If we
+    # didn't do this, we would have a massive breaking change from adding inferred_base_service.
+    if global_service and global_service != int_config.global_config._inferred_base_service:
```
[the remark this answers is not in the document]
wconti27: yes, as discussed in person, we will do a followup PR in order to limit changes for this PR.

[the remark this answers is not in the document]
wconti27: ahhh nice catch, fixed this

ddtrace/internal/schema/span_attribute_schema.py
```diff
@@ -111,5 +115,9 @@ def url_operation_v1(v0_operation, protocol=None, direction=None):
     },
 }
 
+_inferred_base_service: Optional[str] = detect_service(sys.argv)
 
-_DEFAULT_SPAN_SERVICE_NAMES = {"v0": None, "v1": DEFAULT_SERVICE_NAME}
+_DEFAULT_SPAN_SERVICE_NAMES: Dict[str, Optional[str]] = {
+    "v0": _inferred_base_service or None,
+    "v1": _inferred_base_service or DEFAULT_SERVICE_NAME,
```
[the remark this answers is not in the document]
wconti27: I added a check for if its an empty string, shouldn't happen anyways but it will be coerced to None instead of "".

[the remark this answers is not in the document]
wconti27: Not a valid input

ddtrace/internal/telemetry/writer.py
```diff
@@ -47,14 +48,17 @@
...
+
 log = getLogger(__name__)
 
 
 class _TelemetryConfig:
     API_KEY = os.environ.get("DD_API_KEY", None)
     SITE = os.environ.get("DD_SITE", "datadoghq.com")
     ENV = os.environ.get("DD_ENV", "")
-    SERVICE = os.environ.get("DD_SERVICE", "unnamed-python-service")
+    SERVICE = os.environ.get("DD_SERVICE", _inferred_service or "unnamed-python-service")
```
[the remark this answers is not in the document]
wconti27: Will add this in a future PR

ddtrace/settings/config.py
wconti27: Future PR

wconti27:
> LGTM for the CI Vis pieces (but I find it a little curious that the diff for `pytest` snapshots is that the `_dd.base_service` tag is disappearing).

@tabgok and I debugged this yesterday. It's because the command line to run the pytest suite is something along the lines of `python -m pytest` (with a bunch of extra args). We end up getting a global service name of pytest, and since the global service name matches the span's service name of `pytest`, we don't add a base service tag (only if they differ do we add it)

releasenotes/notes/implement-inferred-service-naming-for-default-service-names-56395f9bcb4458de.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    Updates the service naming algorithm to infer the base service name when `DD_SERVICE` is not set, replacing 
+    instances of `'unnamed-python-service'`. Ensures that a more meaningful service name is used whenever
```
brettlangdon, line 5:
```suggestion
    Updates the service naming algorithm to infer the base service name when ``DD_SERVICE`` is not set, replacing 
    instances of ``'unnamed-python-service'``. Ensures that a more meaningful service name is used whenever 
```

tests/telemetry/test_writer.py
```diff
@@ -197,7 +197,7 @@ def test_app_started_event(telemetry_writer, test_agent_session, mock_time):
         assert payload == result["payload"]
 
 
-def test_app_started_event_configuration_override(test_agent_session, run_python_code_in_subprocess, tmpdir):
+def test_app_started_event_configuration_override(test_agent_session, run_python_code_in_subprocess, tmpdir, tracer):
```
brettlangdon, line 200: do we need `tracer` here?

tests/internal/service_name/test_inferred_base_service.py
```diff
@@ -0,0 +1,79 @@
...
+        (base_path / "modules" / "no" / "python_files" / "here.txt").touch()  # Module with no subdirectories
+        (base_path / "modules" / "m1" / "first" / "nice" / "package" / "not_a_python_file.txt").touch()
+
+        yield base_path
+
+
+def test_python_detector(mock_file_system):
+    # Mock the current working directory to the test_modules path
+    with patch("os.getcwd", return_value=str(mock_file_system)):
+        tests = [
```
brettlangdon, line 54: NIT: better to structure as parameterized test?

```python
@pytest.mark.parametrize("cmd,expected", [("python modules/m1/first/nice/package", "m1.first.nice.package"), ...])
def test_python_detector(cmd, expected, mock_file_system):
    with patch("os.getcwd", return_value=str(mock_file_system)):
        cmd_parts = cmd.split(" ")
        detected_name = detect_service(cmd_parts)
        assert detected_name == expected, f"Test failed for command: [{cmd}]"
```


basically same-same. the parametrize approach is more consistent with other tests we have. only meaningful change is each scenario would get reported as a test case, rather than all as one, and every scenario would run every time, instead of failing/stopping after the first failed scenario.

non-blocking comment.

tests/internal/service_name/test_imports.py
```diff
@@ -91,7 +89,7 @@ def test_service_name_import_with_client_service_names_enabled_v0():
     from ddtrace.internal.schema.span_attribute_schema import service_name_v1
     from ddtrace.internal.schema.span_attribute_schema import url_operation_v0
 
-    assert DEFAULT_SPAN_SERVICE_NAME == DEFAULT_SERVICE_NAME
+    assert DEFAULT_SPAN_SERVICE_NAME == "ddtrace_subprocess_dir"
```
brettlangdon, line 92: do we want to have a constant for this? seems to be repeated a bunch.

mabdinur (approved):

ddtrace/settings/_inferred_base_service.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+INIT_PY = "__init__.py"
+ALL_PY_FILES = "*.py"
+
+CACHE: Dict[str, Optional[str]] = {}
+
+
+class DetectionContext:
+    def __init__(self, envs):
```
brettlangdon, line 19: typing missing for `envs`

ddtrace/settings/_inferred_base_service.py
```diff
@@ -0,0 +1,203 @@
...
+        self.envs = envs
+
+
+class ServiceMetadata:
+    def __init__(self, name):
```
brettlangdon, line 24:
```suggestion
    def __init__(self, name: str):
```
?

ddtrace/settings/_inferred_base_service.py
```diff
@@ -0,0 +1,203 @@
...
+        # prevent service name from being an empty string
+        self.name = name if name != "" else None
+
+
+class PythonDetector:
+    def __init__(self, ctx):
```
brettlangdon, line 30: typing missing for `ctx`

ddtrace/settings/_inferred_base_service.py
```diff
@@ -0,0 +1,203 @@
...
+        self.ctx = ctx
+        name = "python"
+        self.name = name
```
brettlangdon, line 33:
```suggestion
        self.name = "python"
```

brettlangdon: `self.name` doesn't appear to be used anywhere? any reason to not just hard code into the pattern?

ddtrace/settings/_inferred_base_service.py
```diff
@@ -0,0 +1,203 @@
...
+    for detector in detectors.values():
+        metadata, detected = detector.detect(args_to_search)
+        if detected and metadata.name:
+            CACHE[cache_key] = metadata.name
+            return metadata.name
+    CACHE[cache_key] = None
+    return None
+
+
+def _is_executable(file_path):
```
brettlangdon, line 191: adding typing for `file_path` and return value

ddtrace/settings/_inferred_base_service.py
```diff
@@ -0,0 +1,203 @@
...
+                if ok:
+                    return ServiceMetadata(value), True
+                return ServiceMetadata(self.find_nearest_top_level(stripped)), True
+
+            if has_flag_prefix and arg == "-m":
+                module_flag = True
+
+            prev_arg_is_flag = has_flag_prefix
+
+        return ServiceMetadata(""), False
```
brettlangdon, line 89: should we just return `Optional[ServiceMetadata]` instead? and then this is just `return None`

ddtrace/settings/_inferred_base_service.py
```diff
@@ -0,0 +1,203 @@
...
+            if not fnmatch.filter(os.listdir(current), ALL_PY_FILES):
+                break
+            last = current
+            current = up
+            up = current.parent
+
+        return last.name
+
+
+def detect_service(args: List[str], detector_classes=[PythonDetector]) -> Optional[str]:
```
brettlangdon, line 123: do we need to support multiple detectors?

if not, it may be easier to just hard code the single detector usage in this function, and then in the future we can make it support multiple if we need them?

ddtrace/settings/_inferred_base_service.py
```diff
@@ -0,0 +1,203 @@
...
+    for detector_class in detector_classes:
+        detector_instance = detector_class(ctx)
+
+        for i in range(len(possible_commands)):
+            detector_name = detector_instance.name
+            detector_pattern = detector_instance.pattern
+
+            command = possible_commands[i]
+
+            if re.search(detector_pattern, command):
```
brettlangdon, line 165: should this functionality live inside of the detector_instance?

e.g. so we have something like this:

```python
for detector_class in detector_classes:
    detector_instance = detector_class(ctx)
    for i, cmd in enumerate(possible_commands):
        if detector_instance.matches(cmd):
            detectors[detector.name] = detector_instance            
            executable_args.add(i)
            continue
        elif _is_executable(command):
            executable_args.add(i)
```

just a suggestion about class design, functionally wouldn't change what you have.

ddtrace/settings/_inferred_base_service.py
```diff
@@ -0,0 +1,203 @@
...
+                detectors.update({detector_name: detector_instance})
+                # append to a list of arg indexes to ignore since they are executables
+                executable_args.add(i)
+                continue
+            elif _is_executable(command):
```
brettlangdon, line 170: is it ok that we may run this check multiple times per-command? since we potentially run this once per-command and per-detector?


it could be worth while to separate the two, doing the `_is_executable()` check separate from the detector checks.

ddtrace/settings/_inferred_base_service.py
```diff
@@ -0,0 +1,203 @@
...
+            Defaults to [PythonDetector].
+
+    Returns:
+        Optional[str]: The name of the detected service, or None if no service was detected.
+    """
+
+    if not args:
+        return None
+
+    cache_key = ";".join(args)
```
brettlangdon, line 144: `args` should work as a dict key since it is hash-able, or else you can cast it to a tuple too, might be cheaper to do than doing the join.

can the order change, do we need to sort args to make it be consistent?

ddtrace/settings/_inferred_base_service.py
```diff
@@ -0,0 +1,203 @@
...
+                detectors.update({detector_name: detector_instance})
+                # append to a list of arg indexes to ignore since they are executables
+                executable_args.add(i)
+                continue
+            elif _is_executable(command):
+                # append to a list of arg indexes to ignore since they are executables
+                executable_args.add(i)
+
+    args_to_search = []
+    for i in range(len(args)):
```
brettlangdon, line 175: consider using `enumerate` for this pattern, such that you get `for i, arg in enumerate(args):`

more pythonic and removes the need for the `arg = args[i]` line.

ddtrace/settings/_inferred_base_service.py
```diff
@@ -0,0 +1,203 @@
...
+
+def _is_executable(file_path):
+    normalized_path = os.path.normpath(file_path)
+    if not os.path.isfile(normalized_path):
+        return False
+
+    # Split the path into directories and check for 'bin'
+    directory = os.path.dirname(normalized_path)
+    while directory and os.path.dirname(directory) != directory:  # Check to prevent infinite loops
+        if os.path.basename(directory) == "bin":
```
brettlangdon, line 199: you might want `.endswith("bin")`? there is also `/usr/sbin/` for example.

ddtrace/settings/_inferred_base_service.py
```diff
@@ -0,0 +1,203 @@
...
+from typing import Tuple
+
+
+INIT_PY = "__init__.py"
+ALL_PY_FILES = "*.py"
+
+CACHE: Dict[str, Optional[str]] = {}
+
+
+class DetectionContext:
```
brettlangdon, line 18: do we need this indirection? do you see more context being added later?

e.g. why have a wrapper class when you can just pass `os.environ` around?

brettlangdon: This could just be a NamedTuple

ddtrace/settings/_inferred_base_service.py
```diff
@@ -0,0 +1,203 @@
...
+    def __init__(self, envs):
+        self.envs = envs
+
+
+class ServiceMetadata:
```
brettlangdon, line 23: same here, do you see more service metadata being added in the future?

Could this just be a NamedTuple instead?

ddtrace/settings/_inferred_base_service.py
wconti27: smart, done.

ddtrace/settings/_inferred_base_service.py
wconti27: makes sense

ddtrace/settings/_inferred_base_service.py
wconti27: Since we've changed there to be only 1 detector, this isn't an issue. We can refactor later if that changes.

tests/internal/service_name/test_inferred_base_service.py
wconti27: switched it be parameterized, works great now thanks!

ddtrace/settings/_inferred_base_service.py
wconti27: makes sense.

ddtrace/settings/_inferred_base_service.py
wconti27: Done.

ddtrace/settings/_inferred_base_service.py
wconti27: Possibly??

wconti27: Im going to leave this, doesn't hurt

ddtrace/settings/_inferred_base_service.py
wconti27: done.

ddtrace/settings/_inferred_base_service.py
wconti27: sorting can't hurt, Ill make this change.

ddtrace/settings/_inferred_base_service.py
wconti27: done.

ddtrace/settings/_inferred_base_service.py
brettlangdon: yeah, I think because of the volume of creation of `ServiceMetadata` is low it isn't a problem.

The reason to consider a `NamedTuple` or alternative is they are lighter weight memory wise, so if you needed to create a ton of these or create them in a tight loop, could be better for memory/GC pressure.

tests/internal/service_name/test_imports.py
wconti27: added one.

tests/conftest.py
```diff
@@ -305,7 +323,14 @@ def run_function_from_file(item, params=None):
     expected_out = marker.kwargs.get("out", "")
     expected_err = marker.kwargs.get("err", "")
 
-    with NamedTemporaryFile(mode="wb", suffix=".pyc") as fp:
+    # Create a temporary dir named `ddtrace_subprocess_dir` that will be used for service naming
+    # consistency
+    temp_dir = gettempdir()
```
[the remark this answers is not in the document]
wconti27: done, do you want to take a look?

brettlangdon (approved):

PR #11274, opened 2024-11-01
chore: set the default of _should_skip to false

wantsui:
While reviewing the flaky marker in gunicorn's `test_span_schematization` in https://github.com/DataDog/dd-trace-py/blob/223c261250ef1bd1a7547cd0d6311efc153a2395/tests/contrib/gunicorn/test_gunicorn.py#L188, @brettlangdon pointed out that the timestamp is long past.

ie:
```
@flaky(until=1706677200)
```

This translates to Jan 31, 2024. 

However, if you run the test for `test_span_schematization` today, it passes fine because it is getting skipped:

```
1 passed with 0 warnings, 0 failed
```

The bug fix here is to set the default to false, but keep the spirit of the original `condition` if statement.

If you rerun the test_span_schematization with this branch, it now doesn't get skip and the test correctly fails:
```
FAILED tests/contrib/gunicorn/test_gunicorn.py::test_span_schematization - AttributeError: module 'ddtrace.internal.compat' has no attribute 'stringify'
```

Leaving this PR in a draft state for now to see what other tests this affects.

## Checklist

## Reviewer Checklist
- Title is accurate


tests/utils.py
```diff
@@ -1339,9 +1339,9 @@ def _should_skip(condition=None, until: int = None):
         until = dt.datetime.fromtimestamp(until)
     if until and dt.datetime.utcnow() < until.replace(tzinfo=None):
         return True
-    if condition is not None and not condition:
-        return False
-    return True
+    if condition is not None and condition:
+        return True
+    return False
```
quinna-h, line 1344:
```suggestion
return condition is not None and condition
```

quinna-h: nit suggestion

erikayasuda (approved):

PR #11310, opened 2024-11-06
fix(profiling): bound memory growth for internal string table

nsrip-dd:
The profiler interns strings in an internal buffer. This buffer is currently not cleared.
But it can contain arbitrary user-generated strings (e.g. asyncio `Task` names),
and thus can grow without bound over the lifetime of the program.
Clear the string buffer after every profile upload to mitigate the problem.

This PR doesn't currently have a regression test. I took a stab at writing one
in de29bfbc07 but it's a little tricky to test reliably. For now, we can rely on the relenv
to catch this since we originally saw the problem there.

## Checklist

## Reviewer Checklist
- Title is accurate

PROF-10824


ddtrace/internal/datadog/profiling/dd_wrapper/src/profile.cpp
```diff
@@ -51,6 +51,14 @@ Datadog::Profile::cycle_buffers()
     return true;
 }
 
+void
+Datadog::Profile::reset()
+{
+    const std::lock_guard<std::mutex> lock(profile_mtx);
+    strings.clear();
+    string_storage.clear();
```
nsrip-dd, line 59: Is this safe to do? What I'm worried about is if something uses `insert_or_get`, gets a string view pointing to the strings we're storing here, and then we call `reset` while something still holds on to the string view. cc @sanchda @taegyunkim

nsrip-dd: Specifically I had wondered about here, where we release the profile lock and _then_ clear it out. Is there a window where other parts of the program can add stuff to the string table (and get references into it) before we clear it?

https://github.com/DataDog/dd-trace-py/blob/2c023d9044bbd40da1ffa40bf78ce24f21d6e3e8/ddtrace/internal/datadog/profiling/dd_wrapper/src/ddup_interface.cpp#L308-L312

taegyunkim: hmm, as far as I know state is cleared when uploading has been done. 

https://github.com/DataDog/dd-trace-py/blob/62727981a6fd702a6fd48f0f35e581aec36ede4f/ddtrace/internal/datadog/profiling/dd_wrapper/src/ddup_interface.cpp#L309-L311

And when clearing state if we clear everything properly in the profile state, it should be ok?

nsrip-dd: So there's the `profile_release` call there that releases the profile lock. Then `profile_clear_state` clears out some data structures, under the lock for each thing it does. Another way to phrase my question is, can samples be added between the times the lock is released here? Like, between lines 310 and 311?

taegyunkim: I believe that can happen, and those samples would be lost, since they would be cleared.

nsrip-dd: Hmm... the more I think about this the less I'm sure there's a safe way to do this without bigger changes. The problem is that when we build samples, we hold references into the string table across multiple critical sections. So for something like this:

```
auto s = dup_start_sample();
ddup_push_task_name(s, "abc");
ddup_flush_sample(s);
```

The `ddup_push_task_name` call adds a copy of `"abc"` to the string table and holds that reference via `s`. Then, `ddup_flush_sample(s)` reads that string table reference when passing the sample into libdatadog. Between those two calls, we can't drop the string from the table or we'll deallocate it and `ddup_flush_sample` will read a dangling pointer. But the profiling lock isn't held between those two calls so the profiler might concurrently drop the string table (regardless of where we put the call to drop the string table). It seems like we need a different approach...

taegyunkim: Would the following let us avoid this issue?
hold the profile lock 
- upload
- clear the state
and release the lock

nsrip-dd: Unfortunately no. The problem is that we build a sample across multiple API calls, and the profile (specifically the string table) is not locked between those calls. So even if we clear the state under the lock, we can get dangling references if we happen to be concurrently building a profile sample. I'm working on a few ideas for how to change this.

emmettbutler (approved):

nsrip-dd:
Closing in favor of #11612

PR #11317, opened 2024-11-06
fix(lib-injection): do not import user installed ddtrace

brettlangdon:
One of the checks we do when auto-injecting into a process is to see if the user already has `ddtrace` installed. If we do, then we will _not_ inject into it.

However, the way we check if the user has `ddtrace` installed is we `import ddtrace`. This may have an unintended consequence of importing `ddtrace` into a process which otherwise wouldn't normally.

The impact of importing `ddtrace` should be fairly low, since we won't instrument anything, but it is better to avoid the import overhead if the user wasn't going to import it anyways.

Instead we move to using `importlib.util.find_spec` to search for the module on the `sys.meta_path`. The downside to doing this is we won't know the installed `ddtrace` version.

## Checklist

## Reviewer Checklist
- Title is accurate


lib-injection/sources/sitecustomize.py
```diff
@@ -205,9 +206,12 @@ def _inject():
     integration_incomp = False
     runtime_incomp = False
     os.environ["_DD_INJECT_WAS_ATTEMPTED"] = "true"
+    spec = None
     try:
-        import ddtrace
-    except ImportError:
+        spec = importlib.util.find_spec("ddtrace")
+        if not spec:
```
brettlangdon, line 212: getting `None` back is a valid response.

I decided to just raise in that case to keep majority of this code the same as it was previously.

P403n1x87:
> The downside to doing this is we won't know the installed ddtrace version.

But once we have the spec we could potentially get hold of the `_version.py` source and retrieve the version from there, no?

brettlangdon:
> Wouldn't we need unit or regression tests to validate this change?

Yes, but I have no idea where this regression test should go. I didn't see any lib-injection tests in this repo, so probably needs to be a system test change?

I can take a quick look into that today, I would like a regression test.

emmettbutler (approved):

PR #11335, opened 2024-11-08
chore(er): introduce exception identity

P403n1x87:
We introduce the computation of exception identity to limit the number of capturing instances for the same exception over time.

## Checklist

## Reviewer Checklist
- Title is accurate

### DIFF
--- a/ddtrace/debugging/_exception/replay.py
+++ b/ddtrace/debugging/_exception/replay.py
@@ -20,6 +20,7 @@ from ddtrace.internal.logger import get_logger
 from ddtrace.internal.packages import is_user_code
 from ddtrace.internal.rate_limiter import BudgetRateLimiterWithJitter as RateLimiter
 from ddtrace.internal.rate_limiter import RateLimitExceeded
+from ddtrace.internal.utils.time import HourGlass
 
 
 log = get_logger(__name__)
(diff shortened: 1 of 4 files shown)


emmettbutler (approved):

ddtrace/debugging/_exception/replay.py
```diff
@@ -47,13 +48,70 @@
...
+
+        # We haven't seen this exception in a while
+        hg.turn()
+
+        return False
+    finally:
+        if len(EXCEPTION_IDENT_LIMITER) > 1024:
+            # We limit the number of exception identities we track to avoid
+            # memory leaks.
+            sorted_keys = sorted(EXCEPTION_IDENT_LIMITER, key=lambda k: EXCEPTION_IDENT_LIMITER[k])
```
GreenMatan, line 105: Shouldn't the order be based on started_at / end_at?

ddtrace/debugging/_exception/replay.py
```diff
@@ -47,13 +48,70 @@
...
+
+def exception_chain_ident(chain: ExceptionChain) -> int:
+    h = 0
+    for exc, tb in chain:
+        h = (h << 1) ^ exception_ident(exc, tb)
+        h &= 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
+    return h
+
+
+def limit_exception(chain: ExceptionChain) -> bool:
```
GreenMatan, line 83: Isn't it confusing there are now two ways to rate-limit exception (can_capture & limit_exception)?

GreenMatan (approved):
Left two small comments, the fingerprinting is looking good to me  👍

ddtrace/debugging/_exception/replay.py
P403n1x87: I thought it made more sense to sort by `end_at`, as these would be the timers that go off first. When the interval is the same for all the timers, these are the "older" ones (and `start_at` would essentially be equivalent to `end_at` modulo a constant time shift). So this will evict the older timers, increasing the odds that we will re-capture an exception that we haven't seen recently (compared to more recent ones).

ddtrace/debugging/_exception/replay.py
P403n1x87: I agree this can be confusing. The reason why I haven't unified the two is that I don't want to unwind the exception chain if the global rate limiter, or the data on the trace that we might have put before, tells us that we don't want to capture. This way we can prevent doing some throw-away work.

ddtrace/debugging/_exception/replay.py
GreenMatan: I was confused about the implicit nature of this call, clarified offline, resolving.

ddtrace/debugging/_exception/replay.py
GreenMatan: We've discussed offline about ways to tackle exceptions fairness to possibly mitigate this subtlety

shatzi (approved):

mabdinur (approved):

PR #11377, opened 2024-11-13
fix(profiling): reduce lock profiler log spam

nsrip-dd:
The lock profiler would log a warning if it couldn't determine a
name for a lock, and it would try determining a name multiple times for
the same lock.  This lead to excessive log spam. Downgrade this to a
debug log and only try to determine the name once.

## Checklist

## Reviewer Checklist
- Title is accurate

PROF-10898

### DIFF
--- a/ddtrace/profiling/collector/_lock.py
+++ b/ddtrace/profiling/collector/_lock.py
@@ -298,7 +298,7 @@ class _ProfiledLock(wrapt.ObjectProxy):
 
             if not self._self_name:
                 self._self_name = ""
-                LOG.warning(
+                LOG.debug(
                     "Failed to get lock variable name, we only support local/global variables and their attributes."
                 )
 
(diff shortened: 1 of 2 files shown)


ddtrace/profiling/collector/_lock.py
```diff
@@ -298,7 +298,7 @@ def _maybe_update_self_name(self):
 
             if not self._self_name:
                 self._self_name = ""
-                LOG.warning(
+                LOG.debug(
```
nsrip-dd, line 301: cc @taegyunkim since we're making this a debug log, is there any useful info we could put in the debug log?

taegyunkim: I want some information regarding whether the lock is used in part of some library or stdlib but that would require some more logic here, so nothing at the moment. We'd eventually want this to be exported to instrumentation telemetry logs.

taegyunkim (approved):

taegyunkim:
Thank you for the fix!

nsrip-dd:
For those watching at home, I've tried the merge queue 3 times and it's failed for different reasons each time :( I'm just going to squash and merge, I'm reasonably confident that the failures are not related to this change.

PR #11390, opened 2024-11-13
chore(profiling): stack collector tests for libdd and stack v2

taegyunkim:
## Checklist

## Reviewer Checklist
- Title is accurate


tests/profiling_v2/collector/test_stack.py
```diff
@@ -558,3 +567,125 @@ def sleep_instance(foobar, self):
...
+    output_filename = pprof_prefix + "." + str(os.getpid())
+
+    assert ddup.is_available
+    ddup.config(env="test", service=test_name, version="my_version", output_filename=pprof_prefix)
+    ddup.start()
+
+    # Start some (green)threads
+    def _dofib():
+        for _ in range(10):
+            # spend some time in CPU so the profiler can catch something
```
nsrip-dd, line 610: I _assume_ this will be okay since we're doing a recursive Fibonacci, which has exponential complexity. But is there a flakiness risk here? As I understand it, the stack collector samples based on elapsed wall time, right? How much time needs to elapse for us to know a sample will be collected?

nsrip-dd (approved):
Change look reasonable to me, just a question about possible flakiness

tests/profiling_v2/collector/test_stack.py
taegyunkim: Our default sampling interval is 10 milliseconds
- stack v2 defined [here](https://github.com/DataDog/dd-trace-py/blob/65d02a08a80b43c8644fa9633a6fb7f5de8e93ba/ddtrace/internal/datadog/profiling/stack_v2/include/constants.hpp#L6)
- stack v1 [here](https://github.com/DataDog/dd-trace-py/blob/65d02a08a80b43c8644fa9633a6fb7f5de8e93ba/ddtrace/profiling/collector/stack.pyx#L462)

Yes, we rely on the fact that `_fib()` function call takes a sufficiently long time. Seems like `_fib(28)` only takes ~50ms on Python.3.11.9 on my mac. To make sure it's less flaky I will change it to take a few seconds, so it takes at least 2 orders of magnitude.

And since this is test only change, I'll also update other tests in this file to spend a little more time to make sure that they're less flaky. 

Thanks for raising this question.

taegyunkim: Oh, actually it already does, because 50ms * 10 times * 10 threads = 5 seconds

taegyunkim: Updated so that each _dofib() takes about 1 second and we call it from 5 threads. I want a test to finish < 10 seconds.

PR #11394, opened 2024-11-13
chore(tracing): fix inferred base service for pytest command ran with --ddtrace optional argument

wconti27:
# Motiviation

Add special case for `--ddtrace` pytest argument, that was causing CI tests to have a different service name than local testing. CI testing adds the `--ddtrace` arg to the command. Previously the code would skip any args that started with `-` as well as the following arg, but that should not be the case for this arg.

## Checklist

## Reviewer Checklist
- Title is accurate


Yun-Kim (approved):
This is just ensuring tests are aware of the inferred test service names right? Wasn't able to find anything else in this PR. Otherwise LGTM from team mlobs

wconti27:
> What was special about the `mako` module in this context?

The change was to preserve current mako span naming behavior in the case that `DD_SERVICE` is unset, resulting in `config.service` being unset. Previously the service would default to mako's default service name, but with this change `config.service` is set with the inferred service name. If we didn't change the mako code it would use this new inferred service name always, which is breaking. Instead, now we use `trace_utils.int_service` to compute the service, and it coalesces correctly to the default value if our config.service is the inferred service name.

mabdinur (approved):
Do we expect this change to have any customer impact (ie do we need a release note?)

emmettbutler:
@wconti27 is that true of any other integrations, or just mako?

wconti27:
> Do we expect this change to have any customer impact (ie do we need a release note?)

@mabdinur The only customer impact should be limited to those using ddtrace with pytest via the following syntax `pytest --ddtrace ...`. If you think we need a release note for that I can add it.

wantsui (approved):

romainkomorndatadog:
> The only customer impact should be limited to those using ddtrace with pytest via the following syntax pytest --ddtrace ...

FWIW, `pytest` has lots of other single-arg options (eg: `pytest -s path/to/my/tests`, or `-x`, etc). It feels a little weird to special-case `--ddtrace`.

gnufede (approved):

P403n1x87 (approved):

releasenotes/notes/fix-pytest-inferred-service-naming-5ba12e991d040a1b.yaml
```diff
@@ -0,0 +1,6 @@
+---
+fixes:
+  - |
+    Updates the inferred base service name algorithm to ensure that arguments following ``--ddtrace`` are no
```
wconti27, line 4:
```suggestion
    ci_visibility: Updates the inferred base service name algorithm to ensure that arguments following ``--ddtrace`` are no 
```

emmettbutler (approved):

PR #11404, opened 2024-11-14
chore(symdb): upload compressed symbol payloads

P403n1x87:
We add support for compressed symbol database payloads.

## Checklist

## Reviewer Checklist
- Title is accurate


ojung (approved):
Looks good, thanks!

ddtrace/internal/utils/http.py
```diff
@@ -418,7 +419,7 @@ def parse_message(msg):
 class FormData:
     name: str
     filename: str
-    data: str
+    data: Union[str, bytes]
```
ojung, line 422: Why keep the str? Do we ever want to upload the payload as plain json? There seems to be no way to disable compression, so I think this should be `bytes`. But then I wonder as well: We are actually sending a string payload (`data="[symbols_placeholder]",`), can't we send raw bytes instead?

P403n1x87: This is a general utility within the library. When the data is `str` it is `encode`d into bytes. Ultimately, the generated payload will be bytes

emmettbutler (approved):

shatzi (approved):

PR #11406, opened 2024-11-14
chore(internal): avoid freezegun in ddtrace

romainkomorndatadog:
`freezegun` (and `pytest-freezegun`) may cause wrong durations to be reported when used, so we add an integration that automatically enforces that `ddtrace` be an ignored module by:
- calling `freezegun.configure()` with `extend_ignore_list` to add `ddtrace`
- wrapping `freezegun.configure()` so that any attempt to reset the `default_ignore_list` includes `ddtrace`

Additionally, since `pytest` does not patch all integrations by default, and because we don't want to force users to use `--ddtrace-patch-all` for `freezegun` to be patched, we update our `pytest`  plugin(s) to patch `freezegun` during the `pytest_load_initial_conftests` hook (which should be early enough).

This is not considered a fix as there are no active issues reported and this (so far) primarily affects telemetry recorded by the CI Visibility service.

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/__init__.py
```diff
@@ -36,6 +39,14 @@
     warnings.simplefilter("ignore")
     from .tracer import Tracer as _
 
+# freezegun may cause strange durations to be reported for spans or stopwatch uses
+try:
+    import freezegun
+
+    freezegun.configure(extend_ignore_list=("ddtrace",))
+except ImportError:
+    pass
```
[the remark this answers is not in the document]
P403n1x87: If we really need to do this, we should consider making this an import hook instead, so we don't force-import `freezegun`

P403n1x87 (commented):
My initial reaction is that we should add `freezegun` alongside `gevent` as trigger for module cloning. This at least as a short-term solution. Long-term we could consider having our native timing module/utilities to separate more from the runtime.

emmettbutler (commented):
I second @P403n1x87's comment, let's avoid importing freezegun.

ddtrace/contrib/freezegun/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+Enabling
+~~~~~~~~
+The freezegun integration is enabled by default. Use :func:`patch()<ddtrace.patch>` to enable the integration:
+    from ddtrace import patch
+    patch(freezegun=True)
+
+
+Configuration
+~~~~~~~~~~~~~
+The freezegun integration is not configurable, but may be disabled using DD_PATCH_MODULES:freezegun=false.
```
vitor-de-araujo, line 13: Isn't it like this?
```suggestion
The freezegun integration is not configurable, but may be disabled using DD_PATCH_MODULES=freezegun:false.
```

vitor-de-araujo (approved):

ddtrace/contrib/freezegun/__init__.py
```diff
@@ -0,0 +1,28 @@
+"""
+The freezegun integration reconfigures freezegun's default ignore list to ignore ddtrace.
+
+Enabling
+~~~~~~~~
+The freezegun integration is enabled by default. Use :func:`patch()<ddtrace.patch>` to enable the integration:
```
romainkomorndatadog, line 6:
```suggestion
The freezegun integration is enabled by default. Use :func:`patch()<ddtrace.patch>` to enable the integration::
```

ddtrace/contrib/pytest/_plugin_v2.py
```diff
@@ -99,7 +100,7 @@ def _handle_itr_should_skip(item, test_id) -> bool:
             InternalTest.mark_itr_forced_run(test_id)
             return False
 
-        InternalTest.mark_itr_skipped(test_id)
+        # InternalTest.mark_itr_skipped(test_id)
```
romainkomorndatadog, line 103:
```suggestion
        InternalTest.mark_itr_skipped(test_id)
```

emmettbutler (approved):

PR #11413, opened 2024-11-14
fix(llmobs): sync and async generators can be decorated

sabrenner:
MLOB-1892

## What does this PR do
Fixes support for decorating sync and async generator functions. 

Some notes:

- While we will still capture function name and inputs (inputs only for non `llm` and `embedding` decorators), we will not automatically capture outputs from the generator
  - This change is intended to allow user annotation to still be captured, as previously, the generator would be captured as the "output", and user annotation would never go through.
- Async generators are not identifiable by `iscoroutine` and `isgeneratorfunction`, so we rely on `isasyncgenfunction` instead, verified in tests.

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/llmobs/decorators.py
```diff
@@ -48,17 +79,32 @@ async def wrapper(*args, **kwargs):
 
             else:
 
+                @wraps(func)
+                def generator_wrapper(*args, **kwargs):
+                    if not LLMObs.enabled:
+                        log.warning(SPAN_START_WHILE_DISABLED_WARNING)
+                        yield from func(*args, **kwargs)
```
lievan, line 86: nice, so `yield from` already handles `throw`, `send`, `GeneratorExit`, `StopIteration`?

tests/llmobs/test_llmobs_decorators.py
```diff
@@ -536,3 +536,121 @@ def f(prompt, arg_2, kwarg_1=None, kwarg_2=None):
                 output_value="my custom output",
             )
         )
+
```
lievan, line 539: should we add some test cases to ensure decorator generators are still compatible with `send`, `throw`..?

sabrenner: good callout, let me try it out manually and then add a test if needed!

ddtrace/llmobs/decorators.py
sabrenner: yeah, it does (from some testing i've done) for sync generators (`yield from` can only be used in sync cases). the `async` versions are a bit different, will push some tests & fixes up today trying to handle `async` generators better for those cases.

ddtrace/llmobs/decorators.py
```diff
@@ -13,6 +16,44 @@
...
+            except GeneratorExit:
+                await gen.aclose()
+                break
+            except StopAsyncIteration as e:
+                await gen.athrow(e)
+                break
+            except Exception as e:
+                await gen.athrow(e)
+                raise
+    except (StopAsyncIteration, GeneratorExit):
```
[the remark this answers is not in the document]
lievan: By this line?
`next_val = await gen.asend(None)`

lievan (approved):
Thanks!

ddtrace/llmobs/decorators.py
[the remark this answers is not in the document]
sabrenner: yeah, from that line. i special-cased these so we don't tag them as errors (as they are just generator lifecycle end indicators)

PR #11430, opened 2024-11-17
chore(crashtracker): add profiler_config tag if profiler enabled

taegyunkim:
The previous attempt set the tag after crashtracker has started, and tags didn't show up in the logs. 

## Checklist

## Reviewer Checklist
- Title is accurate


emmettbutler (approved):

ddtrace/internal/core/crashtracking.py
```diff
@@ -57,6 +59,9 @@ def start() -> bool:
     for key, value in crashtracker_config.tags.items():
         add_tag(key, value)
 
+    if profiling_config.enabled:
+        add_tag("profiler_config", config_str(profiling_config))
```
brettlangdon, line 63: can profiling be enabled/disabled at runtime via remote config?

taegyunkim: I am not aware of such functionality. @sanchda do we have this?

sanchda (commented):
Looks fine to me.  This does change extension loading a bit, but I think the way it does so is pretty tame (from an overhead perspective).

ddtrace/internal/core/crashtracking.py
brettlangdon: not a problem, was mostly trying to explore if we would end up in a state where we don't set this config but profiling is enabled.

as long as it is startup only decision, then I think we are good here?

danielsn (approved):

PR #11438, opened 2024-11-18
fix(profiler): initialize memory collector until after libdatadog

sanchda:
The current implementation of this system has a pretty negligent state initialization sequence.

In particular, the memory allocation profiler relies on libdatadog in some circumstances, and should be initialized _after_ it.

## Checklist

## Reviewer Checklist
- Title is accurate

### DIFF
--- a/ddtrace/profiling/profiler.py
+++ b/ddtrace/profiling/profiler.py
@@ -347,10 +347,6 @@ class _ProfilerInstance(service.Service):
 
             for module, hook in self._collectors_on_import:
                 ModuleWatchdog.register_module_hook(module, hook)
-
-        if self._memory_collector_enabled:
-            self._collectors.append(memalloc.MemoryCollector(r))
-
         exporters = self._build_default_exporters()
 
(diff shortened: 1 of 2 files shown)


ddtrace/profiling/profiler.py
```diff
@@ -364,6 +360,11 @@ def start_collector(collector_class: Type) -> None:
                 before_flush=self._collectors_snapshot,
             )
 
+        # The memory allocation profiler can't be enabled until after the exporters are built, since it optionally
```
taegyunkim, line 363: do we need to do this for stack and lock profilers too?

sanchda: Good call.  Let me refactor this a bit, since our initialization sequence is pretty hairy.

juanjux (approved):

juanjux:
Only semi-related, you can mostly ignore this: When reading `Collector` I thought this was an implementation of some unused/free memory collector, not a class to get memory data, maybe `MemorySampler` or something similar would be a better name?

PR #11442, opened 2024-11-19
chore(ci_visibility): don't incorrectly report tests as new

romainkomorndatadog:
This tweaks the logic for marking tests as "new"  (with the `test.is_new` tag) to not mark tests as new:
- if fetching unique tests failed (treated as "the length of unique tests is 0")
- if the test has parameters (this check is done twice since we allow `pytest` to set parameters after a test has been discovered)

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/internal/ci_visibility/recorder.py
```diff
@@ -868,6 +868,12 @@ def is_unique_test(cls, test_id: Union[TestId, InternalTestId]) -> bool:
         if instance is None:
             return False
 
+        # The assumption that we were not able to fetch unique tests properly if the length is 0 is acceptable
+        # because the current EFD usage would cause the session to be faulty even if the query was successful but
+        # not unique tests exist. In this case, we assume all tests are unique.
```
vitor-de-araujo, line 873: 'unique test' in this context just means 'known test', right? Maybe we should call it that everywhere? (not in this PR, though)

vitor-de-araujo (approved):

ddtrace/internal/ci_visibility/recorder.py
romainkomorndatadog: It's documented as the "unique tests" endpoint in some places, and "known tests" in others, unfortunately, so I tried to name them accordingly (eg: in the API client codebase, I refer to it as unique tests, and in the EFD codebase, I refer to it as known tests). It felt more consistent with the way we named thing in RFCs.

TBH, I don't know if being inconsistent inside the codebase is better or worse than being inconsistent with our docs.

PR #11458, opened 2024-11-20
chore(service-naming): improve inferred service naming algorithm

wconti27:
## Motivation

Updates inferred service naming algorithm. Previously, the service naming algorithm would skip over any flag arguments (args starting with `-`), as well as the following argument as it was assumed to be the argument value. The update changes the algorithm to run again if no service name was found the first time. The second time, the algorithm will not skip arguments that were preceded by a flag argument.

Effect:
-- Previous Behavior: `pytest --no-cov my/file.py` -> service name = `""`
-- New Behavior: `pytest --no-cov my/file.py` -> service name = `"my.file"` 

Additionally adds check to ensure a python module included after `-m` module is importable before using it as the service name. 

Also updates the service naming algorithm to use try-catches to prevent any unforeseen errors.

Adds many more test cases, fixes a few snapshots.

## Checklist

## Reviewer Checklist
- Title is accurate


raphaelgavache (approved):

ddtrace/settings/_inferred_base_service.py
```diff
@@ -145,39 +156,51 @@ def detect_service(args: List[str]) -> Optional[str]:
...
+
+        # Iterate through the matched detectors again, this time not skipping args preceded by flag args
+        for detector in detectors.values():
+            metadata = detector.detect(args_to_search, skip_args_preceded_by_flags=False)
+            if metadata and metadata.name:
+                CACHE[cache_key] = metadata.name
+                return metadata.name
+    except Exception as ex:
+        # Catch any unexpected errors to be extra safe
+        print("Unexpected error during inferred base service detection: ", ex)  # or log the exception
```
brettlangdon, line 202: `/print/log.exception/` ?

brettlangdon: Above you have:

> log.debug("Unexpected error while processing argument: ", arg, "Exception: ", ex)

brettlangdon (approved):
suggestion to replace `print` with a log statement, otherwise lgtm

ddtrace/settings/_inferred_base_service.py
wconti27, line 202:
```suggestion
        log.debug("Unexpected error during inferred base service detection: ", ex)  # or log the exception
```

wconti27: fixed!

ddtrace/settings/_inferred_base_service.py
```diff
@@ -145,39 +156,51 @@ def detect_service(args: List[str]) -> Optional[str]:
...
+
+        # Iterate through the matched detectors again, this time not skipping args preceded by flag args
+        for detector in detectors.values():
+            metadata = detector.detect(args_to_search, skip_args_preceded_by_flags=False)
+            if metadata and metadata.name:
+                CACHE[cache_key] = metadata.name
+                return metadata.name
+    except Exception as ex:
+        # Catch any unexpected errors to be extra safe
+        log.debug("Unexpected error during inferred base service detection: ", ex)
```
tabgok, line 202: Might be worth it to have this as a WARN

ddtrace/settings/_inferred_base_service.py
```diff
@@ -145,39 +156,51 @@ def detect_service(args: List[str]) -> Optional[str]:
...
+
     CACHE[cache_key] = None
     return None
```
tabgok, line 205: Do we want None or default-python-service or something else?

ddtrace/settings/_inferred_base_service.py
```diff
@@ -59,22 +65,27 @@ def detect(self, args: List[str]) -> Optional[ServiceMetadata]:
...
-                if not abs_path.exists():
-                    continue
-                stripped = abs_path
-                if not stripped.is_dir():
-                    stripped = stripped.parent
-                value, ok = self.deduce_package_name(stripped)
-                if ok:
-                    return ServiceMetadata(value)
-                return ServiceMetadata(self.find_nearest_top_level(stripped))
+                try:
```
emmettbutler, line 75: Making this `try` block as small as possible to catch the expected exceptions would make it easier to track down errors based on logging.

wantsui (approved):

ddtrace/settings/_inferred_base_service.py
```diff
@@ -145,39 +156,51 @@ def detect_service(args: List[str]) -> Optional[str]:
...
-        if i not in executable_args:
-            args_to_search.append(arg)
-
-    # Iterate through the matched detectors
-    for detector in detectors.values():
-        metadata = detector.detect(args_to_search)
-        if metadata and metadata.name:
-            CACHE[cache_key] = metadata.name
-            return metadata.name
+    try:
```
emmettbutler, line 159: What sort of exceptions are you expecting from this block? Being more specific about them would make the code easier to reason about and less prone to confusing behavior.

emmettbutler (approved):

ddtrace/settings/_inferred_base_service.py
wconti27: This block I wanted to encompass all the path operations.

ddtrace/settings/_inferred_base_service.py
wconti27: Honestly Im just trying to be extra cautious since any exception thrown in this code will break applications. I wouldn't really expect anything to throw errors here. I can up the log to a warning.

ddtrace/settings/_inferred_base_service.py
wconti27: we fallback to `default-python-service` for V1, None for V0, that logic is within the service naming schema files.

ddtrace/settings/_inferred_base_service.py
wconti27: If Python was typed, I could probably enumerate all the possible exceptions thrown by `Pathlib`, but given it isn't it wouldn't be timely to find and add all those exception types to further scope down these catch blocks.

PR #11460, opened 2024-11-20
fix(profiler): update memalloc guard

sanchda:
Previously, the memory allocation profiler would use Python's builtin thread-local storage interfaces in order to set and get the state of a thread-local guard.

I've updated a few things here.

* I think get/set idioms are slightly problematic for this type of code, since it pushes the responsibility of maintaining clean internal state up to the parent.  A consequence of this is that the propagation of the underlying state _by value_ opens the door for race conditions if execution changes between contexts (unlikely here, but I think minimizing indirection is still cleaner).  Accordingly, I've updated this to use native thread-local storage
* Based on @nsrip-dd's observation, I widened the guard over `free()` operations.  I believe this is correct, and if it isn't then the detriment is performance, not correctness.
* I got rid of the PY37 failovers

We don't have any reproductions for the defects that prompted this change, but I've been running a patched library in an environment that _does_ reproduce the behavior, and I haven't seen any defects.

1. I don't believe this patch is harmful, and if our memory allocation tests pass then I believe it should be fine.
2. I have a reason to believe this fixes a critical defect, which can cause crashes.

## Checklist

## Reviewer Checklist
- Title is accurate


P403n1x87 (approved):

nsrip-dd (approved):

taegyunkim (approved):

sanchda:
@nsrip-dd , could you I ask you to re-review?  Your critical eye was very helpful in the last round and I'd like to enjoy its benefit for the final cut.

ddtrace/profiling/collector/_memalloc_reentrant.h
```diff
@@ -1,50 +1,214 @@
...
+
+#ifdef __linux__
+    // On Linux, we need to make sure we didn't just fork
+    // pthreads will guarantee the lock is consistent, but we at least need to clear it
+    static pid_t my_pid = 0;
+    if (my_pid == 0) {
+        my_pid = getpid();
+    } else if (my_pid != getpid()) {
+        // We've forked, so we need to free the lock
+        memlock_unlock(lock);
```
nsrip-dd, line 134: From the `pthread_mutex_unlock` man page: "If a thread attempts to unlock a mutex that it has not locked or a mutex which is unlocked, undefined behavior results." https://linux.die.net/man/3/pthread_mutex_unlock. Is this okay to do? Or do we need to re-initialize the mutex?

ddtrace/profiling/collector/_memalloc_reentrant.h
```diff
@@ -1,50 +1,214 @@
...
 #else
-        PyThread_set_key_value(memalloc_reentrant_key, NULL);
+    return pthread_mutex_unlock(&lock->mutex) == 0;
 #endif
 }
 
+// Timed lock function
 static inline bool
-memalloc_get_reentrant(void)
+memlock_lock_timed(memlock_t* lock, uint32_t timeout_ms)
```
nsrip-dd, line 121: Also, to capture our in-person discussion, your idea was to change this from a timed lock to an un-timed lock and bail immediately if we can't acquire the lock (just in memalloc_add_event? others?) Plus an option (set via env var) to crash if the lock is contended so we can debug why this lock is even needed.

ddtrace/profiling/collector/_memalloc_heap.c
nsrip-dd, line 1: Also discussed in-persons: we should add locking to the heap profiler too, probably with its own lock. If the allocation profiler can be accessed concurrently, so can this.

ddtrace/profiling/collector/_memalloc_reentrant.h
sanchda: Thank you--done.  The environment variable is `_DD_PROFILING_MEMALLOC_CRASH_ON_MUTEX_PASS` and it'll do a case-insensitive check for truthy (1, true, yes, on enable(d))

ddtrace/profiling/collector/_memalloc_reentrant.h
sanchda: True, fixed by being more precise about the mutex attr (`PTHREAD_MUTEX_ERRORCHECK`)

ddtrace/profiling/collector/_memalloc_heap.c
sanchda: Done in [3f5cdb3](https://github.com/DataDog/dd-trace-py/pull/11460/commits/3f5cdb37f1dc70d8a831e32b4dd5d14f7ab07d79)

ddtrace/profiling/collector/_memalloc_heap.c
```diff
@@ -156,27 +202,39 @@ memalloc_heap_track(uint16_t max_nframe, void* ptr, size_t size, PyMemAllocatorD
     if (global_heap_tracker.sample_size == 0)
         return false;
 
+    // Take the lock
+    if (!memlock_trylock(&g_memheap_lock)) {
```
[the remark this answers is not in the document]
sanchda: Oh wow!  Good catch!  No, I actually spent some time making the structs more uniform and adapting the implementing of the clamped increment to satisfy the requirements of a clamped _add_ just so I could put the heap accounting in an atomic add.

But I ordered it incorrectly in the PR.

I fixed this--thank you.

PR #11471, opened 2024-11-20
chore(profiling): gunicorn test improvements

taegyunkim:
- use `--worker-tmp-dir /dev/shm` as in https://pythonspeed.com/articles/gunicorn-in-docker/, https://docs.gunicorn.org/en/stable/faq.html#how-do-i-avoid-gunicorn-excessively-blocking-in-os-fchmod
- clean up server when the request fails

## Checklist

## Reviewer Checklist
- Title is accurate


tests/profiling_v2/test_gunicorn.py
```diff
@@ -62,7 +62,7 @@ def _test_gunicorn(gunicorn, tmp_path, monkeypatch, *args):
     # DEV: We only start 1 worker to simplify the test
     proc = gunicorn("-w", "1", *args)
     # Wait for the workers to start
-    time.sleep(5)
+    time.sleep(10)
```
sanchda, line 65: Do you know whether proc.poll consumes the proc?  I think it's just a check.

If it's a check, it might be better to factor this out into a bounded wait.  Like check proc.poll for up to 10 seconds, but continue as soon as it passes through the gate.

Not sure whether that solves anything, just thinking about potential issues in adding fixed-time sleeps.

sanchda (approved):

tests/profiling_v2/test_gunicorn.py
taegyunkim: It's just a check, and as far as I know it returns asap.

taegyunkim: hmm, maybe it'd be better to leave this as 5 seconds to see whether using `--worker-tmp-dir` makes any difference

PR #11478, opened 2024-11-21
fix(ci_visibility): get default environment from agent when DD_ENV is not set

romainkomorndatadog:
This fixes an issue where, if we are in EVP mode, and the agent has a custom default environment set (eg: using `DD_APM_ENV`), but `DD_ENV` is not set in the environment, we would incorrectly set the environment to `None`.

Instead, we now query the agent's info page and use the `config.default_env` key to choose the environment value.

In somewhat-related changes, we now also explicitly default the environment to `"none"` in agentless mode if it is not set by the user, whereas before we would use a `None` value which would serialize to `null`, and we would rely on the backend to perform the `null` -> `"none"` change. Since the backend was already enforcing this behavior, it should be a no-op from the user's perspective.

## Checklist

## Reviewer Checklist
- Title is accurate

### DIFF
--- a/ddtrace/internal/ci_visibility/recorder.py
+++ b/ddtrace/internal/ci_visibility/recorder.py
@@ -222,8 +222,14 @@ class CIVisibility(Service):
         self._git_data: GitData = get_git_data_from_tags(self._tags)
 
         dd_env = os.getenv("_CI_DD_ENV", ddconfig.env)
+        dd_env_msg = ""
 
         if ddconfig._ci_visibility_agentless_enabled:
+            # In agentless mode, normalize an unset env to none (this is already done by the backend in most cases, so
+            # it does not override default behavior)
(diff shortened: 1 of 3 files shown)


ddtrace/internal/ci_visibility/recorder.py
```diff
@@ -222,8 +222,14 @@ def __init__(self, tracer=None, config=None, service=None):
         self._git_data: GitData = get_git_data_from_tags(self._tags)
 
         dd_env = os.getenv("_CI_DD_ENV", ddconfig.env)
```
vitor-de-araujo, line 224: `ddconfig.env` comes from the `DD_ENV` enviroment variable?

romainkomorndatadog: Correct.

PR #11490, opened 2024-11-21
fix(iast): add code to filter out ddtrace stuff from dir() on patched modules

juanjux:
Signed-off-by: Juanjo Alvarez <juanjo.alvarezmartinez@datadoghq.com>

## Description

While testing on dd-source CI we found an issue where a module was doing a `dir(other_module)` and the changed results from the patched module were breaking stuff (because our patching would add `ddtrace_aspects`, `ddtrace_sink_points`, et cetera to the results and the original module was expecting all `other_module` symbols to have some members like `id`). 

This PRs fixes this problem by:

- Creating a custom `__dir__` function (that will override any pre-existing ones) removing from the results all the symbols that we add ourselves while patching.
- Renaming all added `ddtrace` symbols to `__ddtrace`.

Also:
- Adds a `_DD_IAST_NO_DIR_PATCH` config var to disable the wrapping of the patched module `__dir__` functions in case the user have some side-effect problem.
- The return type of `visit_ast` has been fixed (it wrongly was `str` while is in fact a `ast.Module` type).

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/appsec/_iast/_ast/ast_patching.py
```diff
@@ -427,30 +456,34 @@ def astpatch_module(module: ModuleType, remove_flask_run: bool = False) -> Tuple
...
-        return "", ""
+        return "", None
 
     if remove_flask_run:
         source_text = _remove_flask_run(source_text)
 
-    new_source = visit_ast(
+    if not asbool(os.environ.get(IAST.ENV_NO_DIR_PATCH, "false")):
+        # Add the dir filter so __ddtrace stuff is not returned by dir(module)
+        source_text += _DIR_WRAPPER
```
gnufede, line 478: I believe we should do this after patching the code, and only if the code was patched, which is done in the next line, right?

gnufede: I mean, this change should go in the AST visitor, just like we introduce the new imports of the aspects if required, we should add this to the end

juanjux: Yes. However, it's not easy to add it after patching because it's not as easy as adding the source code (with is already compiled to AST), but I will take a look at how to do it.

ddtrace/appsec/_iast/_ast/visitor.py
```diff
@@ -38,71 +40,71 @@ def _mark_avoid_convert_recursively(node):
...
-    "alias_module": "ddtrace_aspects",
+    "alias_module": _PREFIX + "aspects",
     "functions": {
-        "StringIO": "ddtrace_aspects.stringio_aspect",
-        "BytesIO": "ddtrace_aspects.bytesio_aspect",
-        "str": "ddtrace_aspects.str_aspect",
-        "bytes": "ddtrace_aspects.bytes_aspect",
-        "bytearray": "ddtrace_aspects.bytearray_aspect",
-        "ddtrace_iast_flask_patch": "ddtrace_aspects.empty_func",  # To avoid recursion
+        "StringIO": _PREFIX + "aspects.stringio_aspect",
```
avara1986, line 45: what about move this to a function?
```
_build_iast_module_path(string):
    return _PREFIX + "aspect." + string
```

and
```
"StringIO": _build_iast_module_path("aspects.stringio_aspect"),
```

juanjux: I don't see the point. If this was more complicated I would see it, but it's just a concat (actually I am thinking of using `f"{_PREFIX}aspects.stringio_aspect"` but the `+` is more readable.

tests/appsec/iast_packages/inside_env_runner.py
```diff
@@ -46,9 +46,10 @@ def try_patched(module_name, expect_no_change=False):
             patched_module
         ), "Patched source is None after patching: Maybe not an error, but something fishy is going on"
         new_code = unparse(patched_module)
+        # JJJ
```
avara1986, line 49: :point_up:

tests/appsec/iast/aspects/test_add_aspect_fixtures.py
```diff
@@ -38,7 +38,8 @@ def test_operator_add_dis(
 
         bytecode = dis.Bytecode(mod.do_operator_add_params)
         dis.dis(mod.do_operator_add_params)
-        assert bytecode.codeobj.co_names == ("ddtrace_aspects", "add_aspect")
+        # JJJ
```
avara1986, line 41: :point_up:

tests/appsec/iast_packages/inside_env_runner.py
juanjux: woopsie

gnufede (approved):

PR #11493, opened 2024-11-21
fix(profiling): `asyncio` task fixes for stack v2

taegyunkim:
This PR fixes 2 issues.

1. Associate `asyncio` tasks with spans

Task rendering begins with calling `render_task_begin`, and it is then followed by call to `FrameStack::render()`. However, we only add span id association via call to `render_thread_begin()` so tasks are missing span id labels. To properly link spans to tasks, we add the label when we create a new sample for task. And while debugging this issue, discovered the following. 

2. Propagate `asyncio` task names properly

Echion adds a dummy frame having task name, code [here](https://github.com/taegyunkim/echion/blob/9d5bcc5867d7aefff73c837adcba4ef46eecebc6/echion/threads.h#L282). When it tries to render tasks, it [peeks at the top of the frame](https://github.com/taegyunkim/echion/blob/9d5bcc5867d7aefff73c837adcba4ef46eecebc6/echion/threads.h#L351) stack's name and uses that. However, there could be frames preceding the frame stack added by [this line](https://github.com/taegyunkim/echion/blob/9d5bcc5867d7aefff73c837adcba4ef46eecebc6/echion/threads.h#L276). Leading to using incorrect task names.

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/internal/datadog/profiling/stack_v2/src/stack_renderer.cpp
```diff
@@ -119,6 +130,13 @@ StackRenderer::render_python_frame(std::string_view name, std::string_view file,
     if (!utf8_check_is_valid(file.data(), file.size())) {
         file = invalid;
     }
+    // DEV: Echion pushes a dummy frame containing task name, and its line
+    // number is set to 0.
+    if (!task_state.pushed_task_name and line == 0 and name != invalid and name != "<invalid>") {
```
sanchda, line 135: How do you feel about skipping the checks for invalid names?  We already propagate invalid symbol and filenames.

Actually, the UX for that might not be great, since it'll erroneously associate a bunch of execution into potentially invalid tasks (like we might have hundreds of tasks all associated to "<invalid>").

Instead:  what if we have a static sequence number that (atomically) increments every time we see an invalid task, and name the task something like "invalid-XXX"?  That'd eliminate the need to check, it'd make sure tasks are labeled when execution is associated to a task (this is important!), and it'd isolate execution away from a common label.

releasenotes/notes/profiling-stack-v2-task-names-1d566a1793280aa0.yaml
```diff
@@ -0,0 +1,8 @@
+---
+fixes:
+  - |
+    Fixes an issue where ``asyncio`` task names are not properly propagated
+    when using stack v2, i.e. when ``DD_PROFILING_STACK_V2_ENABLED`` is set.
+    Fixes an issue where ``asyncio`` tasks are not associated with spans
```
sanchda, line 6: Copy/paste issue?

tests/profiling_v2/collector/test_stack_asyncio.py
```diff
@@ -36,79 +45,89 @@ async def hello():
...
-    t1, t2 = loop.run_until_complete(maintask)
+    with tracer.trace("test_asyncio", resource=resource, span_type=span_type) as span:
+        span_id = span.span_id
+        local_root_span_id = span._local_root.span_id
+
+        loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(loop)
+        if _asyncio_compat.PY38_AND_LATER:
+            maintask = loop.create_task(hello(), name="main")
+        else:
```
sanchda, line 62: If we dump out before on python 3.7, we can probably lose this branch as well

sanchda (approved):
Very important 👍

ddtrace/internal/datadog/profiling/stack_v2/src/stack_renderer.cpp
```diff
@@ -64,7 +66,7 @@ StackRenderer::render_thread_begin(PyThreadState* tstate,
 }
 
 void
-StackRenderer::render_task_begin(std::string_view name)
+StackRenderer::render_task_begin(std::string_view)
```
nsrip-dd, line 69: It's really not obvious to me just from the PR description or any comments in the code why this wasn't right before or why it's right now. As far as I can piece it together, the name here is derived when Echion enumerates tasks during a sample call. Is it not computing the name correctly? Are we adding the same name multiple times? Could you please include some description of the bug and why this fix works? And do we need to fix anything in your Echion fork, given that it's passing a no-longer-useful parameter here?

taegyunkim: It is not computing the name correctly. Updated the PR description. I don't think we need to fix anything in the Echion fork yet.

ddtrace/internal/datadog/profiling/stack_v2/src/stack_renderer.cpp
```diff
@@ -89,9 +91,18 @@ StackRenderer::render_task_begin(std::string_view name)
         ddup_push_walltime(sample, thread_state.wall_time_ns, 1);
         ddup_push_cputime(sample, thread_state.cpu_time_ns, 1); // initialized to 0, so possibly a no-op
         ddup_push_monotonic_ns(sample, thread_state.now_time_ns);
-    }
 
-    ddup_push_task_name(sample, name);
+        // We also want to make sure the tid -> span_id mapping is present in the sample for the task
+        const std::optional<Span> active_span =
+          ThreadSpanLinks::get_instance().get_active_span_from_thread_id(thread_state.id);
```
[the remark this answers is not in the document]
taegyunkim: I am not sure whether I understood your question correctly. From my understanding, there's 1-to-many relationship between threads and tasks, and we have 1-to-1 relationship between threads and spans. So, I'm just connecting the dots to associate span ids to tasks. 

We have `asyncio` machinery defined [here](https://github.com/DataDog/dd-trace-py/blob/67eb77dacca1d8679ca7fa088e3c873cba840208/ddtrace/profiling/_asyncio.py#L33) for both v1 and v2, btw.

releasenotes/notes/profiling-stack-v2-task-names-1d566a1793280aa0.yaml
taegyunkim: Oh this PR fixes these two issues.

ddtrace/internal/datadog/profiling/stack_v2/src/stack_renderer.cpp
P403n1x87: There's always the option of "fixing" echion to report better metadata. For instance, we could make the frame object carry a `kind`, that we could set to a value representing "Task", instead of working around the shortcoming?

ddtrace/internal/datadog/profiling/stack_v2/src/stack_renderer.cpp
```diff
@@ -119,6 +130,13 @@ StackRenderer::render_python_frame(std::string_view name, std::string_view file,
     if (!utf8_check_is_valid(file.data(), file.size())) {
         file = invalid;
     }
+    // DEV: Echion pushes a dummy frame containing task name, and its line
+    // number is set to 0.
+    if (!task_state.pushed_task_name and line == 0 and name != invalid and name != "<invalid>" and file != invalid) {
+        ddup_push_task_name(sample, name);
+        task_state.pushed_task_name = true;
+    }
```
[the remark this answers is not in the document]
taegyunkim: We could. Do we want or not want to push the task name as a frame?

ddtrace/internal/datadog/profiling/stack_v2/src/stack_renderer.cpp
```diff
@@ -119,6 +130,13 @@ StackRenderer::render_python_frame(std::string_view name, std::string_view file,
     if (!utf8_check_is_valid(file.data(), file.size())) {
         file = invalid;
     }
+    // DEV: Echion pushes a dummy frame containing task name, and its line
+    // number is set to 0.
+    if (!task_state.pushed_task_name and line == 0 and name != invalid and name != "<invalid>" and file != invalid) {
```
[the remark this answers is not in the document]
taegyunkim: Removed `file`. I think I added the checks as I've been getting `<invalid>` for task names in the tests. I don't mind showing more of `<invalid>` ask we're already showing this to the users.

ddtrace/internal/datadog/profiling/stack_v2/src/stack_renderer.cpp
taegyunkim: Let me try removing those checks. 

@P403n1x87's suggestion would be the cleanest way here. Propagate a flag from echion and check that instead of relying on `line == 0`. I intentionally avoided changing echion code as I didn't want it to diverge more.

sanchda: Rather than propagating new information in the underlying struct, it might be cleaner to structure the interface to push a different category of information.

We'd have a `render_task_name` interface in the echion fork's Render system.  The Echion implementation of Render would just push it as a frame, and ddtrace would push it into ddup as a task name.

releasenotes/notes/profiling-stack-v2-task-names-1d566a1793280aa0.yaml
sanchda: Oh, sorry.  I saw the single `- |` and forgot how these yamls are parsed.  Carry on, apologies for the noise

ddtrace/internal/datadog/profiling/stack_v2/src/stack_renderer.cpp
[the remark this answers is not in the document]
sanchda: While I _do_ really like the idea of having the tasks all visualized in the flamegraph, it does conflict with our UX at the moment.

If we want to retain the desired output, I think it'd be really interesting to see whether we can add the pseudo-frames in the backend.  That way we'd be able to split things at the root level in a bunch of different ways.

So my uninformed and weakly held opinion is that we populate the task name, but do not submit this as a frame.

ddtrace/internal/datadog/profiling/stack_v2/src/stack_renderer.cpp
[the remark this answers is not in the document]
sanchda: Yeah--I don't think there's a wrong way or a right way.

The only quantitative thing here is that omitting the task name leads to a UX where the underlying execution appears as though it did not occur within a task at all.  It's actually been incredibly useful (for, uh... profiler escalations 😢) to know when things are on a task or not--even if the task name is garbage.

ddtrace/internal/datadog/profiling/stack_v2/src/stack_renderer.cpp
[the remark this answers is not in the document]
taegyunkim: OK, yeah we also have this side panel selector for task names. 

Screenshot 2024-12-04 at 11 58 45 AM

ddtrace/internal/datadog/profiling/stack_v2/src/stack_renderer.cpp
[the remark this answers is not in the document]
taegyunkim: That's a reasonable concern but it would be better to be implemented in echion. Instead of simply using `<invalid>`, it could use context specific `<invalid>`, i.e. <`invalid-task-name>`

ddtrace/internal/datadog/profiling/stack_v2/src/stack_renderer.cpp
taegyunkim: Yeah, we should update echion. I'm just delaying such changes to happen after we merge our fork to the upstream echion.

ddtrace/internal/datadog/profiling/stack_v2/src/stack_renderer.cpp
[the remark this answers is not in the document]
nsrip-dd: Right, I was wondering about the span-to-task/coroutine relationship. I think I see how it works though. We get the active span for a thread through the tracer's context provider: https://github.com/DataDog/dd-trace-py/blob/47c7b5287da25643e46652e6d222a40a52f2382a/ddtrace/profiling/collector/stack.pyx#L527. And I guess it's up to the tracer to update the context provider when a task/coroutine related to a span becomes active on a thread. So it's not directly related to this PR, just curious how it works :)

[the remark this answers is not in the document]
taegyunkim: Oh right, you found the right code location. The tracer simply let us know the span when it's activated, and we use that information to keep track of span <-> thread relation

sanchda (approved):

Issue #11479, opened 2024-11-21
Celery task_postrun is called only after the span has finished in chains (due to a regression)


oranav:
#10676 introduced a regression which causes the `task_postrun` signal to never be able to attach tags to a span when working with [Celery canvases](https://docs.celeryq.dev/en/latest/userguide/canvas.html) (chains, callbacks, chords etc).

When working with a chain for instance, the task always calls `apply_async` on its callback just before it finishes (before `task_postrun` is fired). With the changes introduced in #10676, the span is already finished at this point (and sent to the tracer), so adding new tags doesn't do anything, making the span degenerated.

This means that the following happens:

1. `task_prerun` is called, creating a new span, and [registering it under `prerun_span`](https://github.com/DataDog/dd-trace-py/blob/74e3d515a6a088883d8a2d54f9f880eb4e147d99/ddtrace/contrib/internal/celery/signals.py#L56)
2. The task is run (either successfully or not)
3. Celery sees that there's a callback registered, and [fires the next task in the chain](https://github.com/celery/celery/blob/1b35d1d5966614ce36af75808ee21b0d2db6745d/celery/app/trace.py#L512-L519)
4. `Task.apply_async` is invoked, but it's wrapped with `_traced_apply_async_inner`. Inside its `finally` block it [sees that there's a `prerun_span`](https://github.com/DataDog/dd-trace-py/blob/74e3d515a6a088883d8a2d54f9f880eb4e147d99/ddtrace/contrib/internal/celery/app.py#L150-L155) belonging to the worker span - therefore forcefully finishing it
5. `task_postrun` is fired, and tries to [attach tags to the span](https://github.com/DataDog/dd-trace-py/blob/74e3d515a6a088883d8a2d54f9f880eb4e147d99/ddtrace/contrib/internal/celery/signals.py#L92-L94), but at this point it's doomed, since the span has already finished.

All in all I would recommend to revert #10676 and come up with a better implementation, or at least to remove the `task_postrun` part of it, since it's utterly broken.

# Reproduce

tasks.py:
```python3
from celery import Celery, chain

app = Celery("tasks")


@app.task
def a():
    print("a")


@app.task
def b():
    print("b")
```

fire.py:
```python3
from tasks import a, b

(a.si() | b.si()).delay()
```

Run Celery with:
```
DD_TRACE_ENABLED=1 ddtrace-run celery -A tasks.app worker -l debug -P solo -c 1
```

Then fire the chain with:
```
[...]

PR #11498, opened 2024-11-21
fix(celery): stop closing prerun_span too soon to account for Celery chains scenario

wantsui:
We've made a few changes to handle celery context recently, including: https://github.com/DataDog/dd-trace-py/pull/10676

In particular the goal of https://github.com/DataDog/dd-trace-py/pull/10676 was to handle a scenario where a long running task may run into an exception, preventing it from closing.

Unfortunately, this scenario did not account for cases where tasks are chained and may not close until later.

See: https://github.com/DataDog/dd-trace-py/issues/11479 and https://github.com/DataDog/dd-trace-py/issues/11624

With this PR, the sample app in https://github.com/DataDog/dd-trace-py/issues/11479 would attach the celery specific span back to the root span.

I also need to add tests for the chains scenario.

Related to AIDM-494

## Checklist

## Reviewer Checklist
- Title is accurate


tests/contrib/celery/test_chained_task.py
```diff
@@ -0,0 +1,46 @@
...
+        celery_worker_cmd.split(),
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        preexec_fn=os.setsid,
+        close_fds=True,
+        cwd=str(os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(__file__))))),
+    )
+
+    # Wait a few seconds for the celery worker to start up before trying to run tasks
+    time.sleep(5)
```
tabgok, line 25: oof, can we use something like `celery.task.control.Inspect.stats()` instead?  Sleep-for-external-service are a great source of flakiness.

```
while celery.task.control.Inspect.stats() != worker and max_time not waited:
    sleep_a_small_amount
```

tests/contrib/celery/test_chained_task.py
```diff
@@ -0,0 +1,46 @@
...
+        stderr=subprocess.PIPE,
+        preexec_fn=os.setsid,
+        close_fds=True,
+        cwd=str(os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(__file__))))),
+    )
+
+    task_runner_process.wait()
+
+    # Kill the process so it starts to send traces to the Trace Agent
+    os.killpg(worker_process.pid, signal.SIGKILL)
```
tabgok, line 38: why is this needed?

wantsui: If the worker process doesn't get killed, the process will just run endlessly. I'm killing it so there are less logs to parse through.

tests/contrib/celery/test_chained_task.py
wantsui: Thanks! I'll look into that. Maybe that's why it's failing in CI sometimes but not locally!

tests/contrib/celery/test_chained_task.py
tabgok: ok - this won't work on windows, though (see https://stackoverflow.com/questions/28551180/how-to-kill-subprocess-python-in-windows)

tests/contrib/celery/test_chained_task.py
```diff
@@ -0,0 +1,45 @@
...
+def test_task_chain_task_call_task():
+    celery_worker_cmd = "ddtrace-run celery -A tests.contrib.celery.tasks worker -c 1 -l DEBUG -n uniquename1 -P solo"
+    celery_task_runner_cmd = "ddtrace-run python tests/contrib/celery/run_tasks.py"
+    worker_process = subprocess.Popen(
+        celery_worker_cmd.split(),
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        preexec_fn=os.setsid,
+        close_fds=True,
+        cwd=str(os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(__file__))))),
```
tabgok, line 21: do you want to split and look at [-1] ?

emmettbutler (commented):
There are still some failing tests - I'll wait to review until they're passing

tests/contrib/celery/test_chained_task.py
wantsui: It looks like we call `os.killpg` in several places today: https://github.com/search?q=repo%3ADataDog%2Fdd-trace-py%20os.killpg&type=code

Is there a way to exclude some tests from running on Windows?

wantsui: Update: I think I can achieve this with just the subprocess instead: https://docs.python.org/3/library/subprocess.html#subprocess.Popen.kill, which should make it more stable in windows.

I'll see how it performs in CI though.

tests/contrib/celery/test_chained_task.py
wantsui: I ended up just updating this so that we run the test in the current directory where the test live, instead of the convoluted way of trying to go back to the `dd-trace-py` root directory.

```
    celery_worker_cmd = "ddtrace-run celery -A tasks worker -c 1 -l DEBUG -n uniquename1 -P solo"
    celery_task_runner_cmd = "ddtrace-run python run_tasks.py"

    # The commands need to run from the directory where this test file lives
    current_directory = str(os.path.dirname(__file__))

```

Resolving this.

emmettbutler (approved):

tabgok (approved):

tests/contrib/celery/test_chained_task.py
wantsui: the tests pass in CI now :)

tests/contrib/celery/test_chained_task.py
wantsui: Resolving since this is fixed in the recent commit.

Issue #10203, opened 2024-08-13
[feature] Selenium Integration


saada:
### Summary of problem

Other ecosystems like Java seem to support Selenium integrations. Can we add support in Python as well?

https://docs.datadoghq.com/tests/browser_tests/#selenium

https://github.com/DataDog/dd-trace-java/blob/d81dc1416e98a8ac88ad41578ba393c20a71c198/dd-java-agent/instrumentation/selenium/src/main/java/datadog/trace/instrumentation/selenium/SeleniumUtils.java#L72-L86

Are there ongoing plans to support this?

emmettbutler:
Thanks @saada. We'll add Selenium support to our backlog of feature requests, which means it will be included in an upcoming prioritization pass.

romainkomorndatadog:
This is on our backlog of things to do for Python test visibility.

PR #11609, opened 2024-12-02
ci: add azure and chrome to testrunner image

duncanpharvey:
This updates the `testrunner` image to:
- Install `azure-functions-core-tools-4` for serverless use
- Install `google-chrome` to support the Selenium integration's tests

Additionally, it fixes the install process for MariaDB which was breaking `apt` usage due to outdated behavior from `mariadb_repo_setup`:

```
...
   && ./mariadb_repo_setup   && apt-get install -y --no-install-recommends libmariadb-dev libmariadb-dev-compat   && rm mariadb_repo_setup   && rm -rf /var/lib/apt/lists/*:
#10 108.7 Err:1 https://dlm.mariadb.com/repo/mariadb-server/11.rolling/repo/debian buster/main amd64 mysql-common all 1:11.5.2+maria~deb10
#10 108.7   404  Not Found [IP: 104.17.191.14 443]
#10 108.9 Err:2 https://dlm.mariadb.com/repo/mariadb-server/11.rolling/repo/debian buster/main amd64 mariadb-common all 1:11.5.2+maria~deb10
#10 108.9   404  Not Found [IP: 104.17.191.14 443]
#10 108.9 E: Failed to fetch https://dlm.mariadb.com/repo/mariadb-server/11.rolling/repo/debian/pool/main/m/mariadb/mysql-common_11.5.2+maria~deb10_all.deb  404  Not Found [IP: 104.17.191.14 443]
...
```

## Checklist
- [x] PR author has checked that all the criteria below are met
- The PR description includes an overview of the change
- The PR description articulates the motivation for the change
- The change includes tests OR the PR description describes a testing strategy
- The PR description notes risks associated with the change, if any
- Newly-added code is easy to change
- The change follows the [library release note guidelines](https://ddtrace.readthedocs.io/en/stable/releasenotes.html)
- The change includes or references documentation updates if necessary
- Backport labels are set (if [applicable](https://ddtrace.readthedocs.io/en/latest/contributing.html#backporting))

## Reviewer Checklist
- [x] Reviewer has checked that all the criteria below are met 
- Title is accurate
- All changes are related to the pull request's stated goal
- Avoids breaking [API](https://ddtrace.readthedocs.io/en/stable/versioning.html#interfaces) changes
- Testing strategy adequately addresses listed risks
- Newly-added code is easy to change
- Release note makes sense to a user of the library
- If necessary, author has acknowledged and discussed the performance implications of this PR as reported in the benchmarks PR comment
[...]

PR #11610, opened 2024-12-03
feat(ci_visibility): support selenium in test visibility

romainkomorndatadog:
Introduces support for Selenium, including integration with the RUM product for session recording/replays (requested in #10203 ).

A new `hatch` environment is added which tests support for the feature in both `v1` and `v2` versions of the `pytest` plugin, with a matching addition to test suite and specs.

A few other changes are added:
- properly adding the `test.type` tag when using the test API (used by manual test runners and the new pytest plugin)
- adding the `type` tag when the test span is originally created (so that it may be accessed by the Selenium integration) 
- telemetry for events created/finished for tests is being split out into its own function because the multipurpose single function was becoming unreasonably branch-y with the addition of yet another set of tags 

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/contrib/internal/selenium/patch.py
```diff
@@ -0,0 +1,133 @@
...
+        root_span.set_tag("test.browser.driver", "selenium")
+        root_span.set_tag("test.browser.driver_version", get_version())
+
+        # Submit empty values for browser names or version if multiple are found
+        browser_name = webdriver_instance.capabilities.get("browserName")
+        browser_version = webdriver_instance.capabilities.get("browserVersion")
+
+        existing_browser_name = root_span.get_tag("test.browser.name")
+        if existing_browser_name is None:
+            root_span.set_tag("test.browser.name")
```
vitor-de-araujo, line 88: Is the `set_tag` without a value here intentional? It does not match the logic with `browser_version` below.

Also, can you clarify why this logic is needed (i.e., when does this situation happen)?

ddtrace/contrib/selenium/__init__.py
```diff
@@ -0,0 +1,33 @@
+"""
+The Selenium integration enriches Test Visibility data with extra tags and, if available,
+Real User Monitoring session replays.
+
+Enabling
+~~~~~~~~
+
+The Selenium integration is enabled by default in test contexts (eg: pytest, or unittest).Use
```
vitor-de-araujo, line 8:
```suggestion
The Selenium integration is enabled by default in test contexts (eg: pytest, or unittest). Use
```

docker/Dockerfile
```diff
@@ -59,14 +59,20 @@ RUN \
   && apt-get install -y --no-install-recommends nodejs npm \
   && npm install -g @datadog/datadog-ci \
   # Install Mariadb last because mariadb_repo_setup currently breaks apt
-  && wget https://downloads.mariadb.com/MariaDB/mariadb_repo_setup \
-  && chmod +x mariadb_repo_setup \
-  && ./mariadb_repo_setup \
+  && wget -O- https://mariadb.org/mariadb_release_signing_key.pgp |gpg --dearmor > /etc/apt/trusted.gpg.d/mariadb.gpg \
+  && echo 'deb [arch=amd64,arm64] https://mirror.mariadb.org/repo/11.rolling/debian/ buster main' > /etc/apt/sources.list.d/mariadb.list \
```
vitor-de-araujo, line 63: You should need an `apt-get update` after adding a new repo. Maybe this is installing from the Debian repos and not from the upstream one?

releasenotes/notes/ci_visibility-feat-selenium_rum_support-f1b81d0de8820aaa.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    CI Visibility: adds support for `Selenium and RUM integration <https://docs.datadoghq.com/tests/browser_tests/>`_
```
vitor-de-araujo, line 4: This is the cursedest link syntax I've ever seen.

tests/snapshots/test_selenium_chrome.test_selenium_chrome_pytest_rum_enabled.json
```diff
@@ -0,0 +1,182 @@
...
+      "test.command": "pytest --ddtrace -s",
+      "test.framework": "pytest",
+      "test.framework_version": "8.3.4",
+      "test.module": "",
+      "test.module_path": "",
+      "test.name": "test_selenium_local_pass",
+      "test.source.file": "test_selenium.py",
+      "test.status": "pass",
+      "test.suite": "test_selenium.py",
+      "test.type": "test",
```
vitor-de-araujo, line 165: Shouldn't `test.browser.driver` and friends be showing up here?

ddtrace/contrib/internal/selenium/patch.py
```diff
@@ -0,0 +1,133 @@
+import os
+import time
+import typing as t
+
+import selenium
+from selenium.webdriver.remote.webdriver import WebDriver
```
P403n1x87, line 6: Can we avoid force-importing this and doing something similar to https://github.com/DataDog/dd-trace-py/blob/23ab0e6610ee8a4a1bbd0507ade14011651ea66f/ddtrace/contrib/internal/django/patch.py#L834 instead, that is, register a "when imported" hook to do the patching when that module is actually imported by the target application.

romainkomorndatadog: Will do.

docker/Dockerfile
romainkomorndatadog: Ah, yes, that was probably taken care of by `mariadb_repo_setup` before, but regardless, we're going to move those changes to #11609 anyway.

ddtrace/contrib/internal/selenium/patch.py
romainkomorndatadog: No, it should have an empty value. This is why this PR's still in draft. 😄 I created it so I could refer to it in a comment on another PR.

tests/snapshots/test_selenium_chrome.test_selenium_chrome_pytest_rum_enabled.json
romainkomorndatadog: Yep. See earlier comment re: draft PR.

romainkomorndatadog (commented):
Going to be removing a few changes from this related to image building in favor of #11609 and follow-up PRs.

releasenotes/notes/ci_visibility-feat-selenium_rum_support-f1b81d0de8820aaa.yaml
romainkomorndatadog: Yes.

ddtrace/contrib/internal/selenium/patch.py
```diff
@@ -108,26 +135,50 @@ def _handle_enter(self):
...
+
+    @when_imported("selenium.webdriver.remote.webdriver")
+    def _(m):
+        global _WRAPPING_CONTEXTS_STORES
+        for wrapper_tup in [
+            (SeleniumGetWrappingContext, m.WebDriver.get),
+            (SeleniumQuitWrappingContext, m.WebDriver.quit),
+            (SeleniumQuitWrappingContext, m.WebDriver.close),
+        ]:
+            wrapper = wrapper_tup[0](wrapper_tup[1])
```
vitor-de-araujo, line 167: Isn't this the same as:
```suggestion
        for wrapper in [
            SeleniumGetWrappingContext(m.WebDriver.get),
            SeleniumQuitWrappingContext(m.WebDriver.quit),
            SeleniumQuitWrappingContext(m.WebDriver.close),
        ]:
```

ddtrace/contrib/selenium/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+
+The Selenium integration is enabled by default in test contexts (eg: pytest, or unittest). Use
+:func:`patch()<ddtrace.patch>` to enable the integration:
+
+    from ddtrace import patch
+    patch(selenium=True)
+
+
+When using pytest, the `--ddtrace-patch-all` flag is required in order for this integration to
+be enabled..
```
romainkomorndatadog, line 16:
```suggestion
be enabled.
```

ddtrace/contrib/selenium/__init__.py
```diff
@@ -5,12 +5,16 @@
 Enabling
 ~~~~~~~~
 
-The Selenium integration is enabled by default in test contexts (eg: pytest, or unittest).Use
-:func:`patch()<ddtrace.patch>` to enable the integration::
+The Selenium integration is enabled by default in test contexts (eg: pytest, or unittest). Use
+:func:`patch()<ddtrace.patch>` to enable the integration:
```
vitor-de-araujo, line 9: Wasn't the double colon needed?

docker/Dockerfile
```diff
@@ -53,24 +53,40 @@ RUN apt-get update \
...
 RUN if [ "$TARGETARCH" = "amd64" ]; \
-    then echo 'deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main' > /etc/apt/sources.list.d/google-chrome.list \
-        && wget -O- https://dl.google.com/linux/linux_signing_key.pub |gpg --dearmor > /etc/apt/trusted.gpg.d/google.gpg \
-        && apt update \
-        && apt install -y google-chrome-stable ; \
-    fi
+  then \
+    curl https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor > microsoft.gpg \
+    && mv microsoft.gpg /etc/apt/trusted.gpg.d/microsoft.gpg \
+    && echo "deb [arch=amd64] https://packages.microsoft.com/repos/microsoft-debian-buster-prod buster main" > /etc/apt/sources.list.d/dotnetdev.list \
```
vitor-de-araujo: Can't wait to install Microsoft Debian™ at home.

hatch.toml
```diff
@@ -473,4 +473,10 @@ test = [
 
 [[envs.selenium.matrix]]
 python = ["3.7", "3.10", "3.12"]
-selenium = ["~=3.0", "~=4.0"]
+tested_pytest_plugin_version = ["v1", "v2"]
+
+[envs.selenium.overrides]
+matrix.tested_pytest_plugin_version.env-vars = [
+    { key = "_TESTED_PYTEST_PLUGIN_VERSION", value = "false", if = ["v1"]},
```
vitor-de-araujo, line 480: Can't we just set `DD_PYTEST_USE_NEW_PLUGIN_BETA` directly here? It seems weird to have a "version" variable whose values are true and false. :thinking:

vitor-de-araujo (approved):

hatch.toml
romainkomorndatadog: Setting `DD_PYTEST_USE_NEW_PLUGIN_BETA` would set the version of the plugin that's running in the `hatch` environment, but we want to affect the version of the plugin that's being tested, and since it's being tested in a subprocess, it needs to be passed in to that subprocess' env.

Since we're running the tests with `--no-ddtrace`, I was worried that using `DD_PYTEST_USE_NEW_PLUGIN_BETA` would end up being confusing.

ddtrace/contrib/internal/selenium/patch.py
romainkomorndatadog: Ah yeah, hadn't even thought of that. Will update.

ddtrace/contrib/selenium/__init__.py
romainkomorndatadog: oh, maybe, I'll check.

P403n1x87 (approved):

ddtrace/contrib/selenium/__init__.py
```diff
@@ -0,0 +1,37 @@
+"""
+The Selenium integration enriches Test Visibility data with extra tags and, if available,
+Real User Monitoring session replays.
+
+Enabling
+~~~~~~~~
+
+The Selenium integration is enabled by default in test contexts (eg: pytest, or unittest). Use
+:func:`patch()<ddtrace.patch>` to enable the integration:
```
romainkomorndatadog, line 9:
```suggestion
:func:`patch()<ddtrace.patch>` to enable the integration::
```

quinna-h (approved):

ddtrace/contrib/internal/selenium/patch.py
```diff
@@ -0,0 +1,185 @@
...
+        time.sleep(config.selenium.flush_sleep_ms / 1000)
+
+        if is_rum_active:
+            root_span.set_tag("test.is_rum_active", "true")
+
+        webdriver_instance.delete_cookie("datadog-ci-visibility-test-execution-id")
+
+
+def get_version() -> str:
+    import selenium
```
erikayasuda, line 150: Why do we do a local import for each of the publicly exposed methods instead of globally?

ddtrace/contrib/internal/selenium/patch.py
```diff
@@ -0,0 +1,185 @@
...
+
+    try:
+        return selenium.__version__
+    except AttributeError:
+        log.debug("Could not get Selenium version")
+        return ""
+
+
+def patch() -> None:
```
quinna-h, line 159: Could we add a test that calls this patch method (I see you have a test that calls unpatch)

quinna-h (changes requested):

ddtrace/contrib/internal/selenium/patch.py
romainkomorndatadog: None of the other tests that validate the functionality work if `patch()` doesn't work. Adding a test just for `patch()` doesn't seem valuable to me.

In fact, the `unpatch()` test doesn't test that the methods are patched, it tests that, when the functions are called after unpatching, we don't continue to add tags.

ddtrace/contrib/selenium/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+
+When using pytest, the `--ddtrace-patch-all` flag is required in order for this integration to
+be enabled.
+
+Configuration
+~~~~~~~~~~~~~
+
+The Selenium integration can be configured using the following options:
+
+DD_CIVISIBILITY_RUM_FLUSH_WAIT_MILLIS: The time in milliseconds to wait after flushing the RUM session.
```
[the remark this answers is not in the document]
romainkomorndatadog: Yeah that seems wise.

romainkomorndatadog (commented):
Addressed PR comments in ef7ae79df .

ddtrace/contrib/internal/selenium/patch.py
romainkomorndatadog: I was under the impression that helps avoids having ddtrace itself import the module. Eg: [django](https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/contrib/internal/django/patch.py) does the same thing. Maybe that's a relic of past implementations, though.

quinna-h (approved):

ddtrace/contrib/internal/selenium/patch.py
erikayasuda: Ah yeah you're right. Yeah I'm not too sure which one is the "better" or "newer" way of doing this... so I don't need to block on it. I'll probably try to figure out if there's a difference and if we should be consistent though.

docs/configuration.rst
```diff
@@ -597,6 +597,17 @@ Test Visibility
...
+     type: Integer
+     default: 500
+
+     description: |
+        Configures how long, in milliseconds, the Selenium integration will wait after invoking the RUM flush function
+        during calls to the driver's ``quit()`` or ``close()`` methods. This helps ensure that the call to the
+        asynchronous function finishes before the driver is closed.
+
+     version_added:
+        v2.19.0:
```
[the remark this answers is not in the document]
romainkomorndatadog: I can't count. 😓

romainkomorndatadog, line 610:
```suggestion
        v2.18.0:
```

PR #11612, opened 2024-12-03
fix(profiling): get rid of leaky string table

nsrip-dd:
Profiler samples typically contain several strings, for sample labels
(task name, endpoint/resource, etc) and frame info. Those strings tend
to be repeated. We build samples across several API calls, from Python,
Cython, C Python extensions, and C++ code, making it hard to reason
about the lifetimes of the strings we get. So, it's easier to just copy
the strings we need while building a sample. And it's preferably to only
store one copy. So, our libdatadog wrapper uses a string table, which
addresses both of these concerns. However, nothing is ever removed from
the table. And some things like task names can have new values each
time, meaning we have unbounded memory growth.

I previously tried to fix this by periodically clearing the string
table. However, this comes with its own new problems. We can construct
samples in parallel with profile uploading. So if we try to clear the
table when we upload the profile, or really any other time, we need to
make sure in-progress samples keep their strings around until the sample
is flushed to libdatadog. And we need to handle forking so we don't leak
a whole string table on fork.

It turns out we don't really need that string table at all. Libdatadog
already interns strings once the sample makes it into the libdatadog
profile. That solves the duplicate string problem. We still want to copy
strings while building the sample, but we can just do that on a
per-sample basis, and the strings only need to be alive while we build a
sample. We can use a small amount of scratch space for each sample to
copy strings. We will likely only have a few samples in flight at any
time, and we already pool the samples. So we can reuse this scratch
memory effectively. We'll only use as memory in proportion to the
number of samples being built concurrently.

So, this commit implements a simple string arena for each sample object
to use. We may want to tweak the default buffer size and how much memory
we keep around when resetting the arena. We could even deduplicate,
since we might see some duplicates per-sample from frame info.

PROF-10824

## Checklist

## Reviewer Checklist
- Title is accurate

### DIFF
--- a/ddtrace/internal/datadog/profiling/dd_wrapper/include/sample.hpp
+++ b/ddtrace/internal/datadog/profiling/dd_wrapper/include/sample.hpp
@@ -15,6 +15,28 @@ extern "C"
 
 namespace Datadog {
 
+namespace internal {
+
+struct StringArena
+{
+    static constexpr size_t DEFAULT_SIZE = 1024;
+    // Strings are backed by fixed-size chunks. The chunks can't grow, or they'll
+    // move and invalidate pointers into the arena.  So, if we need more space, we
+    // add more chunks.
+    std::vector<std::vector<char>> chunks;
+
+    StringArena();
+    // Clear the backing data of the arena, except for a smaller initial segment.
+    // Views returned by insert are invalid after this call.
+    void reset();
+    // Copies the contents of s into the arena and returns a view of the copy in
+    // the arena. The returned view is valid until the next call to reset, or
+    // until the arena is destroyed.
+    std::string_view insert(std::string_view s);
+};
+
+} // namespace internal
+
 class SampleManager; // friend
 
 class Sample
@@ -44,6 +66,9 @@ class Sample
 
     // Additional metadata
     int64_t endtime_ns = 0; // end of the event
+
+    // Backing memory for string copies
+    internal::StringArena string_storage{};
 
   public:
     // Helpers
--- a/ddtrace/internal/datadog/profiling/dd_wrapper/include/sample.hpp
+++ b/ddtrace/internal/datadog/profiling/dd_wrapper/include/sample.hpp
@@ -17,19 +17,29 @@ namespace Datadog {
 
 namespace internal {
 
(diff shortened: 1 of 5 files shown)


ddtrace/internal/datadog/profiling/dd_wrapper/include/sample.hpp
```diff
@@ -15,6 +15,34 @@ extern "C"
...
+
+struct StringArena
+{
+    // TODO: pick a reasonable default size and explain it. Goal is to be big
+    // enough to accommodate "average" sample string data without need for much
+    // growth (new allocs), but without too much wasted space in the typical case.
+    // User-provided things like function and task names can be arbitrarily-sized.
+    // Basically want (# of frames) * (name + filename size) + (# of labels) * (reasonable label size)
+    // where the sizes are a reasonable expected value.
+    static constexpr size_t DEFAULT_SIZE = 64 * 1024;
```
taegyunkim, line 28: is there a way to look at ingested profiles (sent from python services)? it would be handy to have a statistics of strings from those and set the default based on that. 

Oh wait, this is per sample not per profile. Then, is instrumentation telemetry the only way?

taegyunkim (approved):
Overall looks good to me. Nice to remove one usage of `std::mutex` from profile.cpp, and it's easier to reason about the lifetime.

ddtrace/internal/datadog/profiling/dd_wrapper/include/sample.hpp
nsrip-dd: Yeah, good idea. I think we can compute how big this would be from a sample in a profile, since we'd have all the labels and frames. I'll collect some data and update this based on what I find 👍

nsrip-dd: I downloaded a bunch of Python relenv profiles and computed the total size of strings associated with each sample (label values and function/file names, basically). The largest I saw for a single sample was ~10000 bytes, but I think that might have been a native profile that my query incorrectly returned. Most of them were actually python profiles, though, so I think the numbers are okay. So I think 16KiB would be a reasonable size to let use usually only have one "chunk" and not need to do more allocations.

PR #11629, opened 2024-12-05
fix(profiling): recreates locks and data structures after fork

taegyunkim:
## Checklist

## Reviewer Checklist
- Title is accurate


nsrip-dd (commented):
I think it makes sense to do this but I'm wondering about the safety of it. Specifically for `string_table_mtx` (which we might get rid of) and `thread_info_map_lock`. As I understand it, while a thread forks, another thread could be in the middle of manipulating the string table or thread info map. And the child process will see those data structures in the exact state they were in from the parent. So they could be in an inconsistent state. And the next time the child process accesses the data structures (which it can since the lock is unlocked) I'm not sure what would happen... Would it be better/necessary to recreate the string table and thread info map in the child process, so they're in a clean state?

Maybe @sanchda could weigh in since he understands the forking concerns better than I do

taegyunkim:
@nsrip-dd Oh, good catch. Yeah, we also need to clear the underlying data structure to make sure we won't be in an inconsistent state. Let me update the code.

releasenotes/notes/profiling-reset-lock-51a032479617f038.yaml
```diff
@@ -0,0 +1,8 @@
+---
+fixes:
+  - |
+    Fixes an issue where calling ``unlock()`` on unlocked mutex could result in
+    an undefined behavior.
```
emmettbutler, line 5:
```suggestion
    an undefined behavior.
  - |
```

emmettbutler (approved):

nsrip-dd:
FYI I'm currently merging https://github.com/DataDog/dd-trace-py/pull/11612, which will get rid of the string table & its lock, so you'll need to update this once that lands

ddtrace/internal/datadog/profiling/stack_v2/src/sampler.cpp
```diff
@@ -66,13 +66,26 @@ _stack_v2_atfork_child()
     // The only thing we need to do at fork is to propagate the PID to echion
     // so we don't even reveal this function to the user
     _set_pid(getpid());
+    // We also need to recreate the lock used by echion
+    thread_info_map_lock.~mutex();
+    new (&thread_info_map_lock) std::mutex();
+    {
+        std::lock_guard<std::mutex> lock(thread_info_map_lock);
+        thread_info_map.clear();
```
nsrip-dd, line 74: 🤔 Is this safe? If the concern is that another thread might have been manipulating `thread_info_map` at the time of fork, couldn't this break as well?

I tried it out with this toy example: https://gist.github.com/nsrip-dd/361372bbd3882532f4fa7167a85e292b. In the toy, the parent creates a map, and then spins off a thread which continually modifies the map. Then the parent forks in a loop, and the child clears the map. With that, the child crashes consistently while clearing the map...

nsrip-dd: To capture what we discussed in person earlier, your idea for how to address this was to have a pre-fork handler in the parent take the needed locks, and then release them post-fork in the parent, so that the child will get the data structures in a clean, protected state.

Looks like we already do this for the `Uploader` object, for example: https://github.com/DataDog/dd-trace-py/blob/d234cdf2b03fc4b05d2c7305b1c689f34fe64774/ddtrace/internal/datadog/profiling/dd_wrapper/src/uploader.cpp#L162-L173.

PR #11631, opened 2024-12-05
feat(propagation): add DD_TRACE_PROPAGATION_BEHAVIOR_EXTRACT to handle x-org propagation

ZStriker19:
Configuration: DD_TRACE_PROPAGATION_BEHAVIOR_EXTRACT
Values:
 - continue (default): The tracing library continues the trace from the incoming headers, if present. Also, incoming baggage is propagated.
- restart: The tracing library always starts a new trace with a new trace-id. If there are distributed tracing headers, a span link will be added to reference the trace context. Also, incoming baggage is propagated.
- ignore: The tracing library always starts a new trace with a new trace-id without creating any span links. Also, incoming baggage is dropped.
Note: We do not need to implement this at the moment because the workaround we have in place DD_TRACE_PROPAGATION_STYLE_EXTRACT=none performs this exact function of completely ignoring incoming trace context headers and baggage headers. If we receive feedback, we can add this new option that we maintain.

Result
✅ Head services: Customers can configure these services with the restart configuration so that they are always the root span, regardless of incoming distributed tracing headers.
✅ Internal services: By default, internal services will remain unchanged and continue to report complete traces to Datadog. If a user has configured an internal service with the restart configuration, this will lead to partial traces. Thankfully with Span Links we can easily direct the customer to the upstream trace (if it was sampled).
❌ Sometimes-head services: This solution does not solve for services that are sometimes head services and sometimes not. The best solution for these services is to set the restart configuration so that a new trace can begin and sampling priority can be recalculated, and in the case that this was not a head service then the upstream service can be found with a span link.
❌ Reentrant systems: This solution does not improve this scenario no matter how the reentrant service is configured. Under the continue configuration, the reentrant service is part of the same trace as the original root span but it is orphaned. Under the restart configuration, a new trace will be started and the span will be decorated with a span link containing the original trace-id but the contained span-id would not point to a span tracked in their Datadog organization.

Note: This contains a fix, allowing only baggage to be propagated.

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/propagation/http.py
```diff
@@ -1129,27 +1134,45 @@ def my_controller(url, headers):
...
             if config._propagation_extract_first:
-                # loop through the extract propagation styles specified in order, return whatever context we get first
+                # use only first propagation style specified, even if it doesn't find a context
+                # we don't try the other styles
                 for prop_style in config._propagation_style_extract:
                     propagator = _PROP_STYLES[prop_style]
-                    context = propagator._extract(normalized_headers)
+                    rec_context = propagator._extract(normalized_headers)
+                    # if configured to restart, append our first context to the current context as a span link
+                    if rec_context:
```
ZStriker19, line 1149: Slightly changed behavior here so that the context never is set to `None`. If this goes against spec I can change back.

.github/workflows/system-tests.yml
```diff
@@ -20,6 +20,7 @@ jobs:
         with:
           persist-credentials: false
           repository: 'DataDog/system-tests'
+          ref: 'zach.montoya/poc/propagation-behavior'
```
ZStriker19, line 23: Going to remove these, just doing because struggling to build tracer locally

ZStriker19:
Ignore the failing system-test, it's just due to telemetry config differences since we added a new config.

ddtrace/_trace/tracer.py
```diff
@@ -773,10 +773,8 @@ def _start_span(
 
         # Update the service name based on any mapping
         service = config.service_mapping.get(service, service)
-
         links = context._span_links if not parent else []
-
-        if trace_id:
+        if trace_id or links:
```
ZStriker19, line 777: Since we now count just span_links as a non-empty context, we need to make sure we initialize the span with them if they're available. This seems a little worrisome because we rely on the span __init__ to make the right choices when we pass in None for some of these values. However, it does work.

ddtrace/_trace/tracer.py
```diff
@@ -775,8 +775,7 @@ def _start_span(
         service = config.service_mapping.get(service, service)
 
         links = context._span_links if not parent else []
-
-        if trace_id:
+        if trace_id or links:
```
mabdinur, line 778: Should this also consider baggage here? ie: we should propagate a context if it has trace info, span links, and/or baggage. A comment would be useful here too. 

```suggestion
        if trace_id or links or context._baggage:
```

mabdinur: Also including links here seems like a separate bug fix, should we call this out in a release note?

ddtrace/contrib/trace_utils.py
```diff
@@ -577,7 +577,10 @@ def activate_distributed_headers(tracer, int_config=None, request_headers=None,
         #     app = Flask(__name__)  # Traced via Flask instrumentation
         #     app = DDWSGIMiddleware(app)  # Extra layer on top for WSGI
         current_context = tracer.current_trace_context()
-        if current_context and current_context.trace_id == context.trace_id:
+
+        # We accept incoming contexts with only baggage or only span_links, however if we
+        # already have a current_context then a baggage only or span_link only context will be tossed out
+        if current_context and (not context.trace_id or current_context.trace_id == context.trace_id):
```
mabdinur, line 583: With the proposed logic 
if:
 - we have a current_context AND 
 - the extracted context does not have a trace_id 
then:
- we deactivate current_context and replace it with the new context

This feels a bit off. Instead I think we should:
- keep the current_context active AND
- add the baggage and span links in the new context to the current context. 

Activating a context with no trace_id will break the existing trace

ddtrace/propagation/http.py
```diff
@@ -988,6 +990,24 @@ def _extract_configured_contexts_avail(normalized_headers):
                 styles_w_ctx.append(prop_style)
         return contexts, styles_w_ctx
 
+    @staticmethod
+    def _context_to_span_link(context: Context, style: str, reason: str) -> Optional[SpanLink]:
+        # encoding expects at least trace_id and span_id
+        if context.span_id and context.trace_id:
```
mabdinur, line 996: What do other languages do here? We should ensure this behavior aligned across tracers. Either we need to allow span links with Trace_ID=0 and SPAN_ID=0 (to record conflicting/ignored baggage headers) or we need a well defined way to merge baggage headers in one format (tracecontext) to another (x-datadog headers).

ddtrace/propagation/http.py
```diff
@@ -1130,24 +1141,29 @@ def my_controller(url, headers):
         :param dict headers: HTTP headers to extract tracing attributes.
         :return: New `Context` with propagated attributes.
         """
-        if not headers:
-            return Context()
+        context = Context()
+        if not headers or config._propagation_behavior_extract == _PROPAGATION_BEHAVIOR_IGNORE:
```
mabdinur, line 1145: Do we need to do this here? 

It will be more performant if we check if `config._propagation_behavior_extract == _PROPAGATION_BEHAVIOR_IGNORE` in `ddtrace.settings.config.py`. This way we can only do it once. Then we disable propagation in one place and log a warning if a customer is setting conflicting settings.

ddtrace/_trace/tracer.py
ZStriker19: I think previously we never would've gotten here with a context that just had a span link via propagation. However I guess you're saying in case users are trying to create a span from a context with just a span link manually? I think that this is such an obscure edge case that it doesn't need a release note but if you disagree I can add.

docs/configuration.rst
```diff
@@ -368,6 +368,26 @@ Trace Context propagation
...
+         After extracting the headers for propagation, this configuration determines what is done with them.
+
+         The default value is ``continue`` which always propagates valid headers.
+         ``ignore`` ignores all incoming headers and ``restart`` turns the first extracted valid propagation header 
+         into a span link and propagates baggage if present.
+
+         Example: ``DD_TRACE_PROPAGATION_STYLE_EXTRACT="ignore"`` to ignore all incoming headers and to start a root span without a parent.
+
+     version_added:
+       v2.20.0:
```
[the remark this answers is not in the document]
erikayasuda: If this isn't gonna be merged by Wed, it'll probably have to wait for 2.20+. Otherwise, I think we can get it in for 2.20rc2. The goal is to be able to promote an rc to patch release by Monday

ddtrace/contrib/trace_utils.py
ZStriker19, line 583: Thanks for the catch, fixing and updating comment

erikayasuda (approved):
Some questions for my own clarity. Otherwise LGTM from LP side

ddtrace/contrib/trace_utils.py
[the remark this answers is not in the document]
ZStriker19: context.trace_id can either be None or it can be an actual trace_id. I'm checking it so we don't compare None to an int. Also about to change the above statement since Munir was correct and it needs to be modified.

ZStriker19: However will continue to throw out an incoming context that lacks a trace_id as the actions around merging baggage and appending span links etc could lead to a mess and I think is basically never what someone would expect to happen.

ddtrace/propagation/http.py
ZStriker19: Since baggage is evaluated separately, it's always added the final context we decide on, whether it has span links decided through `DD_TRACE_PROPAGATION_STYLE=restart`, or through conflicting incoming contexts. So I don't think this is actually an issue. If a context is formed from a propagation style other than baggage, it'll always have a trace_id at least.

docs/configuration.rst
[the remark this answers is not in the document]
ZStriker19: Thanks for the timeline info! If it misses 2.20rc2 then I'll update it.

mabdinur (approved):
Overall looks good to me. We need to confirm with @zacharycmontoya that the distributed tracing edge cases uncovered in this PR are documented somewhere. We should align one implementation across all tracers.

releasenotes/notes/propagation_behavior_extract-3d16765cfd07485b.yaml
```diff
@@ -0,0 +1,8 @@
+---
+features:
+  - |
+    propagation: Introduces the environment variable ``DD_TRACE_PROPAGATION_BEHAVIOR_EXTRACT``
```
mabdinur, line 4: We should also note that this PR introduces support for propagating baggage when trace headers are not received (we could also move this fix to it's own PR).

rachelyangdog (approved):

rachelyangdog:
In terms of baggage, I think it looks good!

mabdinur (approved):

PR #11662, opened 2024-12-10
fix(llmobs): do not ignore global patch env vars

### LINKED ISSUE #11639: LLMObs: DD_PATCH_MODULES is not taken into account
Hello, when enabling LLMObs with an env variable, it appears that `DD_PATCH_MODULES` configuration is at least partially ignored.

With the following docker setup:
```
ENV DD_PATCH_MODULES=langchain:false
ENV DD_LLMOBS_ENABLED=true
CMD ["ddtrace-run", "python", "-u", "service.py"]
```

The service will patch `langchain` on start-up. Changing `DD_LLMOBS_ENABLED` to false fixes the issue.

I haven't fully figured out the code, but I think it is because LLMObs calls `patch` on all supported integrations ([code](https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/llmobs/_llmobs.py#L351)), and this method doesn't check env variables. Changing it to [patch_all](https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/_monkey.py#L217) may fix the issue, but I am unsure of side effects

Yun-Kim:
Fixes #11639.

This PR fixes an issue where `LLMObs.enable()` ignored/overrode global patch env vars including `DD_TRACE_<INTEGRATION>_ENABLED` and `DD_PATCH_MODULES`.

## Checklist

## Reviewer Checklist
- Title is accurate


emmettbutler (approved):

ddtrace/llmobs/_llmobs.py
```diff
@@ -347,8 +348,18 @@ def flush(cls) -> None:
...
+        integrations_to_patch = {integration: True for integration in SUPPORTED_LLMOBS_INTEGRATIONS.values()}
+        dd_patch_modules = os.getenv("DD_PATCH_MODULES")
+        dd_patch_modules_to_str = parse_tags_str(dd_patch_modules)
+        integrations_to_patch.update(
+            {k: asbool(v) for k, v in dd_patch_modules_to_str.items() if k in SUPPORTED_LLMOBS_INTEGRATIONS.values()}
+        )
+        for module, _ in integrations_to_patch.items():
+            env_var = "DD_TRACE_%s_ENABLED" % module.upper()
+            if env_var in os.environ:
+                integrations_to_patch[module] = asbool(os.environ[env_var])
```
sabrenner, line 361: probably a very unlikely scenario, but (more for my own understanding) this means that we discard whatever was set from `DD_PATCH_MODULES` in some cases, ie
```
DD_PATCH_MODULES=openai:False
DD_TRACE_OPENAI_ENABLED=True
```
we would enable OpenAI, even though it is set to `False` from `DD_PATCH_MODULES`. is that the same behavior with the current patching logic in `patch_all` (we merge the values instead of going with `False` if even one is `False`)? again probably a very unlikely case lol (i don't know why someone would have two different values explicitly), but I'm curious if we're consistent with that behavior

sabrenner (commented):
lgtm, thanks for fixing! just left one clarification question. also wondering if we should add a small test for `DD_TRACE_<INTEGRATION>_ENABLED` as well, but the logic does seem like it'd work.

it might be worth leaving a small note in our public docs that the env vars here will take precedence over the options passed into `LLMObs.enable()`, but up to you!

ddtrace/llmobs/_llmobs.py
Yun-Kim: In the Python tracer, `DD_PATCH_MODULES` overrides `DD_TRACE_%_ENABLED` (see [preload](https://github.com/DataDog/dd-trace-py/blob/e04c8acf72186e63998cb4c97df0d3708aaeb15d/ddtrace/bootstrap/preload.py#L114-L117) and [patch_all](https://github.com/DataDog/dd-trace-py/blob/e04c8acf72186e63998cb4c97df0d3708aaeb15d/ddtrace/_monkey.py#L230)). I'll follow the same pattern to avoid inconsistencies. Thanks for the call out!

Yun-Kim:
> it might be worth leaving a small note in our public docs that the env vars here will take precedence over the options passed into LLMObs.enable(), but up to you!

I'm leaning towards passing on this for now as ideally LLM Obs users shouldn't need to mess with these APM settings and this will probably add more complication to our setup instructions. If we run into specific use cases we can revisit adding this warning. WDYT?

sabrenner (approved):
> I'm leaning towards passing on this for now as ideally LLM Obs users shouldn't need to mess with these APM settings and this will probably add more complication to our setup instructions. If we run into specific use cases we can revisit adding this warning. WDYT?

I think that's fair 😄 LGTM!

PR #11669, opened 2024-12-11
refactor(trace): make byte code injection more generic than coverage reporting for Python 3.10, 3.11 (AIDM-380)

labbati:
The current code we have in `ddtrace/internal/coverage` to inject invocation of callback presents the following limitations:
- very specific to coverage reporting: it hard codes the fact that _every_ line should be traced.
- drastically different implementations for 3.10, 3.11 that makes. The team is leaning toward refactoring all to the 3.10 approach which has been measured to be more performant (past work, out of scope for this PR).
- lack (or at least not evident) of testing for a tricky scenario for injection which is try/except/finally blocks, where it is extremely important to not mess around with the line table and the exception table (py 3.11) that are used to determine which code lines are executed in case of exceptions.
- for py3.11, some line table adjustments weren't working for specific patterns (added in tests)

At the same time, we need to capability to inject at some desired code location the invocation of a callback for a new functionality that we are currently adding: automatic reporting of handled exceptions. The only thing that really changes, in this case, is the selection of the injection points.

This PR does (for py 3.10 and 3.11):
- move the business logic for injection, as a generic abstraction, from `ddtrace.internal.coverage` to a more generic `ddtrace.internal.bytecode_injection`.
- add extensive testing specific to byte code manipulation (including opcode, argument, jump, exception verification).
- replace coverage's 3.10 usage of line injection with this more generic version.
- fixes a few inaccuracies with line calculation under specific conditions (e.g. in some cases when there are try/except blocks).

## Checklist

## Reviewer Checklist
- Title is accurate


P403n1x87:
Is the more generic [`inject_hook`](https://github.com/DataDog/dd-trace-py/blob/54b60764e4dc386594ca774eec94dcf4b7c88298/ddtrace/internal/injection.py#L183) not good enough in terms of implementation/performance for the intended use? I'm asking to check whether we can avoid extra complexity in the coverage code, which is quite an ad-hoc implementation.

labbati:
> Is the more generic [`inject_hook`](https://github.com/DataDog/dd-trace-py/blob/54b60764e4dc386594ca774eec94dcf4b7c88298/ddtrace/internal/injection.py#L183) not good enough in terms of implementation/performance for the intended use? I'm asking to check whether we can avoid extra complexity in the coverage code, which is quite an ad-hoc implementation.

On a large number of modules to be instrumented, using `bytecode` (used by `inject_hook`) it was causing a x10 startup overhead (200ms --> 2s). By profiling, I tracked it down to the overhead added by `bytecode`.

Following up in DM with more details.

tests/internal/bytecode_injection/test_core.py
```diff
@@ -4,7 +4,7 @@
 import pytest
 
 
-if sys.version_info[:2] >= (3, 10):
+if sys.version_info[:2] >= (3, 10) and sys.version_info[:2] < (3, 2):
```
labbati, line 7:
```suggestion
if sys.version_info[:2] >= (3, 10) and sys.version_info[:2] < (3, 12):
```

dubloom: Thank you

ddtrace/internal/bytecode_injection/core.py
```diff
@@ -0,0 +1,615 @@
...
+import sys
+from types import CodeType
+import typing as t
+
+
+CallbackType = t.Callable[[t.Any], t.Any]
+
+# This is primarily to make mypy happy without having to nest the rest of this module behind a version check
+# NOTE: the "prettier" one-liner version (eg: assert (3,11) <= sys.version_info < (3,12)) does not work for mypy
+assert sys.version_info >= (3, 10) and sys.version_info < (3, 12)  # nosec
```
P403n1x87, line 12:
```suggestion
assert (3, 10) <= sys.version_info < (3, 12)  # nosec
```

ddtrace/internal/bytecode_injection/core.py
P403n1x87, line 1: We already have some bytecode injection utilities in `ddtrace.internal.injection`. Can we perhaps move this source to `ddtrace/internal/injection/core.py` and move `ddtrace/internal/injection.py` to `ddtrace/internal/injection/__init__.py` for now?

P403n1x87 (commented):
High-level LGTM. Because bytecode manipulations come with the high risk of introducing segmentation faults, I think we must make sure that we are not breaking code objects after instrumentation. I would suggest adding tests similar to the ones run by Dynamic Instrumentation against some framework test suites. There we collect all the code objects on import and inject a hook call at every line. See https://github.com/Datadog/dd-trace-py/blob/79ea26feec8176f80676565279d6da8c16928608/tests/debugging/exploration and https://github.com/DataDog/dd-trace-py/blob/35ce9393e3d1ee8570e0c657f7a674d4bece0489/.github/workflows/test_frameworks.yml#L109-L110

.github/workflows/test_frameworks.yml
```diff
@@ -44,10 +44,6 @@ jobs:
             profiling: 1
             iast: 0
             appsec: 0
-          - suffix: IAST
-            profiling: 0
-            iast: 1
-            appsec: 0
```
P403n1x87, line 50: Was this causing issues? Maybe worth checking with the IAST team if so

PR #11695, opened 2024-12-12
chore(profiling): ddup.upload() optionally gets tracer object

taegyunkim:
profiling tests that check for span ids are flaky. I wonder whether sharing tracer instance across tests causes that problem. 

## Checklist

## Reviewer Checklist
- Title is accurate

### DIFF
--- a/ddtrace/internal/datadog/profiling/ddup/_ddup.pyi
+++ b/ddtrace/internal/datadog/profiling/ddup/_ddup.pyi
@@ -3,6 +3,7 @@ from typing import Optional
 from typing import Union
 from .._types import StringType
 from ddtrace._trace.span import Span
+from ddtrace._trace.tracer import Tracer
 
 def config(
     env: StringType,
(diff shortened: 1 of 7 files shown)


nsrip-dd (approved):
Sorry if I'm missing something obvious, but for the tests, like `def test_push_span(stack_v2_enabled, tmp_path, tracer)`, where does the `tracer` value come from? Does adding that parameter give each test its own new tracer instance?

Approving since the tests pass, so evidently it works, just curious _how_ it works :)

taegyunkim:
> Sorry if I'm missing something obvious, but for the tests, like `def test_push_span(stack_v2_enabled, tmp_path, tracer)`, where does the `tracer` value come from? Does adding that parameter give each test its own new tracer instance?
> 
> Approving since the tests pass, so evidently it works, just curious _how_ it works :)

Yes, each test gets its own new tracer instance. It's a pytest fixture, https://docs.pytest.org/en/6.2.x/fixture.html, and is defined here
https://github.com/DataDog/dd-trace-py/blob/965af0821d82088ac4570042a2a445c1e682b164/tests/profiling_v2/collector/conftest.py#L6-L8

code navigation/intellisense don't work on those and it's a little hard to find the definition..

PR #11730, opened 2024-12-13
feat(langgraph): submit spans from langgraph to APM and LLMObs

sabrenner:
This PR adds a simple integration and corresponding LLMObs plugin for the LangGraph framework.

### Enablement

For both APM and LLMObs, this integration is disabled by default, and only enabled with `DD_TRACE_LANGGRAPH_ENABLED=true` or `patch(langgraph=True)`, along with `_DD_TRACE_LANGGRAPH_ENABLED=true`, due to its experimental nature.

### APM Tracing

The APM integration adds basic tracing for LangGraph `Pregel` graph invocations, as well as the execution of each node/step in the graph (traced as `RunnableSeq`).

### LLM Observability Plugin

The LLMObs plugin submits the traces to LLM Observability.

#### Graph Invocation/Streaming Output Tagging

Because we patch the `(a)stream` method for both streams and `(a)invoke`, since they rely on `(a)stream` under the hood, we are bound to the yields from the stream. The object shape and value depend on the `stream_mode` passed in to the stream function. For `invoke`, it defaults to `"values"`, which is perfect for our use case. However, if one were to call `stream` directly, and either pass in nothing or something like `stream_mode=["updates", "debug"]`, the output we would have for the graph span would appear as:

```
{
  "type": "task_result",
  "timestamp": "2025-01-03T14:45:26.087608+00:00",
  "step": "3",
  "payload": {
    "id": "399cfe0a-aefd-f67c-6909-fcea80ef508b",
    "name": "e",
    "error": "None",
    "result": [
      "('a_list', ['e'])"
    ],
    "interrupts": []
  }
}
```

This is a known limitation for now, and a more thought-out solution for either adding on `"values"` to `stream_mode` and omitting their yields if it wasn't already present, or having some kind of clever joining, can be added in a later PR or by feature request.

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/contrib/internal/langgraph/patch.py
```diff
@@ -0,0 +1,178 @@
...
+def patch():
+    if getattr(langgraph, "_datadog_patch", False):
+        return
+    
+    langgraph._datadog_patch = True
+
+    from langgraph import graph
+
+    Pin().onto(langgraph)
+    integration = LangGraphIntegration(integration_config=config.langgraph) # make integration
```
sabrenner:
```suggestion
    integration = LangGraphIntegration(integration_config=config.langgraph)
```

ddtrace/contrib/internal/langgraph/patch.py
```diff
@@ -0,0 +1,152 @@
...
+config._add(
+    "langgraph",
+    {
+        "span_prompt_completion_sample_rate": float(os.getenv("DD_VERTEXAI_SPAN_PROMPT_COMPLETION_SAMPLE_RATE", 1.0)),
+        "span_char_limit": int(os.getenv("DD_VERTEXAI_SPAN_CHAR_LIMIT", 128)),
+    },
+)
+
+# @with_traced_module
+# def traced_runnable_seq_invoke(langgraph, pin, func, instance, args, kwargs):
```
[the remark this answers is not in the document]
sabrenner: yea just kept it in case we wanted to use it but we're probably good to get rid of it 👍

ddtrace/contrib/internal/langgraph/patch.py
```diff
@@ -0,0 +1,152 @@
...
+@with_traced_module
+def traced_runnable_callable_invoke(langgraph, pin, func, instance, args, kwargs):
+    # used for tracing function executions
+    node_name = instance.name
+    integration: LangGraphIntegration = langgraph._datadog_integration
+
+    # inputs = get_argument_value(args, kwargs, 0, "input")
+    config = get_argument_value(args, kwargs, 1, "config")
+    metadata = config.get("metadata", {}) if isinstance(config, dict) else {}
+    if node_name in ("_write", "_route", "_control_branch") or metadata.get("visited", False):
```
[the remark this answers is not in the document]
sabrenner: it's the janky way of saying the "node" associated with this config has already been traced. from what i've seen, the config object will be shared for all parts/`RunnableCallable`s of a sequence, so `_write`, `_branch`, `_control_branch`, and then the node itself and any routing function are included. we exclude the special case ones starting with `_`, but we don't want to trace the routing function. so, i opted for now to add `metadata["visited"]` to hint that if there is another non-special-cased callable as part of the runnable sequence for the node invoke, that we don't want to trace it.

but, and i think there's a comment below in the code for it, it really just means we need a better way of identifying the actual function for the node vs the function called for routing.

ddtrace/contrib/internal/langgraph/patch.py
```diff
@@ -0,0 +1,152 @@
...
+#     print("running node", steps[0].name, steps[-1].name)
+
+#     result = func(*args, **kwargs)
+#     # print("---traced_runnable_seq_invoke end---")
+#     return result
+
+
+@with_traced_module
+def traced_runnable_callable_invoke(langgraph, pin, func, instance, args, kwargs):
+    # used for tracing function executions
```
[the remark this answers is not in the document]
sabrenner: yep good callout!

ddtrace/llmobs/_integrations/langgraph.py
```diff
@@ -0,0 +1,168 @@
...
+    ):
+        if not self.llmobs_enabled:
+            return
+
+        inputs = get_argument_value(args, kwargs, 0, "input")
+        span_name = kw.get("name", span.name)
+
+        span._set_ctx_items(
+            {
+                SPAN_KIND: "agent",  # should nodes be workflows? should it be dynamic to if a subgraph is included?
```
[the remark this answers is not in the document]
sabrenner: yep. i think what could be confusing though is that a `tool` node might (will? could? should??) call a LangChain `tool` under the hood, which we already trace as a `tool`, so maybe in that case we'd want the tool node to be a workflow, and the underlying tool call to be the tool, if that makes sense

ddtrace/contrib/internal/langgraph/patch.py
```diff
@@ -0,0 +1,152 @@
...
+
+    try:
+        result = func(*args, **kwargs)
+    except Exception:
+        span.set_exc_info(*sys.exc_info())
+        integration.metric(span, "incr", "request.error", 1)
+        raise
+    finally:
+        integration.llmobs_set_tags(
+            span, args=args, kwargs=kwargs, response=result, operation="graph", name=instance.name
```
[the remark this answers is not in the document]
sabrenner: i had that originally 😆 i'll add it back, wasn't sure what looked best lol

ddtrace/llmobs/_integrations/langgraph.py
```diff
@@ -0,0 +1,168 @@
...
+                INPUT_VALUE: inputs,
+                OUTPUT_VALUE: response,
+                NAME: span_name,
+            }
+        )
+
+        if operation != "node":
+            return  # we set the graph span links in handle_pregel_loop_tick
```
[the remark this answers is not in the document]
sabrenner: it doesn't, but i did do some `span._set_ctx_item` for `SPAN_LINKS` for the graph, specifically for the case where we don't have any more tasks to queue up, we need to link all of the finished tasks to the graph's output (`output -> output`)

ddtrace/llmobs/_integrations/langgraph.py
```diff
@@ -0,0 +1,168 @@
...
+from ddtrace.llmobs._constants import OUTPUT_VALUE
+from ddtrace.llmobs._constants import SPAN_KIND
+from ddtrace.llmobs._constants import SPAN_LINKS
+from ddtrace.llmobs._integrations.base import BaseLLMIntegration
+from ddtrace.llmobs._utils import _get_llmobs_parent_id
+from ddtrace.llmobs._utils import _get_nearest_llmobs_ancestor
+from ddtrace.span import Span
+
+
+node_invokes: Dict[str, Any] = {}
```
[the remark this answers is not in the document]
sabrenner: also concerns I had, sounds good 😄

ddtrace/llmobs/_integrations/langgraph.py
```diff
@@ -0,0 +1,168 @@
...
+
+        span_links = (
+            [
+                {
+                    "span_id": str(_get_llmobs_parent_id(span)) or "undefined",
+                    "trace_id": "{:x}".format(span.trace_id),
+                    # we assume no span link means it is the first node of a graph
+                    "attributes": {
+                        "from": "input",
+                        "to": "input",
```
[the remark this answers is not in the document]
sabrenner: that's the assumption for now (i think we should try and challenge that to make sure)

ddtrace/contrib/internal/langgraph/patch.py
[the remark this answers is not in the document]
Yun-Kim: Does this also trace conditional routing functions?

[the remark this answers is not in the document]
sabrenner: no, that is ignored by the 
```python
if node_name in ("_write", "_route", "_control_branch") or metadata.get("visited", False):
```
check. But, we can definitely leverage that here if/when we want to.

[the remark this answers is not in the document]
sabrenner: added some rough docstrings that have our current understanding of how we're interacting with LangGraph. i figure with more offline discussions we can expand as needed on them

ddtrace/llmobs/_integrations/langgraph.py
```diff
@@ -0,0 +1,185 @@
...
+        """
+        if not self.llmobs_enabled:
+            return
+
+        graph_span = (
+            tracer.current_span()
+        )  # since we're running the the pregel loop, and not in a node, the graph span should be the current span
+        graph_caller = _get_nearest_llmobs_ancestor(graph_span) if graph_span else None
+
+        if not more_tasks and graph_span is not None:
```
Yun-Kim, line 101: The graph is finished, we are now setting span links from the overarching graph span to the outputs of this graph.

ddtrace/contrib/internal/langgraph/patch.py
```diff
@@ -0,0 +1,153 @@
...
+    """
+    Trace the invocation of a Pregel (CompiledGraph) instance.
+    This operation represents the parent execution of an individual graph.
+    This graph could be standalone, or embedded as a subgraph in a node of a larger graph.
+    Under the hood, this graph will `tick` through until all computed tasks are completed.
+    """
+    integration: LangGraphIntegration = langgraph._datadog_integration
+    span = integration.trace(
+        pin,
+        "%s.%s.%s" % (instance.__module__, instance.__class__.__name__, instance.name),
```
Yun-Kim, line 87: Let's get a better name for this other than `LangGraph`

ddtrace/_monkey.py
```diff
@@ -146,6 +147,7 @@
     "httplib": ("http.client",),
     "kafka": ("confluent_kafka",),
     "google_generativeai": ("google.generativeai",),
+    "langgraph": ("langgraph.graph",),
```
Yun-Kim, line 150: Will this not be patched if someone imports `langgraph.pregel/func/channels/prebuilt` or other top-level langgraph directories?

sabrenner: ah hmm i suppose so, i forgot i had this here. without this, the patching was never triggered - i only added it in after trying to figure out, without too much understanding, how exactly patching is triggered in python. so, i'm not sure what i'd change this to here

i think i might defer to you as i don't have too much knowledge on how the patching hooks are triggered 😅 but we can run thru it offline too, i'd love to learn how it works.

ddtrace/llmobs/_integrations/langgraph.py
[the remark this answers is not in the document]
Yun-Kim: Going to default to `task` if a runnable sequence (node)

ddtrace/contrib/internal/langgraph/patch.py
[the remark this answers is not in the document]
Yun-Kim: Update: not needed anymore as we are tracing RunnableSeq, for which the first step will almost always be a node.

ddtrace/_monkey.py
Yun-Kim: Yeah let's talk about this offline once you're back, happy to walk through patching hooks and underlying instrumentation logic in Python 😄

ddtrace/_monkey.py
```diff
@@ -48,6 +48,7 @@
     "grpc": True,
     "httpx": True,
     "kafka": True,
+    "langgraph": False,
```
sabrenner, line 51: defaulting this to `False`, as it's an experimental integration and we want it gated behind explicit enablement (via env var or direct `patch`)

tests/llmobs/suitespec.yml
```diff
@@ -92,3 +95,16 @@ suites:
...
+      - '@bootstrap'
+      - '@core'
+      - '@tracing'
+      - '@contrib'
+      - '@langgraph'
+      - '@requests'
+      - '@llmobs'
+      - tests/contrib/langgraph/*
+    runner: riot
+    snapshot: false
```
Yun-Kim, line 110: Let's set this to true, I believe the patch test case for get_version() requires a testagent

erikayasuda (approved):
LGTM for guild/core files 👍

tests/contrib/langgraph/test_langgraph_llmobs.py
```diff
@@ -0,0 +1,187 @@
...
+            assert span_link["attributes"] == {"from": from_io, "to": to_io}
+            found = True
+            break
+    assert found
+
+
+@pytest.mark.parametrize(
+    "ddtrace_global_config",
+    [{"_llmobs_enabled": True, "_api_key": "<not-a-real-key>", "_llmobs_ml_app": "unnamed-ml-app"}],
+)
```
Yun-Kim, line 23: I know I wrote this but... is there a way we can just write a default fixture to include these config values for all of the below tests and avoid having to write all of the tests in a test class?

ddtrace/contrib/internal/langgraph/patch.py
```diff
@@ -0,0 +1,226 @@
...
+
+
+config._add("langgraph", {})
+
+
+@with_traced_module
+def traced_runnable_seq_invoke(langgraph, pin, func, instance, args, kwargs):
+    """
+    Traces an invocation of a RunnableSeq, which represents a node in a graph.
+    Although this API is usable elsewhere, internal to LangGraph it is used to represent the
```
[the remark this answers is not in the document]
sabrenner: yeah it is - not really sure why I added this line in the comment, idt it adds a ton of value. will remove this part of the comment.

ddtrace/contrib/internal/langgraph/patch.py
```diff
@@ -0,0 +1,226 @@
...
+            except Exception:
+                span.set_exc_info(*sys.exc_info())
+                span.finish()
+                raise
+
+    return _astream()
+
+
+@with_traced_module
+def patched_pregel_loop_tick(langgraph, pin, func, instance, args, kwargs):
```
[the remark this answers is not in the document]
sabrenner: nope, only `loop.tick` sync. there is an [AsyncPregelLoop](https://github.com/langchain-ai/langgraph/blob/a03f1f74692ec078a7affa57415744df5950ed2c/libs/langgraph/langgraph/pregel/loop.py#L901), but it only implements async context manager methods, looks like extends the sync `PregelLoop` and doesn't change `tick` at all

ddtrace/_monkey.py
Kyle-Verhoog: I'd rather we omit it from being in `patch_all()` and use `_DD_TRACE_LANGGRAPH_ENABLED` just to be extra clear that it's experimental

tests/contrib/langgraph/conftest.py
```diff
@@ -0,0 +1,147 @@
...
+        import langgraph
+
+        pin = Pin.get_from(langgraph)
+        pin.override(langgraph, tracer=mock_tracer)
+        yield langgraph
+        unpatch()
+
+
+@pytest.fixture
+def mock_llmobs_writer():
```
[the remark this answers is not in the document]
sabrenner: re-worked the `conftest` a bit to include this a simplify it a little, let me know what you think! https://github.com/DataDog/dd-trace-py/pull/11730/commits/0370e96b7cb19c951a2d3c4bd731ddf0240f982d

tests/contrib/langgraph/test_langgraph_llmobs.py
sabrenner: attempted in https://github.com/DataDog/dd-trace-py/commit/0370e96b7cb19c951a2d3c4bd731ddf0240f982d

ddtrace/_monkey.py
sabrenner: added in https://github.com/DataDog/dd-trace-py/pull/11730/commits/ec996b0fd2c2daa60e9f639d8c832dcd2339914f, you'd now need to either do `patch(langgraph=True)` or `DD_TRACE_LANGGRAPH_ENABLED=1` **along with** `_DD_TRACE_LANGGRAPH_ENABLED=1` in order for patching to happen.

ddtrace/llmobs/_integrations/langgraph.py
```diff
@@ -0,0 +1,174 @@
...
+        self, finished_tasks: dict, next_tasks: dict, more_tasks: bool, is_subgraph_node: bool = False
+    ):
+        """Compute incoming and outgoing span links between finished tasks and queued tasks in the graph."""
+        if not self.llmobs_enabled:
+            return
+        graph_span = (
+            tracer.current_span()
+        )  # we're running between nodes, so the current span should be the pregel graph
+        if graph_span is None or graph_span.span_type != SpanTypes.LLM:
+            return
```
Yun-Kim, line 72: Just curious here - should this be an LLMOBS-specific span? If not, we would be creating a span link with a regular APM span right?
I think I could see a case if a user manually wraps/starts a span between the graph and node execution then we would break. Once the context management PR is merged, should we change this logic to specifically look for the active LLMObs span via `LLMObs._current_span()`?

ddtrace/llmobs/_integrations/langgraph.py
```diff
@@ -0,0 +1,174 @@
...
+                OUTPUT_VALUE: format_langchain_io(response),
+                NAME: self._graph_nodes_by_task_id.get(instance_id, {}).get("name") or kwargs.get("name", span.name),
+                SPAN_LINKS: current_span_links + span_links,
+            }
+        )
+        if operation == "graph" and not _is_subgraph(span):
+            self._graph_nodes_by_task_id.clear()
+
+    def llmobs_handle_pregel_loop_tick(
+        self, finished_tasks: dict, next_tasks: dict, more_tasks: bool, is_subgraph_node: bool = False
```
Yun-Kim, line 63: Question - do we need `more_tasks` to be passed? Can we not just check if `len(next_tasks) == 0`?

ddtrace/llmobs/_integrations/utils.py
```diff
@@ -163,3 +166,41 @@ def get_system_instructions_from_google_model(model_instance):
...
+
+
+LANGCHAIN_ROLE_MAPPING = {
+    "human": "user",
+    "ai": "assistant",
+    "system": "system",
+}
+
+
+def format_langchain_io(
```
Yun-Kim, line 178: These helpers look the same as the ones in `langchain.py` ([ref](https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/llmobs/_integrations/langchain.py#L540-L568)) - can we move those out into the shared _utils.py and reuse them for both integrations?

tests/contrib/langgraph/conftest.py
```diff
@@ -0,0 +1,161 @@
...
+@pytest.fixture
+def llmobs_span_writer():
+    yield TestLLMObsSpanWriter(interval=1.0, timeout=1.0)
+
+
+@pytest.fixture
+def llmobs(
+    tracer,
+    llmobs_span_writer,
+):
```
Yun-Kim, line 58:
```suggestion
def llmobs(tracer, llmobs_span_writer):
```

tests/contrib/langgraph/test_langgraph_patch.py
```diff
@@ -0,0 +1,102 @@
...
+
+        self.assert_not_double_wrapped(RunnableSeq.invoke)
+        self.assert_not_double_wrapped(RunnableSeq.ainvoke)
+        self.assert_not_double_wrapped(Pregel.stream)
+        self.assert_not_double_wrapped(Pregel.astream)
+        self.assert_not_double_wrapped(PregelLoop.tick)
+
+    def test_ddtrace_run_patch_on_import(self):
+        # We check that the integration's patch function is called only
+        # after import of the relevant module when using ddtrace-run.
```
Yun-Kim, line 57:
```suggestion
        # Overriding the base test case due to langgraph's code structure not allowing langgraph to be patched by a direct import.
```

Yun-Kim (commented):
Final review for me, everything looks good except for a couple small comments!

ddtrace/llmobs/_integrations/utils.py
sabrenner: ah yeah i should reuse them for langchain as well, thanks for the callout! i have them currently in the `_integrations/`-level `utils.py` (we have some other google utils there too). just wanna confirm if that's the right spot, or if you're referencing the outer `_utils.py` that has like general SDK utils

Yun-Kim: yup that sounds good to me!

ddtrace/llmobs/_integrations/langgraph.py
sabrenner: hm yeah i don't think we technically do. i think i did it this way because i wasn't sure if for some reason there'd ever be a case where `len(next_tasks) != 0` and `more_tasks == True` (although like, why would that ever be the case??). it even looks like that in the [source](https://github.com/langchain-ai/langgraph/blob/aab6fdf3f3c5ff6f695335cc0896b32da4b1dc6e/libs/langgraph/langgraph/pregel/loop.py#L486-L488). i'll do that other check for now, i doubt they'd change this logic.

Yun-Kim: Hmm if they expose the boolean for us then I'm fine with continuing to use that, but thanks for looking into this!

ddtrace/llmobs/_integrations/langgraph.py
sabrenner: ah hmm good point. we do want it to be the current active llmobs span. i think i'll want to change this to `LLMObs._current_span()` once that lands. 

we'll still have the issue of manually generated spans though; all of this logic here is predicated on the only spans involved being langgraph-generated spans. but, i think we can utilize the beta-nature of this integration to see how exactly we'd wanna handle that case (i'm pretty sure manually generated spans would end up being disconnected linking-wise)

Yun-Kim: Cool, since this is experimental for now I'm happy merging this as is and adding next steps once context management is merged + other misc fixes as we find them!

Yun-Kim (approved):

PR #11768, opened 2024-12-18
fix(profiling): Reset all native mutices on fork

sanchda:
## Checklist

## Reviewer Checklist
- Title is accurate

### DIFF
--- a/ddtrace/internal/datadog/profiling/dd_wrapper/src/uploader_builder.cpp
+++ b/ddtrace/internal/datadog/profiling/dd_wrapper/src/uploader_builder.cpp
@@ -186,3 +186,9 @@ Datadog::UploaderBuilder::build()
 
     return Datadog::Uploader{ output_filename, ddog_exporter };
 }
+
+void
+Datadog::UploaderBuilder::postfork_child()
+{
+    tag_mutex.unlock();
+}
--- a/ddtrace/internal/datadog/profiling/stack_v2/src/sampler.cpp
+++ b/ddtrace/internal/datadog/profiling/stack_v2/src/sampler.cpp
@@ -67,6 +67,7 @@ _stack_v2_atfork_child()
     // so we don't even reveal this function to the user
     _set_pid(getpid());
     ThreadSpanLinks::postfork_child();
+    thread_info_map_lock.unlock();
 }
 
 __attribute__((constructor)) void
--- a/ddtrace/internal/datadog/profiling/dd_wrapper/include/uploader_builder.hpp
+++ b/ddtrace/internal/datadog/profiling/dd_wrapper/include/uploader_builder.hpp
@@ -43,6 +43,8 @@ class UploaderBuilder
     static void set_output_filename(std::string_view _output_filename);
 
     static std::variant<Uploader, std::string> build();
+
+    static void postfork_child();
 };
 
 } // namespace Datadog
--- a/ddtrace/internal/datadog/profiling/dd_wrapper/src/ddup_interface.cpp
+++ b/ddtrace/internal/datadog/profiling/dd_wrapper/src/ddup_interface.cpp
@@ -24,6 +24,7 @@ ddup_postfork_child()
     Datadog::Uploader::postfork_child();
     Datadog::SampleManager::postfork_child();
     Datadog::CodeProvenance::postfork_child();
+    Datadog::UploaderBuilder::postfork_child();
 }
 
 void
--- a/ddtrace/internal/datadog/profiling/dd_wrapper/src/sample.cpp
+++ b/ddtrace/internal/datadog/profiling/dd_wrapper/src/sample.cpp
@@ -408,7 +408,6 @@ Datadog::Sample::push_absolute_ns(int64_t _timestamp_ns)
     return true;
 }
 
-
 bool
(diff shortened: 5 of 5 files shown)


ddtrace/internal/datadog/profiling/dd_wrapper/src/uploader_builder.cpp
```diff
@@ -186,3 +186,9 @@ Datadog::UploaderBuilder::build()
 
     return Datadog::Uploader{ output_filename, ddog_exporter };
 }
+
+void
+Datadog::UploaderBuilder::postfork_child()
+{
+    tag_mutex.unlock();
```
nsrip-dd, line 193: I see that the sanitizers are already flagging this ([example](https://github.com/DataDog/dd-trace-py/actions/runs/12394585911/job/34598388160?pr=11768#step:5:819)), but this is undefined behavior if the mutex is already unlocked. I think we need to do what the `Datadog::Uploader` already does and lock the mutexes before forking, fork, and then unlock them in the parent and child.

sanchda: Ah, I just saw this.  Right on--thanks

I hate how std::mutex made pthreads _worse_.  The error-checking type pthreads mutex wraps up these issues.

ddtrace/internal/datadog/profiling/stack_v2/src/sampler.cpp
```diff
@@ -67,6 +67,11 @@ _stack_v2_atfork_child()
     // so we don't even reveal this function to the user
     _set_pid(getpid());
     ThreadSpanLinks::postfork_child();
+
+    // `thread_info_map_lock` is a global lock held in echion; it needs to be destroyed and re-initialized to ensure
+    // it has a consistent state
+    thread_info_map_lock.~mutex();
+    new (&thread_info_map_lock) std::mutex;
```
taegyunkim, line 74:
```suggestion
    // `thread_info_map_lock` and `task_link_map_lock` are global locks held in echion; they need to be destroyed and re-initialized to ensure
    // they have a consistent state
    thread_info_map_lock.~mutex();
    new (&thread_info_map_lock) std::mutex;
    task_link_map_lock.~mutex();
    new (&task_link_map_lock) std::mutex;
```

Used here https://github.com/DataDog/dd-trace-py/blob/e46e3b5785ee26aec4fbf9e7cd2c13c40a435697/ddtrace/internal/datadog/profiling/stack_v2/src/sampler.cpp#L180

releasenotes/notes/fix-profiling-native-mutices-62440b5a3d9d6c4b.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    Fixes a bug where profiling mutexes were not cleared on fork in the child process.  This could
+    cause deadlocks in certain configurations.
```
taegyunkim, line 5:
```suggestion
    profiling: Fixes a bug where profiling mutexes were not cleared on fork in the child process.  This could
    cause deadlocks in certain configurations.
```

ddtrace/internal/datadog/profiling/dd_wrapper/src/profile.cpp
```diff
@@ -203,6 +203,7 @@ Datadog::Profile::collect(const ddog_prof_Sample& sample, int64_t endtime_ns)
 void
 Datadog::Profile::postfork_child()
 {
-    profile_mtx.unlock();
+    profile_mtx.~mutex();
```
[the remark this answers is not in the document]
sanchda: As for the second point, you're correct.  We can elide the explicit destructor.  Not sure why I did it the way I did.

As for the first point, it's UB in the standard but the linux implementation appears to work in all cases I tested (pthreads is better here than generic NPTL).  That doesn't answer how things will work on Windows, but given that this code doesn't support Windows yet at all, I'd prefer to punt on that.  This does leave the question of macos open, and I'd be inclined to believe that macos will indeed summon nasal demons.

danielsn:
For future, we should upstream these changes to echion

ddtrace/internal/datadog/profiling/dd_wrapper/src/profile.cpp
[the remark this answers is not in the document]
sanchda: OK, leaked the forks and rebuilt them as discussed.  I think we're better now?

ddtrace/internal/datadog/profiling/dd_wrapper/src/code_provenance.cpp
```diff
@@ -14,9 +14,8 @@ namespace Datadog {
 void
 Datadog::CodeProvenance::postfork_child()
 {
-    get_instance().mtx.~mutex();            // Destroy the mutex
```
taegyunkim, line 17: Does destroying the mutex result in an UB?

sanchda: I think whether or not it's UB is UB.

When a thread "expires" because of a fork, it is _not_ destroyed.  So the part of the standard that discusses the fact that destroying a mutex held by a destroyed thread is UB doesn't apply.

idk lol

taegyunkim (approved):

PR #11791, opened 2024-12-19
chore(tracer): remove asm properties from the tracer class

mabdinur:
## Description

Removes the following tracer fields and replaces all usages with the corresponding `asm_config`:
- Tracer._asm_enabled
- Tracer._iast_enabled
- Tracer._appsec_standalone_enabled

Next steps:
- Remove all asm, profiling, and serverless specific logic from `ddtrace/_trace/tracer.py`.
  - Define a product level configure methods
  - Introduce product level processors and then product specific logic.
  - Deprecate calling tracing.configure(....) for non tracing features.
 
## Motivation

- Decouples asm configurations from tracing
- Avoids storing duplicate references for global configurations. Ideally global configurations should have once source of truth. This will help simplify remote configuration.

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/pin.py
```diff
@@ -6,6 +6,7 @@
 import wrapt
 
 import ddtrace
+from ddtrace.settings.asm import config as asm_config
```
mabdinur, line 9:
```suggestion
```

ddtrace/pin.py
```diff
@@ -144,7 +145,7 @@ def override(
     def enabled(self):
         # type: () -> bool
         """Return true if this pin's tracer is enabled."""
-        return bool(self.tracer) and (self.tracer.enabled or self.tracer._apm_opt_out)
+        return bool(self.tracer) and (self.tracer.enabled or asm_config._apm_opt_out)
```
mabdinur, line 148:
```suggestion
       # inline to avoid circular imports
       from ddtrace.settings.asm import config as asm_config

        return bool(self.tracer) and (self.tracer.enabled or asm_config._apm_opt_out)
```

avara1986 (approved):

avara1986:
LGTM! great job Munir! :clap:

ddtrace/_trace/tracer.py
```diff
@@ -489,16 +480,15 @@ def configure(
...
-            self._iast_enabled = asm_config._iast_enabled = iast_enabled
+            asm_config._iast_enabled = iast_enabled
 
         if appsec_standalone_enabled is not None:
-            self._appsec_standalone_enabled = asm_config._appsec_standalone_enabled = appsec_standalone_enabled
+            asm_config._appsec_standalone_enabled = appsec_standalone_enabled
 
-        if self._appsec_standalone_enabled and (self._asm_enabled or self._iast_enabled or config._sca_enabled):
-            self._apm_opt_out = True
+        if asm_config._apm_opt_out:
```
erikayasuda, line 491: I saw there's already ASM approval on the PR, but are these actually equivalent conditions (and this opt out isn't a consequence of these conditions being met)? Can we safely rely on the ASM config to update this for us, or should we just keep the same checks but on `asm_config.*` instead of `self.*`?

ddtrace/appsec/_remoteconfiguration.py
```diff
@@ -52,12 +52,6 @@ def enable_appsec_rc(test_tracer: Optional[Tracer] = None) -> None:
 
     Parameters `test_tracer` and `start_subscribers` are needed for testing purposes
     """
-    # Import tracer here to avoid a circular import
```
erikayasuda, line 55: 🎉

ddtrace/pin.py
erikayasuda: Sorry, what exactly is this checking for if we're also allowing a disabled tracer due to ASM standalone to return True too?

ddtrace/settings/asm.py
```diff
@@ -269,6 +269,12 @@ def _eval_asm_can_be_enabled(self):
     def _api_security_feature_active(self) -> bool:
         return self._asm_libddwaf_available and self._asm_enabled and self._api_security_enabled
 
+    @property
+    def _apm_opt_out(self) -> bool:
+        return (
+            self._asm_enabled or self._iast_enabled or tracer_config._sca_enabled is True
```
erikayasuda, line 275: Do we need this? Can we not just do:
```suggestion
            self._asm_enabled or self._iast_enabled or tracer_config._sca_enabled
```

erikayasuda (approved):
Just a nit and a few clarifying questions. Otherwise LGTM from the core/LP side

ddtrace/settings/asm.py
mabdinur: tracer_config._sca_enabled can be None. 

If the _sca_enabled is None and _asm_enabled=True and _iast_enabled=True then `_asm_enabled or _iast_enabled or _sca_enabled == None`. This is wrong. We always want `_apm_opt_out` to return a boolean. Enusuring `tracer_config._sca_enabled` is always a boolean is beyond the scope of this PR.

ddtrace/_trace/tracer.py
mabdinur: This change is equivalent if there is only one instance of the tracer OR are all tracers expected to share the same ASM configurations. I believe this condition holds for all users.

rachelyangdog (approved):

PR #11817, opened 2024-12-20
fix(lib-injection): remove python-json-logger from library compatibility check

brettlangdon:
This library is only used in our testing, we do not have an integration for it, so the version used by any customer should not matter when evaluating library compatibility.

## Checklist

## Reviewer Checklist
- Title is accurate


lib-injection/sources/min_compatible_versions.csv
```diff
@@ -24,13 +24,15 @@ asyncpg,~=0.23
 asynctest,==0.13.0
 austin-python,~=1.0
 avro,0
+azure.functions,0
```
vitor-de-araujo, line 27: Are these other additions to these files expected?

emmettbutler (approved):

lib-injection/sources/min_compatible_versions.csv
emmettbutler: I don't think it's necessary to mention most changes to this file in release notes, because it's intended to be invisible to users.

PR #11823, opened 2024-12-20
chore(tracer): deprecate multiple initializations of ddtrace.Tracer [3.0]

mabdinur:
- Logs a deprecation warning if the Tracer is initialized more than once. The global tracer.
- Ensures that multiple instances of the DummyTracer and CIVisibility Tracer can be created (maintain the current behavior). This will help reset tracer configurations between tests.
- Update some tests that require multiple tracer instances. This will make the 3.0 upgrade easier to do.

## Checklist

## Reviewer Checklist
- Title is accurate


erikayasuda (approved):

erikayasuda:
I think this PR should be a `feat` not `chore`?

Also I wanna push this branch into `staging` and see how it does in rel env before we merge if that's OK?

mabdinur:
> Would this be a breaking change?

Kinda of. 

The Tracer currently does not support initializing multiple tracers. Every instance of the tracer uses the same global ContextVar for span parenting. This leads to a broken traces. With this we have two options:
1) Fix this ContextVar bug and officially support multiple tracer instances
2) Prevent users from initializing multiple tracers. 

I went with the second option since historically we advised customers against initializing their tracers and strongly recommended using the global tracer. 

This change falls in a weird grey area where it's technically a breaking change but if you squint hard enough you can convince yourself it is a bug fix/enhancement.

> Next steps

I will update this PR to deprecate the initialization of multiple tracers. We can remove support for multiple tracer instances in 3.0.

ddtrace/_trace/tracer.py
```diff
@@ -195,6 +195,24 @@ class Tracer(object):
     """
 
     SHUTDOWN_TIMEOUT = 5
+    _instance = None
+
+    def __new__(cls, *args, **kwargs):
```
P403n1x87, line 200: Can't we do this check in `__init__`? `__new__` is more for metaclasses

P403n1x87:
Would we be able to manage the global tracer with the product plugin interface now? If so, can we make that change in this PR as well?

ddtrace/_trace/tracer.py
mabdinur: I used __new__ becuase I can easily override this method in the Dummy, CiVis, and MLobs tracers. Overriding `__init__` would require having custom `__init__(...)` for each tracer type (ex: calling `super(DummyTracer, self).__init__()` would raise deprecation warning).

P403n1x87: But that would be a problem in tests only, no? I don't think we should avoid doing the right thing because of our tests 🤔

mabdinur: You're right. I will fix this

P403n1x87 (approved):

tests/integration/test_integration_snapshots.py
```diff
@@ -203,8 +214,6 @@ def test_trace_with_wrong_meta_types_not_sent(encoding, meta, monkeypatch):
     ],
 )
 @pytest.mark.parametrize("encoding", ["v0.4", "v0.5"])
-@snapshot()
-@pytest.mark.xfail
 def test_trace_with_wrong_metrics_types_not_sent(encoding, metrics, monkeypatch):
```
ZStriker19, line 217: If this test is no longer a snapshot, how is it asserting things?

mabdinur:
> I think this should have a deprecation release note. Other than that just one question about testing changes.

It already does 😄

tests/integration/test_integration_snapshots.py
mabdinur, line 217:
```suggestion
@snapshot()
@pytest.mark.xfail
def test_trace_with_wrong_metrics_types_not_sent(encoding, metrics, monkeypatch):
```

mabdinur: Nice catch. This was a miss on my part. I will revert this

ZStriker19 (approved):

PR #11839, opened 2024-12-27
feat(config): support libdatadog's library_config for config through file

BaptisteFoy:
## Description of changes
- Bumps `libdatadog` to v15.0.0 & handles the required changes to crashtracking
- Starts using the library_config component of `libdatadog` that allows reading configuration from a file **with a higher precedence than environment variables**

### Supported config entries
As of libdatadog v15.0.0 ([source](https://github.com/DataDog/libdatadog/pull/807/files#diff-6011d3279618b70cecf54936d7241711522b09210d2d590a72255357a4698adbR120-R129)):
- `DD_TRACE_DEBUG`
- `DD_SERVICE`
- `DD_ENV`
- `DD_VERSION`
- `DD_PROFILING_ENABLED`

## Testing strategy
Added a unit test.

To test this PR, you can:
1. Create a Linux VM
2. Start this tracer version & ensure everything works well (non-regression)
3. Create a file at the following path: `/etc/datadog-agent/managed/datadog-apm-libraries/stable/libraries_config.yaml`
4. Write the following content in the file:
```yaml
rules:
  - selectors:
      - origin: language
        matches:
          - python
        operator: equals
    configuration:
      DD_TRACE_DEBUG: true
```
5. Start this tracer version
6. Verify that debug logs are properly enabled

## Additional concerns
This change does increase the size of the serverless layer by 380KB (14204KB vs 13824 as the limit) even after optimization. I had to go from `lto=true` to `lto=fast`, as `lto=true` somehow prevents release builds on older machines. That accounts for 200KB, while the 200 other KB are mostly coming from the libdatadog bump.

## Checklist

## Reviewer Checklist
- Title is accurate


taegyunkim (commented):
👋 I'd like to suggest not adding new modules under /profiling as the functionality is not profiling. You should be able to expose the libdatadog API through src/core, or add a new rust/python module under src/

taegyunkim:
Adding @brettlangdon to the reviewers early so you could get feedback sooner than later

ddtrace/internal/datadog/profiling/cmake/FindLibdatadog.cmake
taegyunkim, line 1: Thank you for bumping up libdatadog version for profiling!

brettlangdon (commented):
can we break up this PR?

I think there is the libdatadog version bump, core package renaming, and then the config file changes which can all get broken up.

tests/internal/test_native.py
```diff
@@ -0,0 +1,41 @@
...
+    config_2.write_text(
+        """
+rules:
+  - selectors:
+    - origin: language
+      matches:
+        - nodejs
+      operator: equals
+    configuration:
+      DD_SERVICE: my-service
```
mabdinur, line 35: In the python tracer the DD_SERVICE environment variable can be overridden by [integrations](https://github.com/DataDog/dd-trace-py/blob/v2.20.0rc1/ddtrace/contrib/internal/flask/patch.py#L66C9-L66C25) and other ddtrace products (ex: [serverless](https://github.com/DataDog/dd-trace-py/blob/v2.20.0rc1/ddtrace/settings/config.py#L483)). 

We will need to audit all the configurations supported by libdatadog and ensure the value set via APM Stable Configurations take precedence.

ddtrace/__init__.py
```diff
@@ -5,6 +5,12 @@
 
 LOADED_MODULES = frozenset(sys.modules.keys())
 
+# Configuration for the whole tracer from file. Do it before anything else happens.
+from ddtrace.internal.native import get_configuration_from_disk
+
+for key, value in get_configuration_from_disk().items():
```
mabdinur, line 11: Can we add an "internal" environment variable to gate-keep this feature. Something like `_DD_APM_STABLE_CONFIG_ENABLED`. This way we can make this feature opt-in until it's fully tested and validated. It will also allow us to introduce breaking changes and iterate on the initial design.

ddtrace/internal/native/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+    configurator = PyConfigurator(debug_logs)
+    configurator.set_envp(["%s=%s" % (k, v) for k, v in os.environ.items()])
+    configurator.set_args(sys.argv)
+
+    # Set the file override if provided. Only used for testing purposes.
+    if file_override:
+        configurator.set_file_override(file_override)
+
+    return configurator.get_configuration()
```
mabdinur, line 23: Does this configurator handle sending instrumentation telemetry? If so we may end up sending duplicate instrumentation telemetry events ([here for dd_service](https://github.com/DataDog/dd-trace-py/blob/v2.20.0rc1/ddtrace/settings/config.py#L483)). 

Also it would be nice if we could distinguish between configurations set by customer's via env_vars and configurations that are set by stable configs. We could address this limitation in a future PR.

mabdinur (commented):
Overall looks great! High level comments:
- Setting environment variables directly is good enough for this POC now but we will need to revisit this design decision to better track the origin of configurations.
- We need to identify all scenarios where the value set by stable configurations can be overriden by the ddtrace library.
- We need to document a clear precedence for configurations and ensure all libraries follow it (rc -> stable_config -> in_code -> envar -> reading from yaml file)
- We need to document which configurations are supported by `15.0`. Including this in the PR description is fine.
- We should split up this PR (@brettlangdon's comment)

ddtrace/internal/datadog/profiling/dd_wrapper/src/crashtracker.cpp
```diff
@@ -207,8 +207,8 @@ Datadog::Crashtracker::start()
 
     auto result = ddog_crasht_init(config, receiver_config, metadata);
     ddog_Vec_Tag_drop(tags);
-    if (result.tag != DDOG_CRASHT_RESULT_OK) { // NOLINT (cppcoreguidelines-pro-type-union-access)
-        auto err = result.err;                 // NOLINT (cppcoreguidelines-pro-type-union-access)
+    if (result.tag != DDOG_VOID_RESULT_OK) { // NOLINT (cppcoreguidelines-pro-type-union-access)
```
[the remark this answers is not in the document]
BaptisteFoy: Indeed, this is https://github.com/DataDog/dd-trace-py/pull/12063 now

ddtrace/__init__.py
BaptisteFoy: I'm not sure this is necessary here; given that this change is a no-op if the stable config YAML files don't exist.

ddtrace/internal/native/__init__.py
BaptisteFoy: It doesn't afaik, at least it wasn't added in the [libdatadog PR](https://github.com/DataDog/libdatadog/pull/807). 

I agree that we need to display the source here; especially as in the very near future we'll have two sources for stable config (manually set by customer // sent through Fleet Automation). This will be done in another PR, hopefully in the next week or so, and we'll return a list of tuples (key, value, source) instead of a dict. 

It won't matter much as long as we don't have a better way to apply stable configs than env vars, but at least the data will be there for future improvements.

tests/internal/test_native.py
BaptisteFoy: Agreed. I also think that we need to have a discussion on the order of precedence. For instance in the agent, config values set by the runtime (ie inside the agent code) take precedence over Fleet Automation. One could argue that we should replicate that behaviour here; but imo this is a product decision.

src/native/library_config.rs
```diff
@@ -0,0 +1,70 @@
...
+#[pyclass(name = "PyConfigurator", module = "ddtrace.internal._native")]
+pub struct PyConfigurator {
+    configurator: Box<Configurator>,
+    file: String,
+    envp: Vec<String>,
+    args: Vec<String>,
+}
+
+#[pymethods]
+impl PyConfigurator {
```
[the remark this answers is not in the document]
BaptisteFoy: The Rust FFI code is built as part of the wheel (there is a `RustExtension` in `setup.py`). We have build tests for major environments (and I hope all supported environments). So if the build passes we can be confident that nothing will break for customers. Even if the build is skipped the class here will simply do nothing and return an empty dict every time; which is a noop.

releasenotes/notes/libdatadog-library_config-3c36c4f868d2c86f.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    config: Support configuration from a file on disk. Use the libdatadog native library.
```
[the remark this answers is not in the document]
BaptisteFoy: I definitely don't mind putting the `no-changelog` label here; but if we do not notify customers now in the release notes now, will we be able to in the future? Or should we rely on the GA announcement instead?

ddtrace/internal/native/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+import sys
+from typing import Dict
+
+from ._native import DDSketch  # noqa: F401
+from ._native import PyConfigurator
+
+
+def get_configuration_from_disk(debug_logs: bool = False, file_override="") -> Dict[str, str]:
+    """
+    Retrieves the tracer configuration from disk. Calls the PyConfigurator object
```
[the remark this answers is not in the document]
BaptisteFoy: Very good point; thanks

BaptisteFoy:
@brettlangdon @mabdinur Could you please review https://github.com/DataDog/dd-trace-py/pull/12063 & https://github.com/DataDog/dd-trace-py/pull/12065 as new prerequisites to this PR, as suggested?

releasenotes/notes/libdatadog-library_config-3c36c4f868d2c86f.yaml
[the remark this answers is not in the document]
mabdinur: We can announce this feature in a future release (even if it isn't the first library version with the change). Let's put no-changelog for now and then document this feature after the GA announcement.

ddtrace/__init__.py
```diff
@@ -5,6 +5,11 @@
 
 LOADED_MODULES = frozenset(sys.modules.keys())
 
+# Configuration for the whole tracer from file. Do it before anything else happens.
+from ddtrace.internal.native import apply_configuration_from_disk
+
+apply_configuration_from_disk()
```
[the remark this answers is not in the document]
BaptisteFoy: Done in [485a7fb](https://github.com/DataDog/dd-trace-py/pull/11839/commits/485a7fbd34d5862822077df516e3d406870f81b1)!

ddtrace/internal/native/__init__.py
```diff
@@ -1 +1,42 @@
...
+
+    Currently here is the order of precedence (higher takes precedence):
+    1. Dynamic remote configuration
+    2. Runtime configuration (ie fields set manually by customers / from the ddtrace code)
+    3. Configuration from disk
+    4. Environment variables
+    5. Default values
+    """
+    for key, value in get_configuration_from_disk().items():
+        os.environ[key] = str(value).lower()
```
brettlangdon, line 42: not sure I love this 🤔 

I don't have a good suggestion, but modifying system state in this way just _feels_ weird.

would we expect that a user should be able to access/see the configuration applied by accessing it at `os.environ[key]` ?

tests/internal/test_native.py
```diff
@@ -0,0 +1,59 @@
...
+
+    # Check if both modules are in sys.modules
+    if module_a in sys.modules and module_b in sys.modules:
+        # Get the position of the modules in sys.modules (which is an OrderedDict in Python 3.7+)
+        modules = list(sys.modules.keys())
+        return modules.index(module_a) < modules.index(module_b)
+    return False  # If one or both modules are not imported, return False
+
+
+def test_native_before_settings():
```
brettlangdon, line 55: we should probably have a test that applying a configuration via file gets applied to the application.

do we expect to have system-tests for this stuff as well? (if not, can we?)

ddtrace/internal/native/__init__.py
BaptisteFoy: It does really feel weird to me too; and to be fair it's a hack to make it work quickly. There is no global configuration object that is used everywhere that we can use.

Hopefully in the not-so-far future we'll be able to properly apply this configuration and track its origin; but this is a whole other problem to tackle.

src/native/Cargo.toml
```diff
@@ -4,13 +4,14 @@ version = "0.1.0"
 edition = "2021"
 
 [profile.release]
-lto = true
+lto = false
```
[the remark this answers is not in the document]
BaptisteFoy: I'm investigating with folks that are more well-versed in all this optimization stuff than I am; but when I set it to `true` the whole build breaks so...

tests/internal/test_native.py
brettlangdon: > Besides, yes we want to add system-tests to test this behaviour on customer-like systems

Can we add those _before_ we merge this? It should be fully tested before we merge, since as soon as it is merged it can be released.

ddtrace/internal/native/__init__.py
BaptisteFoy: Well; looks like something else changed because now it passes. I re-enabled it 😕

src/native/Cargo.toml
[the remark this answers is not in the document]
BaptisteFoy: Well; looks like something else changed because now it passes. I re-enabled it 😕

tests/internal/test_native.py
paullegranddc: > It should be fully tested before we merge, since as soon as it is merged it can be released.

Would you be ok with merging this under an experimental flag that disables calling to this code by default, so we are sure no customer uses it? It's hard to bootstrap system-tests if we don't have any merged language implementation to run the actual tests against.

We need to make sure that this code doesn't crash stuff obviously but this is something that can be tested in integration tests in ddtrace-py rather than in system-tests.

brettlangdon: You should be able to write tests against a dev version of this library pretty easily: https://github.com/DataDog/system-tests/blob/main/docs/execute/binaries.md#python-library

Merging tests in system-tests can be tricky since we need to ensure they are expected to fail on older versions.


I'd prefer the approach of writing at least the first pass of tests against this branch, but otherwise, a private/hidden feature flag would be ok.

BaptisteFoy: I've written a first system test and made it pass locally against this PR: https://github.com/DataDog/system-tests/pull/3982. Happy to get your review @brettlangdon. Let me know if there is anything to change on this PR to run this test.

BaptisteFoy:
Closing in favour of https://github.com/DataDog/dd-trace-py/pull/12260

PR #11848, opened 2025-01-02
fix(profiling): remove slow getpid call from memalloc path

nsrip-dd:
memalloc uses getpid to detect whether the process has forked, so that
we can unlock the memalloc lock in the child process (if it isn't
already locked). Unfortunately the getpid call is quite slow. From the
man page: "calls to getpid() always invoke the actual system call,
rather than returning a cached value." Furthermore, we _always_ attempt
to take the lock for allocations, even if we aren't going to sample
them. So this is basically adding a syscall to every allocation.

Move this logic out of the allocation path. Switch to using
pthread_atfork handlers to ensure that the lock is held prior to
forking, and unlock it in the parent and child after forking. This
(maybe) has the added benefit of making sure the data structures are in
a consistent state in the child process after forking. Unclear if that's
an issue prior to this change, though. I may be missing some code that
resets the profiler on fork anyway?

## Checklist

## Reviewer Checklist
- Title is accurate

### DIFF
--- a/ddtrace/profiling/collector/_memalloc.c
+++ b/ddtrace/profiling/collector/_memalloc.c
@@ -57,6 +57,28 @@ static alloc_tracker_t* global_alloc_tracker;
 static void
 memalloc_init(void);
 
+static void
+memalloc_prefork(void)
+{
+    // Lock the mutex prior to forking. This ensures that the memory profiler
+    // data structures will be in a consistent state in the child process.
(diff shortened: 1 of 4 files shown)


sanchda (approved):

nsrip-dd:
2.17 backport is blocked on #11801. 

2.19 backport failed as well. Seems like #11460 isn't on the 2.19 branch?

PR #11856, opened 2025-01-03
chore(llmobs): assert using span events instead of mocks

Yun-Kim:
Follow up of #11781 to further clean up LLMObs tests, specifically replacing potentially flaky LLMObs span writer mocks and assertions with the test LLMObsSpanWriter dummy class. Also clean up the `tests/llmobs/conftest.py` file which previously contained a ton of rarely used and sometimes redundant fixtures.

## Checklist

## Reviewer Checklist
- Title is accurate

### DIFF
--- a/tests/llmobs/test_llmobs_decorators.py
+++ b/tests/llmobs/test_llmobs_decorators.py
@@ -19,7 +19,7 @@ def mock_logs():
         yield mock_logs
 
 
-def test_llm_decorator_with_llmobs_disabled_logs_warning(LLMObs, mock_logs):
+def test_llm_decorator_with_llmobs_disabled_logs_warning(llmobs, mock_logs):
     for decorator_name, decorator in (("llm", llm), ("embedding", embedding)):
 
         @decorator(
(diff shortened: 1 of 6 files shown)


tests/llmobs/test_llmobs_decorators.py
```diff
@@ -47,227 +47,207 @@ def test_non_llm_decorator_with_llmobs_disabled_logs_warning(LLMObs, mock_logs):
...
         pass
 
     f()
-    span = LLMObs._instance.tracer.pop()[0]
-    mock_llmobs_span_writer.enqueue.assert_called_with(
-        _expected_llmobs_llm_span_event(
-            span, "llm", model_name="test_model", model_provider="test_provider", session_id="test_session_id"
-        )
+    span = llmobs._instance.tracer.pop()[0]
+    assert llmobs_events[0] == _expected_llmobs_llm_span_event(
```
ncybul, line 63: Discussed this in person, but I think it is a bit convoluted that we have to pass in arguments like `model_name`, `model_provider`, etc. into this helper to get the expected llmobs span event. I think it would be better if we pulled these values from the span that is popped from the tracer (otherwise, why even get this span from the tracer when it seems like you are just reconstructing the span event based on the decorator args).

ncybul (approved):
Besides the one comment I left, this LGTM. Thanks for the thorough explanation! Also, feel free to ping for a final pass.

tests/llmobs/test_llmobs_decorators.py
Yun-Kim: You are right and I will fix this, but this will be a huge change that I would rather do in a standalone PR than this one. since this is already a hefty PR.

PR #11872, opened 2025-01-07
ci(botocore): fix flaky botocore tests

mabdinur:
Botocore sensitive data redaction sets `AWSPayloadTagging.request_redaction_paths` and `AWSPayloadTagging.response_redaction_paths` the first time [AWSPayloadTagging.expand_payload_as_tags(...)](https://github.com/DataDog/dd-trace-py/blob/4970e6d0e8f6787e841576e6229ee572d743e9e8/ddtrace/_trace/utils_botocore/aws_payload_tagging.py#L83) is called.
 
In tests overriding `config.botocore.payload_tagging_request` and `config.botocore.payload_tagging_request` using `tests.utils.override_config(..)` is not sufficient. We also need to set `ddtrace._trace.utils_botocore.span_tags._PAYLOAD_TAGGER.validated` to `False`. This ensures the overriden botocore configurations are actually used.

Note: This PR does not fix the underlying issue of `AWSPayloadTagging` configurations becoming out of sync with `ddtrace.config.botocore`. We will need to address this bug in a future PR.

## Checklist

## Reviewer Checklist
- Title is accurate

### DIFF
--- a/tests/contrib/botocore/test.py
+++ b/tests/contrib/botocore/test.py
@@ -22,8 +22,8 @@ from moto import mock_stepfunctions
 import pytest
 
 from ddtrace._trace._span_pointer import _SpanPointer
-from ddtrace._trace.utils_botocore import span_tags
 from ddtrace._trace._span_pointer import _SpanPointerDirection
+from ddtrace._trace.utils_botocore import span_tags
 from tests.utils import get_128_bit_trace_id_from_headers
 
 
--- a/tests/contrib/botocore/test.py
+++ b/tests/contrib/botocore/test.py
@@ -22,6 +22,7 @@ from moto import mock_stepfunctions
 import pytest
 
 from ddtrace._trace._span_pointer import _SpanPointer
+from ddtrace._trace.utils_botocore import span_tags
 from ddtrace._trace._span_pointer import _SpanPointerDirection
 from tests.utils import get_128_bit_trace_id_from_headers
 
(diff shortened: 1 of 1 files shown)


tests/contrib/botocore/test.py
```diff
@@ -3953,6 +3958,12 @@ def test_aws_payload_tagging_s3(self):
     @pytest.mark.snapshot(ignores=snapshot_ignores)
     @mock_s3
     def test_aws_payload_tagging_s3_invalid_config(self):
+        # Setting the validated flag to False ensures the payload config is re-validated
```
mabdinur, line 3961:
```suggestion
        # Setting the validated flag to False ensures the redaction paths configurations are re-validated
```

brettlangdon (commented):
Does this need to be set for all botocore tests? would an auto-use fixture be better? e.g. if someone adds a new test do they need to copy/pasta this setting as well?


example:

```python

@pytest.fixture(autouse=True)
def reset_payload_tagger():    
    span_tags._PAYLOAD_TAGGER.validated = False
```

bouwkast:
> Does this need to be set for all botocore tests? would an auto-use fixture be better? e.g. if someone adds a new test do they need to copy/pasta this setting as well?
> 
> example:
> 
> ```python
> @pytest.fixture(autouse=True)
> def reset_payload_tagger():    
>     span_tags._PAYLOAD_TAGGER.validated = False
> ```

This feature is disabled by default so most of these tests will not use this, but if it is as simple as that then I think that is reasonable.

mabdinur:
> ```python
> @pytest.fixture(autouse=True)
> def reset_payload_tagger():    
>     span_tags._PAYLOAD_TAGGER.validated = False
> ```

Yup that's a cleaner approach but this is supposed to be a change. Ideally we should address the underlying issue and not just re-write tests to avoid the broken edge case.

erikayasuda (approved):

PR #11888, opened 2025-01-09
ci(serverless): force reimport of modules during serverless test

purple4reina:
Commit https://github.com/DataDog/dd-trace-py/commit/12655eebaa93ec7eb9ce6858ddf5e07ae4e4dae7 removed the `@pytest.mark.subprocess()` from the serverless slow imports test. This effectively rendered the test useless as `ddtrace` and all required modules had already been imported in the parent process.

Adding the subprocess mark back will ensure that the sys modules are fresh. We must still go through and delete entries because `@pytest.mark.subprocess()` relies on `multiprocessing`, so it must be removed so as to ensure it can be reimported if needed.

## Checklist

## Reviewer Checklist
- Title is accurate


emmettbutler (approved):

.gitlab/tests.yml
```diff
@@ -85,13 +85,14 @@ build_base_venvs:
...
     - !reference [.testrunner, before_script]
-    - pip install riot==0.20.1
+    - pip install riot==0.20.1 "setuptools_scm[toml]>=4"
     - unset DD_SERVICE
     - unset DD_ENV
     - unset DD_TAGS
     - unset DD_TRACE_REMOVE_INTEGRATION_SERVICE_NAMES_ENABLED
   script:
     - |
+      python -m setuptools_scm --force-write-version-files
```
purple4reina, line 95: This line will create the `ddtrace/_version.py` file. Without it, importing `ddtrace` will attempt to fetch the version, won't find it, then will import `importlib_metadata` as a fallback mechinism.

emmettbutler:
This change is especially tough to merge because the change to `tests.yml` causes the entire test suite to run, even stuff that a human can tell is clearly unrelated.

emmettbutler:
fixing the profile suite issue here https://github.com/DataDog/dd-trace-py/pull/12187

emmettbutler:
here's a fix for the llmobs failures https://github.com/DataDog/dd-trace-py/pull/12191

.gitlab/tests.yml
```diff
@@ -84,13 +84,14 @@ build_base_venvs:
...
```
brettlangdon, line 94: This will get handled by https://github.com/DataDog/dd-trace-py/pull/12314 as well

.gitlab/tests.yml
```diff
@@ -84,13 +84,14 @@ build_base_venvs:
   retry: 2
   before_script:
     - !reference [.testrunner, before_script]
-    - pip install riot==0.20.1
+    - pip install riot==0.20.1 "setuptools_scm[toml]>=4"
```
brettlangdon, line 87:
```suggestion
    - pip install riot==0.20.1
```

.gitlab/tests.yml
brettlangdon:
```suggestion
```

brettlangdon (commented):
https://github.com/DataDog/dd-trace-py/pull/12314 was merged, so the rewriting of version file should not be necessary anymore.

tests/internal/test_serverless.py
```diff
@@ -78,7 +79,40 @@ def test_not_azure_function_consumption_plan():
...
+    "package,blocklist",
+    [
+        ("ddtrace", expanded_blocklist),
+        ("ddtrace.contrib.internal.aws_lambda", expanded_blocklist),
+        ("ddtrace.contrib.internal.psycopg", expanded_blocklist),
+        # requests imports urlib3 which imports importlib.metadata
+        ("ddtrace.contrib.requests", standard_blocklist),
+    ],
+)
+def test_slow_imports(package, blocklist):
```
brettlangdon, line 115: should we be running this in a subprocess? is it going to leak state of imported and unloaded modules between tests?

purple4reina: Good question. It looks like `@pytest.mark.subprocess` doesn't work for parametrized tests. And even when removing the parametrization, the test is unable to find `expanded_blocklist` because only the test function itself is run in a subprocess, not the whole file. At least that is what I'm assuming.

That said, this test now will put `sys.modules` back to its original list after every run, so there shouldn't be a problem with leaked state.

tests/internal/test_serverless.py
```diff
@@ -78,7 +79,40 @@ def test_not_azure_function_consumption_plan():
...
+
+
+@pytest.mark.parametrize(
+    "package,blocklist",
+    [
+        ("ddtrace", expanded_blocklist),
+        ("ddtrace.contrib.internal.aws_lambda", expanded_blocklist),
+        ("ddtrace.contrib.internal.psycopg", expanded_blocklist),
+        # requests imports urlib3 which imports importlib.metadata
+        ("ddtrace.contrib.requests", standard_blocklist),
```
purple4reina, line 112: @brettlangdon I'm wondering if it would be possible to run this test importing all the packages in `ddtrace.contrib`. Right now, none of these import paths are tested and I'm concerned that our build scripts might be removing important binaries which these require. Right now, we'd have no way of knowing if say `import ddtrace.contrib.flask` attempts to import one of the removed binaries listed above.

tests/internal/test_serverless.py
brettlangdon: `@pytest.mark.subprocess` is interesting. It takes the contents of the test function, dumps it to a file, and then runs it in a subprocess. It is expected that 100% of the necessary code to execute in the subprocess is defined in the test function body.

This means things like imports and etc also need to be imported explicitly in the test function. You cannot rely on any global state. (TIL it doesn't work for parametrized tests too, which I guess makes sense).


You might want to check the `run_python_code_in_subprocess` test fixture, that might be better for this use case:

```python
def test_func(run_python_code_in_subprocess):

    code = """
import ddtrace.auto
import sys


assert "threading" not in sys.modules
"""

    stderr, stdout, status, _ = run_python_code_in_subprocess(code)
    assert status == 0, stderr.decode()
    assert stderr == b""
    assert stdout == b""
```


This will work with parametrized test, and will need formatting or env vars or something to pass parametrized values into the test code, but maybe a bit more explicit and easier to reason about?


https://github.com/DataDog/dd-trace-py/blob/cd96af659ee68336919d3efcd6f525c237d21d1d/tests/internal/test_settings.py#L244-L328

tests/internal/test_serverless.py
brettlangdon: 🤔 I am not sure to be honest, because just importing `ddtrace.contrib.requests` might not actually import anything. Since most of the `ddtrace.contrib.<integration>.__init__` files will check if the necessary packages are installed first before trying to load the integration patching code. So it might basically be a no-op and then not actually test what you need.


e.g. you might need to actually install all of these packages first to then test them.


if we consider this, we might want to move the serverless tests out of `tests/internal` and re-think how they are structured/run. There are some `appsec_iast_packages` tests (`tests/appsec/iast_packages`) which basically install a ton of packages to test against (not sure we should copy/pasta that code, but example to consider).

tests/internal/test_serverless.py
purple4reina: Awesome! That works perfectly! I like this solution a lot. Gonna push the change shortly.

tests/internal/test_serverless.py
```diff
@@ -78,47 +79,67 @@ def test_not_azure_function_consumption_plan():
...
+standard_blocklist = [
+    "ddtrace.appsec._api_security.api_manager",
+    "ddtrace.appsec._iast._ast.ast_patching",
+    "ddtrace.internal.telemetry.telemetry_writer",
+    "email.mime.application",
+    "email.mime.multipart",
+    "logging.handlers",
+    "multiprocessing",
+    "importlib_metadata",
+    # These modules must not be imported because their source files are
```
ZStriker19, line 91: Is the removal done on a case by case basis? Does something bad happen if they do end up being imported, or we just start running code/features we shouldn't be? I'm just wondering if there's any way to future proof the list against any new packages that we add that shouldn't be in the serverless layer, but I'm guessing this is on a human decision level and is on serverless' radar and is quickly discovered by serverless tests failing with a new tracer version?

ZStriker19 (approved):

tests/internal/test_serverless.py
purple4reina: You can see exactly which files are being removed looking [here](https://github.com/DataDog/datadog-lambda-python/blob/main/Dockerfile). We're not removing whole swaths of binary files, just one at a time. Adding new files to that list is done on a case by case basis, decision made by humans only.

We also have a fairly robust testing and monitoring system (you can see that [here](https://github.com/DataDog/serverless-self-monitoring)). Before releasing any new layer version, we'll push a release candidate, let it run for a day, then check a set of dashboards to ensure its health. We have a ton of sample lambda functions running all different configurations and traffic patterns (you can see a list of those [here](https://github.com/DataDog/serverless-self-monitoring/blob/main/self_monitor/aws/CURRENT_STACKS.md)).

Lastly, should we happen to put out a bad release, and a customer's lambda function crashes. They can easily roll back to a previous Datadog Python layer version. Additionally, this only impacts customers using one of our layers. We do not remove binary files when, for instance, a customer pip installs datadog-lambda and packages this dependency on their own.

But in general, if we've removed a file, and ddtrace tries to import it, bad things will happen. What do we do to detect those before releasing? This PR is a key part of that.

PR #11915, opened 2025-01-13
chore(llmobs): implement answer relevancy ragas metric

lievan:
Implements answer relevancy metric for ragas integration.

### About Answer Relevancy

Answer relevancy metric focuses on assessing how pertinent the generated answer is to the given prompt. A lower score is assigned to answers that are incomplete or contain redundant information and higher scores indicate better relevancy. This metric is computed using the question, the retrieved contexts and the answer.

The Answer Relevancy is defined as the mean cosine similarity of the original question to a number of artificial questions, which where generated (reverse engineered) based on the response.

## Example trace
image

## Checklist

## Reviewer Checklist
- Title is accurate


tests/llmobs/conftest.py
```diff
@@ -159,6 +168,17 @@ def mock_ragas_evaluator(mock_llmobs_eval_metric_writer, ragas):
     patcher.stop()
 
 
+@pytest.fixture
```
lievan, line 171: we patch the embedding similarity calculation so we do not need to rely on a cassette recording fetching embedding encodings from openai. The cassette would result in +200k plus lines

ddtrace/llmobs/_evaluators/ragas/answer_relevancy.py
```diff
@@ -0,0 +1,143 @@
...
+        # initialize data we annotate for tracing ragas
+        score, answer_classifications = (math.nan, None)
+
+        with self.llmobs_service.workflow(
+            "dd-ragas.answer_relevancy", ml_app=_get_ml_app_for_ragas_trace(span_event)
+        ) as ragas_ar_workflow:
+            try:
+                evaluation_metadata[EVALUATION_SPAN_METADATA] = self.llmobs_service.export_span(span=ragas_ar_workflow)
+
+                cp_inputs = self._extract_evaluation_inputs_from_span(span_event)
```
Yun-Kim, line 90: what does `cp` stand for here?

ddtrace/llmobs/_evaluators/ragas/answer_relevancy.py
```diff
@@ -0,0 +1,143 @@
...
+                if cp_inputs is None:
+                    logger.debug(
+                        "Failed to extract question and contexts from "
+                        "span sampled for `ragas_answer_relevancy` evaluation"
+                    )
+                    return "fail_extract_answer_relevancy_inputs", evaluation_metadata
```
Yun-Kim, line 96: I'm seeing a potential for deduplicating code, looks like all of the evaluators' code runs more or less this section in `evaluate()` (and evaluator-specific names/instance I'm sure we can account for). What if we move this (skeleton for setting up llm instance, tracing, and getting eval inputs) to the base class `evaluate()` and have each evaluator implement a more specific `_evaluate()` method?

tests/llmobs/conftest.py
```diff
@@ -150,6 +150,15 @@ def reset_ragas_faithfulness_llm():
     ragas.metrics.faithfulness.llm = previous_llm
 
 
+@pytest.fixture
+def reset_ragas_answer_relevancy_llm():
```
Yun-Kim, line 154: Let's move this to the ragas evaluator test file or a new conftest under the ragas directory, I'd like this top-level conftest to be general llmobs and avoid polluting it

tests/llmobs/conftest.py
Yun-Kim: Nice! Let's also move this to a new conftest under the ragas directory since this is ragas specific.

ddtrace/llmobs/_evaluators/ragas/answer_relevancy.py
```diff
@@ -0,0 +1,143 @@
...
+            - evaluation metadata (dict)
+        If the ragas answer relevancy instance does not have `llm` set, we set `llm` using the `llm_factory()`
+        method from ragas which currently defaults to openai's gpt-4o-turbo.
+        """
+        self.ragas_answer_relevancy_instance = self._get_answer_relevancy_instance()
+        if not self.ragas_answer_relevancy_instance:
+            return "fail_answer_relevancy_is_none", {}
+
+        evaluation_metadata = {}  # type: dict[str, Union[str, dict, list]]
+        trace_metadata = {}  # type: dict[str, Union[str, dict, list]]
```
Yun-Kim, line 79: What's the difference between trace metadata and evaluation metadata here? Is this needed to be on the trace vs the evaluation?

tests/llmobs/test_llmobs_ragas_evaluators.py
```diff
@@ -238,3 +240,236 @@ def test_llmobs_with_faithfulness_emits_traces_and_evals_on_exit(mock_writer_log
...
+
+
+def test_ragas_answer_relevancy_has_modified_answer_relevancy_instance(
+    ragas, mock_llmobs_submit_evaluation, reset_ragas_answer_relevancy_llm, llmobs
+):
+    """Context precision instance used in ragas evaluator should match the global ragas context precision instance"""
+    from ragas.llms import BaseRagasLLM
+    from ragas.metrics import answer_relevancy
+
+    class FirstDummyLLM(BaseRagasLLM):
```
Yun-Kim, line 279: Idea: if we are not concerned with testing the actual LLM calls themselves, then I would suggest we do the same thing here and replace cassettes with this mocking. Would this work?

tests/llmobs/test_llmobs_ragas_evaluators.py
```diff
@@ -238,3 +240,236 @@ def test_llmobs_with_faithfulness_emits_traces_and_evals_on_exit(mock_writer_log
...
+            return "second dummy llm"
+
+    answer_relevancy.llm = SecondDummyLLM()
+
+    rar_evaluator = RagasAnswerRelevancyEvaluator(llmobs)
+
+    assert rar_evaluator.ragas_answer_relevancy_instance.llm.generate_text() == "second dummy llm"
+
+
+@pytest.mark.vcr_logs
```
Yun-Kim, line 312: If we are going to continue with cassette testing, then I would recommend following the same comment as in the context precision PR and rely on 2-3 cassettes based on the LLM input span (`_llm_span_with_expected_ragas_inputs_in_prompt() vs _llm_span_with_expected_ragas_inputs_in_messages() vs custom`) instead of having a new cassette for each test

tests/llmobs/test_llmobs_ragas_evaluators.py
```diff
@@ -238,3 +240,236 @@ def test_llmobs_with_faithfulness_emits_traces_and_evals_on_exit(mock_writer_log
...
+    assert root_span["parent_id"] == "undefined"
+    assert root_span["meta"] is not None
+
+    root_span_trace_id = root_span["trace_id"]
+    for child_span in ragas_spans[1:]:
+        assert child_span["trace_id"] == root_span_trace_id
+        assert child_span["parent_id"] == root_span_id
+
+
+def test_llmobs_with_answer_relevancy_emits_traces_and_evals_on_exit(mock_writer_logs, run_python_code_in_subprocess):
```
Yun-Kim, line 428: Feel free to remove this test, I don't really see the value added by these exit tests

ddtrace/llmobs/_evaluators/ragas/answer_relevancy.py
lievan: Yep, evaluation metadata stores metadata for the eval itself that must be visualized (e.g. quotes in the answer that were flagged as problematic) and trace metadata is just the generic meta.metadata field we have for spans

tests/llmobs/test_llmobs_ragas_evaluators.py
lievan: I think it _could_ work but it could be quite nasty. To mock the LLM calls we would need to intercept the ragas logic at some point in the middle of some relatively complex string parsing. We would need to make sure the string output that we mock 1. matches well with the input 2. is in the correct format such that it doesn't break the downstream logic, in other words its a json string that matches the data structures in `ragas/models.py`

So i'm more comfortable recording a cassette here for different chains of LLM calls we would expect and re-using them across tests, which is what you suggested in the context relevancy PR

ddtrace/llmobs/_evaluators/ragas/answer_relevancy.py
lievan: Ah copy and paste mistake (cp = context precision), renamed to answer_relevancy_inputs for clarity

ddtrace/llmobs/_evaluators/ragas/answer_relevancy.py
lievan: That makes sense to separate the logic for extracting inputs, tracing setup, etc. vs the core evaluation workflow itself. I can add this as refactor as a follow up TODO, along with improving the code for how we detect what is a ragas span vs user-application span, that i'll work on in a separate PR which shouldn't block the release

tests/llmobs/test_llmobs_ragas_evaluators.py
Yun-Kim: Yeah, if it looks like a lot of work/non-trivial I'm fine with the reduced cassette testing. Thanks for looking into this!

tests/llmobs/test_llmobs_ragas_evaluators.py
```diff
@@ -238,3 +245,186 @@ def test_llmobs_with_faithfulness_emits_traces_and_evals_on_exit(mock_writer_log
...
+    rar_evaluator = RagasAnswerRelevancyEvaluator(llmobs)
+    failure_msg, _ = rar_evaluator.evaluate(_llm_span_without_io())
+    assert failure_msg == "fail_extract_answer_relevancy_inputs"
+    assert rar_evaluator.llmobs_service.submit_evaluation.call_count == 0
+
+
+def test_ragas_answer_relevancy_has_modified_answer_relevancy_instance(
+    ragas, mock_llmobs_submit_evaluation, reset_ragas_answer_relevancy_llm, llmobs
+):
+    """Context precision instance used in ragas evaluator should match the global ragas context precision instance"""
```
Yun-Kim, line 280:
```suggestion
    """Answer relevancy instance used in ragas evaluator should match the global ragas context precision instance"""
```

Yun-Kim (approved):
As a next step/PR I think I'd like llmobs eval tests to avoid mocking the `submit_evaluation` function and instead create a dummy eval writer and use eval events to assert finished evals ([like here for spans](https://github.com/DataDog/dd-trace-py/blob/main/tests/llmobs/conftest.py#L184-L214)). WYDT?

lievan:
>As a next step/PR I think I'd like llmobs eval tests to avoid mocking the submit_evaluation function and instead create a dummy eval writer and use eval events to assert finished evals ([like here for spans](https://github.com/DataDog/dd-trace-py/blob/main/tests/llmobs/conftest.py#L184-L214)). WYDT?

@Yun-Kim sounds good, will add this as a todo for leaning up tests/ragas integration!

PR #11929, opened 2025-01-14
ci: add healthcheck while setting up test http server process

gnufede:
CI: Add healthcheck to test http server to avoid failures

## Checklist

## Reviewer Checklist
- Title is accurate


tests/appsec/iast/conftest.py
```diff
@@ -162,11 +165,27 @@ def check_native_code_exception_in_each_python_aspect_test(request, caplog):
...
+    status = None
+    retries = 0
+    while status != 200 and retries < 5:
+        cmd = [
+            "python",
+            os.path.join(current_dir, "fixtures", "integration", "http_config_server.py"),
+            CONFIG_SERVER_PORT,
+        ]
+        process = subprocess.Popen(cmd, cwd=current_dir)
+        time.sleep(0.1)
```
brettlangdon, line 177: is half a second enough time for subprocess to come up?

brettlangdon (approved):
generic/non-blocking comment, overall lgtm

tests/appsec/iast/conftest.py
gnufede: If I understand correctly, [subprocess.Popen returns when the child process has been created](https://github.com/python/cpython/blob/main/Lib/subprocess.py#L1035). We're adding 0.1 seconds to wait until it's really ready to respond to requests, maybe it's not enough, but we're currently not waiting at all and it works most of the time.

gnufede, line 177:
```suggestion
        time.sleep(0.2)
```

PR #11952, opened 2025-01-15
chore(llmobs): update ragas trace ml app

lievan:
Update the ml application of any ragas evaluation span to be the same as the span being evaluated.

#### Refresher:
- Spans are generated by datadog for RAGAS operations
- These spans need to be identified as RAGAS-specific to avoid an infinite eval loop and tagged with 'runner.integration:ragas' for backend purposes
- Some of these spans are auto-instrumented through our langchain integration, meaning they can't be manually tagged

#### Previous Implementation:
- Added a `dd-ragas-` prefix to the ML application name for RAGAS spans
- This prefix was used to identify which spans came from RAGAS when spans are processed 

#### Now:
- add `_dd_ragas` prefix on ragas span's ml app for identification purposes, except only temporarily.
- **Just before sending these spans to the backend, remove the prefix**

An alternative solution is to utilize `annotation_contexts` to set a tag on auto-instrumented ragas spans. This is a larger refactor that i would like to explore when we clean up the ragas evaluators (more cleanly separate tracing setup vs actual eval logic).

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/llmobs/_llmobs.py
```diff
@@ -290,6 +286,10 @@ def _stop_service(self) -> None:
         except ServiceStatusError:
             log.debug("Error stopping LLMObs writers")
 
+        # Remove listener hooks for span events
```
lievan, line 289: this makes sure hooks aren't removed prematurely before while ragas is still running as triggered by `self._evaluator_runner.stop()`

ddtrace/llmobs/_constants.py
```diff
@@ -51,8 +51,9 @@
 RUNNER_IS_INTEGRATION_SPAN_TAG = "runner.integration"
 
 # The ml app of all ragas traces have this prefix that we use to detect
-# whether a span is generated from the ragas evaluation itself.
-RAGAS_ML_APP_PREFIX = "dd-ragas"
+# whether a span is generated from the ragas evaluation itself. We then
+# remove this prefix from the ml app before we submit the span.
+TEMP_RAGAS_ML_APP_PREFIX = "_dd_ragas_"
```
lievan, line 56: changed the prefix name to be more 'private' given it's not exposed to the user

ddtrace/llmobs/_llmobs.py
```diff
@@ -184,13 +184,13 @@ def _llmobs_span_event(cls, span: Span) -> Tuple[Dict[str, Any], bool]:
...
         metrics = span._get_ctx_item(METRICS) or {}
         ml_app = _get_ml_app(span)
+        span._set_ctx_item(ML_APP, ml_app)
 
         is_ragas_integration_span = False
-
-        if ml_app.startswith(constants.RAGAS_ML_APP_PREFIX):
+        if ml_app.startswith(constants.TEMP_RAGAS_ML_APP_PREFIX):
             is_ragas_integration_span = True
+            ml_app = ml_app.replace(constants.TEMP_RAGAS_ML_APP_PREFIX, "")
```
Yun-Kim, line 192: I'm not a fan of this temporary set/replace/remove behavior. What is the ultimate goal of using the ragas temp prefix? Can we not set a ragas identifier onto the span's internal store object instead of relying on a temporary ml app name?

lievan: thanks for the suggestion, just added in logic to use the span's store instead

ddtrace/llmobs/_constants.py
```diff
@@ -50,9 +50,9 @@
 # Used to differentiate traces of Datadog-run operations vs user-application operations.
 RUNNER_IS_INTEGRATION_SPAN_TAG = "runner.integration"
 
-# The ml app of all ragas traces have this prefix that we use to detect
-# whether a span is generated from the ragas evaluation itself.
-RAGAS_ML_APP_PREFIX = "dd-ragas"
+# All ragas traces have this context item set so we can differentiate
+# spans generated from the ragas integration vs user application spans.
+IS_EVALUATION_SPAN = "_is_evaluation_span"
```
Yun-Kim, line 55:
```suggestion
IS_EVALUATION_SPAN = "_ml_obs.evaluation_span"
```
Let's use the _ml_obs prefix to make it clear this is coming from us (for future reviewers/users)

ddtrace/llmobs/_llmobs.py
```diff
@@ -185,10 +187,8 @@ def _llmobs_span_event(cls, span: Span) -> Tuple[Dict[str, Any], bool]:
         metrics = span._get_ctx_item(METRICS) or {}
         ml_app = _get_ml_app(span)
 
-        is_ragas_integration_span = False
-
-        if ml_app.startswith(constants.RAGAS_ML_APP_PREFIX):
-            is_ragas_integration_span = True
+        is_evaluation_span = _is_evaluation_span(span)
+        span._set_ctx_item(IS_EVALUATION_SPAN, is_evaluation_span)
```
Yun-Kim, line 191: Do we need this check in this function scope?

ddtrace/llmobs/_llmobs.py
```diff
@@ -210,9 +210,9 @@ def _llmobs_span_event(cls, span: Span) -> Tuple[Dict[str, Any], bool]:
             llmobs_span_event["session_id"] = session_id
 
         llmobs_span_event["tags"] = cls._llmobs_tags(
-            span, ml_app, session_id, is_ragas_integration_span=is_ragas_integration_span
+            span, ml_app, session_id, is_ragas_integration_span=is_evaluation_span
```
Yun-Kim, line 213: Seems like we don't necessarily need to pass this information in to `_llmobs_tags()` or even return it from this function. Thoughts on just doing the checks individually when needed instead of passing it around?

ddtrace/llmobs/_llmobs.py
Yun-Kim: Was this triggering any bugs?

ddtrace/llmobs/_utils.py
```diff
@@ -127,6 +128,23 @@ def _get_span_name(span: Span) -> str:
...
+    nearest LLMObs span ancestor. Default to 'False'
+    """
+    is_evaluation_span = span._get_ctx_item(IS_EVALUATION_SPAN)
+    if is_evaluation_span is not None:
+        return is_evaluation_span
+    llmobs_parent = _get_nearest_llmobs_ancestor(span)
+    while llmobs_parent:
+        is_evaluation_span = llmobs_parent._get_ctx_item(IS_EVALUATION_SPAN)
+        if is_evaluation_span is not None:
+            return is_evaluation_span
```
Yun-Kim, line 143: Let's do a bool check to be defensive here

ddtrace/llmobs/_llmobs.py
lievan: we can move it outside of the span creation function and have a check in `_submit_llmobs_span`

ddtrace/llmobs/_llmobs.py
lievan: Yea, it was triggering a bug where traces of ragas evaluations were not being successfully traced in when an application was exiting since the listener hooks were removed before the tracing happened

ddtrace/llmobs/_llmobs.py
lievan: Just do a small refactor - lmk what you think!

ddtrace/llmobs/_llmobs.py
```diff
@@ -124,22 +125,28 @@ def _submit_llmobs_span(self, span: Span) -> None:
...
         is_llm_span = span._get_ctx_item(SPAN_KIND) == "llm"
-        is_ragas_integration_span = False
+        is_evaluation_span = False
         try:
-            span_event, is_ragas_integration_span = self._llmobs_span_event(span)
+            span_event = self._llmobs_span_event(span)
+
+            is_evaluation_span = _is_evaluation_span(span)
+            if is_evaluation_span:
+                span._set_ctx_item(IS_EVALUATION_SPAN, is_evaluation_span)
```
Yun-Kim, line 134: are we doing this for performance reasons?

Yun-Kim (commented):
Last small comments

ddtrace/llmobs/_llmobs.py
```diff
@@ -124,22 +125,28 @@ def _submit_llmobs_span(self, span: Span) -> None:
...
+                span_event["tags"].append("{}:ragas".format(constants.RUNNER_IS_INTEGRATION_SPAN_TAG))
```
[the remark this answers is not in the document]
lievan: Actually i just saw your other comment. So you recommend calling `_is_evaluation_span` twice?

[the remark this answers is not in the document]
lievan: I think that's cleaner & the function is constant time anyway because of gauranteed ragas structure, i'll update it, ty

ddtrace/llmobs/_llmobs.py
lievan: I'm going to remove it. there's little to no perf adv and if we remove it then the code inside the "try" block for generating span events would be simplified & ragas-free

ddtrace/llmobs/_llmobs.py
```diff
@@ -239,8 +237,6 @@ def _llmobs_tags(
             tags["error_type"] = err_type
         if session_id:
             tags["session_id"] = session_id
-        if is_ragas_integration_span:
-            tags[constants.RUNNER_IS_INTEGRATION_SPAN_TAG] = "ragas"
         existing_tags = span._get_ctx_item(TAGS)
         if existing_tags is not None:
             tags.update(existing_tags)
```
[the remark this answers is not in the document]
lievan: sounds good

Yun-Kim (approved):

PR #11962, opened 2025-01-15
chore(tracing): deprecates FilterRequestsOnUrl [3.0]

mabdinur:
In recent ddtrace versions spans can be sampled by tags and resource names via the `DD_TRACE_SAMPLING_RULES` configuration. Setting this configuration allows ddtrace to compute accurate trace metrics and prevents the generation of partial traces. 

To encourage the use of `DD_TRACE_SAMPLING_RULES`, `FilterRequestsOnUrl` will be deprecated.

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/trace/__init__.py
```diff
@@ -1,8 +1,7 @@
 from ddtrace._trace.context import Context
-from ddtrace._trace.filters import FilterRequestsOnUrl
```
mabdinur, line 2: This isn't a breaking change, this modifies an unreleased change.

ZStriker19: It looks like in the advanced_usage.rst it's a documented feature and we're changing it. Is that not breaking?

erikayasuda: Shouldn't we keep this in for 2.20, and then completely remove for 3.0+?

mabdinur: customers can still import `FilterRequestsOnUrl` from the `ddtrace.filters` module. They will just get a deprecation warning.

mabdinur: Here I am just removing it from `ddtrace.trace` package. Exposing FilterRequestsOnUrl in `ddtrace.trace` is something I just introduced and is unreleased.

erikayasuda (approved):

tests/tracer/test_filters.py
```diff
@@ -1,43 +1,8 @@
-from unittest import TestCase
```
ZStriker19, line 1: Could we keep these tests but just change the import path? Or make a private method these use for now? I'm worried about stripping out tests before something is actually removed.

mabdinur: I'll revert this. We don't need to change this test at all. It will just log a deprecation warning in ci.

ZStriker19 (approved):

itssimon:
@mabdinur Just stumbled upon this deprecation, but couldn't find any migration notes. How can DD_TRACE_SAMPLING_RULES be used to achieve the same as FilterRequestsOnUrl? This should be documented.

mabdinur:
> @mabdinur Just stumbled upon this deprecation, but couldn't find any migration notes. How can DD_TRACE_SAMPLING_RULES be used to achieve the same as FilterRequestsOnUrl? This should be documented.

`DD_TRACE_SAMPLING_RULES` allows you to sample **root spans** via resource names, service names, tags, and operation names. For example: `DD_TRACE_SAMPLING_RULES='[{"tags": {"http.url": "url_regex"}, "sample_rate": 0}]'`. 
Note: Sampling decisions are set using the first span in a trace. If the span containing the `http.url` tag is NOT a root span then this rule will not be applied. 


We highly recommend using sampling rules to filter spans. This ensures trace metrics and agent sampling decisions are correctly computed. However if you would like the ability to filter traces by non-root spans you can configure a custom trace filter/trace processor. For example:

```
import re
from ddtrace.trace import tracer
from ddtrace.trace import TraceFilter
from ddtrace.ext import http
from typing import List, Union

class FilterRequestsOnUrl(TraceFilter):
    r"""Filter out traces from incoming http requests based on the request's url.

    This class takes as argument a list of regular expression patterns
    representing the urls to be excluded from tracing. A trace will be excluded
    if its root span contains a ``http.url`` tag and if this tag matches any of
    the provided regular expression using the standard python regexp match
    semantic (https://docs.python.org/3/library/re.html#re.match).

    :param list regexps: a list of regular expressions (or a single string) defining
                         the urls that should be filtered out.

    Examples:
    To filter out http calls to domain api.example.com::

        FilterRequestsOnUrl(r'http://api\\.example\\.com')

    To filter out http calls to all first level subdomains from example.com::

        FilterRequestOnUrl(r'http://.*+\\.example\\.com')

    To filter out calls to both http://test.example.com and http://example.com/healthcheck::

        FilterRequestOnUrl([r'http://test\\.example\\.com', r'http://example\\.com/healthcheck'])
    """

    def __init__(self, regexps: Union[str, List[str]]):
        if isinstance(regexps, str):
            regexps = [regexps]
        self._regexps = [re.compile(regexp) for regexp in regexps]

    def process_trace(self, trace):
        """
        When the filter is registered in the tracer, process_trace is called by
        on each trace before it is sent to the agent, the returned value will
        be fed to the next filter in the list. If process_trace returns None,
        the whole trace is discarded.
        """
        for span in trace:
            url = span.get_tag(http.URL)
            if span.parent_id is None and url is not None:
                for regexp in self._regexps:
                    if regexp.match(url):
                        return None
        return trace

tracer.configure(trace_processors=[FilterRequestsOnUrl(["regex1", "regex2"])])
```

PR #11986, opened 2025-01-16
chore(llmobs): fix behavior for running multiple evaluators

lievan:
There was unexpected behavior when turning on multiple evaluators where only the last evaluator in the evaluators list was being run multiple times for one span.

Fix it by replacing `executor.map` with just individual calls to `executor.submit`, which should do the same thing but it avoids the the unexpected behavior above

To reproduce: the added test case in this PR fails if we switch back to `executor.map`. [See this CI run for the reproduced error](https://github.com/DataDog/dd-trace-py/commit/a6279dfa20745d29e46795bdc1e408227f5a0e0c)

## Checklist

## Reviewer Checklist
- Title is accurate


ddtrace/llmobs/_evaluators/runner.py
```diff
@@ -117,14 +117,9 @@ def periodic(self, _wait_sync=False) -> None:
...
-                        [
-                            span_event
-                            for span_event, span in span_events_and_spans
-                            if self.sampler.sample(evaluator.LABEL, span)
-                        ],
-                    )
+                    for span_event, span in span_events_and_spans:
+                        if self.sampler.sample(evaluator.LABEL, span):
+                            self.executor.submit(evaluator.run_and_submit_evaluation, span_event)
             else:
```
Yun-Kim, line 123: Can we refactor this + the sync execution to avoid duplicate code?

Yun-Kim (approved):

ddtrace/llmobs/_evaluators/runner.py
lievan: Ah yes, ty

PR #11987, opened 2025-01-16
chore(asm): improve dependency for api security

christophe-papazian:
In a rare occurence on system tests, api security could create an import error due to circular dependency.
This PR fixes that by delaying some import at object creation.

## Checklist

## Reviewer Checklist
- Title is accurate

### DIFF
--- a/ddtrace/appsec/_api_security/api_manager.py
+++ b/ddtrace/appsec/_api_security/api_manager.py
@@ -7,10 +7,6 @@ from typing import Optional
 
 from ddtrace import constants
 from ddtrace._trace._limits import MAX_SPAN_META_VALUE_LEN
-from ddtrace.appsec import _processor as appsec_processor
-from ddtrace.appsec._asm_request_context import add_context_callback
-from ddtrace.appsec._asm_request_context import call_waf_callback
-from ddtrace.appsec._asm_request_context import remove_context_callback
 from ddtrace.appsec._constants import API_SECURITY
 from ddtrace.appsec._constants import SPAN_DATA_NAMES
 from ddtrace.internal.logger import get_logger
@@ -55,6 +51,7 @@ class APIManager(Service):
         log.debug("Enabling %s", cls.__name__)
         cls._instance = cls()
         cls._instance.start()
+
         log.debug("%s enabled", cls.__name__)
 
     @classmethod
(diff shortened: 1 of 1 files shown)


ddtrace/appsec/_api_security/api_manager.py
```diff
@@ -75,12 +72,18 @@ def __init__(self) -> None:
         log.debug("%s initialized", self.__class__.__name__)
         self._hashtable: collections.OrderedDict[int, float] = collections.OrderedDict()
 
+        from ddtrace.appsec import _processor as appsec_processor
+        import ddtrace.appsec._asm_request_context as _asm_request_context
```
gnufede, line 76: Both imports trigger the ctypes init?

ddtrace/appsec/_processor.py
```diff
@@ -434,3 +436,10 @@ def on_span_finish(self, span: Span) -> None:
                     del self._span_to_waf_ctx[s]
                 except Exception:  # nosec B110
                     pass
+
+
+# load waf at the end only to avoid possible circular imports with gevent
+import ddtrace.appsec._ddwaf as ddwaf  # noqa: E402
+from ddtrace.appsec._metrics import _set_waf_init_metric  # noqa: E402
+from ddtrace.appsec._metrics import _set_waf_request_metrics  # noqa: E402
+from ddtrace.appsec._metrics import _set_waf_updates_metric  # noqa: E402
```
gnufede, line 445: For these imports that are not used here, maybe worth just importing them directly wherever they are used instead of using this class as a proxy?

gnufede (approved):

ddtrace/appsec/_api_security/api_manager.py
christophe-papazian: At first yes, but after the changes not anymore. But it's still allow to reduce the direct dependencies between the modules.

ddtrace/appsec/_processor.py
christophe-papazian: As they are used inside the AppsecSpanProcessor where everything is critical, I prefer not to import them locally, that's why I kept the import global, but at the end to avoid circular errors.

avara1986 (approved):

avara1986:
LGTM!

PR #12004, opened 2025-01-20
fix: set span error metadata only for escaped errors

dubloom:
Before this change we set some span error metadata for every call of record_exception. This would set the span as an errored span. However, we can call several times record_exception within a span and we can pass handled exceptions as arguments. In the later case, we want the span not to be errored.

This change aims at addressing this. We will set the relevant metadata only if the error is escaped (which means unhandled).

## Checklist

## Reviewer Checklist
- Title is accurate

### DIFF
--- a/releasenotes/notes/fix-otel-record-exception-api-17f3a4b23948ec79.yaml
+++ b/releasenotes/notes/fix-otel-record-exception-api-17f3a4b23948ec79.yaml
@@ -1,5 +1,4 @@
 ---
 fixes:
   - |
-    tracing(otel-api): Ensures a call to otel-api record-exception sets
-    the span error metadata only if escaped is True.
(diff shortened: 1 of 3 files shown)


emmettbutler (approved):

releasenotes/notes/fix-otel-record-exception-api-17f3a4b23948ec79.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    tracing(otel-api): Ensures a call to otel-api record-exception sets
+    the span error metadata only if escaped is True.
```
mabdinur, line 5:
```suggestion
    opentelemetry: Ensures ``Span.record_exception(...)`` sets the span error metadata only if escaped is True.
```

mabdinur (approved):
I clean up the commit history and left one suggestion. 

Overall looks good to me!

